PR #40, opened 2026-06-30
Hide the fullscreen control when running inside an iframe (#1062) (replicates mxtommy/Kip #1075)

mairas:
Replicates upstream PR https://github.com/mxtommy/Kip/pull/1075 ("Hide the fullscreen control when running inside an iframe (#1062)") by dillan.

Method: commit-by-commit cherry-pick (the upstream PR had a linear history with no merge commits).

This is an experimental replica carried in the SKip fork for evaluation. It should be reviewed and tested before merge.

### DIFF
--- a/src/app/core/services/uiEvent.service.spec.ts
+++ b/src/app/core/services/uiEvent.service.spec.ts
@@ -1,6 +1,6 @@
 import { TestBed } from '@angular/core/testing';
 import { beforeEach, describe, expect, it } from 'vitest';
-import { uiEventService } from './uiEvent.service';
+import { uiEventService, isEmbeddedInIframe } from './uiEvent.service';
 
 describe('GestureService', () => {
   let service: uiEventService;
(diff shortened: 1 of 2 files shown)


mairas:

**Verdict:** Safe to adopt. Byte-identical faithful replica of upstream #1075; correct iframe detection, tests assert the fixed behavior, no marine-safety surface.

- CI: all 13 checks green
- No actionable findings.

mairas:

**Independent verdict:** correct — The iframe detection `window.self !== window.top` (with a catch-all returning true) correctly hides the fullscreen control in every embedded context and correctly shows it at top level. Judged purely on the code: at top level self and top reference the same WindowProxy so it returns false (control shown); in any framed context (same-origin, cross-origin, or sandboxed opaque-origin) the iframe's window differs from the host's so it returns true (control hidden and toggleFullScreen no-ops). The change is isolated to the fullscreen affordance and touches no instrument-data code path, so it cannot produce wrong or stale heading/wind/depth/speed/position/autopilot values.

Confidence: `high`.

- [nit] Doc comment overstates when the SecurityError is thrown

PR #41, opened 2026-06-30
[codex] Fix autopilot standby button label (replicates mxtommy/Kip #1068)

mairas:
Replicates upstream PR https://github.com/mxtommy/Kip/pull/1068 ("[codex] Fix autopilot standby button label") by macjl.

Method: commit-by-commit cherry-pick (the upstream PR had a linear history with no merge commits).

This is an experimental replica carried in the SKip fork for evaluation. The change fixes the autopilot standby button label and adds a unit test for the widget-autopilot component.

### DIFF
--- a/src/app/widgets/widget-autopilot/widget-autopilot.component.html
+++ b/src/app/widgets/widget-autopilot/widget-autopilot.component.html
@@ -15,7 +15,7 @@
         <svg width="100%" height="100%" viewBox="0 0 100 14" preserveAspectRatio="xMidYMid meet" class="svg-element svg-modes">
           <text xml:space="preserve"
             x="50.090435"
-            y="10.743966">Disengage</text>
+            y="10.743966">{{ standbyButtonLabel() }}</text>
         </svg>
       </mat-icon>
     </button>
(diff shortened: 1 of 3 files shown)


mairas:
## Automated review — issue #69 tier-2 adoption pass

**Verdict:** Safe to adopt. Faithful, byte-identical replica of upstream #1068; the fix is correct, the label tracks the actual engage/disengage REST action for API v2, and the three new tests assert the fixed behavior.

- CI: all 13 checks green

One non-blocking observation, **not actioned** (theoretical + pre-existing, byte-identical to upstream): during the transient `apEngaged === null` window before data arrives, the standby label can briefly read "Disengage" while a click sends Engage. Pre-dates this PR. Closing that gap would be an intentional divergence from upstream — left as a maintainer decision.

mairas:
## Correction — finding re-classified as valid and fixed

My earlier comment deferred the null-state label issue as an "upstream divergence." That was wrong: the upstream PR itself wasn't reviewed, so faithfulness to it doesn't imply correctness. On reconsideration this is a **valid marine-safety-relevant bug** and has been fixed (commit `8f7586d0`).

**Bug:** the v2 standby toggle POSTs *engage* whenever `apEngaged()` is falsy (`!this.apEngaged()`), but `standbyButtonLabel()` only returned "Engage" when `apEngaged() === false`. During the transient `apEngaged === null` window before autopilot state arrives, the button read **"Disengage" while a click actually engaged** the autopilot.

**Fix:** the label now uses the same falsy check as the action (`!this.apEngaged()`), so null and false both read "Engage" and only a known-engaged state reads "Disengage". Added a regression test for the null case. All 13 CI checks green.

PR #42, opened 2026-06-30
Restore TTL reset to dashes after a data timeout (#1069) (replicates mxtommy/Kip #1076)

mairas:
Replicates upstream PR https://github.com/mxtommy/Kip/pull/1076 ("Restore TTL reset to dashes after a data timeout (#1069)") by dillan.

Method: commit-by-commit cherry-pick (linear upstream history, no merge commits).

This is an experimental replica carried in the SKip fork for evaluation. It restores the behaviour where widget values reset to dashes (the "bootstrap-null" display) after a data timeout has elapsed, fixing a regression introduced in a prior refactor. The change adds a failing regression test first (RED) and then makes the bootstrap-null suppression stateful so it is only applied on the very first bootstrap, not on subsequent timeout-driven null values (GREEN).

### DIFF
--- a/src/app/core/directives/widget-streams.directive.spec.ts
+++ b/src/app/core/directives/widget-streams.directive.spec.ts
@@ -1,6 +1,6 @@
 import { TestBed } from '@angular/core/testing';
 import { beforeEach, afterEach, describe, expect, it, vi } from 'vitest';
-import { Subject, Observable } from 'rxjs';
+import { Subject, BehaviorSubject, Observable } from 'rxjs';
 import { WidgetStreamsDirective } from './widget-streams.directive';
 import { DataService, IPathUpdate } from '../services/data.service';
 import { UnitsService } from '../services/units.service';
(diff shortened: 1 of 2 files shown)


mairas:
## Automated review — issue #69 tier-1 adoption pass

**Verdict:** Safe to adopt. Byte-identical faithful replica of upstream #1076; the TTL-reset fix is correct against SKip's real DataService and the regression test genuinely asserts the fixed safety behavior (values reset to "--" after timeout).

- CI: all 13 checks green
- No actionable findings.

mairas:

Following a review-depth concern, this tier-1 PR got a second, **content-only** review that treats the upstream PR as *unreviewed*, bans "matches upstream / byte-identical" as a correctness argument, and requires every claim to be backed by concrete input→code→output traces (5 traces here), with emphasis on boundary/transient states.

**Independent verdict:** correct — The stateful `filter` with an externally-captured `seenNonNull` latch correctly drops only leading (bootstrap) nulls and, once any real value has been observed, lets a later timeout-driven null pass through so the widget resets to '--'. Because the closure is created once (outside the operator) and every retry/resubscription re-enters the same closure, the TTL null is no longer swallowed on resubscribe. It cannot reset a live value (nulls only appear via timeoutPathObservable, which only fires when no data arrives within dataTimeout) and it cannot fail to reset a dead value on the default source. Value 0 / negative angles are correctly treated as real (not suppressed).

Confidence: `high`.

- [minor] TTL reset targets the first path registration regardless of source (pre-existing, not introduced by this PR)
- [nit] Retry loop re-emits null every ~10s while a path stays dead (harmless)

<sub>Reviewed post-merge for the record.</sub>

PR #47, opened 2026-06-30
Add MCP dashboard-schema generator (replicates mxtommy/Kip #1072)

mairas:
Replicates upstream PR https://github.com/mxtommy/Kip/pull/1072 ("Add MCP dashboard-schema generator") by dillan.

Method: commit-by-commit cherry-pick (10 commits, linear history) preserving original authorship.

This is an experimental replica carried in the fork for evaluation only. It has not been built or test-run as part of replication.


mairas:
## Content review: accept-with-fixes — sound tooling, confirmed inert at runtime

Read the generator (`generate.ts`, `ast.ts`, `serialize.ts`) in full. Confirmed runtime-inertness: nothing under `src/` imports the tool or the artifact, and `package.json` adds only two npm scripts (no `bin`/`postinstall`/`prepare`). The generator is genuinely careful — no `child_process`/`exec`/`eval`/dynamic require; all fs reads are hardcoded paths; parsing is pure syntactic `ts.createSourceFile` (Kip source is never executed); `literalToValue` fails loud on anything it can't statically resolve rather than guessing; output is deterministic with a CI drift gate.

- `[low · design]` the ~92 KB `src/assets/kip-dashboard-schema.json` is swept into the Angular assets glob → shipped to every device and served at `/assets/...`, though only the external MCP server consumes it. Move it out of `src/assets` or exclude it from the glob.
- `[low · correctness]` `SCHEMA_VERSION`/`CONFIG_FILE_VERSION`/theme constants are hardcoded with no test binding them to their source of truth → silent drift if Kip bumps them (the rest of the generator reads from source and fails on drift; these don't).
- `[info]` stale doc comment (references `ts-morph`; the tool uses the TS compiler API), and unrelated d3/branding changes ride along (the d3-selection/d3-shape dep removals are verified safe — `src/` imports only the umbrella `d3`).

Sound dev-only tooling; the fixes are hygiene, none touch on-device data paths.

mairas:
## Fixes pushed

- **Bundle bloat / served artifact:** the ~91 KB `kip-dashboard-schema.json` is now excluded from the Angular assets glob, so it no longer ships to every device or gets served at `/assets/...`. It stays in the repo for the generator and the external MCP-server consumer.
- **Doc comment:** corrected the stale "ts-morph" reference (the tool uses the TypeScript compiler API).

Verified: build excludes the artifact from `public/`, generator test suite 24/24 still green, app builds.

**Not changed:** the hardcoded `SCHEMA_VERSION` / `CONFIG_FILE_VERSION` / theme constants (low, drift risk). Binding them to their source of truth would need a source-reading assertion in the drift-guard spec; deferred as this is external-only dev tooling with no on-device runtime footprint.

PR #48, opened 2026-06-30
Fix Help images 404 when KIP is served under a base href (#1056) (replicates mxtommy/Kip #1074)

mairas:
Replicates upstream PR https://github.com/mxtommy/Kip/pull/1074 ("Fix Help images 404 when KIP is served under a base href (#1056)") by dillan.

Method: commit-by-commit cherry-pick of the two upstream commits (RED guard + GREEN fix) onto `origin/main`, preserving the original author. No conflicts arose.

This is an experimental replica carried in the fork for evaluation. It should be built and reviewed before merge.


mairas:
## Content review: accept — correct and complete

Scanned all help-docs markdown on the branch: the two parent-relative paths (`chartplotter.md`, `welcome.md`) are the only offenders and both are fixed to base-relative `assets/...`, consistent with how `index.html` and `menu.json` resolve under a sub-path base href. The added vitest guard catches `../` and `/`-leading forms.

- `[info · test-quality]` the guard is a path-*shape* check, not an existence check, and doesn't catch a `./`-prefixed relative form — minor, worth tightening the regex.

Clean, well-scoped fix with a useful regression guard.

PR #51, opened 2026-06-30
Add regression tests for solar widget config persistence (#1061) (replicates mxtommy/Kip #1079)

mairas:
Replicates upstream PR https://github.com/mxtommy/Kip/pull/1079 ("Add regression tests for solar widget config persistence (#1061)") by dillan.

Method: commit-by-commit cherry-pick (linear history, no merge commits).

This is an experimental replica carried in the fork for evaluation. It has not been independently tested in the SKip build environment and should be reviewed before merge.


mairas:
## Content review: accept-with-fixes — the two component tests are real; the round-trip tests are not

Judged test quality, not count. The two `bootThenData` tests genuinely exercise the component (mock a no-data reboot → feed live panel power → assert `gaugeProgress` derives from the persisted `arrayRatedPowerW`) and would catch the #1061 regression returning.

- `[medium · test-quality]` the four "config persistence round-trip" tests build literal fixtures and round-trip them through `JSON.stringify/parse` + lodash `cloneDeep/merge` **directly** — they assert JSON/lodash round-trip objects, not Kip's own save/load code, so they wouldn't catch the real persistence path breaking.
- `[low · test-quality]` the "config merge" test only validates `DEFAULT_CONFIG`'s shape, not the index-wise array-merge behavior that actually caused #1061.

Keep the two component tests; either point the round-trip tests at the real persistence functions or drop them as they give false coverage confidence.

PR #52, opened 2026-06-30
Cut per-delta work on the data hot path (replicates mxtommy/Kip #1085)

mairas:
Replicates upstream PR https://github.com/mxtommy/Kip/pull/1085 ("Cut per-delta work on the data hot path") by dillan.

Method: commit-by-commit cherry-pick (linear history, single commit).

This is an experimental replica carried in the SKip fork for evaluation. It has not been independently tested in the fork context and needs a build + review before merge.

### DIFF
--- a/src/app/core/directives/widget-runtime.directive.ts
+++ b/src/app/core/directives/widget-runtime.directive.ts
@@ -1,4 +1,4 @@
-import { Directive, computed, effect, input, output, signal, untracked } from '@angular/core';
+import { Directive, computed, effect, input, signal, untracked } from '@angular/core';
 import { cloneDeep, merge } from 'lodash-es';
 import type { IWidgetSvcConfig } from '../interfaces/widgets-interface';
 
@@ -17,7 +17,6 @@ export class WidgetRuntimeDirective {
 
   // Default config input (typically from widget manifest DEFAULT_CONFIG)
   protected defaultConfig = signal<IWidgetSvcConfig | undefined>(undefined);
-  protected runtimeConfig = output<IWidgetSvcConfig | undefined>();
 
   // Internal runtime config
   private _runtimeConfig = signal<IWidgetSvcConfig | undefined>(undefined);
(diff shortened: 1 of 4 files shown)


mairas:

**Verdict:** Safe to adopt. Byte-identical, faithful replica of upstream #1085; the reordering and lazy-timestamp changes are behaviorally equivalent on all displayed values/timestamps, all 13 CI checks green, no maintainer fix commit needed.

- CI: all 13 checks green

One non-blocking nit, **not actioned** (theoretical): the added test doesn't actually distinguish lazy from eager `Date` allocation, but the production change is behaviourally equivalent on all displayed values/timestamps.

mairas:

Following a review-depth concern, this tier-3 PR got a second, **content-only** review that treats the upstream PR as *unreviewed*, bans "matches upstream / byte-identical" as a correctness argument, and requires every claim to be backed by concrete input→code→output traces (7 traces here), with emphasis on boundary/transient states.

**Independent verdict:** correct — On the code's own merits, the reorder emits identical values and timestamps for realistic inputs. Unit conversion is a pure per-element function, and sampleTime/take(1) SELECT elements rather than aggregate, so convert(sample(x)) == sample(convert(x)); moving convert after the sampler changes only how often it runs, not what is emitted. skipWhile relocated onto the raw stream still selects the same first delta because conversion preserves null-ness (null->null, non-null->non-null). The lazy memoized timestamp yields the same Date|null value; the sentinel (undefined) can never collide with a resolved value (Date|null), and the raw timestamp is captured by value as an immutable string, so no staleness/aliasing. $source bucket selection is upstream and untouched. The removed runtimeConfig output has zero consumers and was a side-effect-in-computed. The only divergence is a misconfiguration-only edge (unknown convert unit + suppressBootstrapNull), which changes hang-forever vs emit-null, never a wrong instrument value.

Confidence: `high`.

- [minor] skipWhile-on-raw + unknown convert unit changes suppression behavior

PR #53, opened 2026-06-30
Trim startup bundle and unblock first paint (replicates mxtommy/Kip #1084)

mairas:
Replicates upstream PR https://github.com/mxtommy/Kip/pull/1084 ("Trim startup bundle and unblock first paint") by dillan.

Method: commit-by-commit cherry-pick (linear history, single commit, applied cleanly with no conflicts).

This is an experimental replica carried in the fork for evaluation. It has not been built or runtime-tested here; review and a build are recommended before merge.


mairas:

**Verdict:** Safe to adopt. Byte-identical faithful replica of upstream #1084 (bundle/startup trim); all 13 CI checks green including production webapp-build, lint, and the vitest matrix. No correctness or marine-safety impact.

- CI: all 13 checks green
- No actionable findings.

mairas:

**Independent verdict:** correct — Judged purely on the code, the change trims the eager startup bundle without removing or wrongly deferring anything needed at/before first paint. (1) The d3 narrowing is a pure alias swap: all 7 widgets only ever used d3.select/d3.Selection/d3.arc, which d3-selection.select / d3-shape.arc provide identically (same v3 sub-packages d3@7 already re-exports); instance .select() calls are unchanged; PR head has zero dangling full-d3 references. (2) prism/clipboard are moved from eager global scripts (every route) into the lazily-loaded app-help chunk, which is the ONLY ngx-markdown consumer; ngx-markdown reads global Prism (mjs:227) and global ClipboardJS (mjs:325), and the help chunk sets both at module-eval before <markdown> renders — no other component reads those globals. (3) defer on tween/steelseries still runs them before Angular's module bundles (head precedes body-end in deferred document order), so the steelseries/Tween globals exist before gauge-steel/widget-horizon evaluate, and both consumers additionally guard typeof steelseries==='undefined'. (4) Removing core-js/es/reflect is inert on evergreen browsers (native ES Reflect; it never provided reflect-metadata) and emitDecoratorMetadata:false is safe under Angular AOT + inject()-based DI with no metadata-dependent libraries. main.ts eager providers and both appInitializers are untouched.

Confidence: `high`.

- [nit] Production initial-bundle budget error (3.25mb) is a build-time gate, not a runtime risk

PR #54, opened 2026-06-30
Image assets (1/3): Signal K plugin secure upload, processing, worker pool, cache (replicates mxtommy/Kip #1080)

mairas:
Replicates upstream PR https://github.com/mxtommy/Kip/pull/1080 ("Image assets (1/3): Signal K plugin secure upload, processing, worker pool, cache") by dillan.

Method: commit-by-commit cherry-pick (5 commits, linear history, no merge commits). Each commit's original author was preserved. The package.json and package-lock.json changes auto-merged cleanly with no conflict; SKip identity (`@halos-org/skip`, version 4.8.0) was retained while the PR's new dependencies (`heic-convert`, `isomorphic-dompurify`, `multer`, `sharp`) and the `test:plugin` script change were incorporated.

Note: `package-lock.json` was auto-merged by git and will need regeneration (`npm install`) before a clean build, since dependency installation/build was intentionally not run during replication.

This is an experimental replica carried in the fork for evaluation. It needs a build + review before merge.


mairas:
## Content review: accept-with-fixes — strong ingest/sanitization, but verified DoS + authz gaps to close before merge

Reviewed the full implementation (`image-router.ts`, `image-store.ts`, `worker-pool.ts`, `image-processing.ts`, `image-worker.ts`) across three lenses — upload/sanitization, processing-DoS/filesystem, authz/reliability — reading the code and attempting exploits. Every finding below was independently re-verified against the code. This is **not** a clean adopt: the sanitization is genuinely strong, but the serving path has DoS vectors that can OOM the Signal K host process, and the write gate is authentication-only.

**What's genuinely solid (verified in code):**
- **Type smuggling is closed.** `detectImageType` keys off magic bytes and ignores the client MIME/extension; every raster (and HEIC, transcoded at ingest) is unconditionally re-encoded to WebP by sharp before serving, so a JPEG/HTML or WebP/HTML polyglot loses any payload. Only SVG is served as-is, behind layered defense.
- **SVG XSS defense is layered and I couldn't break it:** DOMPurify at ingest + serve-time CSP `sandbox` + `X-Content-Type-Options: nosniff`.
- **Decompression-bomb caps run before full decode:** 10 MB byte cap (multer + re-checked in ingest), a `width*height*pages > 50 MP` raster check via `sharp.metadata()`, and a stricter 24 MP budget before the pure-JS heic-convert decode.
- **The worker-pool crash-recovery state machine is well-engineered:** idempotent `failWorker` via a `dead` WeakSet (a native crash firing both `error`+`exit` doesn't double-spawn), per-job wall-clock timeout with `timer.unref()`, poison job rejected once and not retried.

**Findings:**

`[high · reliability]` **Serving flood can OOM the Signal K host process** — `image-store.ts:281-289`, `worker-pool.ts:66-77`, `image-router.ts:124-142`. The worker-pool job queue is unbounded and `GET /images/:id` is unauthenticated with no rate/concurrency limit. On a cache miss, `getServable` reads the full original (≤10 MB) into main-thread heap *then* queues the job; coalescing only dedupes identical `id:width` keys. A burst of `GET /images/<id>?w=<width>` across all widths × all ids parks each original in the unbounded queue → the main SK process (which also serves the boat's navigation data) OOM-kills. Verified real.

`[high · reliability]` **Concurrent cache-miss GETs multiply full-size buffers in memory** — same root; ~1000 concurrent misses ≈ multiple GB of heap on a 2–8 GB Pi. Verified real. Fix both with a bounded queue + a concurrency cap on the serve route.

`[medium · security]` **All read routes are unauthenticated** — `image-router.ts:86,101,133`. `GET /images` (list — includes `uploadedBy` principal ids), `GET /images/:id`, `GET /images/cache` have no `isAuth`. On a *secured* boat with no anonymous-read allowance, a guest on the (often open) WiFi enumerates and downloads the entire shared library. Verified real.

`[medium · security]` **Write gate checks authentication, not authorization** — `image-router.ts:19-24`. `isAuthenticatedRequest` returns true for any principal with an identifier; it never checks permission level, so a read-only SK account (e.g. crew given view-only access) can `POST`/`DELETE` images and purge the cache. Verified real.

`[medium · reliability]` **multer leaves non-file fields unbounded** — `image-router.ts:54`. `limits` sets only `fileSize`+`files:1`; `fields`/`parts`/`fieldSize` default to Infinity/1 MB. A multipart body with a tiny valid file + a stream of 1 MB text fields buffers into `req.body` (memoryStorage) until OOM. Verified real — set `fields`/`parts` limits.

`[medium · design]` **Auth rests on an unverified SK assumption** — `image-router.ts:20-22` fail-opens (returns true) when both `skPrincipal` and `skIsAuthenticated` are undefined. If SK doesn't populate those on plugin-router requests on the deployed version, a secured server silently fail-opens on writes. Needs verification against a real security-enabled SK server before unattended deployment — not a deferring code comment.

`[low · reliability]` **`destroy()` leaks pending jobs** — `worker-pool.ts:161-165` terminates workers without rejecting queued/in-flight job promises; awaiting Express handlers hang until socket timeout on plugin stop/restart. Verified real (downgraded to low).

Plus lows: variant cache has no size eviction (unauth GETs can force ~3.5 GB across 500 images × 7 widths on the shared SK data volume), TOCTOU-racy quota under concurrent uploads, orphaned originals never reclaimed on partial write, non-atomic cache/original writes, and a clean-idle-exit zombie worker burning one job into a timeout.

**Bottom line:** the ingest/sanitization path is well-built and I couldn't smuggle a malicious file or stored XSS through it. But before this runs unattended, the two HIGH OOM vectors (unbounded queue + no serve-path backpressure), the open read routes, and the authn-only write gate should be fixed — taking down the Signal K server is the worst failure mode here.

_(Resolution applied earlier: conflict-free; CI fix — unquoted the plugin-test glob so Node 20's `node --test` finds the files. Green on 20/22/24.)_

mairas:
## Fixes pushed

Addresses the DoS + reliability findings from the review:
- **Serve-path OOM (both HIGH):** `getServable` coalesces concurrent identical variant generations and caps concurrent *distinct* generations (a `GenerationLimiter`), so a serve flood can no longer multiply full-original buffers in memory. The worker-pool queue is bounded and `destroy()` now rejects pending jobs instead of leaking them. Excess load returns 503.
- **multer:** `fields: 0` closes the unbounded-text-field memory DoS.
- **Read gating:** `GET /images` (list, exposes uploader ids) and `/images/cache` now require auth — both go through the app's authenticated HTTP client. `GET /images/:id` (serve) stays open by necessity (a browser `` can't carry the bearer token); its DoS is bounded above rather than by auth.
- **Atomic writes:** originals + cached variants written via temp + rename.

Tests added: coalescing, concurrency-cap 503, queue bound, destroy rejection, read-gating, serve-open, 503 mapping. Plugin suite 100/100 on Node 24, 0 fail on Node 20; CI green.

**Deliberately not fixed here** (flagging so it isn't lost): the write gate still checks *authentication, not write permission* — a read-only SK account can still upload/delete/purge — and it fail-opens if SK sets no principal fields on plugin routes. `@signalk/server-api` exposes no principal/permission surface to key these off, so this needs an SK-server integration decision and an end-to-end check against a security-enabled server, which can't be done from this repo. Left for the e2e step.

PR #55, opened 2026-06-30
Fix resize, data-inspector and race-timer hot spots (replicates mxtommy/Kip #1086)

mairas:
Replicates upstream PR https://github.com/mxtommy/Kip/pull/1086 ("Fix resize, data-inspector and race-timer hot spots") by dillan.

Method: commit-by-commit cherry-pick (linear history, no merge commits in the upstream PR).

This is an experimental replica carried in the SKip fork for evaluation. It has not been independently verified against the SKip build — a build and functional review are required before merging.

### DIFF
--- a/src/app/core/components/dashboard/dashboard.component.html
+++ b/src/app/core/components/dashboard/dashboard.component.html
@@ -7,7 +7,6 @@
   (swipeup)="previousDashboard()"
   (swipedown)="nextDashboard()"
   (press)="addNewWidget($event)"
-  (window:resize)="resizeGridColumns()"
 >
 </gridstack>
 @if (gridIsEmpty()) {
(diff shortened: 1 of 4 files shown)


mairas:

**Verdict:** Safe to adopt. Byte-identical faithful cherry-pick of upstream #1086; all three fixes (race-timer subscription lifecycle, data-inspector throttle, redundant resize handler removal) are correct in SKip, the leak test genuinely pins the fix, and lint is clean. Nothing blocks merge.

- CI: all 13 checks green
- No actionable findings.

mairas:

**Independent verdict:** correct — judged on the code itself. (1) The race-timer fix eliminates a real subscription-stacking defect: ensureTimer runs from the config effect (every options() change) and from resetTimer() against a shared BehaviorSubject, so the old un-stored subscribe stacked observers, multiplying beeps/draws per tick. The new code unsubscribes the prior sub, resubscribes once, and binds via takeUntilDestroyed(destroyRef); traced mount/re-arm/reset/destroy all yield exactly one observer, preserve the running value on re-arm, reset to configured length, and produce zero callbacks after destroy. No stale/wrong displayed value, no leak, no double-count. (2) data-inspector throttleTime(500, async, leading+trailing) rate-limits a diagnostic table while trailing:true guarantees eventual convergence to the latest full-tree state within 500ms — no permanent staleness. (3) Removing (window:resize) drops nothing: the pre-existing ResizeObserver on the dashboard host already covers window-driven host resize (and resizeGridColumns early-returns when host height is unchanged), making the binding strictly redundant.

Confidence: `high`.

No findings — confirmed correct on the code's own merits.

PR #57, opened 2026-06-30
Lazy-load widget components and the history chart dialog (replicates mxtommy/Kip #1088)

mairas:
Replicates upstream PR https://github.com/mxtommy/Kip/pull/1088 ("Lazy-load widget components and the history chart dialog") by dillan.

**Method:** commit-by-commit cherry-pick onto `origin/main` (linear upstream history, single commit, no conflicts).

This is an experimental replica carried in the fork for evaluation. It has not been built or runtime-tested here; treat it as a candidate for review before merge.

### DIFF
--- a/src/app/core/components/widget-embedded/widget-embedded.component.html
+++ b/src/app/core/components/widget-embedded/widget-embedded.component.html
@@ -1,3 +1,9 @@
 <mat-card appearance="raised" class="widget-container kip-widget-card-colors" style="border: 1px dashed transparent">
   <ng-container #childOutlet></ng-container>
+  @if (loadFailed()) {
+    <button type="button" class="widget-load-error" (click)="retryWidgetLoad()">
+      <span class="widget-load-error__title">Widget failed to load</span>
+      <span class="widget-load-error__hint">Tap to retry</span>
+    </button>
+  }
 </mat-card>
--- a/src/app/core/components/widget-embedded/widget-embedded.component.scss
+++ b/src/app/core/components/widget-embedded/widget-embedded.component.scss
@@ -15,3 +15,25 @@
 .kip-widget-card-colors {
   background-color: var(--kip-widget-card-background-color);
 }
+
+.widget-load-error {
+  display: flex;
+  flex-direction: column;
+  gap: 4px;
+  align-items: center;
+  justify-content: center;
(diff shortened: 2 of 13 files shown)


mairas:

**Verdict:** Safe to adopt. Faithful cherry-pick of upstream #1088 (byte-identical at the +/- level) plus one correct maintainer CI-fix commit; lazy-loading lifecycle is sound, lint clean, CI green. Nothing blocks.

- Replica faithfulness: `adapted-correctly`
- CI: all 13 checks green

**Maintainer CI-fix commit** (added one extra microtask flush in the `widget-host2` auto-open specs, needed because lazy-loading made child creation async) was independently verified as correct; the lazy-loading lifecycle itself is sound.

mairas:
## Correction + fixes from content re-review

My earlier comment called this PR "sound, no findings." That was too shallow — it leaned on upstream-faithfulness. A deeper content re-review (upstream treated as unreviewed, numeric/state traces required) found two real issues, both now **fixed** on this branch (CI green):

**F1 — major, fixed (`7341e9f6`... history-sync commit).** Lazy-loading made each widget's `DEFAULT_CONFIG` available only after its chunk rendered. `DashboardHistorySeriesSyncService` reads those defaults to compute history series, so widgets on **non-active dashboards** (whose chunks never load) had series mis-registered — opt-out widgets (e.g. windsteer) got **spurious** recorded series, default-path widgets (e.g. horizon) **lost** theirs — self-healing only on an unrelated dashboard edit. Live instrument display was unaffected; the damage was to history recording. Fix: the reconcile flush now warms each in-use widget type's `DEFAULT_CONFIG` (`getComponentType`) before extracting series, restoring determinism. It runs after first paint, so the initial-bundle / first-paint win this PR targets is preserved. Added a regression test. (The full alternative — extracting every widget's `DEFAULT_CONFIG` into a static registry — was rejected as a 35-component refactor for no extra benefit here.)

**F2 — minor, fixed (widget-host commit).** A failed widget chunk import (flaky wifi + cold cache after an app update; no service worker to mitigate) left a silent blank tile with no recovery. Now `widget-host2` and `widget-embedded` show a "Widget failed to load — tap to retry" placeholder that re-imports on tap.

PR #61, opened 2026-06-30
Video widget (3/3): accessibility, theming and UX overhaul (replicates mxtommy/Kip #1094)

mairas:
Replicates upstream PR https://github.com/mxtommy/Kip/pull/1094 ("Video widget (3/3): accessibility, theming and UX overhaul") by dillan.

Method: commit-by-commit cherry-pick (linear upstream history, 65 commits, original authorship preserved). The cherry-pick applied cleanly with no conflicts; SKip's package.json identity (`@halos-org/skip`, version) was retained automatically and the PR's new dependencies (`hls.js`, `piexifjs`, `@types/piexifjs`) were added.

Note: `package-lock.json` carries the upstream-replicated changes and will need regeneration before a build.

This is an experimental replica carried in the fork for evaluation. It has not been built or runtime-verified here.


mairas:
## Content review: accept-with-fixes — streaming security is solid; real marine-safety gaps remain

Read the widget, config, and stream utils in full across three lenses (streaming-security, resource-lifecycle, correctness+marine), attempting URL-bypass, leak, and stale-state exploits. Findings re-verified against the code.

**Genuinely solid (verified):**
- **URL scheme validation is done right and I couldn't bypass it:** the direct-URL source runs the raw string through the WHATWG `URL` parser and gates on the *normalized* `protocol`, binding the normalized `href`. `javascript:`/`data:`/`blob:`/`file:`, mixed-case, and embedded-whitespace tricks all rejected.
- **PTZ command construction is injection-safe:** `cameraId` gated through `/^[A-Za-z0-9-]+$/`, path built from a fixed `TPtzPath` union via `new URL(path, dir)`; preset token in the JSON body, not the URL. Velocities clamped to [-1,1].
- **Teardown is thorough:** `dispose()` bumps a generation token, clears timers/watchdog, removes listeners, `hls.destroy()`, stops receiver tracks + closes the PC, DELETEs the WHEP resource, resets `<video>`; `attach()` disposes at its head so a source change can't stack pipelines.

**Findings:**

`[high → medium · marine-safety]` **No liveness watchdog after the first frame — a frozen stream shows stale footage as live** — the only health monitor is a one-shot 8 s first-frame watchdog; once `onPlaying` sets `painted=true` it's cleared and nothing further tracks frame progress. No `stalled`/`waiting`/`timeupdate` on `<video>`, no stall handling on the MJPEG ``; WebRTC only reconnects on ICE `failed`. A camera that freezes while the PC stays alive holds the last frame with no spinner/error/reconnect — a helmsman reads a stale engine-room or dock-approach feed as current. Verified real. Fix this first for helm use (bind `timeupdate`/`stalled`, or an age indicator).

`[medium · marine-safety]` **PTZ keepalive defeats the gateway auto-stop and isn't cancelled on visibility change → camera runaway** — `beginPtz` starts a 1200 ms `setInterval` that re-sends the move "to defeat the gateway's auto-stop safety timeout." `stopPtzKeepAlive` fires on pointerup/cancel/blur but **not** on `visibilitychange`. Touch-hold pan, then the dashboard auto-switches page / the OS backgrounds the tab → no pointer event, the pipeline is disposed but the interval keeps POSTing move, defeating the failsafe; the camera slews to its mechanical limit. Verified real.

`[medium · reliability]` **WHEP negotiation and plugin/preset probes have no request timeout** — `whep.util.ts` fetches with no `AbortController`. If the gateway accepts the socket but never responds (wedged go2rtc), the fetch never settles → the "Couldn't connect" catch never runs → permanent "Connecting…" spinner, no error, no Retry. Verified real.

`[medium · reliability]` **WHEP `resourceUrl` assigned before the stale-generation check → leaks the go2rtc consumer** — `widget-video.component.ts:595-596`: `this.whepResource = resourceUrl` runs unconditionally, then the `generation !== gen` branch bails *without restoring* the prior value. With two negotiations in flight (preset change / reconnect) answered out of order, the superseded session's URL overwrites the active one; on teardown the active resource is never DELETEd and go2rtc keeps the consumer. Verified real.

Plus lows: `snapshotMessageTimer` and the `waitForIceGathering` fallback `setTimeout` aren't cleared on destroy.

**Bottom line:** security and teardown are well done; the marine-safety items (frozen-as-live liveness, PTZ runaway) and the no-timeout hang are what to fix before putting a camera feed in front of a helmsman.

_(Resolution applied earlier: Video widget re-expressed in main's lazy widget registry, +`hls.js`. CI green.)_

mairas:
## Fixes pushed

Addresses the marine-safety + reliability findings:
- **Frozen-as-live (HIGH):** after the first frame, a liveness watch tracks `timeupdate` progress; a stream that stops advancing (frozen without erroring) now forces a bounded reconnect, so the operator sees a reconnecting state instead of stale footage presented as live.
- **PTZ runaway:** the camera move is stopped (and an explicit stop sent) on `visibilitychange` when the page is hidden, and on destroy — not just clearing the keep-alive interval. Backgrounding / a page-switch delivers no pointerup/blur, so the keep-alive would otherwise defeat the gateway auto-stop.
- **WHEP hang:** negotiate/DELETE now carry an `AbortSignal` timeout, so a gateway that accepts the socket but never answers can't wedge the widget on "Connecting…".
- **WHEP leak:** `resourceUrl` is assigned only after the stale-generation check, so an out-of-order superseded negotiation can't overwrite/leak the active session's go2rtc consumer.
- Cleared the ICE-gathering fallback timer on early completion and the snapshot-message timer on destroy.

Tests: WHEP timeout signal, PTZ-stop-on-hide. Full vitest suite 705/705; CI green.

PR #65, opened 2026-06-30
Await server writes before reload/refresh (fix misused-promise bugs) (replicates mxtommy/Kip #1089)

mairas:
Replicates upstream PR https://github.com/mxtommy/Kip/pull/1089 ("Await server writes before reload/refresh (fix misused-promise bugs)") by dillan.

**Method:** commit-by-commit cherry-pick (the upstream PR is linear, no merge commits), preserving each original author.

This is an experimental replica carried in the fork for evaluation. It has not been built or test-run here; treat it as a candidate for review rather than a finished change.

## ⚠️ Conflict resolution applied

SKip has diverged substantially from upstream in the config/storage write path (profile-based configuration UI, a reworked patch queue with pending/failure accounting). The following files required hand-resolution:

- **src/app/core/services/storage.service.ts** — Kept SKip's existing patch queue (`_pendingPatches$` / `_patchFailures` / `enqueuePatch` / `awaitQueueDrain`) and layered the PR's per-patch deferred `resolve`/`reject` on top. Critically, the PR's `patch()` `catchError` swallowed the error with `of(null)`; that would have hidden failures from SKip's outer queue `catchError`, leaving `_patchFailures` stuck at 0 and making `awaitQueueDrain` falsely report success. Changed it to fire `arg.reject` then **re-throw**, so SKip's outer layer still counts the failure and keeps the queue alive. `removeItem` now returns `Promise<void>` but routes through `enqueuePatch` (not raw `patchQueue$.next`) to preserve the pending-patch accounting.
- **src/app/core/components/options/configuration/config.component.ts** — Kept SKip's version entirely. SKip replaced the upstream `saveConfig`/`deleteConfig`/restore flow with a profile-based service that already `await`s every server operation before reload/refresh, so the PR's intent is already satisfied here by different means; the methods the PR patched no longer exist in SKip.
- **src/app/core/components/options/configuration/config.component.spec.ts** — Kept SKip's version (the PR's new tests target the removed `saveConfig`/`deleteConfig` methods).
- **src/app/core/services/settings.service.ts** — Applied the PR's await-then-reload intent to `loadDemoConfig`, keeping SKip's `useServerStorage` naming and dropping the PR's now-redundant inner readiness check (SKip already early-returns when storage is not ready).
- **src/app/core/services/settings.service.spec.ts** — Kept SKip's existing `loadDemoConfig` tests, fixed SKip's "writes when ready" test whose `setConfig` mock returned `undefined` (now broken because `loadDemoConfig` chains `.then`), and ported the PR's await-sequencing tests in SKip's `createService`/real-`StorageService`-spy style (adapted `useSharedConfig` → driven via `createService`).
- **src/app/core/services/storage.service.spec.ts** — Kept SKip's `HttpTestingController` harness and re-added the PR's two `removeItem` completion tests adapted to that harness (the PR used a different `HttpClient` mock setup incompatible with SKip's).
- **src/app/core/components/options/display/display.component.spec.ts** — Resolved a `SettingsServiceMock` overlap so `getDisablePathValidation`/`setDisablePathValidation`/`getBrowserTabTitle`/`setBrowserTabTitle` each appear exactly once.

Needs a build + review before merge.

The remaining commits (history-series reconcile retry, configuration-upgrade start-fresh, display.component result reporting) applied cleanly against SKip.

### DIFF
--- a/src/app/core/components/options/display/display.component.spec.ts
+++ b/src/app/core/components/options/display/display.component.spec.ts
@@ -41,8 +41,8 @@ class SettingsServiceMock {
     public getSplitShellSide() { return 'left' as const; }
     public getSplitShellSwipeDisabled() { return false; }
     public getWidgetHistoryDisabled() { return false; }
+    public getBrowserTabTitle() { return 'KIP'; }
     public getDisablePathValidation() { return false; }
-    public setDisablePathValidation(): void { }
     public setAutoNightMode(): void { }
     public setRedNightMode(): void { }
     public setNightModeBrightness(): void { }
@@ -53,6 +53,8 @@ class SettingsServiceMock {
     public setSplitShellSide(): void { }
     public setSplitShellSwipeDisabled(): void { }
     public setWidgetHistoryDisabled(): void { }
+    public setBrowserTabTitle(): void { }
+    public setDisablePathValidation(): void { }
 }
 
 class PluginConfigClientServiceMock {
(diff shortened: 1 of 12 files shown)


mairas:
## Automated review — issue #69 tier-3 adoption pass (⚠️ conflict-resolved)

**Verdict:** Safe to adopt. The conflict-resolved replica correctly achieves upstream #1089's "await writes before reload/refresh" intent against SKip's storage/profile rewrite. Faithfulness: `adapted-correctly`. The await ordering and the sequential JSON-Patch write queue were checked; every formerly-unawaited write is now correctly sequenced.

**Pre-review CI fix** (independently verified correct): a missing `TestBed` import in `storage.service.spec.ts` had failed compilation and *hidden* the storage tests — once added, those tests run and pass.

**One confirmed minor finding — fixed in this PR (commit `a60cc938`):**

`removeItem` now returns a `Promise<void>` that rejects on a failed server write, but `renameProfile`/`deleteProfile` discarded the return value. On a real failure path (offline, 401, SK server restart) that produced an *unhandled promise rejection*. The conflict resolution updated `removeItem`'s contract but not its two profile-service callers.

Fix: swallow the per-write rejection at both call sites (`void this.storage.removeItem(...).catch(() => undefined)`) — the failure is still surfaced via the existing `awaitQueueDrain()` check, so `deleteProfile` still throws and `renameProfile` still warns. Also updated the `profile.service.spec` storage mock so `removeItem` returns `Promise<void>` (it had drifted to the old `void` contract). All 13 CI checks green.

mairas:

Following a review-depth concern, this tier-3 PR got a second, **content-only** review that treats the upstream PR as *unreviewed*, bans "matches upstream / byte-identical" as a correctness argument, and requires every claim to be backed by concrete input→code→output traces (9 traces here), with emphasis on boundary/transient states.

**Independent verdict:** correct — Every applicationData write in the save/restore/delete/rename/switch/demo/startFresh paths is either directly awaited (setConfig/getConfig via lastValueFrom) or correctly sequenced through the JSON-Patch queue (removeItem) before any reload/refresh reads it back. The deferred resolve/reject wiring on removeItem settles on HTTP completion, and the queue's _pendingPatches$/_patchFailures accounting is preserved (rethrow after arg.reject), so awaitQueueDrain reports true only when the server actually confirmed the write with no failure. No config is silently lost or read stale in any real-deployment scenario I could construct.

Confidence: `high`.

- [nit] removeItem promise dangles unresolved if StorageService is torn down mid-request

Issue #18, opened 2026-06-30
[SEC-01] Session JWT and never-expiring device tokens stored in localStorage (XSS-exfiltratable)


mairas:
**Severity:** medium | **Confidence:** confirmed | **Area:** security | **Source:** codebase review 2026-06-30, finding `SEC-01`

## Problem

In token mode, KIP's authentication service persists the session JWT to `localStorage` under the key `authorization_token` and re-hydrates it on every page load. Device access tokens — which may be created with an expiry of `NEVER` — are likewise stored in `localStorage` indefinitely. This is a deliberate design tradeoff, documented in a code comment at lines 118–119 of `authentication.service.ts` ("Option A — the session JWT is the persisted credential now that the plaintext password is not stored"). `localStorage` is readable by any script executing on the same origin, meaning any same-origin XSS vulnerability, present or future, can exfiltrate these credentials. While no concrete same-origin XSS vector was identified in the current codebase (Angular's template escaping and controlled `DomSanitizer` bypasses reduce the immediate risk), the storage exposure and credential lifetime are confirmed security concerns independent of an active exploit.

## Impact

The primary concern is credential lifetime: a never-expiring device token stored in `localStorage` means that a single same-origin XSS event grants an attacker permanent, revocation-resistant Signal K server access. KIP is commonly deployed as a marine dashboard, often with iframe or embed widgets that enlarge the effective attack surface on the origin. Cookie-based auth (already implemented and `httpOnly`-scoped) avoids this exposure entirely, but token mode — used for cross-origin deployments — leaves long-lived credentials client-side. HaLOS's same-origin SSO deployment already favors the safer cookie mode, but operators running KIP in token mode on their own are exposed.

## Affected code

- `src/app/core/services/authentication.service.ts:98` — boot re-load of persisted token from `localStorage`
- `src/app/core/services/authentication.service.ts:396` — never-expiring user session token persisted to `localStorage`
- `src/app/core/services/authentication.service.ts:405` — expiring user session token also persisted to `localStorage`
- `src/app/core/services/authentication.service.ts:491` — device access token with `NEVER` expiry persisted to `localStorage`
- `src/app/core/services/authentication.service.ts:501` — expiring device token persisted to `localStorage`

## Suggested direction

[...]

Issue #17, opened 2026-06-30
[RED-05] settings.service rewrite: signals-first API over a single persistence path


mairas:
**Plan v2.1, re-derived 2026-07-03 against post-#46/#66/#82/#105/#146 main.** Supersedes the 2026-07-01 draft (kept in comments, with its document-review). Open questions settled with @mairas 2026-07-03; **document-review of v2 passed** (4 personas, approve-with-fixes, no must-fixes) and its fixes are folded in below. **All four PRs merged 2026-07-03** (#160, #166, #168, #172); follow-ups: #79 (bridge retirement, now unblocked), #170 (hydration ordering), #159 (stub inertness).

## Background (original RED-05 finding, condensed)

`settings.service.ts` duplicated a per-setter persistence block (rebuild whole config blob → route server vs localStorage) across every setter, wrapped in `getX`/`getXAsO`/`setX` triad ceremony. The original counts (858 lines, 65 members) are stale — intervening merges already shrank the file — but the structural defect stands: **duplication of the write path, not LOC, is the problem.** This is explicitly *not* a "generic configuration store" project.

## Ground truth (2026-07-03)

- File is 761 lines / 55 public members: 19 sync getters, 9 `*AsO()` observable getters, 15 set/save methods, 10 BehaviorSubjects + 1 signal (`configUpgrade`). (PR #157, merged mid-planning, has since removed `widgetHistoryDisabled` entirely — its counts and the `getWidgetHistoryDisabledAsO` deletion item are already satisfied; 8 toSignal sites and the hydration gate note shrink accordingly. Re-verify counts at each PR start.)
- `useServerStorage` is a `readonly = true` constant (#82): **13 dead `else` branches and 3 dead localStorage save helpers** (app/theme/dashboards). `saveConnectionConfigToLocalStorage` is **live by design** (device scope). The old plan's routing matrix is unrepresentable; ARCH-13 survives only as deletion plus version-const scatter.
[...]

PR #82, opened 2026-07-01
refactor(auth): session-only cookie authentication

### LINKED ISSUE #22: [SK-03] Autopilot/course v2 REST calls hard-code same-origin paths, breaking cross-origin token mode
**Severity:** medium | **Confidence:** confirmed | **Area:** signalk | **Source:** codebase review 2026-06-30, finding `SK-03`

## Problem

The autopilot widget hard-codes its Signal K v2 REST endpoints as app-origin-relative paths (e.g. `/signalk/v2/api/vessels/self/autopilots/...` and `/signalk/v2/api/vessels/self/navigation/course`). Because these paths start with `/`, the browser resolves them against `window.location.origin` — the host serving Kip — rather than the actual Signal K server host. This is only correct when Kip is served same-origin or behind a reverse proxy. The codebase explicitly supports cross-origin token authentication (when the Signal K server origin differs from the Kip origin), but in that mode all autopilot and course v2 POST/PUT/DELETE commands are sent to the Kip host and silently fail. No URL rewriting is performed by the authentication interceptor; it only appends the JWT header.

## Impact

Any deployment where Kip is hosted on a different origin from the Signal K server and authentication mode is `token` (cross-origin) will have a non-functional autopilot widget — operators cannot engage, disengage, or adjust the autopilot or course through Kip. This is a correctness regression relative to the rest of the app: the History API client already derives its URL correctly from the discovered `httpServiceUrl`, so the pattern to fix this is established. The bug is localized and fixable upstream.

## Affected code

- `src/app/widgets/widget-autopilot/widget-autopilot.component.ts:62-64` — defines the hard-coded `V2_AUTOPILOTS` and `V2_COURSE` constants
- `src/app/widgets/widget-autopilot/widget-autopilot.component.ts:955-969` — `executeRestRequest` passes `cmd.path` directly to `http.post`/`put`/`delete` without prepending a base URL
- `src/app/core/services/history-api-client.service.ts:62-63` — reference implementation that correctly derives its URL from the discovered `httpServiceUrl`

## Suggested direction

Build the autopilot and course v2 base URLs from the discovered `httpServiceUrlV2` (exposed by `SignalKConnectionService`) rather than from hard-coded absolute paths — the same approach `HistoryApiClientService` uses with `httpServiceUrl.replace('/v1/', '/v2/')`. This ensures requests target the Signal K server's origin in all deployment modes, including cross-origin token deployments.

## Verification

Verified against the codebase: `widget-autopilot.component.ts` confirmed to define `V2_AUTOPILOTS` and `V2_COURSE` as hard-coded leading-`/` paths, with `executeRestRequest` passing them unmodified to the HTTP client. Token-mode cross-origin support confirmed in `authentication.service.ts`. Discovered `httpServiceUrlV2` confirmed in `signalk-connection.service.ts`. Confidence: confirmed.

### LINKED ISSUE #46: Support only Signal K session authentication; stop storing and manipulating usernames/passwords
## Position

SKip should authenticate **only** through Signal K server session authentication. The app should not own a credential entry UI, and should not store, hold (even transiently), or transmit raw usernames and passwords. The server owns the login — SKip obtains a session from it and uses only that session/token thereafter.

The fork already started down this path: the plaintext password is no longer persisted (`SettingsService` keeps `loginPassword` in memory only and purges any legacy persisted value), and same-origin **cookie mode** delegates entirely to the server's session/SSO login. What remains is the cross-origin **token mode**, which still runs a KIP-managed username/password flow. This issue is to finish the direction and make session authentication the only supported model.

## Why

- **Smaller credential surface.** Every place SKip collects, holds, or POSTs a raw password is another spot an XSS payload or a logging mistake can leak it. Delegating to the server removes SKip from the credential path.
- **Don't duplicate the server's job.** The Signal K server already owns the login form, password policy, rate limiting, and — critically — SSO/OIDC. A KIP-managed `username`/`password` form can't carry an OIDC redirect flow; only server-side session login can.
- **Matches the HaLOS deployment.** SKip is served by the SK server same-origin at `/@halos-org/skip/`, behind the reverse proxy and Authelia. That is cookie/session mode by definition — the cross-origin KIP-managed credential form is never exercised in our topology.

## Current state

Already aligned:
- `SettingsService` — `loginPassword` is transient in-memory only; legacy persisted passwords are purged on load (`settings.service.ts`).
- Cookie mode — session comes from `GET /skServer/loginStatus`; no SKip-held credential (`authentication.service.ts`).

Still KIP-managed credential handling (token mode):
- `ModalUserCredentialComponent` — a SKip-owned "Sign in to Signal K" modal that collects `user` + `password`.
- `signalk.component.ts` (Connectivity tab) — keeps `loginName`/`loginPassword` in `connectionConfig`, opens the credential modal, then logs in with `auth.login({ usr, pwd })`.
- `app-initNetwork.service.ts` — startup token-mode auto-login from `config.loginName`/`config.loginPassword`.
- `AuthenticationService.login({ usr, pwd })` — POSTs `{ username, password }` to the server's `auth/login` endpoint.
- `widget-login.component.ts` — a login widget that opens the same credential modal.
- `SignalkRequestsService.requestUserLogin(userId, userPassword)` — sends `username`/`password` over the SK WS delta channel. **No live callers** — dead code.
- `IConnectionConfig` / `v10-config-interface.ts` — still carry `loginName`/`loginPassword` fields.

## Proposed direction

- Remove SKip's own credential entry, storage, and transmission paths (the modal, the login widget, the `connectionConfig` login fields, the dead `requestUserLogin`).
- Obtain a session exclusively from the Signal K server: redirect to the server's session/SSO login the way cookie mode already does, and consume the resulting session/token.
- Keep device access tokens as the separate unattended/headless-display mechanism — that is not a stored user password and is out of scope here.

## Open question (the crux)

Cross-origin token mode is the case that exists *because* there is no server session to ride. If SKip stops collecting credentials itself, how does a genuinely cross-origin deployment get a session?

- **Option A — require same-origin/reverse-proxy (cookie mode) as the only supported topology.** This is already how HaLOS ships SKip; cross-origin KIP-managed login goes away entirely.
- **Option B — OIDC/redirect-based session for cross-origin.** Heavier, but keeps a non-proxied deployment working without SKip ever seeing a password.

This needs a decision before implementation. For HaLOS, Option A looks sufficient.

## Related

- #18 (SEC-01) — token storage in localStorage. Related but distinct: that's about *where the session token lives*; this is about *not handling raw credentials at all*.
- #12 (AUTH-01), #32 (AUTH-02) — existing auth-flow bugs in the same area.

mairas:
## Why

SKip is served **same-origin** by the Signal K server (behind the reverse proxy / SSO), so the httpOnly session cookie is the only credential it needs. Every place SKip collected, held, or transmitted a raw username/password or a JWT/device token was extra credential surface an XSS payload or logging mistake could leak — and it duplicated the server's own login/SSO. This finishes the direction agreed in #46 (Option A, **strictly same-origin**): server session authentication is the only supported model.

## What changed

- **`AuthenticationService`** — removed token load/renewal, `login`/`logout`/`setSession`, device tokens, and the cookie-vs-token auth mode. Session state now derives solely from `GET /skServer/loginStatus`.
- **HTTP interceptor** — same-origin credentials only; never attaches a JWT header.
- **Delta service** — no `&token=` on the WS upgrade; (re)connect driven off the session signal.
- **`signalk-requests`** — removed `requestUserLogin` and `requestDeviceAccessToken` (+ the device-token delta handling).
- **Connectivity tab & login route** — no in-app credential modal; sign-in redirects to the server's SSO. `ModalUserCredentialComponent` deleted.
- **Freeboard-SK iframe** — loads same-origin with no token.
- **Config** — dropped `loginName` / `loginPassword` / `useDeviceToken` from the connection config; `SettingsService` purges these legacy keys from stored `connectionConfig` on load and always persists to server applicationData.

Removing the localStorage token surface also resolves the SEC-01 concern for those keys (#18 narrows to any residual device-token history, now gone).

## Topology decision → #22 moot

Same-origin is the only supported topology, so `window.origin` always equals the SK origin and the autopilot's hard-coded same-origin v2 paths are correct everywhere. #22 closes as moot.

## Verification

- `npm run lint` — clean
- `npm run build:dev` — clean
- `npm test` — **759 tests / 134 files passing**. Token-mode and device-token tests removed with their code; cookie session-state coverage retained; interceptor, delta, settings, and app-init specs updated to the single mode.

Closes #46
Closes #22


mairas:
## Code review — session-only cookie authentication (#46)

**Scope:** 26 files, +164 / −1460 vs `main`.
**Intent:** Remove SKip's token/device-token/credential surface; authenticate only through the same-origin SSO session cookie (Option A, strictly same-origin).
**Reviewers (8):** correctness, testing, maintainability, project-standards (always-on) + security, reliability, adversarial, api-contract (selected: diff touches auth, permissions, exported type signatures, bootstrap error handling, and is >50 LOC).

Findings were cross-checked between reviewers; the two highest-agreement items were independently flagged by 2–3 personas each. All actionable findings are addressed in commit `1dbbf1b`.

### P2 — fixed

| # | File | Issue | Reviewers |
|---|------|-------|-----------|
| 1 | `settings.service.ts` (load) | The legacy top-level `authorization_token` localStorage key was no longer read or deleted by any code path — a pre-upgrade token (often a **never-expiring device token**) would persist at rest as a still-server-valid bearer credential, defeating part of this issue's own goal. **Fix:** purge it unconditionally on load + test. | security, adversarial |
| 2 | `settings.service.ts:465` | `setBrowserTabTitle()` routed on the now-frozen `useSharedConfig` instead of `useServerStorage` like every other setter, so on a default install the tab title wrote to localStorage and was silently lost on the next (server-loaded) reload. **Fix:** route on `useServerStorage` + routing regression test. | correctness, maintainability, testing |
| 3 | `CLAUDE.md:38` | Architecture note still described "two modes" auth (cookie + JWT/device token). **Fix:** rewritten to the single same-origin session model. | project-standards |

### P3 — fixed

| # | File | Issue | Reviewers |
|---|------|-------|-----------|
| 4 | `widget-tutorial.component.html` | Onboarding tutorial instructed users to toggle "Login to server", enter credentials and "Sign in" — controls deleted in this PR. **Fix:** rewritten to describe the SSO sign-in redirect. | adversarial |
| 5 | `app-initNetwork.service.ts`, `config.component.ts` | Doc/inline comments still framed logic around removed token/device "modes". **Fix:** updated to the single-session model. | maintainability |
| 6 | `signalk-delta.service.ts` | WS-drop comment claimed the loginStatus re-check happens "rather than retrying blindly", but WS reconnect is driven independently by the (untouched) connection state machine. **Fix:** comment corrected. | reliability |

### Deferred to #17 (settings.service generic-store rewrite)

These are pre-existing consequences of `useServerStorage` collapsing to a constant `true`; both the maintainability reviewer and the wave plan recommend sweeping them in #17 rather than a partial cleanup here (which #17 would redo):

- ~17 now-unreachable `else { …localStorage… }` branches and three dead `save*ToLocalStorage()` helpers.
- `useSharedConfig` is no longer UI-editable but still gates read-side paths: `config.component` export/backup reads localStorage when `useSharedConfig` is false while writes go to the server, and the `missing-shared-config` 404 recovery UI is unreachable for fresh installs.

### Residual risks (accepted — consistent with Option A)

- **Cross-origin upgrade:** a pre-existing install whose `signalKUrl` is cross-origin has no in-app recovery (the credential form is gone; `loginStatus` probes the serving origin). This is the explicit intent of Option A — HALOS ships SKip same-origin — but there is no migration for the unsupported topology.
- **WS retry after session expiry:** the untouched `ConnectionStateMachine` retries the WS every ~3s indefinitely; after a cookie expiry this is a slow re-check loop, not a storm, and is arguably desirable kiosk auto-recovery. Pre-existing, out of this PR's scope.
- Config schema change is unmarked for downgrade (acceptable — the fork has no supported downgrade path).

All P2/P3 findings addressed; deferred items tracked in #17; residuals are accepted Option-A limitations. Full suite green (763 tests / 134 files), lint + prod build clean, all 13 CI checks passing on the pre-fix commit.

<sub>Review orchestrated via mairas-workflow:review — 8 independent persona sub-agents, findings merged and deduplicated.</sub>

PR #88, opened 2026-07-02
test(signalk-requests): cover requestId correlation and status branches

### LINKED ISSUE #8: [TST-03] SignalkRequestsService has smoke-only tests; requestId correlation and response routing are uncovered
**Severity:** high | **Confidence:** confirmed | **Area:** testing | **Source:** codebase review 2026-06-30, finding `TST-03`

## Problem

`SignalkRequestsService` is a 251-line service that implements the Signal K REQUESTS API — covering PUT requests, login requests, device access-token requests, and asynchronous response routing keyed on `requestId`. Despite this substantial scope, its spec file contains only a single smoke test that verifies the service can be instantiated via Angular's dependency injection. The core logic — `requestId` generation and correlation in `putRequest()`, the `updateRequest()` response-routing path that matches incoming deltas to pending requests by `requestId`, the six status-code branches (200/202/400/401/403/405), error and flush handling, and the `widgetUUID`-based subscriber dispatch — is entirely uncovered by automated tests.

## Impact

PUT requests via the Signal K REQUESTS API are the mechanism by which control widgets (e.g., autopilot controls) send commands to the vessel's data bus. A regression in `requestId` correlation or response routing could silently drop confirmations, misroute error responses to the wrong widget, or leave pending requests in a leaked state. Because this is a correctness- and safety-relevant flow, the lack of unit test coverage means such regressions would only surface at runtime on hardware. This is an upstream gap that applies equally to any fork shipping control or PUT functionality.

## Affected code

- `src/app/core/services/signalk-requests.service.spec.ts:1-13`
- `src/app/core/services/signalk-requests.service.ts`

## Suggested direction

Add unit tests that exercise `putRequest()` requestId generation, the `updateRequest()` correlation logic (matching a synthetic delta response back to the originating request by `requestId`), each status-code branch (200, 202, 400, 401, 403, 405), the unknown-requestId warning path, and the error/flush handling in subscriber cleanup. Tests should drive the service through the delta service's request stream as the stimulus, keeping them free of real network calls.

Note: the service does not implement client-side timeout logic (504 is only a server-reported status code string); tests need not cover a Kip-side timeout mechanism.

## Verification

Verified against the codebase: the spec file is confirmed to be 13 lines with a single `'should be created'` instantiation test and no further cases (confidence: confirmed).

mairas:
## Why

`SignalkRequestsService` had a single DI smoke test — requestId correlation and response routing (the safety-relevant logic) were uncovered (TST-03).

## What

Expanded the spec into 12 behavioral tests, driving the service through a mocked delta stream and toast:

- `putRequest()` generates a requestId, strips a leading `self.`, and publishes a `vessels.self` PUT delta; rejects undefined value / empty path without publishing.
- `updateRequest()` correlates a response to its originating request by requestId and routes to the correct widget, leaving other pending requests untouched.
- The 200/401/403/405 pass-through branches dispatch the correct status description with no error toast; 400 surfaces the server message as an error toast; the 202 interim ack stays pending until a final response; an unknown requestId fires a warn toast without dispatching.

No production code changed.

**Known coverage gaps (P3, deliberate):** the `updateRequest()` error/flush branch is effectively-dead defensive code (reachable only under a deprecated RxJS synchronous-error flag), and the 500/502/504 → "unknown status" else branch is documented but not exercised. Both noted for a follow-up rather than faked here.

Closes #8

### DIFF
--- a/src/app/core/services/signalk-requests.service.spec.ts
+++ b/src/app/core/services/signalk-requests.service.spec.ts
@@ -1,12 +1,144 @@
 import { TestBed } from '@angular/core/testing';
-import { beforeEach, describe, expect, it } from 'vitest';
-import { SignalkRequestsService } from './signalk-requests.service';
+import { Subject } from 'rxjs';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { ISignalKDeltaMessage } from '../interfaces/signalk-interfaces';
+import { SignalkRequestsService, skRequest } from './signalk-requests.service';
+import { SignalKDeltaService } from './signalk-delta.service';
+import { ToastService } from './toast.service';
 
 describe('SignalkRequestsService', () => {
-  beforeEach(() => TestBed.configureTestingModule({}));
+  let service: SignalkRequestsService;
+  let requestUpdates$: Subject<ISignalKDeltaMessage>;
+  let publishDelta: ReturnType<typeof vi.fn>;
+  let toastShow: ReturnType<typeof vi.fn>;
+
+  beforeEach(() => {
+    requestUpdates$ = new Subject<ISignalKDeltaMessage>();
+    publishDelta = vi.fn();
+    toastShow = vi.fn();
+
+    TestBed.configureTestingModule({
+      providers: [
+        SignalkRequestsService,
+        {
+          provide: SignalKDeltaService,
+          useValue: {
+            subscribeRequestUpdates: () => requestUpdates$.asObservable(),
+            publishDelta,
+          },
+        },
+        { provide: ToastService, useValue: { show: toastShow } },
+      ],
+    });
+
+    service = TestBed.inject(SignalkRequestsService);
+  });
 
   it('should be created', () => {
-    const service: SignalkRequestsService = TestBed.inject(SignalkRequestsService);
     expect(service).toBeTruthy();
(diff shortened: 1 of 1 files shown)


mairas:
## Independent review — verdict: minor (mergeable)

Real behavioral tests, not smoke stubs — they drive the service through a mocked delta stream and pin requestId generation + `self.` stripping, the null-return guards, requestId correlation in `updateRequest()` (matching one of two pending requests, leaving the other pending), verbatim status descriptions for 200/401/403/405 with no error toast, the 400 error-toast path, the 202 interim-ack staying pending, and the unknown-requestId warn path. 12 tests pass; assertions are behavioral, not implementation-coupled.

Two deliberate P3 coverage gaps:

- **P3 (documented)** — `updateRequest()`'s error/flush branch (`requestStatus$.error` + array flush) is uncovered. It's effectively-dead defensive code, reachable only under RxJS's deprecated synchronous-error flag, so a regression there wouldn't be caught.
- **P3 (documented)** — 500/502/504 exist in `deltaStatusCodes` but aren't in the allowed set, so they fall to the 'unknown status code' else branch. The `it.each` covers 200/401/403/405 only, so that surprising treatment isn't pinned.

Neither blocks merge.

PR #90, opened 2026-07-02
refactor(history): drop duplicated history interfaces in kip-series client

### LINKED ISSUE #33: [HF-01] Dead `getValues()` method in KipSeriesApiClientService and duplicated history interfaces
**Severity:** low | **Confidence:** confirmed | **Area:** history | **Source:** codebase review 2026-06-30, finding `HF-01`

## Problem

`KipSeriesApiClientService.getValues()` issues a GET request to `${kipPluginServiceUrl}history/values`, but the plugin router registers no such route — only `/series`, `/series/:id`, `/series/reconcile`, and `/displays/*` routes are wired under `/plugins/kip`. Any call to this method would result in a 404. The method has zero callers in the codebase; history-value reads are exclusively performed through `HistoryApiClientService.getValues()`. To compound the confusion, the method's log strings and JSDoc even reference `HistoryApiClientService`, confirming it is a stale copy-paste artifact. In addition, the interfaces `IHistoryValueMetadata`, `IHistoryValuesResponse`, and `IHistoryValuesQueryParams` are declared verbatim in both `kip-series-api-client.service.ts` and `history-api-client.service.ts`, creating two competing exported types of the same name.

## Impact

The dead method poses no runtime risk in the current state because it is never invoked, but it creates maintenance confusion: a future developer may attempt to use it, observe 404s, and waste time diagnosing a routing problem that does not exist. The duplicate interface declarations risk divergence — a type fix applied to one file may silently leave the other stale, causing subtle type-mismatch issues in consumers. The issue is pure cleanup with no effect on SKip-specific marine functionality; resolving it reduces confusion without requiring a fork divergence.

## Affected code

- `src/app/core/services/kip-series-api-client.service.ts:220-261`
- `src/app/core/services/kip-series-api-client.service.ts:30-59`
- `src/app/core/services/history-api-client.service.ts:19-46`
- `kip-plugin/src/index.ts:1349-1513`

## Suggested direction

Delete `getValues()` and the duplicated interface block (`IHistoryValueMetadata`, `IHistoryValuesResponse`, `IHistoryValuesQueryParams`) from `KipSeriesApiClientService`, keeping that service focused on series CRUD and reconcile operations. The shared types should live in a single location — either imported from `history-api-client.service.ts` or extracted to a dedicated shared contract module — so there is exactly one canonical definition.

## Verification

Verified against the codebase (confidence: confirmed) — the route table in `kip-plugin/src/index.ts` contains no `history/values` entry, `grep "\.getValues("` over `src/` shows callers only on `HistoryApiClientService`, and the duplicate interface text is present in both service files.

mairas:
## Why

`kip-series-api-client.service.ts` re-declared three history interfaces (`IHistoryValueMetadata`, `IHistoryValuesResponse`, `IHistoryValuesQueryParams`) that already have canonical definitions in `history-api-client.service.ts` (HF-01).

## What

Removed the duplicated declarations from the kip-series client. Every consumer already imports these types from `history-api-client`, and the `IKipSeriesDefinition` re-exports are untouched, so this is contract-safe. The dead `getValues()` the issue also named was already gone.

Closes #33

### DIFF
--- a/src/app/core/services/kip-series-api-client.service.ts
+++ b/src/app/core/services/kip-series-api-client.service.ts
@@ -1,11 +1,10 @@
 import { DestroyRef, inject, Injectable } from '@angular/core';
-import { HttpClient, HttpParams } from '@angular/common/http';
+import { HttpClient } from '@angular/common/http';
 import { firstValueFrom } from 'rxjs';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { SignalKConnectionService } from './signalk-connection.service';
 import { PluginConfigClientService } from './plugin-config-client.service';
 import type { IKipSeriesDefinition } from '../contracts/kip-series-contract';
-import type { AggregateMethod, TimeRangeQueryParams } from '@signalk/server-api/history';
 
 export type {
   IKipConcreteSeriesDefinition,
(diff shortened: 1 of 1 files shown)


mairas:

Pure type-level dead-code removal, zero behavior change. No spec change, and that's correct: `kip-series-api-client.service.spec.ts` never referenced the removed interfaces (it only exercises `reconcileSeries`/`getSeriesDefinitions`), and the surviving canonical definitions in `history-api-client.service.ts` are still exercised by the existing history specs. An assert-absence test would be an anti-pattern. No findings.

PR #91, opened 2026-07-02
fix(history): parameterize SQLite history storage statements

### LINKED ISSUE #34: [HF-03] SQLite history storage uses string interpolation instead of parameterized statements
**Severity:** low | **Confidence:** confirmed | **Area:** history | **Source:** codebase review 2026-06-30, finding `HF-03`

## Problem

Every SQL query and insert in the SQLite history storage layer is assembled by concatenating values directly into SQL text rather than using bound parameters. String values are escaped via a hand-rolled `escape()` function that doubles single quotes, and numeric values are coerced through `Math.trunc`/`Number`. While single-quote doubling is technically sufficient for SQLite string literals, this approach is a fragile maintenance hazard: a single missed call site or an edge case in the escaping logic could introduce a SQL injection vulnerability. Notably, `db.prepare(sql)` is already used but `.all()` is called with no bound parameters, meaning the API for parameterized queries is in place but not utilized. Path, context, and source values originate from network-delivered Signal K deltas, and requested paths come from History API query strings, making the inputs externally influenced.

## Impact

The immediate risk is low because the current escaping logic covers the known cases, but hand-rolled escaping creates ongoing maintenance risk: future changes to the storage layer may introduce new interpolation sites without escaping, and the pattern sets a precedent that is easy to follow incorrectly. Additionally, building SQL strings dynamically defeats SQLite's statement caching, which is a performance concern for the hot insert and select paths that run on every data flush and history query. For a production deployment relying on this plugin for Signal K history data, hardening the storage layer is a worthwhile investment.

## Affected code

- `kip-plugin/src/sqlite-history-storage.service.ts:937-967`
- `kip-plugin/src/sqlite-history-storage.service.ts:986-1011`
- `kip-plugin/src/sqlite-history-storage.service.ts:1222-1238`
- `kip-plugin/src/sqlite-history-storage.service.ts:409-447`

## Suggested direction

Switch all inserts and queries to parameterized prepared statements by passing bound values to `.all()` and `.run()` instead of interpolating them into the SQL string. Remove the `escape()`, `nullableString()`, and `nullableNumber()` helpers once they have no remaining call sites. Reuse cached prepared statement objects for the per-flush insert and per-query select paths to benefit from SQLite's statement caching.

## Verification

Verified against the codebase: all cited locations confirm SQL is built by string interpolation rather than bound parameters (confidence: confirmed).

mairas:
## Why

The SQLite history storage layer built its insert/upsert/select SQL by string interpolation (HF-03) — fragile and an injection footgun even for internal data.

## What

Converted the insert/upsert and select paths to parameterized prepared statements, binding values through `.run()`/`.all()` instead of interpolating. Removed the now-unused `escape()`/`nullableString()`/`nullableNumber()` helpers. Public method signatures are unchanged, so the sole importer (`kip-plugin/src/index.ts`) needs no edit.

Verified by CI (`build:plugin` + `test:plugin`, Node ≥22.5).

Closes #34

### DIFF
--- a/kip-plugin/src/sqlite-history-storage.service.ts
+++ b/kip-plugin/src/sqlite-history-storage.service.ts
@@ -66,8 +66,11 @@ interface TLogger {
   error: (msg: string) => void;
 }
 
+type TSqlBindValue = string | number | null;
+
 interface ISqliteStatement {
-  all: () => unknown[];
+  all: (...params: TSqlBindValue[]) => unknown[];
+  run: (...params: TSqlBindValue[]) => unknown;
 }
 
 interface ISqliteDatabase {
@@ -411,8 +414,8 @@ export class SqliteHistoryStorageService {
       return;
     }
 
-    await this.runSql(`DELETE FROM history_series WHERE series_id = ${this.escape(series.seriesId)}`);
-    await this.runSql(`
+    await this.runSqlWithParams('DELETE FROM history_series WHERE series_id = ?', [series.seriesId]);
+    await this.runSqlWithParams(`
       INSERT INTO history_series (
         series_id,
         dataset_uuid,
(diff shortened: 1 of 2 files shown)


mairas:

The added `index.test.cjs` test is a genuine injection-safety regression guard, not a smoke stub: it enqueues a sample whose path/seriesId/source carry SQL metacharacters and a `'); DROP TABLE history_samples;--` payload, flushes, then asserts the value round-trips verbatim (4.25), the tricky path is still listed, and the table survives — exercising the parameterized insert and select paths end to end. Runs under `testRequiresNodeSqlite` (CI-only).

- **Note** — the assertions pin 'injection-safe + verbatim round-trip' rather than 'parameterized specifically' (a correctly-escaped string-interpolation implementation would also pass). Acceptable behavior-testing, not a defect.

PR #92, opened 2026-07-02
refactor(signalk-options): make status fields signals and enable OnPush

### LINKED ISSUE #29: [ANG-01] Zoneless app mutates plain fields in subscribe callbacks, skipping change detection
**Severity:** low | **Confidence:** confirmed | **Area:** angular | **Source:** codebase review 2026-06-30, finding `ANG-01`

## Problem

The app runs in zoneless mode (`provideZonelessChangeDetection()` in `src/main.ts`), which means only signal writes, `markForCheck()` calls, async-pipe unwrapping, or event bindings schedule change detection — bare RxJS `subscribe` callbacks assigning plain class fields do not. `signalk.component.ts` is not `OnPush` and updates two plain fields (`this.endpointServiceStatus` and `this.streamStatus`) directly inside `subscribe` callbacks. The component template binds these fields with `{{ endpointServiceStatus.serverDescription }}` and `{{ streamStatus.message }}` without any reactive wrapper. The view is correct on first render only because the underlying observables are `BehaviorSubject`s that replay synchronously before the first CD pass; any subsequent emission would not by itself trigger a re-render. A scan of the codebase found roughly 43 of 103 `@Component` files lack `OnPush`, and several widget-config dialogs follow the same plain-field-mutation-in-subscribe pattern, making this a systemic issue rather than an isolated one.

## Impact

In practice, the stale-view window is narrow for `signalk.component` because other reactive signals in the same template (such as `app.appVersion()`) incidentally trigger CD and refresh the plain-field bindings. However, the pattern is correct by accident and will surface as `NG0100` errors in dev mode (the app enables exhaustive `checkNoChanges` with a 500 ms interval). Widget-config dialogs that follow the same pattern and have fewer incidental signal changes are more likely to exhibit genuinely stale UI. The issue is flagged as neutral for fork relevance — it is a maintenance and correctness fix aimed at the upstream codebase.

## Affected code

- `src/app/core/components/options/signalk/signalk.component.ts:113-138`
- `src/app/core/components/options/signalk/signalk.component.html:73-74`
- `src/main.ts` (`provideZonelessChangeDetection` and `provideCheckNoChangesConfig`)

## Suggested direction

Convert async-driven template state to Angular signals: use `toSignal()` when constructing the observable chain, or call `.set()` inside the subscribe callback on a `signal()`-backed field. Alternatively, inject `ChangeDetectorRef` and call `markForCheck()` inside each subscribe callback as a lower-effort stopgap. Standardise all affected components on `ChangeDetectionStrategy.OnPush`. `remote-control.component.ts` already uses `this.screens.set(...)` and `this.activeScreenIdx.set(...)` correctly and serves as the reference pattern to follow.

## Verification

Verified against the codebase — all cited file paths and line references were confirmed present in the working tree (confidence: confirmed).

mairas:
## Why

The SignalK options component mutated two plain fields (`endpointServiceStatus`, `streamStatus`) inside RxJS subscribe callbacks. In the zoneless app that skips change detection, so the connectivity status block could render stale (ANG-01).

## What

- Convert `endpointServiceStatus` and `streamStatus` to `toSignal`, and `isConnecting` to a `signal`; add `ChangeDetectionStrategy.OnPush`.
- Both status streams re-emit the **same mutated object** on each update, so `toSignal` uses `equal: () => false` — the default `Object.is` equality would treat a same-reference re-emit as unchanged and drop it under OnPush.
- `isConnecting` had to become a signal too: OnPush otherwise strands its post-`await` reset, leaving the Connect button stuck on "Validating…" after a failed validation.
- Specs emit a same-reference mutation (not a fresh literal) so they stay red without the `equal` override.

The `equal:()=>false` and `isConnecting` handling address findings from the branch's own review pass. Verified locally: component spec green, lint + app build clean.

Closes #29

### DIFF
--- a/src/app/core/components/options/signalk/signalk.component.html
+++ b/src/app/core/components/options/signalk/signalk.component.html
@@ -12,8 +12,8 @@
         </mat-form-field>
       </td>
       <td class="connect-cell">
-        <button mat-flat-button class="connect-btn" type="submit" [disabled]="!connectionForm.form.valid || isConnecting">
-          {{ isConnecting ? 'Validating...' : 'Connect' }}
+        <button mat-flat-button class="connect-btn" type="submit" [disabled]="!connectionForm.form.valid || isConnecting()">
+          {{ isConnecting() ? 'Validating...' : 'Connect' }}
         </button>
       </td>
     </tr>
@@ -55,8 +55,8 @@
     <div class="info-container">
       <pre>KIP: {{ app.appVersion() }}
 {{ app.browserVersion() }}, {{ app.osVersion() }}<br />
-{{ endpointServiceStatus.serverDescription }}
-{{ streamStatus.message }}, {{ isUserSession() ? 'Signed in' : 'Not signed in' }}
+{{ endpointServiceStatus().serverDescription }}
+{{ streamStatus().message }}, {{ isUserSession() ? 'Signed in' : 'Not signed in' }}
 Internet: {{ internetAvailabilityLabel() }}</pre>
     </div>
   </div>
--- a/src/app/core/components/options/signalk/signalk.component.spec.ts
+++ b/src/app/core/components/options/signalk/signalk.component.spec.ts
@@ -1,10 +1,15 @@
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { beforeEach, describe, expect, it } from 'vitest';
 import { SettingsSignalkComponent } from './signalk.component';
+import { SignalKConnectionService } from '../../../services/signalk-connection.service';
+import { SignalKDeltaService } from '../../../services/signalk-delta.service';
 
 describe('SettingsSignalkComponent', () => {
   let component: SettingsSignalkComponent;
   let fixture: ComponentFixture<SettingsSignalkComponent>;
+
+  const statusText = (): string =>
+    (fixture.nativeElement as HTMLElement).querySelector('pre')?.textContent ?? '';
 
   beforeEach(async () => {
     await TestBed.configureTestingModule({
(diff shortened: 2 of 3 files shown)


mairas:
## Independent review — verdict: changes-needed → **resolved in the pushed commit**

The first pass of this branch had three real P2s; all are fixed in the current commit. Recording findings and fixes for the trail:

- **P2 (fixed)** — `streamStatus` `toSignal` never reacted to real delta emissions. `SignalKDeltaService.streamEndpoint$` re-emits the *same mutated object reference* on every update, and `toSignal`'s internal `Object.is` equality drops same-reference re-emits, so under OnPush the stream line could keep showing a stale `Connected` after a `WebSocket closed`. → Fixed with `equal: () => false` on both status `toSignal`s.
- **P2 (fixed)** — the spec emitted a *fresh object literal*, which `Object.is` accepts, so it passed while production stayed stale — false confidence masking the defect above. → Spec rewritten to re-emit a same-reference mutation, which stays red without the `equal` override.
- **P2 (fixed)** — adding OnPush stranded `isConnecting` (a plain field reset after `await validateSignalKUrl` in a microtask; nothing marks the component dirty), so the Connect button could stick on `Validating…` after a failed validation. → `isConnecting` converted to a `signal`.
- **P3 (addressed)** — the specs used `fixture.detectChanges()`, which masks whether the emission auto-schedules CD. → The same-reference test now asserts the rendered value actually changes, which stays stale under the broken default-equality path.

Verified locally after the fix: component spec green (3 tests), `npm run lint` + `npm run build:dev` clean. CI green.

PR #93, opened 2026-07-02
fix(auth): stop unbounded sign-in loop under hash routing

### LINKED ISSUE #32: [AUTH-02] Self-route loop guard is ineffective under hash routing; sign-in can loop unboundedly
**Severity:** low | **Confidence:** confirmed | **Area:** auth | **Source:** codebase review 2026-06-30, finding `AUTH-02`

## Problem

KIP uses hash-based routing (`withHashLocation()`), meaning its login page lives at `/#/login` and `window.location.pathname` is always `/`. The self-route guard in `isSafeReturnTo` checks whether the `returnTo` URL matches `SELF_ROUTE_PATHS = ['/login']` by comparing against `resolved.pathname` — but for a hash-routed URL like `/#/login`, `resolved.pathname` resolves to `/`, never `/login`. The guard therefore never fires and `/#/login` is incorrectly treated as a safe return destination. Separately, `WidgetLoginComponent.ngOnInit` in cookie mode unconditionally calls `ssoRedirect.manualSignIn()` without checking whether the user is already authenticated, and `manualSignIn` itself resets the redirect budget without consulting it. Together these form a loop: landing on `/#/login` (cookie mode) with an active IdP session causes the SSO to bounce back to `returnTo=/#/login`, the widget remounts, and `manualSignIn` fires again — indefinitely.

## Impact

A user who reaches `/#/login` via a bookmark or stale link while already authenticated in cookie mode will be caught in an unbounded SSO redirect loop with no escape. The trigger is narrow — KIP itself never routes to `/login` in cookie mode — but the guard that is supposed to prevent exactly this scenario is silently inert under the app's actual router configuration. The defect is a correctness issue in the auth flow rather than a security vulnerability, but it degrades the user experience to an unrecoverable loop.

## Affected code

- `src/app/core/utils/login-redirect.util.ts:3` — `SELF_ROUTE_PATHS` definition
- `src/app/core/utils/login-redirect.util.ts:38-41` — `isSafeReturnTo` pathname check
- `src/app/widgets/widget-login/widget-login.component.ts:30-36` — unconditional `manualSignIn()` in cookie mode
- `src/app/core/services/sso-redirect.service.ts:104-111` — `manualSignIn` resets budget with no budget check
- `src/main.ts:88` — `withHashLocation()` establishes hash routing

## Suggested direction

Two independent fixes close the loop. First, make the self-route check hash-aware: when evaluating whether a `returnTo` URL is KIP's own login route, inspect `resolved.hash` (or the full reconstructed path) rather than `resolved.pathname` alone, so `/#/login` is correctly identified as a self-route. Second, guard `manualSignIn` against re-entry when the user is already authenticated: in `WidgetLoginComponent.ngOnInit`, check `loginStatus` before calling `manualSignIn()` and skip the bounce if the session is already `loggedIn`. Either fix alone breaks the loop; both together also restore the intended budget semantics.

## Verification

Verified against the codebase: all cited mechanisms (hash routing config, dead pathname guard, unconditional `manualSignIn`, budget bypass) confirmed present and the loop chain confirmed reachable (confidence: confirmed).

mairas:
## Why

Under hash routing the self-route loop guard was ineffective, so a sign-in could loop unboundedly (AUTH-02).

## What

- Make `isSafeReturnTo` hash-aware so KIP's own `/#/login` route is recognized as a self-route and not treated as an external return target.
- Guard `WidgetLoginComponent.ngOnInit` to skip `manualSignIn()` when the session is already logged in (e.g. a bookmarked `/login`).

Either fix alone breaks the loop; both are included for defense in depth.

**Theoretical residual (P3):** the self-route fragment match is exact-string, so a matrix-param or trailing-slash `/#/login` variant would evade it — not reachable in practice, since `returnTo` is self-generated and the logged-in case is covered by the `ngOnInit` guard.

Closes #32

### DIFF
--- a/src/app/core/utils/login-redirect.util.spec.ts
+++ b/src/app/core/utils/login-redirect.util.spec.ts
@@ -43,6 +43,15 @@ describe('isSafeReturnTo', () => {
   it('rejects the login self-route (avoids a redirect loop)', () => {
     expect(isSafeReturnTo('/login')).toBe(false);
   });
+
+  it('rejects the hash-routed login self-route (hash routing puts /login in the fragment)', () => {
+    expect(isSafeReturnTo('/#/login')).toBe(false);
+    expect(isSafeReturnTo('/#/login?returnTo=%2Fdashboard')).toBe(false);
+  });
+
+  it('accepts a non-login hash route', () => {
+    expect(isSafeReturnTo('/#/dashboard/0')).toBe(true);
+  });
 });
 
 describe('buildLoginRedirectUrl', () => {
--- a/src/app/core/utils/login-redirect.util.ts
+++ b/src/app/core/utils/login-redirect.util.ts
@@ -2,6 +2,19 @@
 // SSO bounce instead of returning the user to real content.
 const SELF_ROUTE_PATHS = ['/login'];
 
+// The app uses hash-based routing (main.ts withHashLocation), so KIP's own routes live in the URL
+// fragment: /#/login resolves to pathname '/' with hash '#/login'. Checking only the pathname would
+// let /#/login pass as a safe returnTo and loop the SSO bounce, so the fragment's route is checked too.
+function isSelfRoute(resolved: URL): boolean {
+  if (SELF_ROUTE_PATHS.includes(resolved.pathname)) {
+    return true;
+  }
+  const fragment = resolved.hash.startsWith('#') ? resolved.hash.slice(1) : '';
+  const queryIndex = fragment.indexOf('?');
+  const fragmentPath = queryIndex >= 0 ? fragment.slice(0, queryIndex) : fragment;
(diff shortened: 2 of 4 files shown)


mairas:

Both specs pin the fix: `login-redirect.util.spec` asserts `/#/login` and `/#/login?returnTo=…` resolve to `false` while `/#/dashboard/0` stays `true`; `login.component.spec` verifies both branches (not-authenticated → redirecting, `manualSignIn` called once; loggedIn → no redirect, `manualSignIn` not called).

- **P3 (theoretical)** — the self-route fragment match is exact-string, so `/#/login;x=1` or `/#/login/` would evade it. Not reachable in practice: `returnTo` is self-generated from `window.location` and never carries matrix params or a trailing slash, and the logged-in bookmark case is independently covered by the `ngOnInit` guard. No change.

PR #43, opened 2026-06-30
Fix True Wind Angle offset in Wind/Sail Steer simple compass mode (#1066, #1063) (replicates mxtommy/Kip #1073)

mairas:
Replicates upstream PR https://github.com/mxtommy/Kip/pull/1073 ("Fix True Wind Angle offset in Wind/Sail Steer simple compass mode (#1066, #1063)") by dillan.

Method: commit-by-commit cherry-pick (linear history, 2 commits).

This is an experimental replica carried in the SKip fork for evaluation. The change fixes True Wind Angle offset calculation in the Wind/Sail Steer widget when using simple compass mode — upstream issues #1066 and #1063. The PR includes a failing test (RED) followed by the fix (GREEN), preserving the TDD history.

PR #94, opened 2026-07-02
fix(windsteer): recompute TWA base immediately on compass-mode toggle

### LINKED ISSUE #73: Wind/Sail-Steer: TWA needle transient on live compass-mode toggle (#43 follow-up)
## Summary

In the Wind/Sail-Steer widget, toggling compass mode **live** (via widget options) briefly displays the True Wind Angle needle offset by ~heading (~90° in typical cases) for about one wind sample (~1s), until the next TWA update arrives. Steady-state readings are correct; this is a transient, self-correcting, user-action-triggered glitch.

Introduced by PR #43 (replicates mxtommy/Kip#1073), which correctly fixed the persistent ~90° offset in simple compass mode. Surfaced by an independent content re-review; deferred from that PR as a follow-up (minor).

## Mechanism (verified)

After the fix, the parent's computed TWA base is mode-dependent (`heading + T` in enhanced, `T` in simple), recomputed only inside `onTrueWindAngle` on a new stream emission. Toggling `compassModeEnabled` at runtime does **not** re-fire `onTrueWindAngle`: it isn't part of the stream path signature, and the widget is not recreated on options-save (config flows in-place via `WidgetRuntimeDirective`), so `WidgetStreamsDirective.observe()` early-returns without replay. Meanwhile the `svg-windsteer` TWA effect re-runs on the `compassModeEnabled()` input change and applies the **new** heading offset to the **stale** parent base → needle displaced by ~heading until the next TWA sample recomputes the base.

- File: `src/app/widgets/widget-windsteer/widget-windsteer.component.ts`

## Suggested fix

Recompute the displayed true-wind base on a mode/heading change, not only on new TWA data. Minimal approach: cache the last raw boat-relative TWA value in `onTrueWindAngle`, then in the existing `runtime.options()` effect (which already re-runs on `compassModeEnabled` change) recompute `base = computeTrueWindBaseAngle(path, lastRawTWA, currentHeading(), compassModeEnabled)` and `trueWindAngle.set(normalizeAngle(base))`. Read `currentHeading()` untracked to avoid a per-heading-tick recompute. Parent effects flush before child effects, so the SVG then applies the new offset to a fresh base and no transient appears.

## Severity

Minor — only the TWA needle, only just after a deliberate settings change, self-correcting within ~1s. No wrong steady-state data.

mairas:
## Why

Toggling compass mode live left the True Wind Angle needle showing a transient wrong value until the next wind sample arrived (#73, follow-up to #43).

## What

Cache the last raw boat-relative TWA in `onTrueWindAngle`. In the existing `runtime.options()` effect (which already re-runs when `compassModeEnabled` changes), recompute the true-wind base via `computeTrueWindBaseAngle` using an `untracked` read of `currentHeading()` and set `trueWindAngle` — so a live toggle recomputes the base immediately instead of waiting for the next sample. The child `svg-windsteer` already reacts to the input.

**Informational (P3):** an unrelated options save after the heading drifted will recompute against the live heading (a self-correcting one-frame shift, arguably more current) — noted, no change needed.

Closes #73

### DIFF
--- a/src/app/widgets/widget-windsteer/widget-windsteer.component.spec.ts
+++ b/src/app/widgets/widget-windsteer/widget-windsteer.component.spec.ts
@@ -1,5 +1,11 @@
-import { describe, it, expect } from 'vitest';
-import { computeTrueWindBaseAngle } from './widget-windsteer.component';
+import { WritableSignal, signal } from '@angular/core';
+import { TestBed } from '@angular/core/testing';
+import { beforeEach, describe, expect, it } from 'vitest';
+import { WidgetWindComponent, computeTrueWindBaseAngle } from './widget-windsteer.component';
+import { WidgetRuntimeDirective } from '../../core/directives/widget-runtime.directive';
+import { WidgetStreamsDirective } from '../../core/directives/widget-streams.directive';
+import { IPathUpdate } from '../../core/services/data.service';
+import { IWidgetSvcConfig } from '../../core/interfaces/widgets-interface';
 
 /**
  * Regression tests for #1066 / #1063.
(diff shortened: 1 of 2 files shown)


mairas:

Real regression test: it pushes heading=90 and TWA=45 (asserts 45 in simple mode), toggles `compassModeEnabled`, ticks the effect, and asserts 135 (=90+45) — the exact fixed path, which fails on `origin/main` where an options toggle doesn't recompute the cached base.

- **P3 (informational)** — `applyTrueWindBase()` runs on every options change; in compass mode with a boat-relative path, an unrelated options save after the heading drifted recomputes against the live heading rather than the sampled one — a self-correcting one-frame shift (arguably more current). No change needed.

PR #95, opened 2026-07-02
fix(connection): escalate WebSocket retry backoff instead of fixed interval

### LINKED ISSUE #10: [ARCH-02] WebSocket reconnection is unbounded and uses fixed backoff while HTTP retry is capped and escalating
**Severity:** medium | **Confidence:** confirmed | **Area:** architecture | **Source:** codebase review 2026-06-30, finding `ARCH-02`

## Problem

The connection state machine applies two different, inconsistent retry policies to HTTP discovery failures and WebSocket failures. The HTTP path in `onHTTPDiscoveryError()` compares `_httpRetryCount` against `config.httpRetryCount` (3) and transitions to `PermanentFailure` once the cap is reached, using index-based exponential backoff through `retryIntervals`. The WebSocket path in `onWebSocketError()` unconditionally increments `_webSocketRetryCount` and calls `scheduleWebSocketRetry()` with no comparison against `config.webSocketRetryCount` (5), so WebSocket failures never reach `PermanentFailure` and retry indefinitely. Furthermore, `scheduleWebSocketRetry()` hard-codes `retryIntervals[1]` (3000 ms) for every attempt regardless of retry count, making the delay fixed rather than exponential — contradicting the comment `// Exponential backoff: 2s, 3s, 5s` at the config definition site.

## Impact

The `webSocketRetryCount: 5` config field and the `maxRetries` value carried in status messages are display-only and never enforced, misleading anyone reading the code or the config. The divergent model makes the state machine harder to reason about: it is unclear whether the unbounded WebSocket behavior is an intentional design decision (keep reconnecting to a boat server indefinitely) or an oversight. Until the intent is documented and the code made consistent, contributors risk introducing bugs when modifying retry logic, and operators cannot rely on `webSocketRetryCount` as an effective knob.

## Affected code

- `src/app/core/services/connection-state-machine.service.ts:208-229`
- `src/app/core/services/connection-state-machine.service.ts:357-378`
- `src/app/core/services/connection-state-machine.service.ts:46-51`

## Suggested direction

Pick one explicit policy and implement it consistently. Option A: cap WebSocket retries at `webSocketRetryCount` and transition to `PermanentFailure` the same way the HTTP path does, and drive the retry delay from `retryIntervals` indexed by `_webSocketRetryCount` for real exponential backoff. Option B: if indefinite reconnection is the intended behavior for WebSocket (reconnecting to a boat server that may be temporarily offline), make that explicit — remove the misleading `webSocketRetryCount` cap and `maxRetries` field, add a comment explaining the deliberate asymmetry, and still fix the fixed-delay issue by indexing into `retryIntervals` with a clamped count so backoff still escalates. Either way the delay calculation should mirror the HTTP path's use of `Math.min(_webSocketRetryCount - 1, ...)` as the index.

## Verification

Verified against the codebase; every assertion in the finding was confirmed against `connection-state-machine.service.ts` on branch `profiles` (confidence: confirmed).

mairas:
## Why

WebSocket reconnection used a hard-coded `retryIntervals[1]` (a fixed backoff), while the HTTP discovery path escalates and caps (ARCH-02).

## What

Drive the WebSocket retry delay from `retryIntervals` indexed by a clamped `_webSocketRetryCount`, so it escalates through the configured intervals and holds at the longest once exhausted. The spec asserts escalating delays `[2s, 3s, 5s, 5s]` via fake timers.

**Design decision (needs your eyes):** WebSocket reconnection stays **indefinite** — a marine display should keep reaching for a boat server that may be intermittently offline — rather than capping to `PermanentFailure` like the HTTP path. The `webSocketRetryCount` / `maxRetries` fields are retained as informational display values, not enforced caps. If you'd rather cap it, say so.

**Known cosmetic follow-up:** the retry log still prints `N/webSocketRetryCount` (e.g. `N/5`), now misleading under unbounded retry. Left out of scope to keep the change minimal.

Closes #10

### DIFF
--- a/src/app/core/services/connection-state-machine.service.spec.ts
+++ b/src/app/core/services/connection-state-machine.service.spec.ts
@@ -1,3 +1,4 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { ConnectionStateMachine } from './connection-state-machine.service';
 
 describe('ConnectionStateMachine', () => {
@@ -21,5 +22,33 @@ describe('ConnectionStateMachine', () => {
     const retryWindowMs = service.getHttpRetryWindowMs(2000);
 
     expect(retryWindowMs).toBe(12000);
+  });
+
+  it('escalates WebSocket retry backoff through the configured intervals and then holds', () => {
+    vi.useFakeTimers();
+    try {
+      const retry = vi.fn();
+      service.setWebSocketRetryCallback(retry);
+      service.enableWebSocketMode();
+      service.startHTTPDiscovery();
+      service.onHTTPDiscoverySuccess();
+
+      // Delays follow retryIntervals (2s, 3s, 5s) and clamp to the longest once exhausted,
+      // proving the backoff escalates rather than repeating a fixed interval.
+      const expectedDelays = [2000, 3000, 5000, 5000];
+
+      expectedDelays.forEach((delay, attempt) => {
+        service.onWebSocketError('socket dropped');
+
+        vi.advanceTimersByTime(delay - 1);
+        expect(retry).toHaveBeenCalledTimes(attempt);
+
+        vi.advanceTimersByTime(1);
+        expect(retry).toHaveBeenCalledTimes(attempt + 1);
+      });
+    } finally {
+      service.ngOnDestroy();
+      vi.useRealTimers();
+    }
   });
 });
--- a/src/app/core/services/connection-state-machine.service.ts
+++ b/src/app/core/services/connection-state-machine.service.ts
@@ -218,6 +218,10 @@ export class ConnectionStateMachine implements OnDestroy {
       return;
     }
 
+    // Unlike HTTP discovery, WebSocket reconnection is intentionally unbounded: a
+    // marine display must keep reaching for a boat server that may be intermittently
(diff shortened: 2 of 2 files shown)


mairas:
## Independent review — verdict: minor

The production diff is correct: the WS retry delay now escalates via `retryIntervals[min(count-1, len-1)]` and holds at the longest interval; both the log line and the timer use it.

- **P2 'test gap' — assessed, not actionable.** The reviewer flagged a test gap without a rationale. On inspection the added spec does cover the fixed behavior: it drives four `onWebSocketError` calls under fake timers and asserts delays `[2s, 3s, 5s, 5s]` at exact boundaries (`advanceTimersByTime(delay-1)` then `+1`), pinning both escalation and clamp-and-hold. Treating the finding as unsubstantiated.

Two items for the merger, both in the PR body:

- **Design decision** — reconnection stays *indefinite* (a marine display should keep retrying) rather than capping to `PermanentFailure` like the HTTP path. Say if you'd rather cap it.
- **Cosmetic follow-up** — the retry log still prints `N/webSocketRetryCount` (e.g. `N/5`), misleading under unbounded retry; left out of scope to keep the change minimal.

PR #96, opened 2026-07-02
fix(bootstrap): trigger recovery when Signal K returns an empty config

### LINKED ISSUE #3: [PCS-01] Missing-config recovery broken: 200 {} from SK triggers no recovery prompt
**Severity:** high | **Confidence:** confirmed | **Area:** storage | **Source:** codebase review 2026-06-30, finding `PCS-01`

## Problem

The codebase holds two contradictory assumptions about what Signal K returns when a config slot is missing. `SettingsService.startup()` explicitly documents that a missing server config comes back as an empty object `{}` (not a 404), and guards against it with a silent `console.warn` and early return. At the same time, `AppNetworkInitService` raises the `missing-shared-config` bootstrap issue — which triggers the "Create a default configuration now?" recovery prompt in `AppComponent` — only when the HTTP call throws with `error?.status === 404`. Because `getConfig()` does not convert an empty response body into an error, a `200 {}` response never reaches the 404 catch block. The two assumptions are mutually exclusive: whichever one reflects actual SK behaviour, the other recovery path is permanently dead.

## Impact

If SK returns `200 {}` for a missing slot (as the `SettingsService` comment asserts), a first-run shared-config user will land on a completely blank application with no recovery prompt and no visible explanation. The bootstrap completes successfully (an empty object is truthy), `SettingsService` silently early-returns, and the user has no path forward other than knowing to inspect browser console output. This is a silent data-loss / stuck-state bug for any deployment using the shared-config feature.

## Affected code

- `src/app/core/services/settings.service.ts:97-101`
- `src/app/core/services/app-initNetwork.service.ts:168-183`
- `src/app/core/services/app-initNetwork.service.ts:200-209`
- `src/app/app.component.ts:204-214`
- `src/app/core/services/storage.service.ts:294-302`
- `src/app/core/services/storage.service.ts:585-595`

## Suggested direction

Pick one authoritative source of truth for what SK returns when a config slot is absent. If the answer is `200 {}` (as `settings.service.ts` asserts), detect the empty/appless config in the bootstrap path — after `getConfig()` resolves — and emit `IBootstrapIssue{reason:'missing-shared-config'}` there, keeping the 404 branch only as a defensive fallback. The `bootstrapRemoteContext` truthy check should be tightened so that `{}` (missing `app` key) is treated as an absent config, not a successful bootstrap. A test driving `getConfig` returning `{}` through the full bootstrap should assert that the recovery prompt is shown.

## Verification

Verified against the codebase: both contradictory assumptions are present at the cited locations exactly as described (confidence: confirmed).

mairas:
## Why

When Signal K returned `200 {}` (an empty/appless config) rather than a 404, the missing-config recovery prompt never fired — first-run or wiped-config users were left with a silent broken state (PCS-01).

## What

- In `app-initNetwork.service.ts`, after `getConfig()` resolves, detect an empty/appless config and emit `IBootstrapIssue{ reason: 'missing-shared-config' }`, keeping the existing 404 catch branch as a defensive fallback.
- In `storage.service.ts`, tighten `bootstrapRemoteContext`'s truthy `initConfig` check so `{}` is treated as an absent config rather than a successful bootstrap.

The consumers (`app.component` bootstrap-issue effect, `settings.service` startup) already handle the `missing-shared-config` reason, so no change there.

Closes #3

### DIFF
--- a/src/app/core/services/app-initNetwork.service.spec.ts
+++ b/src/app/core/services/app-initNetwork.service.spec.ts
@@ -60,9 +60,11 @@ describe('AppNetworkInitService', () => {
         navigate: vi.fn().mockResolvedValue(true)
     };
 
+    const validRemoteConfig = (): IConfig => ({ app: { configVersion: 11 }, theme: null, dashboards: [] } as unknown as IConfig);
+
     const mockStorage = {
         waitUntilReady: vi.fn().mockResolvedValue(true),
-        getConfig: vi.fn().mockResolvedValue({}),
+        getConfig: vi.fn().mockResolvedValue(validRemoteConfig()),
         bootstrapRemoteContext: vi.fn()
     };
 
@@ -92,6 +94,8 @@ describe('AppNetworkInitService', () => {
         mockConnectionStateMachine.isHTTPConnected.mockClear();
         mockConnectionStateMachine.enableWebSocketMode.mockClear();
         mockConnectionStateMachine.startWebSocketConnection.mockClear();
+        mockStorage.getConfig.mockReset().mockResolvedValue(validRemoteConfig());
+        mockStorage.bootstrapRemoteContext.mockClear();
         mockAuth.loginStatusValue = null;
         mockAuth.refreshLoginStatus.mockClear();
         mockRouter.navigate.mockClear();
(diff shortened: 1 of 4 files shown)


mairas:

Both tests pin the fix. The app-initNetwork spec drives `getConfig` resolving `{}` through the bootstrap and asserts `reason==='missing-shared-config'`, that `bootstrapRemoteContext` was NOT called (early return), and the degraded path (`resetBudget` not called). The storage spec pins the tightened guard: throws on an appless config when `isRemoteContextBootstrapped()===false`, accepts an app-bearing config. The default-mock change from `{}` to an app-bearing config (with a `beforeEach` reset) is necessary so the pre-existing clean-bootstrap tests stay on the happy path now that `{}` is a recovery signal.

- **Minor (accepted)** — no test asserts app.component actually renders the recovery toast, but that wiring is pre-existing and keyed solely on the asserted reason.

PR #97, opened 2026-07-02
fix(auth): preserve session on transient loginStatus probe failure

### LINKED ISSUE #12: [AUTH-01] Transient loginStatus probe failure leaves cookie session stuck in false logged-out state
**Severity:** medium | **Confidence:** confirmed | **Area:** auth | **Source:** codebase review 2026-06-30, finding `AUTH-01`

## Problem

When a WebSocket connection closes uncleanly in cookie-auth mode, the app probes `loginStatus` via HTTP to check whether the session cookie has expired. If that probe fails for any reason — network timeout, unreachable server, non-2xx response, or unparseable body — `refreshLoginStatus` unconditionally maps the failure to `applyLoginStatus(null)`, which sets `_IsLoggedIn$` to false and `_loginStatus$` to null. This conflates a transient transport failure with a definitive logged-out response. There is no re-probe when the WebSocket subsequently reconnects; the reconnect handler does not call `refreshLoginStatus`. As a result, after a brief connectivity blip the WebSocket session self-heals (the cookie still authenticates the upgrade), but the app's session state remains stale-false until the user forces a full page reload.

## Impact

On a reconnected session that is technically still authenticated, write controls and profile affordances are hidden because `isUserSession$` and `canWriteUserData$` are both derived from `_loginStatus$` and evaluate to false when that value is null. Users lose the ability to write data or manage their profile with no visible explanation. This is particularly impactful in marine deployments where transient WiFi drops are routine — any brief network interruption while at sea will trigger the stale-false state, requiring a manual reload to restore normal operation.

## Affected code

- `src/app/core/services/signalk-delta.service.ts:184-188`
- `src/app/core/services/authentication.service.ts:207-226`

## Suggested direction

`refreshLoginStatus` should distinguish between a transport/timeout failure (HTTP status 0 or an RxJS `timeout` error) and an authoritative not-logged-in response (e.g. 401/403 with a parseable body). On a transport failure, the last known-good `loginStatus` value should be preserved rather than replaced with null. Additionally, the WebSocket `open` handler should call `refreshLoginStatus` (or an equivalent re-probe) so that session state self-heals automatically after a reconnect, without requiring a full page reload.

## Verification

Verified against the codebase — control flow confirmed as described (confidence: confirmed).

mairas:
## Why

A transient `loginStatus` probe failure (e.g. a brief WiFi drop) wedged a live cookie session into a false logged-out state, hiding write controls until a full reload (AUTH-01).

## What

- In `AuthenticationService.refreshLoginStatus`, distinguish transport/timeout failures (HTTP status 0 or an RxJS timeout) from an authoritative not-logged-in response, and on transport failure preserve the last known-good status instead of overwriting it with null.
- Re-probe `refreshLoginStatus` in the `SignalKDeltaService` WebSocket-open handler so session state self-heals after a reconnect.

**Noted (P3):** a non-2xx transient (e.g. a 503 during a server restart) is still classified as logout and momentarily clears the session, recovering on the WS-reconnect re-probe. This follows the issue's narrowed direction (preserve only status 0 / timeout); the primary WiFi-drop case surfaces as status 0 and is covered.

Closes #12

### DIFF
--- a/src/app/core/services/authentication.service.spec.ts
+++ b/src/app/core/services/authentication.service.spec.ts
@@ -114,6 +114,70 @@ describe('AuthenticationService', () => {
       }
     });
 
+    it('transport failure (no HTTP response) preserves the last known-good session', async () => {
+      const service = createService();
+      const httpTesting = TestBed.inject(HttpTestingController);
+
+      // Establish a known-good writable session.
+      const first = service.refreshLoginStatus();
+      expectLoginStatusRequest(httpTesting).flush({ status: 'loggedIn', userLevel: 'admin' });
+      await first;
+
+      // A transient network error carries no logout verdict: session state must survive it.
+      const second = service.refreshLoginStatus();
+      expectLoginStatusRequest(httpTesting).error(new ProgressEvent('network error'));
+      const result = await second;
+
+      expect(result).toEqual({ status: 'loggedIn', userLevel: 'admin' });
+      expect(await firstValueFrom(service.isLoggedIn$)).toBe(true);
+      expect(await firstValueFrom(service.isUserSession$)).toBe(true);
+      expect(await firstValueFrom(service.canWriteUserData$)).toBe(true);
+      httpTesting.verify();
+    });
+
+    it('transport timeout preserves the last known-good session', async () => {
+      vi.useFakeTimers();
+      try {
+        const service = createService();
+        const httpTesting = TestBed.inject(HttpTestingController);
+
+        const first = service.refreshLoginStatus();
+        expectLoginStatusRequest(httpTesting).flush({ status: 'loggedIn', userLevel: 'admin' });
+        await first;
+
+        const second = service.refreshLoginStatus();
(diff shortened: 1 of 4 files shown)


mairas:

Three auth specs pin the fix: a status-0 network error and a 5001ms timeout both preserve `loggedIn`/`isUserSession`/`canWriteUserData`; an authoritative 200/notLoggedIn after a good session still clears it (guards against over-preservation). The delta spec asserts `refreshLoginStatus` is invoked on `socketWSOpenEvent$`. Pre-existing 503 fail-closed and mid-session tests stay green.

- **P3 (informational)** — a non-2xx transient (e.g. a 503 while the SK server restarts) is classified as authoritative logout and momentarily clears the session, healing on the WS-reconnect re-probe rather than being preserved. This follows the issue's narrowed direction (preserve only status 0 / timeout); the primary WiFi-drop case surfaces as status 0 and is covered.

PR #98, opened 2026-07-02
fix(settings): deep-clone default config before returning

### LINKED ISSUE #35: [PCS-05] Default-config singletons are mutated by reference instead of cloned
**Severity:** low | **Confidence:** confirmed | **Area:** storage | **Source:** codebase review 2026-06-30, finding `PCS-05`

## Problem

The `getDefaultAppConfig` and `getDefaultConnectionConfig` methods in `settings.service.ts` assign the exported module-level constant objects directly by reference and then mutate them (e.g. setting `config['configVersion'] = latestConfigVersion`, `config.kipUUID = UUID.create()`, `config.signalKUrl = ...`). No deep clone is performed before mutation. Because `config.blank.const.ts` binds `defaultConfig.app` to the same `DefaultAppConfig` object, these getters mutate the exact object that `profile.service.ts` reads as the authoritative import-validation baseline when checking `configVersion`. Currently the mutations are effectively no-ops (the assigned values equal the existing literals), so there is no live bug, but the cross-module coupling is real and fragile: any future getter that changes a meaningful value would silently corrupt the shared baseline used throughout the application.

## Impact

The fragile coupling means that any extension to a `getDefault*` method that alters a field — even a seemingly innocuous assignment — will invisibly modify the `defaultConfig` singleton. Code paths that read from `defaultConfig` for validation or fallback (such as profile import) would then see a runtime-mutated object rather than the true compile-time default, potentially causing incorrect config-version checks, unexpected value propagation, or hard-to-trace bugs. This is a pre-existing upstream smell, but SKip carries and builds on this code, so the risk accumulates as the codebase evolves.

## Affected code

- `src/app/core/services/settings.service.ts:783-810`
- `src/default-config/config.blank.const.ts:6-28`
- `src/app/core/services/profile.service.ts:90-95`

## Suggested direction

Each `getDefault*` method should `cloneDeep` the corresponding exported constant before mutating any fields on the result, ensuring that the shared module-level singleton is never modified at runtime. The `config.blank.const.ts` constants should be treated as strictly immutable references so that `defaultConfig.app` always reflects the true compile-time default. Introducing a lint rule or TypeScript `Readonly` type on these exports would make the immutability contract explicit.

## Verification

Confirmed against the codebase: `settings.service.ts:783` assigns `DefaultAppConfig` by reference and mutates the result; `config.blank.const.ts` binds `defaultConfig.app` to the same object; `profile.service.ts:90` reads `defaultConfig.app?.configVersion` as the baseline — confirming the cross-module coupling described above (confidence: confirmed).

mairas:
## Why

`getDefaultAppConfig()` / `getDefaultConnectionConfig()` assigned the module-level `DefaultAppConfig` / `DefaultConnectionConfig` singletons **by reference** and then mutated them, corrupting the shared defaults across calls — `profile.service` reads `defaultConfig.app.configVersion` as its import-validation baseline (PCS-05).

## What

- Deep-clone the constant (`cloneDeep` from `lodash-es`, matching `profile.service`) before mutating/returning, so the singleton is never mutated at runtime.
- Mark the two leaf constants `Readonly<…>` in `config.blank.const.ts` to make the immutability contract explicit. Scoped to the two leaves; the `defaultConfig` composite stays mutable because `profile.service` cloneDeeps-and-mutates it.
- New `default config isolation` specs assert each getter returns a fresh pristine object and that mutating a result never leaks into the singleton or a later call.

## Note

Review surfaced a **sibling instance** of the same anti-pattern in `app-initNetwork.service.ts:369` (`this.config = DefaultConnectionConfig; this.config.signalKUrl = …`) that the `Readonly` type can't catch (it's laundered through a mutable alias). Out of this PR's footprint — filed as a follow-up.

Verified locally: lint + `build:dev` clean, settings spec green (21 tests).

Closes #35

### DIFF
--- a/src/app/core/services/settings.service.spec.ts
+++ b/src/app/core/services/settings.service.spec.ts
@@ -4,6 +4,13 @@ import { MatSnackBar } from '@angular/material/snack-bar';
 import { SettingsService } from './settings.service';
 import { StorageService } from './storage.service';
 import { ensureLocalStorage } from '../../../test-helpers/local-storage.test-helper';
+import { DefaultAppConfig, DefaultConnectionConfig } from '../../../default-config/config.blank.const';
+import { IAppConfig, IConnectionConfig } from '../interfaces/app-settings.interfaces';
+
+interface DefaultConfigGetters {
+  getDefaultAppConfig(): IAppConfig;
+  getDefaultConnectionConfig(): IConnectionConfig;
+}
 
 interface SeedOpts {
   sharedConfigName?: string;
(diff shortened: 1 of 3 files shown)


mairas:

The deep-clone fix is correct and the specs are real, not smoke stubs: `expect(second).not.toBe(first)` and the singleton-stays-pristine assertions (`signalKUrl` null, `kipUUID`/`sharedConfigName` unchanged after mutating a returned clone) each genuinely fail against the pre-fix reference-sharing code. `Readonly` on the two leaf constants is verified assignable to existing consumers (readonly → mutable assignment is allowed in TS). Verified locally: 21 settings specs pass, lint + build clean.

- **P3 (out of scope, tracked)** — the same mutate-the-shared-singleton pattern still exists in `app-initNetwork.service.ts:369` (`this.config = DefaultConnectionConfig; this.config.signalKUrl = window.location.origin;`). The new `Readonly` type can't catch it — it's laundered through the mutable `this.config` alias — so on a fresh boot with no stored connection config the singleton is still corrupted in place. Correctly left out of this PR's footprint (settings/config only); filed as a follow-up issue rather than silently expanding scope.

PR #99, opened 2026-07-02
refactor(data): remove unsafe unused unsubscribePath

### LINKED ISSUE #31: [ARCH-07] unsubscribePath destructively completes shared subjects, making it unsafe and uncalled
**Severity:** low | **Confidence:** confirmed | **Area:** architecture | **Source:** codebase review 2026-06-30, finding `ARCH-07`

## Problem

`subscribePath()` in `DataService` deduplicates registrations by `(path, source)` and returns the same shared `BehaviorSubject` (`pathDataUpdate$`) to every caller that requests the same path and source — a deliberate fan-out design. However, `unsubscribePath(path)` locates the first matching registration by path alone (ignoring source) and calls `.complete()` on all four shared subjects. Because the subject instance is shared across all co-subscribers of that `(path, source)` pair, any single consumer calling `unsubscribePath` would terminate the stream for every other consumer on that path. The mismatch between the sharing model in `subscribePath` and the destructive teardown in `unsubscribePath` makes the latter unsafe under the service's own invariants. To make matters worse, a repo-wide search confirms that `unsubscribePath` has zero callers outside its own definition — the main pipeline (`WidgetStreamsDirective`) instead relies on `takeUntilDestroyed` and never invokes it — so per-`(path, source)` registrations accumulate for the lifetime of the session.

## Impact

The accumulation of registrations is bounded by the number of unique `(path, source)` pairs a session encounters (tens to low-hundreds for a typical Signal K dashboard), so this is not a runaway memory leak. The more significant concern is the public API surface: `unsubscribePath` is advertised as a teardown method but would silently break all co-subscribers if ever called. Any future contributor adding a widget that calls `unsubscribePath` on unmount — a perfectly reasonable expectation — would introduce a hard-to-debug stream termination affecting the entire dashboard. This is a latent footgun that is neutral to current SKip deployment behaviour but worth hardening before the pattern is repeated.

## Affected code

- `src/app/core/services/data.service.ts:211-301`
- `src/app/core/directives/widget-streams.directive.ts:103`

## Suggested direction

Either introduce reference counting in `unsubscribePath` so it only tears down the underlying subjects when the last consumer unregisters, or remove `unsubscribePath` entirely in favour of the `takeUntilDestroyed`/`finalize` model already used by `WidgetStreamsDirective`. If the latter path is chosen, document explicitly that per-`(path, source)` registrations are app-lifetime objects bounded by unique path/source combinations, so future contributors do not reintroduce manual teardown.

## Verification

Verified against the codebase (confidence: confirmed): the deduplication logic in `subscribePath`, the destructive `.complete()` calls in `unsubscribePath`, the path-only `findIndex`, and the absence of any callers of `unsubscribePath` outside `data.service.ts` were all confirmed in the working tree.

mairas:
## Why

`DataService.unsubscribePath` destructively `.complete()`d the shared per-`(path, source)` subjects — terminating the stream for **every** co-subscriber of that path — and it was called by nobody (ARCH-07). `WidgetStreamsDirective` uses `takeUntilDestroyed`/`finalize` instead.

## What

- Delete `unsubscribePath` entirely (KISS/YAGNI — no reference-counting for an uncalled method). A repo-wide grep confirmed zero callers.
- Remove the now-dead supporting code it alone required: the `IPathRegistration.combinedSub` field and its assignment (the `combined$` subscription is now anonymous — identical runtime behavior, it was never unsubscribed in practice) and the unused `Subscription` import.
- Update the two stale JSDoc references; document that path registrations are app-lifetime objects, bounded by unique `(path, source)` pairs, intentionally never torn down individually.
- Spec pins the fan-out invariant: the same `(path, source)` returns one shared subject, distinct sources return distinct subjects, and co-subscribers both keep receiving from the shared stream.

## Note

The spec is an **invariant guard**, not a regression test: because the removed method had zero callers there's no buggy behavior to fail against, so it locks in the fan-out property that would make any reintroduced destructive teardown unsafe.

Verified locally: lint + `build:dev` clean, data spec green (7 tests).

Closes #31

### DIFF
--- a/src/app/core/services/data.service.spec.ts
+++ b/src/app/core/services/data.service.spec.ts
@@ -42,6 +42,33 @@ describe('DataService', () => {
     expect(service).toBeTruthy();
   });
 
+  it('shares one live subject per (path, source) so co-subscribers cannot tear each other down', () => {
+    const first$ = service.subscribePath('self.electrical.batteries.10.voltage', 'default');
+    const second$ = service.subscribePath('self.electrical.batteries.10.voltage', 'default');
+    const otherSource$ = service.subscribePath('self.electrical.batteries.10.voltage', 'test-source');
+
+    // Deduplication is keyed by (path, source): identical pairs share one subject, distinct sources do not.
+    expect(second$).toBe(first$);
+    expect(otherSource$).not.toBe(first$);
+
+    const firstValues: unknown[] = [];
+    const secondValues: unknown[] = [];
+    first$.subscribe(update => firstValues.push(update.data.value));
+    second$.subscribe(update => secondValues.push(update.data.value));
+
+    dataPathUpdates$.next({
+      context: 'self',
+      path: 'electrical.batteries.10.voltage',
+      source: 'test-source',
+      timestamp: '2026-01-01T00:00:01.000Z',
+      value: 12.5,
+    });
+
+    // Both co-subscribers keep receiving from the shared stream; there is no teardown that could complete it.
+    expect(firstValues.at(-1)).toBe(12.5);
+    expect(secondValues.at(-1)).toBe(12.5);
+  });
+
   it('applies notification state to path value updates', () => {
     let latest: IPathUpdate | undefined;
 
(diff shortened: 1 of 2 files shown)


mairas:

Deletion is safe: a repo-wide grep confirms `unsubscribePath` had zero real callers (only the definition + two JSDoc mentions), and `WidgetStreamsDirective` tears down via `takeUntilDestroyed`/`finalize`. Removing the entailed dead code (`combinedSub` field/assignment, unused `Subscription` import) is justified — the `combined$` subscription is now anonymous with identical runtime behavior since it was never unsubscribed. The added spec's data-flow was traced (`updatePathData` → `defaultSource._pathData$.next` → `combined$` → `pathDataUpdate$`); its identity-dedup and fan-out assertions genuinely pass.

- **P3 (accepted)** — the spec is an invariant guard, not a regression test: with zero callers there's no buggy state to fail against. Inherent to fixing an uncalled footgun by deletion; it correctly locks in the fan-out property that makes any reintroduced destructive teardown unsafe.

PR #102, opened 2026-07-02
fix(storage): surface config save failures to the user

### LINKED ISSUE #15: [PCS-03] Routine config saves have no user-facing failure reporting
**Severity:** medium | **Confidence:** confirmed | **Area:** storage | **Source:** codebase review 2026-06-30, finding `PCS-03`

## Problem

All `SettingsService` set methods — covering theme, dashboards, units, night mode, split shell, notifications, and datasets — call `storage.service.ts`'s `patchConfig()`, which returns `void` and enqueues the write as a fire-and-forget operation. When a patch fails, the failure handler only calls `console.error` and increments an internal `_patchFailures` counter. That counter is consumed exclusively by `awaitQueueDrain()`, which is only invoked during profile switch, rename, or delete operations — never during routine saves. Additionally, `patchConfig` silently early-returns (with only a `console.error`) when `sharedConfigName` is unset, making that path an invisible no-op. The net result is that a transient failure saving a dashboard layout, theme, or unit preference is silently discarded; the user only discovers the loss after a page reload reverts their change.

## Impact

Users who experience a storage failure during routine configuration changes receive no notification that their change was not persisted. This creates a confusing experience where settings appear to apply immediately but revert on reload, with no indication anything went wrong. In a marine deployment context where users rely on dashboard and unit settings being stable across sessions, silent data loss is particularly disruptive because it may not be noticed until the operator is underway and settings are reverted to unexpected values.

## Affected code

- `src/app/core/services/storage.service.ts:359-364`
- `src/app/core/services/storage.service.ts:96-100`
- `src/app/core/services/settings.service.ts:372-382`
- `src/app/core/services/settings.service.ts:576-585`

## Suggested direction

Surface patch failures to the user for all routine saves. One approach is to expose a per-write completion observable or failure stream from `patchConfig` so callers can react, and show a snackbar notification on failure (consistent with what `resetSettings` already does). Alternatively, settings writes could `await` a write-completion signal rather than discarding the `void` result. The `sharedConfigName`-unset early-return path in `patchConfig` should also be treated as a reportable error rather than a silent no-op.

## Verification

Verified against the codebase: both cited files exist in the fork and the described behavior is confirmed (confidence: confirmed).

mairas:
## Why

Routine config saves (SettingsService's ~10 `set*` methods → `StorageService.patchConfig`) were fire-and-forget and silent on failure: the JSON-Patch queue's `catchError` only bumped a private counter, and a `patchConfig` call with `sharedConfigName` unset silently early-returned (a no-op). A failed dashboard/theme/unit save was invisible until a reload reverted it (PCS-03).

## What

- StorageService now exposes a `patchFailure$` stream (`Subject<IPatchFailure>` = `{ key?, error }`). The queue's `catchError` emits on it — but **only for fire-and-forget patches**; deferred/profile ops settle their own promise and are excluded to avoid double-reporting. The unset-`sharedConfigName` refusal now emits too, instead of a silent no-op. `patchConfig`'s `void` signature and its ~10 call sites are untouched.
- SettingsService subscribes once at construction and shows a MatSnackBar on failure, reusing the existing reset/demo snackbar pattern.
- **Read-only sessions are not alarmed:** expected 401/403 write-denials are filtered out (they mean "no write access," not "broken") — only genuine failures (network / 5xx) raise the persistent toast. (Addresses the branch review's P3.)

## Tests

Failed routine patch emits on the stream naming the target key and triggers the snackbar end-to-end; unset-slot reported without POSTing; a deferred `removeItem` failure stays silent on the stream (rejects its promise); and a **401 denial does not raise the toast**.

Verified locally: lint + `build:dev` clean, settings + storage specs green (55 tests).

Closes #15

### DIFF
--- a/src/app/core/services/settings.service.spec.ts
+++ b/src/app/core/services/settings.service.spec.ts
@@ -1,4 +1,5 @@
 import { TestBed } from '@angular/core/testing';
+import { HttpTestingController } from '@angular/common/http/testing';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { MatSnackBar } from '@angular/material/snack-bar';
 import { SettingsService } from './settings.service';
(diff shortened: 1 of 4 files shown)


mairas:
## Independent review — verdict: minor → **P3 resolved**

Solid design and real tests. The `patchFailure$` stream is emitted only for fire-and-forget patches (deferred/profile ops keep their promise-based handling, verified via grep that all 10 `patchConfig` callers are fire-and-forget), and the specs pin it precisely: a 500 emits on the stream with the target key and triggers the snackbar end-to-end (real StorageService + HttpTestingController), the unset-slot refusal emits without POSTing, and a deferred `removeItem` failure stays silent on the stream while rejecting its promise.

- **P3 (fixed)** — the initial pass would have raised the persistent (`duration:0`) alarm toast on *every* expected 401/403 save in a legitimately read-only session (the queue's own comment cites the read-only 401 as expected). Fixed: 401/403 write-denials are now filtered out of the alarm (they mean "no write access," not "broken"); only genuine failures raise it. Added a spec asserting a 401 does not open the toast. Verified locally (55 settings+storage specs green, lint + build clean).

Issue #101, opened 2026-07-02
SK-01 remaining: bound _skData growth + optional fine-grained SK subscriptions (deferred from #20)


mairas:
Deferred from #20, which removed the dead per-path subscription-policy fields and documented the firehose model. The substantive performance half of SK-01 remains:

1. **Bound `_skData` growth** — prune timed-out entries from `DataService.timeoutPathObservable()` so the per-path map does not grow unbounded under the subscribe=all firehose.
2. **Optional real per-path server-side SK subscriptions** — emit context/subscribe on first widget registration for a path, unsubscribe on last release, forwarding a per-path minimum period.

Both are performance optimizations not currently warranted on a small MFD (the firehose is acceptable at current scale); tracked so the concern is not lost.

PR #103, opened 2026-07-02
refactor(signalk): remove dead IWidgetPath subscription-policy fields

### LINKED ISSUE #20: [SK-01] Firehose subscription: SK fine-grained subscribe/policies never used
**Severity:** medium | **Confidence:** confirmed | **Area:** signalk | **Source:** codebase review 2026-06-30, finding `SK-01`

## Problem

Kip never issues Signal K `subscribe`/`unsubscribe` delta messages to the server. Instead it opens the WebSocket with a single coarse query (`?subscribe=self` or `?subscribe=all` plus `&sendMeta=all`) and performs all filtering and rate-limiting entirely client-side via RxJS `sampleTime`. The per-widget `subscribePath`/`unsubscribePath` calls in `data.service` only mutate in-memory registers; no server message is ever emitted. The subscription policy fields exposed in the configuration UI — `policy`, `format`, `minPeriod`, `smoothingPeriod` — are each explicitly annotated `NOT IMPLEMENTED` in the codebase and are dead fields with no effect.

## Impact

Running a permanent firehose subscription means the server streams every path (or every vessel under `subscribe=all`) continuously, regardless of what the active dashboard widgets actually need. On constrained links (marine embedded hardware, satellite, slow cellular) this wastes bandwidth and CPU on both ends. Additionally, the `_skData` map in `data.service` grows unbounded under `subscribe=all`: the timeout handler nulls a path's value but never removes its entry, so AIS-heavy deployments accumulate entries indefinitely until `resetSignalKData()` is called. The dead policy/format/minPeriod fields in the config UI mislead users into believing server-side rate control is available when it is not.

## Affected code

- `src/app/core/services/signalk-delta.service.ts:59-61`
- `src/app/core/services/signalk-delta.service.ts:284-293`
- `src/app/core/interfaces/widgets-interface.ts:621-628`
- `src/app/core/directives/widget-streams.directive.ts:135-137`

## Suggested direction

Two approaches are viable. The simpler path is to remove the dead `policy`, `format`, `minPeriod`, and `smoothingPeriod` fields from `IWidgetSvcConfig` (and the matching UI controls) to stop implying capability that does not exist, and to add pruning logic to `timeoutPathObservable()` so that timed-out entries are deleted from `_skData` rather than left as nulled stubs. The more complete path is to implement real per-path SK subscriptions: emit a `context/subscribe` delta when the first widget registers a path and an `unsubscribe` delta when the last widget releases it, with `policy` and `minPeriod` forwarded to the server, moving rate-limiting and path selection server-side. At minimum, the firehose model and its `_skData` growth risk under `subscribe=all` should be documented.

## Verification

Verified against the codebase by tracing all outbound socket writes and all callers of `subscribePath`/`unsubscribePath`; confidence: confirmed.

mairas:
## Why

The app consumes the entire Signal K delta firehose and never uses SK's fine-grained subscribe/policies, yet `IWidgetPath` carried four dead fields — `policy`, `format`, `minPeriod`, `smoothingPeriod` — implying per-path subscription control that does not exist (SK-01).

## What

- Remove the four dead fields from `IWidgetPath` (grep-confirmed zero readers and zero writers — no member access, no object-literal writers, no HTML/config controls) and trim the now-unused `TFormat`/`TPolicy` imports.
- Document the firehose model in `signalk-delta.service.ts`: a single coarse subscription (self/all), path selection in `DataService`, per-widget rate limiting in `WidgetStreamsDirective` (`sampleTime`), and no per-path SK subscription policies. (The doc pointer was corrected per the branch review — the rate-limiting lives in `WidgetStreamsDirective`, not `DataService`.)

Pure type-level dead-field removal — no runtime behavior change. Verified locally: lint + `build:dev` clean.

## Deferred (the performance half of SK-01) → #101

Bounding `_skData` growth (pruning timed-out entries) and optional real per-path server-side subscriptions are perf optimizations not currently warranted on a small MFD — tracked in #101 so the concern isn't lost.

Closes #20

### DIFF
--- a/src/app/core/interfaces/widgets-interface.ts
+++ b/src/app/core/interfaces/widgets-interface.ts
@@ -1,6 +1,6 @@
 import { ITheme } from '../services/app-service';
 import { TValidSkUnits } from '../services/units.service';
-import { TFormat, TPolicy, TScaleType } from './signalk-interfaces';
+import { TScaleType } from './signalk-interfaces';
 import type { IElectricalCardModeConfig } from '../contracts/electrical-topology-card.contract';
 
 /**
@@ -689,14 +689,6 @@ export interface IWidgetPath {
   suppressBootstrapNull?: boolean;
   /** Used as a reference ID when path is an Array and array index is not appropriate. */
   pathID?: string | null | '';
-  /** NOT IMPLEMENTED - Signal K - smoothingPeriod=[milliseconds] becomes the transmission rate, e.g. every smoothingPeriod/1000 seconds. Default: 1000 */
-  smoothingPeriod?: number;
-  /** NOT IMPLEMENTED - Signal K - format=[delta|full] specifies delta or full format. Default: delta */
-  format?: TFormat;
-  /** NOT IMPLEMENTED - Signal K - policy=[instant|ideal|fixed]. Default: ideal */
-  policy?: TPolicy;
-  /** NOT IMPLEMENTED - Signal K - minPeriod=[milliseconds] becomes the fastest message transmission rate allowed, e.g. every minPeriod/1000 seconds. This is only relevant for policy='instant' to avoid swamping the client or network. */
-  minPeriod?: number;
   /**
    * Optional: Indicates if the path control must have a path value or not.
    * - If true: The control is valid if the path is either a valid (non-empty) path or empty (null or '').
--- a/src/app/core/services/signalk-delta.service.ts
+++ b/src/app/core/services/signalk-delta.service.ts
@@ -52,7 +52,11 @@ export class SignalKDeltaService implements OnDestroy {
   }
   public streamEndpoint$: BehaviorSubject<IStreamStatus> = new BehaviorSubject<IStreamStatus>(this.streamEndpoint);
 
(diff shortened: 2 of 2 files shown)


mairas:
## Independent review — verdict: minor → **both P3s resolved**

The removal is safe: `policy`/`format`/`minPeriod`/`smoothingPeriod` on `IWidgetPath` were grep-confirmed to have zero readers and zero writers (the `smoothingPeriod` in `dataset-stream.service` is a different field on a different interface, untouched), and the unused `TFormat`/`TPolicy` imports were trimmed. Pure type-level dead-code removal, no runtime change; lint + build clean.

- **P3 (fixed)** — the firehose doc comment pointed at `DataService sampleTime`, but the client-side rate limiting actually lives in `WidgetStreamsDirective`. Corrected.
- **P3 (addressed)** — this branch does only the dead-field removal; SK-01's performance half (bounding `_skData` growth + optional real per-path subscriptions) is deferred. Rather than silently closing SK-01, that half is now tracked in #101, referenced from the PR.

PR #104, opened 2026-07-02
fix(storage): serialize setConfig through the patch queue

### LINKED ISSUE #14: [PCS-02] setConfig bypasses the patch queue, allowing races with in-flight autosave patches
**Severity:** medium | **Confidence:** confirmed | **Area:** storage | **Source:** codebase review 2026-06-30, finding `PCS-02`

## Problem

The storage layer maintains a serialized patch queue (`patchQueue$` via `concatMap`) specifically because Signal K cannot handle concurrent `applicationData` writes. The queue comment states this explicitly, and `patchConfig`, `patchGlobal`, and `removeItem` all correctly route through `enqueuePatch`. However, `setConfig` bypasses the queue entirely and issues a direct, un-queued `http.post`. This means that whenever `setConfig` is called while autosave patches are queued or in-flight against the same `applicationData` document, a full-file POST and a JSON-Patch POST can hit the Signal K server concurrently — exactly the collision scenario the queue was built to prevent. The callers that trigger this gap include `resetSettings`, `loadDemoConfig`, `createProfile`, `duplicateProfile`, `importProfile`, and `renameProfile`, none of which drain the queue before invoking `setConfig`.

## Impact

Any operation that replaces a config slot (profile reset, demo load, profile creation, import, or rename) while an autosave cycle is active can corrupt the active configuration document on the Signal K server. Because the race window is narrow — it requires a concurrent in-flight autosave patch — failures are intermittent and difficult to reproduce, yet the consequence (applicationData collision) is precisely what the serialization queue was designed to avoid. On a live marine deployment where KIP is running continuously and autosave is active, a profile operation can silently discard or partially apply changes.

## Affected code

- `src/app/core/services/storage.service.ts:324-349`
- `src/app/core/services/storage.service.ts:89-107`
- `src/app/core/services/settings.service.ts:624-647`
- `src/app/core/services/settings.service.ts:666-686`

## Suggested direction

Either route `setConfig` through the same `enqueuePatch` serialization pipeline used by `patchConfig`/`patchGlobal`/`removeItem`, so full-file writes and JSON patches never overlap; or introduce an `awaitQueueDrain()` call at every `setConfig` call site before the POST is issued. At minimum, `resetSettings` and `loadDemoConfig` — which overwrite the active slot while autosave targets the same document — must drain before calling `setConfig`. The existing `awaitQueueDrain` already used in `switchProfile` and `deleteProfile` is the correct pattern to extend.

## Verification

Verified against the codebase (confidence: confirmed): the queue, its rationale, all compliant call sites, and every unguarded `setConfig` call site were confirmed to exist as described.

mairas:
## Why

The storage layer serializes every applicationData write through a JSON-Patch queue (`patchQueue$` + `concatMap`) because Signal K can't handle concurrent applicationData writes. But `setConfig` bypassed the queue with a direct `http.post`, so any slot-replacing operation (profile reset, demo load, profile create/import/rename, config migration) could race an in-flight autosave patch against the same document — the exact collision the queue exists to prevent (PCS-02). The window is narrow (needs a concurrent in-flight autosave), so failures are intermittent, but the consequence is a corrupted applicationData document.

## What

- Route `setConfig`'s full-file POST through `enqueuePatch`, the same serial pipeline `patchConfig` / `patchGlobal` / `removeItem` already use. `patch()` is a generic `http.post(url, document)`, so the full config document flows through it unchanged; the queue guarantees it never overlaps a JSON-Patch write.
- `setConfig` stays `async` and awaits a Promise wrapped around the queued write (mirroring `removeItem`), so its contract is preserved: `assertSlotName`'s synchronous throw still rejects, and a queued-write failure still rejects with the raw error (`patch()` logs it). Because the deferred patch sets `reject`, it settles its own promise and is excluded from `patchFailure$` (no double report).
- No change to the six call sites — they already `await setConfig`.

## Tests

New `setConfig queue serialization` describe: a full-file write queues behind an in-flight autosave patch instead of firing a concurrent POST (this test fails on the old direct-POST code — TDD); the promise resolves only after the queued write completes; and it rejects on failure while the queue keeps processing later writes.

Verified locally: storage (35), settings + profile (55), configuration-upgrade (4) specs green; lint + `build:dev` clean.

Closes #14

### DIFF
--- a/src/app/core/services/storage.service.spec.ts
+++ b/src/app/core/services/storage.service.spec.ts
@@ -178,6 +178,58 @@ describe('StorageService', () => {
     });
   });
 
+  describe('setConfig queue serialization', () => {
+    beforeEach(() => {
+      service.storageServiceReady$.next(true);
+      service.activeConfigFileVersion = 11;
+      service.sharedConfigName = 'p';
+    });
+
+    afterEach(() => http.verify());
+
+    it('queues behind an in-flight autosave patch instead of racing a direct POST', async () => {
+      // An autosave JSON-Patch is in flight (dispatched, not yet answered).
+      service.patchConfig('Dashboards', [{ id: 'd1' }]);
+      const patchReq = http.expectOne((r) => Array.isArray(r.body));
+
+      // A full-file write must NOT fire concurrently — it waits its turn in the queue.
+      const done = service.setConfig('user', 'default', blankConfig());
+      http.expectNone((r) => !Array.isArray(r.body));
+
+      patchReq.flush(null);
+      await Promise.resolve();
+
+      // Only now does the full-file write go out, on its own.
+      const setReq = http.expectOne((r) => !Array.isArray(r.body));
+      expect(setReq.request.method).toBe('POST');
+      setReq.flush(null);
+      await done;
+    });
+
+    it('resolves only after the queued full-file write completes', async () => {
+      let resolved = false;
+      const done = service.setConfig('user', 'default', blankConfig()).then(() => { resolved = true; });
+
+      const req = http.expectOne((r) => r.method === 'POST');
+      await Promise.resolve();
+      expect(resolved).toBe(false);
+
+      req.flush(null);
+      await done;
+      expect(resolved).toBe(true);
+    });
+
+    it('rejects on failure yet the queue keeps processing later writes', async () => {
+      const first = service.setConfig('user', 'first', blankConfig());
+      http.expectOne((r) => r.method === 'POST').flush('boom', { status: 500, statusText: 'err' });
+      await expect(first).rejects.toBeTruthy();
+
+      const second = service.setConfig('user', 'second', blankConfig());
+      http.expectOne((r) => r.method === 'POST').flush(null);
+      await expect(second).resolves.toBeNull();
+    });
+  });
+
   describe('bootstrapRemoteContext appless-config guard', () => {
     it('rejects an appless config (SK 200 {}) as an absent bootstrap', () => {
       expect(() => service.bootstrapRemoteContext({
--- a/src/app/core/services/storage.service.spec.ts
+++ b/src/app/core/services/storage.service.spec.ts
@@ -1,6 +1,6 @@
 import { TestBed } from '@angular/core/testing';
 import { HttpTestingController } from '@angular/common/http/testing';
-import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { BehaviorSubject } from 'rxjs';
 import { StorageService, IPatchFailure } from './storage.service';
 import { IConfig } from '../interfaces/app-settings.interfaces';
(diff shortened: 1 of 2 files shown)


mairas:
## Independent multi-persona review — PR #104 (setConfig → serial patch queue)

Reviewers: correctness, testing, maintainability, project-standards, reliability, adversarial (6, report-only). Intent: route `setConfig`'s full-file POST through the serial patch queue so it can't race an in-flight autosave against the same applicationData document (#14 / PCS-02).

The core fix is correct and well-tested — the serialization holds, the reject-on-error contract is preserved (verified: `assertSlotName` sync-throw still rejects; a queued-write failure rejects with the raw error and is excluded from `patchFailure$` — no double report), `_pendingPatches$` accounting stays balanced, and the race test genuinely fails on the old direct-POST code. Standards clean; no VERSION bump applies to this repo.

One finding is the real decision point.

### P2 — Moderate

| # | Where | Issue | Reviewers | Conf |
|---|-------|-------|-----------|------|
| 1 | `storage.service.ts` `patch()` (~71-91) | **No per-request timeout on the queue → this PR widens a head-of-line stall to awaited full-file writes.** `patch()` has no `timeout()` (unlike `getConfig`, which uses `REMOTE_CONFIG_TIMEOUT_MS`). A single *hung* applicationData POST (accepted, never answered — half-open TCP / proxy stall) blocks `concatMap`, so it now also hangs `setConfig` and every later queued write. Cascade: `resetSettings`/`loadDemoConfig` silently never reload and show no error; `ProfileService.createProfile/import/duplicate` hang with `mutationInFlight` latched `true`, so **all** later profile ops throw "Another profile operation is still in progress" until a full page reload. Pre-existing gap (patch() never had a timeout), but **widened by this PR** — the old direct POST couldn't be blocked by an unrelated write. | reliability (0.82), adversarial (0.68), correctness (residual) | ~0.85 |

Fix: add `timeout(REMOTE_CONFIG_TIMEOUT_MS)` inside `patch()`'s pipe (mirror `getConfig`). A stalled write then fails the queue item, `concatMap` advances, `setConfig` rejects (snackbar fires), `finalize` decrements the count. Note it changes behaviour for *all* queue writes (autosave patches gain a timeout too) — a scope call, hence flagged rather than auto-applied.

### P3 — Low

| # | Where | Issue | Reviewers | Conf |
|---|-------|-------|-----------|------|
| 2 | `storage.service.ts` (~118) | Stale queue comment still says "JSON Patch requests"; the queue now also carries full-file writes. Trivial comment update. | maintainability | 0.62 |
| 3 | `storage.service.ts` (125-133, 234-253) | A failing `setConfig` increments the shared `_patchFailures`, which can flip a *concurrent* `awaitQueueDrain` (profile delete/rename) to a spurious "could not confirm — please retry" even though that op succeeded. Narrow (needs two overlapping flows + a failure); the safe direction (over-report, no data loss). | adversarial (0.6), correctness, testing | 0.6 |
| 4 | `storage.service.ts:373` | Deferred-patch Promise idiom (`new Promise((resolve,reject)=>enqueuePatch(...))`) now duplicated in `setConfig` + `removeItem`; a private `enqueueDeferredPatch` would DRY it and protect the reject-wiring contract. YAGNI judgment (2 callers). | maintainability | 0.6 |

### Testing gaps (advisory, non-blocking)
- The double-report exclusion isn't asserted `setConfig`-specifically — but the identical shared branch is covered by `removeItem`'s existing test, so it's a one-line optional add, not a real gap.
- No head-of-line-stall test (tied to finding #1; awkward to unit-test without a timeout to assert against).

Finding #1 (queue timeout) is the one that matters — a narrow but real failure-coupling regression this PR introduces. Recommend addressing it (fold in, or fast-follow) plus the trivial comment (#2) before merge. #3 and #4 are optional. No P0/P1 blockers; the corruption race the PR set out to fix is genuinely fixed.

mairas:
## Review response

Addressed the review findings:

- **#1 (P2, queue timeout) — fixed.** Added `timeout(REMOTE_CONFIG_TIMEOUT_MS)` to `patch()` (mirrors `getConfig`), so a hung applicationData POST now fails the queue item instead of wedging the serial queue and every awaited `setConfig` behind it. A TimeoutError flows through the existing reject/count/advance path. New test pins it: a stalled queued write rejects after the timeout and the queue keeps processing later writes.
- **#2 (P3, stale comment) — fixed.** The queue comment now describes it as serializing all applicationData writes (JSON patches and full-file replacements).
- **#3 (drain cross-contamination on concurrent setConfig failure)** and **#4 (DRY `enqueueDeferredPatch` helper) — left as noted.** Both are narrow/judgment-call and outside this PR's minimal scope; not folding them in.

Re-verified locally: storage spec 36/36, settings + profile + configuration-upgrade specs 59/59, lint + build clean. The timeout applies to all queue writes (autosave patches included), which is the intended hardening.

Issue #118, opened 2026-07-02
AIS: throttle + value-dedupe own-ship updates (Kip#1101)


mairas:
Part of adopting the freeze-fix series from mxtommy/Kip#1101 (open upstream; five sibling issues, one per fix area). Upstream measured −63% handler-wait on AIS radar per-fix churn.

## Problem

`applyOwnShipUpdate` sets the `_ownShip` signal with a fresh object on every `self.navigation.*` delta, unthrottled. The AIS radar's render effect guards re-renders by `ownShipRef` reference equality, which every fresh object defeats — so each position/heading delta triggers downstream work even when nothing changed.

Skip state: exact pre-fix code at `src/app/core/services/ais-processing.service.ts:229` (fresh object per delta) and `:265` (unconditional `set`); the only throttle in the file is the pre-existing `targetsDirty$` one (:177).

## Plan

Cherry-pick (with `-x`, preserving upstream authorship) from the fetched `upstream-pr-1101` branch:

- `b75a2315` — failing test for own-ship update throttling (RED)
- `b63859d4` — buffer updates, throttle 250 ms leading+trailing, set the signal only on a field-level change so unchanged fixes keep the same object reference (GREEN)

## Verification

- Cherry-picked spec additions pass (`ng test --include='src/app/core/services/ais-processing.service.spec.ts'`)
- `npm run lint`, `npm run build:prod`
- CI green on the PR

## Notes

Stacked on #116 (target eviction) because both change `ais-processing.service.ts`: this branch is based on the eviction branch and its PR should merge after the eviction PR (retarget to main once #116's PR merges). Upstream #1101 is still open; if it changes before merge, reconcile before closing this issue.

PR #120, opened 2026-07-02
Bound AIS target growth with hard cap + TTL eviction (replicates part of mxtommy/Kip#1101)

### LINKED ISSUE #116: AIS: bound target growth with hard cap + TTL eviction (Kip#1101)
Part of adopting the freeze-fix series from mxtommy/Kip#1101 (open upstream; five sibling issues, one per fix area). This fix bounds heap, flush and render growth over a voyage — particularly relevant to always-on HaLOS Pi deployments.

## Problem

AIS targets are only removed on an explicit `sensors.ais.status === 'remove'`, which many servers never send, so the `tracks` map grows without bound over long uptimes and every flush/render scales with it.

Skip state: `src/app/core/services/ais-processing.service.ts` matches upstream's pre-patch shape — `tracks` map with no cap/TTL fields (:136), constructor with only the `targetsDirty$` throttle (:166-180), `removeTrack` invoked solely on explicit remove (:399-406). No eviction logic anywhere in Skip (`evict|TARGET_TTL|MAX_TARGETS` have zero hits).

## Plan

- `b44f8685` — failing tests for target eviction (RED)
- `5ebe0092` — hard cap (500 targets) + TTL eviction (10 min, swept every 15 s) via the existing `removeTrack` (GREEN)
- `ca1df062` — make cap enforcement a cheap single pass (fixes a churn regression)

Note: `ca1df062` lands after the own-ship throttle commits upstream, so it may conflict when cherry-picked directly onto this branch; resolve against the eviction-only context.

## Verification

- Cherry-picked spec additions pass (`ng test --include='src/app/core/services/ais-processing.service.spec.ts'`)
- CI green on the PR

## Notes

Shares `ais-processing.service.ts` with the own-ship throttle sibling issue; that branch stacks on this one and this PR merges first. Upstream #1101 is still open; if it changes before merge, reconcile before closing this issue.

mairas:
## Motivation

AIS targets are only removed on an explicit `sensors.ais.status === 'remove'`, which many Signal K servers never send. Over a long uptime the `tracks` map grows without bound, and every flush and radar render scales with it. The upstream freeze audit measured this as multi-second main-thread stalls after long uptime — particularly relevant to always-on HaLOS Pi deployments.

## Approach

Cherry-picked from the open upstream PR mxtommy/Kip#1101 with `git cherry-pick -x`, preserving upstream authorship and messages:

- `b44f8685` — failing tests for target eviction (RED)
- `5ebe0092` — hard cap (`MAX_TARGETS = 500`) enforced on every track insert, plus a 15 s sweep (`EVICTION_SWEEP_MS`) dropping targets not heard from within a 10-minute TTL (`TARGET_TTL_MS`), all routed through the existing `removeTrack` so the context/mmsi indexes are pruned with the tracks (GREEN)
- `ca1df062` — replaces the full sort in `enforceTargetCap` with a single O(n) min-find pass per removed target, fixing an O(n log n)-per-insert churn regression at the cap

All three applied cleanly — no conflicts, no adaptations needed. `ca1df062` sits after the own-ship-throttle commits upstream, but git auto-merged it against the eviction-only context; the own-ship throttle stays out of this PR (sibling issue #118 stacks on this branch). The eviction `interval` subscription is torn down via `takeUntilDestroyed(this.destroyRef)`, consistent with the service's other subscriptions.

Note: upstream #1101 is still open; if it changes before merge, reconcile before closing the issue.

## Verification

- `npx ng test --include='src/app/core/services/ais-processing.service.spec.ts'` — 1 file, 13 tests, all passed (10 pre-existing + 3 new eviction tests)
- `npm run lint` — all files pass
- `npm run build:prod` — completes; only the pre-existing `piexifjs` CommonJS warning

Closes #116

### DIFF
--- a/src/app/core/services/ais-processing.service.spec.ts
+++ b/src/app/core/services/ais-processing.service.spec.ts
@@ -183,3 +183,64 @@ describe('AisProcessingService applyAisUpdate dispatch', () => {
     expect(track!.ais.status).toBe('confirmed');
   });
 });
+
+/**
+ * Eviction bounds unbounded target growth (the "unresponsive after a while"
+ * freeze): many Signal K setups never send an explicit `status: 'remove'`, so
+ * every distinct MMSI ever heard used to accumulate for the app lifetime,
+ * making every flush + radar render O(targets) and heavier over uptime.
+ */
+describe('AisProcessingService target eviction (bounds unbounded growth)', () => {
+  let stream$: Subject<IPathUpdateWithPath>;
+  let service: AisProcessingService;
+  const EVENT_TS = new Date('2026-06-24T00:00:00Z').getTime();
+
+  function makeEvent(fullPath: string, value: unknown): IPathUpdateWithPath {
+    return { path: fullPath, update: { data: { value, timestamp: new Date(EVENT_TS) }, state: 'normal' } } as IPathUpdateWithPath;
+  }
+  function push(fullPath: string, value: unknown): void {
+    stream$.next(makeEvent(fullPath, value));
+    vi.advanceTimersByTime(300);
+  }
+  const internals = () => service as unknown as {
+    tracks: Map<string, unknown>;
+    contextIndex: Map<string, string>;
+    mmsiIndex: Map<string, Set<string>>;
+    maxTargets: number;
+    evictStaleTracks: (nowMs: number) => void;
+  };
+
+  beforeEach(() => {
+    vi.useFakeTimers();
+    stream$ = new Subject<IPathUpdateWithPath>();
+    TestBed.configureTestingModule({
+      providers: [AisProcessingService, { provide: DataService, useValue: { subscribePathTree: () => stream$.asObservable() } as Partial<DataService> }]
+    });
+    service = TestBed.inject(AisProcessingService);
+  });
+  afterEach(() => vi.useRealTimers());
+
+  it('caps the retained track set at maxTargets, evicting the oldest, and prunes the indexes', () => {
+    internals().maxTargets = 5;
+    for (let i = 0; i < 12; i++) {
+      push(`vessels.urn:mrn:imo:mmsi:${100000000 + i}.navigation.position.latitude`, 10 + i * 0.001);
+    }
+    expect(internals().tracks.size).toBe(5);
+    // indexes must not leak entries for evicted tracks
+    expect(internals().contextIndex.size).toBe(5);
+  });
+
+  it('evicts tracks not updated within the TTL when the sweep runs', () => {
+    push(`vessels.urn:mrn:imo:mmsi:123456789.navigation.position.latitude`, 10);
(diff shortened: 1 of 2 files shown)


mairas:
## Review synthesis

Two independent reviewers examined this PR:

1. **Upstream fidelity + eviction interaction** — verified the three cherry-picks (b44f8685, 5ebe0092, ca1df062 from mxtommy/Kip#1101) against upstream with `git range-diff` and a full-file diff: byte-identical apart from cherry-pick trailers, with the two interleaved own-ship-throttle commits correctly excluded and no fragments of them imported by the ca1df062 conflict resolution. Also checked eviction interplay with the `targetsDirty$` 250 ms throttle and the radar widget's cache/selection handling of evicted targets — all correct.
2. **Testing** — verified the cherry-picked spec is byte-identical to upstream b44f8685, confirmed RED-commit behavior (exactly the 3 new eviction tests fail on pre-fix code), and audited fake-timer usage (correct; no real 15 s waits leak into the suite).

### Findings and disposition

| Severity | Finding | Disposition |
|---|---|---|
| minor | Track evicted at birth by cap enforcement leaves a dangling `contextIndex` entry (`resolveTrack` registers the context after `createTrack` has already run `enforceTargetCap`) | **Not fixed here** — confirmed byte-identical to upstream ca1df062, so it is an inherited upstream quirk, not a replication defect; fixing it in this PR would diverge from upstream. Pathological trigger (older-than-all timestamps while at the 500-track cap) and self-healing on the context's next update. Tracked in #123 with a note to prefer an upstream-first fix. |
| minor | The 15 s interval → `evictStaleTracks` wiring had zero coverage: deleting the interval subscription left all 13 tests green | **Fixed** in 8aaedc38 — new test anchors the fake clock with `vi.setSystemTime`, advances past the TTL, and asserts eviction through the interval path (with a mid-TTL keep assertion proving the clock anchor took). |
| nit | Cap test claims "evicting the oldest" but all events share one timestamp, so ordering was unasserted | **Fixed** in 8aaedc38 — new test pushes staggered timestamps over the cap and asserts the least-recently-updated contexts are the ones evicted. |
| nit | `mmsiIndex` declared in the spec's internals accessor but never asserted, so an index leak on eviction would go uncaught | **Fixed** in 8aaedc38 — the sweep test creates the track via its `mmsi` path and asserts `mmsiIndex` is pruned after eviction. |

The three coverage gaps were upstream-identical, so the fixes are additive fork-authored tests only; no cherry-picked content was modified. Both new tests were mutation-checked: deleting the interval wiring and inverting the min-find comparison each fail exactly one new test while the 13 upstream tests stay green.

### Verification after fixes

- `npx ng test --include='src/app/core/services/ais-processing.service.spec.ts'` — 15/15 pass
- `npm run lint` — clean
- `npm run build:prod` — succeeds

PR #121, opened 2026-07-02
Coalesce + time-slice canvas ResizeObserver work (replicates part of mxtommy/Kip#1101)

### LINKED ISSUE #114: Canvas resize storm: coalesce + time-slice ResizeObserver, skip no-op reallocs (Kip#1101)
Part of adopting the freeze-fix series from mxtommy/Kip#1101 (open upstream; five sibling issues, one per fix area). Upstream measured this fix eliminating a 1097 ms main-thread block (median @10× CPU throttle) during fullscreen/grid-relayout resize storms — the primary "can't exit fullscreen" freeze.

## Problem

`CanvasService`'s shared `ResizeObserver` callback processes every entry synchronously (`setHighDPISize` + per-canvas resize callback per entry), so a layout change that resizes many canvases at once forms one long blocking task. `setHighDPISize` also unconditionally reassigns `canvas.width`/`height`, which reallocates and clears the backing store even when the device-pixel size is unchanged.

Skip state: `src/app/core/services/canvas.service.ts:83-127` (synchronous per-entry processing, including debug-log blocks upstream deletes) and `:254-262` (unconditional realloc). Both regions match upstream's pre-patch shape, so the upstream commits should apply near-mechanically. About 10 components register through `CanvasService`, but `WidgetTitleComponent` is one of them and is embedded in ~20 widgets, so a grid relayout still storms many canvases. Note the fix does *not* cover gauge-steel, ais-radar, or horizon resize paths — those bypass `CanvasService`'s observer.

## Plan

- `7e3327d5` — failing tests for resize coalescing (RED)
- `5de5810e` — coalesce entries into a pending map, flush under an 8 ms/frame rAF budget, skip no-op backing-store reallocs (GREEN)

Adaptation to check: Skip's `setHighDPISize` does a `ctx.setTransform` + `clearRect` after sizing; the upstream early-return must skip that too when the size is unchanged (that is the intended upstream behavior).

## Verification

- Cherry-picked spec additions pass under Skip's vitest builder (`ng test --include='src/app/core/services/canvas.service.spec.ts'`)
- CI green on the PR (CI is the source of truth for the full suite)

## Notes

Upstream #1101 is still open; if it changes before merge, reconcile before closing this issue.

mairas:
## Motivation

`CanvasService`'s shared `ResizeObserver` callback processed every entry synchronously (`setHighDPISize` + per-canvas resize callback per entry), so a fullscreen/grid-relayout that resizes many canvases at once formed one long blocking main-thread task — the primary "can't exit fullscreen" freeze. `setHighDPISize` also unconditionally reassigned `canvas.width`/`height`, reallocating and clearing the backing store even when the device-pixel size was unchanged. Upstream measured this fix eliminating a 1097 ms main-thread block (median @10x CPU throttle) during resize storms.

## Approach

Cherry-picked from the open upstream PR mxtommy/Kip#1101 with `git cherry-pick -x`, preserving upstream authorship (Dillan Laughlin):

- `7e3327d5` — failing tests pinning the two mitigations (RED)
- `5de5810e` — coalesce ResizeObserver entries into a pending map, flush under an 8 ms/frame rAF budget with the remainder deferred to the next frame, and skip `setHighDPISize`'s backing-store realloc when the device-pixel size is unchanged (GREEN)

Note: upstream #1101 is still open; if it changes before merge there, we reconcile per the tracking issue.

## Adaptations

Both commits applied cleanly with no conflicts — Skip's pre-patch observer callback and `setHighDPISize` matched upstream's shape. Two Skip-specific divergences were verified rather than adapted:

- Skip's `setHighDPISize` does `ctx.setTransform` + `clearRect` after sizing (unlike upstream Kip). The cherry-picked early return lands *before* the realloc, so the unchanged-size path now also skips the transform reset and clear — which is the upstream intent (the comment even says "Skip it (and the transform reset)"). Skip's block is preserved on the genuine-change path.
- Skip's observer callback carried debug-log blocks that the upstream patch deletes; deleting them with the patch is correct (they were part of the storm's synchronous work).

No fork-authored commits were needed.

## Verification

- `npx ng test --include='src/app/core/services/canvas.service.spec.ts'` — 1 file, 4 tests passed (2 pre-existing + 2 new)
- `npm run lint` — all files pass
- `npm run build:prod` — completes (only the pre-existing `piexifjs` CommonJS warning)

CI is the source of truth for the full suite.

Closes #114

### DIFF
--- a/src/app/core/services/canvas.service.spec.ts
+++ b/src/app/core/services/canvas.service.spec.ts
@@ -1,5 +1,5 @@
 import { TestBed } from '@angular/core/testing';
-import { beforeEach, describe, expect, it } from 'vitest';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { CanvasService } from './canvas.service';
 
 describe('CanvasService', () => {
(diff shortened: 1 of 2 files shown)


mairas:
## Review synthesis

Two independent review lenses examined this PR:

1. **Upstream-fidelity correctness** — hunk-by-hunk comparison against mxtommy/Kip PR #1101 (commits `7e3327d5`, `5de5810e`).
2. **Testing** — do the cherry-picked specs execute meaningfully under Skip's vitest builder, and would they fail on pre-fix code (RED/GREEN verification).

### Fidelity verdict

Both branch commits (`f4960bc7`, `77bc3547`) produce patches byte-identical to the upstream commits, and the resulting `canvas.service.ts` / `canvas.service.spec.ts` are byte-identical to upstream's post-commit files (`git diff 5de5810e 77bc3547 -- <both files>` is empty). No hunks dropped, no conflicts mis-resolved, cherry-pick `-x` trailers and upstream authorship intact. The upstream PR branch contains no later canvas-service commits, so the replication scope is complete.

### Testing verdict

RED verified at `f4960bc7` (tests without fix): the skip-guard test fails with a behavioral assertion (`expected 2 to be 1` — pre-fix code reallocates on unchanged size) and the time-slice test fails on the missing `pendingResize` member. GREEN verified at branch head: 4/4 via `npx ng test --include='src/app/core/services/canvas.service.spec.ts'`. The time-slice test guards the budget check, not just member existence. The one consumer spec exercising `registerCanvas` (widget-race-timer) passes under the new rAF-deferred resize path.

### Findings and dispositions

| Severity | Finding | Disposition |
|---|---|---|
| minor | `unregisterCanvas` doesn't remove the canvas from `pendingResize`, so a canvas destroyed mid-resize-storm gets one stale-size backing-store realloc after `releaseCanvas` zeroed it. The map self-drains (no leak) and the resize callback can't fire (already deleted). | **Dismissed — inherited verbatim from upstream** (file is byte-identical to `5de5810e`). Cost is one wasted, GC-reclaimable realloc per destroyed canvas per storm. Fixing here would break the byte-fidelity this replication PR exists to preserve; the correct venue is an upstream contribution. Deliberately not tracked in this fork — negligible impact, and the fork tracks upstream on this file. |
| nit | The flushResizes test restores its `performance.now`/`requestAnimationFrame` mocks inline at the end of the test body instead of in `afterEach`/try-finally, so a mid-test assertion failure would leak mocks to later tests in the file. | **Dismissed — inherited verbatim from upstream** (`7e3327d5`). Latent-only: it's the last test in the file and vitest isolates test files. Same fidelity reasoning as above. |

No blockers, no fork-authored changes required.

### Verification at head (`77bc3547`)

- `npx ng test --include='src/app/core/services/canvas.service.spec.ts'` — 4/4 passed
- `npm run lint` — all files pass
- `npm run build:prod` — succeeds (pre-existing piexifjs CommonJS warning only)

PR #122, opened 2026-07-02
Decouple AIS radar rotation from target rendering (replicates part of mxtommy/Kip#1101)

### LINKED ISSUE #115: AIS radar: decouple course-up rotation from target rendering (Kip#1101)
Part of adopting the freeze-fix series from mxtommy/Kip#1101 (open upstream; five sibling issues, one per fix area). Upstream measured −54% handler-wait (full stack, 3436→1585 ms) on the AIS radar under course-up rotation.

## Problem

The AIS radar rebuilds rings, targets, vectors and selection on every rotation frame: `rotationPending` bypasses the render-signature short-circuit, and the signature uses `ownShipRef` object identity, which every own-ship heading delta invalidates. During a turn the render cost is O(targets) per frame.

Skip state: `src/app/widgets/widget-ais-radar/widget-ais-radar.component.ts` is near-identical to upstream's pre-patch file (flat SVG layers at :291-296, `rotationPending` bypass at :345-346, `ownShipRef` identity at :360/:377, full rebuild chain at :397-408). Skip's diff vs upstream on this file is ~30 cosmetic lines (type strictness), so the upstream commits should apply with one trivial conflict.

## Plan

- `dd5b59f2` — rotate a single SVG group per frame instead of rebuilding targets; build targets north-up; position-only render signature (`ownPosKey`); split own-ship vector out of target vectors
- `9a200994` — skip `renderRings` on pure-rotation frames

Known conflict: Skip reads `getColors(cfg.color, theme)` where upstream's patched line reads `getColors(cfg.color ?? 'contrast', theme)` — keep Skip's stricter form.

Porting hazard (breaks silently if lost in conflict resolution): `toLocalPoint()` must inverse-rotate click coordinates by `lastViewRotation` into the north-up frame, otherwise target click-selection breaks in course-up mode.

## Verification

Upstream verified this with its perf-harness screenshot/click tooling, which we are not adopting; there are no unit tests for the render change. Compensate with:

- Hunk-by-hunk comparison of the applied diff against the upstream patch after conflict resolution
- Review focus on the rotation math, the `dataChanged` gating, and click hit-testing
- CI green on the PR

## Notes

Upstream #1101 is still open; if it changes before merge, reconcile before closing this issue.

mairas:
## Motivation

The AIS radar rebuilt rings, targets, vectors and selection on every course-up rotation frame: `rotationPending` bypassed the render-signature short-circuit, and the signature compared `ownShipRef` by object identity, so every own-ship heading delta invalidated it. During a turn the render cost was O(targets) per animation frame — upstream's rank-4 render freeze. Upstream measured −54% handler-wait on the radar (full stack, 3436→1585 ms) after this series.

## Approach

Cherry-picked from the open upstream PR mxtommy/Kip#1101 with `git cherry-pick -x`, preserving upstream authorship and messages:

- `dd5b59f2` — move targets/vectors/selection into a `radar-rotation` SVG group so course-up rotation is a single O(1) group transform; build targets north-up; replace `ownShipRef` identity in the render signature with position-only `ownPosKey` (heading no longer forces a rebuild); split `renderOwnShipVector` (per-frame, O(1)) out of `renderTargetVectors` (data-change only); un-rotate click coordinates by `lastViewRotation` back into the north-up frame so hit-testing keeps working in course-up mode.
- `9a200994` — rebuild `renderRings` (rings, labels, own-ship icon) only on data changes; pure-rotation frames just re-orient the own-ship icon.

Note: upstream #1101 is still open; if it changes before merge there, we reconcile per #115.

## Adaptations

- **Known conflict (second pick), resolved keeping Skip's form:** upstream's patched line reads `getColors(cfg.color ?? 'contrast', theme)`; Skip's stricter typing makes the fallback redundant, so the resolution keeps `getColors(cfg.color, theme)` inside the new `if (dataChanged)` block. All other Skip divergences (non-null-assertion removals, `status: string`, dropped `snapshotAisTrack`) were untouched by the merge and preserved.
- **Fork-authored commit `chore(radar): drop upstream perf-harness files`:** both upstream commits carried perf-harness result JSONs and `verify-click.mjs` from upstream's measurement tooling, which Skip does not adopt (no `perf-harness/` infrastructure here). Removed in a separate commit rather than amending the cherry-picks.

## Verification

There are no upstream unit tests for this change (upstream verified visually with its perf harness, not adopted here), so verification is hunk fidelity plus lint/build:

- Hunk-by-hunk diff of the resulting file against upstream's post-patch file: the only remaining deltas are Skip's pre-existing cosmetic type-strictness divergences plus the `getColors` resolution above.
- Click-mapping math checked: the rotation group applies `rotate(-viewRotation)`; `eventToRadarPoint` applies the SVG rotation matrix for `+viewRotation` — the exact inverse — so north-up hit-testing stays correct in course-up mode.
- `npm run lint` — all files pass.
- `npm run build:prod` — succeeds (only the pre-existing `piexifjs` CommonJS warning).
- `npx ng test --include='src/app/core/services/ais-processing.service.spec.ts'` (nearest AIS spec) — 10/10 pass.

Closes #115


mairas:
## Review synthesis

Three independent reviewers examined this PR: (1) hunk-by-hunk replication fidelity against upstream dd5b59f2/9a200994, (2) behavior adversary on the rotation decoupling, (3) test-coverage feasibility for the ported logic.

### Replication fidelity (no findings)

The combined diff on `widget-ais-radar.component.ts` is byte-identical to the upstream `dd5b59f2^..9a200994` diff except the one documented adaptation (`getColors(cfg.color, theme)` vs upstream's `?? 'contrast'`, behaviorally identical since Skip's `getColors` falls back to `'contrast'` internally). Residual full-file divergence vs upstream's post-patch file equals Skip main's pre-existing divergence line-for-line. Per-commit diffs match their upstream counterparts; perf-harness rider files are cleanly dropped in 5237b1a3.

### Fixed: coverage gap on the click un-rotate math (minor)

Upstream verified `eventToRadarPoint` — the path upstream itself called "the one correctness path a static screenshot can't cover" — via `perf-harness/verify-click.mjs`, which this fork correctly drops. That left the porting hazard with no automated verification on a file that will keep absorbing upstream cherry-picks.

Addressed in 0383d9d6: a TestBed-free vitest spec with geometry-hand-derived expectations (5 cases). Verified here: 5/5 pass via `npx ng test --include=...`; a sign-flip mutant at the rotation line fails exactly the two rotation cases, so the spec discriminates the regression class this port risks.

### Tracked, not fixed: settled-rotation own-ship staleness (minor) → #124

Confirmed real: the `!dataChanged && rotationSettled` early return can briefly freeze the own-ship icon/vector (north-up after a north-settled course-up stint; SOG-only changes in settled course-up). But the code is inherited byte-identical from upstream dd5b59f2 (verified by diffing against the upstream post-patch file — the region is absent from the residual diff), it self-heals on the next signature-touching update (~1s on live data), and fixing here would diverge from upstream on a file receiving continued cherry-picks. Filed #124 to track reporting it upstream; not a merge blocker.

### Verification after changes

`npm run lint` passes, `npm run build:prod` succeeds, the new spec passes 5/5.

Issue #87, opened 2026-07-02
History-chart backfill plots linear (:avg) means for circular/direction paths — wrong at the 0/360° wrap


mairas:
## Defect

For angle paths (`unit: 'rad'`, `domain: 'direction'`/`'signed'`), the History-API chart engine backfills each plotted point from the server's `:avg` aggregate, which is a **linear** arithmetic mean. Linear averaging is invalid for circular quantities: the mean of 350° and 10° is 0° (north), but a linear mean returns 180° (south) — the opposite heading. This is worse than showing raw aliased samples, because it fabricates a plausible-looking but reversed value.

## Where

- `HistoryChartStreamService.fetchBackfill()` (`src/app/core/services/history-chart-stream.service.ts`) requests `<path>:avg` (and `:sma`, `:min`, `:max`) and plots the `:avg` column as each point's value.
- `HistoryToChartMapperService.mapValuesToChartDatapoints()` (`src/app/core/services/history-to-chart-mapper.service.ts`) applies only `normalizeToDirection` / `normalizeToSigned` — a modulo wrap into range. It **cannot** correct the circular error, because the server already collapsed each bucket's raw sub-samples into one linear number; the raw values needed for a circular mean are gone by the time the client sees the response.
- The error compounds at coarse resolution: a 60-min window buckets at ~30 s, so each `:avg` folds several raw samples, and any bucket crossing the wrap is wrong.

## Inconsistent with the live tail

The live tail (`startLive`) is circular-correct — it keeps raw values and computes a circular `sma` via `computeWindowStats`. So the backfilled left portion of a direction chart and the live right portion use incompatible semantics, and only the backfill is wrong. The mapper's dataset-wide `lastAverage` (a proper `atan2` circular mean, line ~150) is also contaminated, since it averages the already-linear per-bucket values.

## Direction (not prescriptive)

Options to weigh:

- If the SK2 History API exposes a circular/angular aggregation method, request it for `rad` paths instead of `:avg`. (Open question — needs confirming against the History API surface / the InfluxDB provider's capabilities.)
- Otherwise, for `rad` paths request raw (or fine-resolution) values and aggregate circularly client-side (`circularMeanRad`), accepting a larger backfill payload for angle paths only. This mirrors what the live tail already does and keeps both halves consistent.

Whichever path, the backfill and live-tail aggregation for angle domains must share the same circular semantics so the seam is continuous.

## Minor, same code path

[...]

Issue #86, opened 2026-07-02
Chart density: fixed 120-point target is too coarse for wide and long-duration charts


mairas:
## Observation

Chart sampling density is pinned to a single constant, `TARGET_POINTS_PER_WINDOW = 120`, in `src/app/core/utils/chart-window.util.ts`. From it, `deriveDataSourceInfo()` derives `sampleTime = round(windowMs / 120)`, `maxDataPoints ≈ 120`, and `smoothingPeriod = 25% of maxDataPoints`. This util is shared by **both** the client-side recorder (`DatasetStreamService`) and the History-API engine, so the limit applies everywhere.

120 points is visibly coarse in two regimes:

- **Wide charts.** On a large/high-DPI display a chart canvas can be well over 1000 CSS px wide; 120 points spreads to ~8+ px per point, so the trace looks chunky and stair-stepped relative to the available resolution.
- **Long-duration charts.** A 24h window collapses to one point every 12 minutes (`86.4M ms / 120`); a 7-day window to ~84 min/point. Short-lived features (gusts, tacks, transients) vanish into the aggregation.

## Direction (not prescriptive)

Either raise the constant substantially, or make the target dynamic. The principled dynamic target is the chart's rendered width in device pixels (`canvasCssWidth * devicePixelRatio`), capped — there is little value drawing more points than the axis has pixels. That naturally gives dense wide charts and avoids over-fetching for small ones.

## Tradeoffs to weigh

The target is not free to raise — it scales three things together:

- **Backfill payload / server query.** The History backfill resolution is derived from `sampleTime`, so a higher target means finer aggregation and more rows returned per backfill (though InfluxDB scans the same raw data either way).
- **Live buffer size** and the per-sample `computeWindowStats` cost (O(maxDataPoints) per tick) — both bounded and cheap at 120, still cheap at ~1000, worth a cap beyond that.
- **Recorder memory**, since the recorder holds the same ring buffer client-side.

A cap in the low thousands keeps all three modest while removing the visible coarseness. Chart.js decimation / the streaming plugin can absorb the extra points on the render side.

## Scope

Affects both chart engines (shared util). Independent of the History-engine prototype work in #64.

Issue #85, opened 2026-07-02
History-chart engine: live tail does not re-backfill after WebSocket reconnect (data gap)


mairas:
## Context

The History-API chart engine prototype (issue #64 spike, branch `feat/history-chart-prototype`) fills a chart with **one** History API backfill request, then rides the existing Signal K delta WebSocket for the live tail. See `HistoryChartStreamService.getBackfillThenLive()` / `startLive()` in `src/app/core/services/history-chart-stream.service.ts`.

## Problem

The backfill happens exactly once, at stream construction. If the delta WebSocket drops and reconnects (laptop sleep, network blip, server restart), the live tail resumes emitting new samples but the interval that elapsed *during* the disconnect is never re-fetched. The chart carries a silent gap until that span scrolls off the left edge.

Only a full stream rebuild (data-config change or page reload) currently re-triggers a backfill.

Note: the client-side recorder engine has the same limitation. This issue targets the History engine; the recorder can be tracked separately if we decide it matters there too.

## Desired behavior

A WebSocket reconnect should re-trigger a backfill so the gap is filled from the History API, then resume the live tail.

## Implementation notes (not prescriptive)

A reconnect signal already exists and can drive the re-backfill:

- `ConnectionStateMachine.state$` emits `ConnectionState.Connected` on (re)connect, or
- `SignalKDeltaService.streamEndpoint$` emits `operation: 2` ("Connected").

The re-backfill only needs to cover `from = lastLiveTimestamp` (or `now - windowMs`, whichever is later) rather than the full window, and must de-duplicate against buffered points so the seam is clean.

## Scope

Only affects the History engine path, which is behind the `chartEngine` config / `kip.chartEngine` localStorage flag and not yet the default. Not urgent; captured so the prototype does not ship this gap silently if promoted.

Issue #64, opened 2026-06-30
Consolidate trend charting on the SK History API; investigate whether a client-side recorder is still warranted


mairas:
## Position

SKip carries two parallel mechanisms for historical/trend chart data. We should minimize the overlap and the custom implementation, leaning on the SK v2 History API as the source of truth where it can serve the need. But removing the client-side recorder outright is **not** a foregone conclusion — there's a real performance/UX question, plus two hard constraints, that this issue exists to investigate before any code moves.

## The two mechanisms today

**1. Custom client-side recorder — `DatasetStreamService` (`src/app/core/services/dataset-stream.service.ts`).** The legacy Kip dataset engine, still the live charting path:
- Subscribes to the **live SK delta stream** via `DataService.subscribePath` (not the History API), sampling on a `timer(sampleTime)`.
- Keeps **capped in-memory rolling buffers** per dataset (`historicalData: number[]`, shift/push at `maxDataPoints`).
- Computes its **own statistics locally** — arithmetic avg/SMA/min/max for scalars, circular mean / circular min-max for radian paths. (`ema`/`doubleEma` fields are vestigial, always `null`.)
- Emits to widgets via a batch-then-live `ReplaySubject` (`getDatasetBatchThenLiveObservable`).
- Consumers: `widgets/minichart`, `widgets/widget-data-chart`, `widgets/widget-windtrends-chart`, plus `WidgetDatasetOrchestratorService` and the widget-config dataset list.

**2. SK v2 History API client stack.** `HistoryApiClientService`, `KipSeriesApiClientService`, `HistoryToChartMapperService`, `DashboardHistorySeriesSyncService`. Consumers: the history-chart dialog, `widget-host2`, display options.

**Current point of contact:** the recorder optionally **seeds** a dataset's initial window from the History API (`seedHistoryData`), gated to `sampleTime ≥ 1000ms` and falling back to "live-only mode" on any failure or empty response. So the recorder's buffering and stats are otherwise independent of the History API and run whether or not it exists.

The overlap: both produce time-windowed, aggregated (avg/min/max/SMA) trend series for chart widgets. The History API already supports those aggregations server-side (the recorder even requests `:sma:N`, `:avg`, `:min`, `:max` when seeding).

## Why this is worth doing

- Two engines for "trend chart data" is duplicated surface — two stats implementations, two windowing schemes, two notions of a dataset.
[...]

PR #127, opened 2026-07-02
feat(charts): History-API chart engine behind an opt-in toggle (#64 spike)

mairas:
## What this is

The History-API chart engine from the #64 de-risking spike, rebased onto current `main` and cleaned up for review. It's an alternative data path for `widget-data-chart` that leans on the Signal K v2 History API instead of the client-side recorder (`DatasetStreamService`).

It ships **behind a default-off toggle**, so default behaviour is unchanged. This PR lands the infrastructure the history-chart cluster issues build on; it does **not** promote the engine or retire the recorder (that's #64).

## Approach

- **`HistoryChartStreamService`** — one History-API backfill for the initial window, then a thin SK delta-stream live tail with a minimal rolling buffer and the shared stats util. No second recording engine, and no reach into `ReplaySubject._buffer` (the fragility #13 tracks in the recorder).
- **Shared window/stats utils** (`chart-window.util`, `chart-stats.util`) — extracted from the recorder's inline logic so both engines derive an identical window and identical circular/scalar statistics from the same widget config. The recorder now consumes these utils with behaviour preserved.
- **Graceful degradation** — with no history provider (`getWidgetHistoryDisabled()`), the stream emits an "unavailable" sentinel and the widget shows a clean empty state instead of a blank chart. There is no recorder-style live-only fallback, matching the decision to make the history provider a required part of the marine stack.
- **A/B toggle + instrumentation** — per-widget `chartEngine` config field (`'recorder'` default | `'history'`), a `localStorage['skip.chartEngine']` global override to flip every data-chart at once for live testing, and opt-in perf instrumentation (`localStorage['skip.chartPerf']='1'`) logging backfill cost and live cadence/jitter/freshness. Both keys live in `config-storage.const.ts` under the `skip.` namespace.

## Known follow-ups (do not merge-close)

These are captured against this engine and should be weighed before it's promoted to default:

- **#85** — live tail doesn't re-backfill after a WebSocket reconnect (silent data gap).
- **#86** — the shared 120-point density target is too coarse for wide/long-duration charts.
- **#87** — backfill plots linear `:avg` for circular/direction paths; wrong at the 0/360° wrap (the live tail is already circular-correct, so the two halves currently disagree).
- **#13** — this engine sidesteps the `ReplaySubject._buffer` access, but #13 only closes when the recorder is retired (#64), not here.

## Testing

- `npm run lint` — clean.
- `npm run build:prod` — compiles clean against `main`.
- Chart specs (`chart-stats.util`, `chart-window.util`, `history-chart-stream.service`) — 17 tests pass, covering backfill mapping, the no-provider sentinel, and the live tail.
- Live rendering quality (smoothness, latency, server load vs. the recorder) is what a real-server A/B deployment measures next; that's the point of the toggle + perf flag.

Refs #64.

### DIFF
--- a/src/app/core/constants/config-storage.const.ts
+++ b/src/app/core/constants/config-storage.const.ts
@@ -28,3 +28,9 @@ export const SSO_REDIRECT_BUDGET_KEY = 'skip.ssoRedirectAttempts';
 
 /** Gesture-diagnostics debug flag key. */
 export const GESTURES_DEBUG_KEY = 'skip.gesturesDebug';
+
+/** Chart-engine A/B override flag key (#64 History-API prototype). */
+export const CHART_ENGINE_OVERRIDE_KEY = 'skip.chartEngine';
+
+/** Chart performance instrumentation flag key (#64 History-API prototype). */
+export const CHART_PERF_FLAG_KEY = 'skip.chartPerf';
--- a/src/app/core/interfaces/widgets-interface.ts
+++ b/src/app/core/interfaces/widgets-interface.ts
@@ -402,6 +402,8 @@ export interface IWidgetSvcConfig {
   timeScale?: string;
   /** Used by datachart & windtrend chart Widget to set period configuration */
   period?: number;
+  /** widget-data-chart data engine: 'recorder' (client-side sampler, default) or 'history' (SK History API backfill + live delta tail). #64 prototype toggle. */
+  chartEngine?: 'recorder' | 'history';
   /** Specifies if the chart should track against the average dataset instead of the value (default setting) */
   trackAgainstAverage?: boolean;
   /** Specifies which average data points property (1=avg, 2=ema or 3=dema) the chart dataset will be built with */
--- a/src/app/core/services/dataset-stream.service.ts
+++ b/src/app/core/services/dataset-stream.service.ts
@@ -219,7 +219,7 @@ export class DatasetStreamService implements OnDestroy {
    * then starts building the historicalData values, and pushes them to the Subject.
    *
    * This method handles the process that takes SK data and feeds the Subject. Clients/Observers,
-   * (widgets mostly), will use the getDatasetObservable() method to receive data from the Subject.
+   * (widgets mostly), will use the getDatasetBatchThenLiveObservable() method to receive data from the Subject.
    *
    * Concept: SK_path_values -> datasource -> (ReplaySubject) <- Widget observers
    *
--- a/src/app/core/services/dataset-stream.service.ts
+++ b/src/app/core/services/dataset-stream.service.ts
@@ -5,6 +5,7 @@ import { DataService, IPathUpdate } from './data.service';
 import { HistoryApiClientService, IHistoryValuesResponse, IHistoryValuesQueryParams } from './history-api-client.service';
 import { HistoryToChartMapperService } from './history-to-chart-mapper.service';
 import { UUID } from '../utils/uuid.util'
+import { resolveWindowMs, deriveDataSourceInfo } from '../utils/chart-window.util';
 import { cloneDeep } from 'lodash-es';
 import { NgGridStackWidget } from 'gridstack/dist/angular';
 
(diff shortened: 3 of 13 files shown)


mairas:
## Code review — multi-persona (correctness, reliability, adversarial, maintainability, testing, project-standards)

Six independent Opus reviewers, each in a fresh context, over `main...HEAD` (550 non-test LOC). No P0/P1 — consistent with a default-off, opt-in prototype. Lint, prod build, and the 3 chart specs (17 tests) are green. The findings below are what to resolve **before #64 promotes this engine to default**; several are new issues distinct from the already-filed #85/#86/#87.

Two trivial `safe_auto` fixes were already applied to the branch (see P3 table).

### P2 — Moderate

| # | File:line | Issue | Reviewer(s) | Conf | New? |
|---|-----------|-------|-------------|------|------|
| 1 | `history-chart-stream.service.ts:75` | Transient History-API failure is indistinguishable from no-provider (`getValues` returns `null` for both). On `null` the stream emits `HISTORY_UNAVAILABLE` and `complete()`s **without starting the live tail** — a brief blip during the one-shot backfill permanently blanks the chart until an unrelated rebuild. No retry. | reliability | 0.85 | ✅ (≠ #85 reconnect) |
| 2 | `chart-stats.util.ts:24` | The stats "extraction" is a **copy, not a shared source**. `circularMeanRad`/`circularMinMaxRad`/`normalize*` now exist in 3 places (util, `dataset-stream.service.ts:638`, `history-to-chart-mapper.service.ts:179`); only the History engine uses the util. The "both engines derive identical stats" design claim isn't actually realized — angle math must now be edited in 3 spots to stay consistent. | maintainability | 0.85 | ✅ |
| 3 | `history-chart-stream.service.ts:158` | **Seam discontinuity**: the `Value` series is the server `:avg` (smoothed) during backfill but the raw instantaneous delta once live. Same visual trace switches statistical meaning at the join (smooth → jagged); inconsistent with the recorder, whose `Value` is always instantaneous. | adversarial | 0.80 | ✅ |
| 4 | `history-chart-stream.service.ts:128` | History path-string builder is duplicated from `dataset-stream.service.ts:725` **and already diverged**: this uses `/^self\./`, the recorder uses `/^(vessels\.)?self\./`. A `vessels.self.`-prefixed path is normalized differently by the two engines. | maintainability | 0.80 | ✅ |
| 5 | `history-chart-stream.service.ts:121` | **Clock skew blanks the chart.** Backfill and live points are stamped with *server* time, but the chart.js `realtime` axis prunes/scrolls on the *client* `Date.now()`. A marine tablet without NTP (server GPS-synced) puts every point outside the visible window → successful backfill + live stream, yet a permanently blank chart. Recorder is immune (stamps `Date.now()`). | adversarial + correctness | 0.80 | ✅ |
| 6 | `widget-data-chart.component.ts:780` | Live tail treats angular (`rad`) paths as **scalar** unless the user set an explicit `datachartAngleRange` (default `null`). The recorder auto-detects circular domain from the path's base unit; this path hardcodes `baseUnit: ''` and defaults rad→scalar, so client-side live avg/min/max are computed with arithmetic (not circular) mean. Distinct from #87 (which is the server `:avg`). | correctness | 0.66 | ✅ (≠ #87) |
| 7 | `history-chart-stream.service.ts:112` | Live resampling defects: `merge(path$.take(1), timer+withLatestFrom)` over a `BehaviorSubject` **double-emits the first sample** (duplicate point, double-weighted stat), and because held resamples reuse the value's frozen server timestamp, a silent sensor makes the trace freeze at old-x then get pruned — the chart **blanks during a dropout** instead of holding the last value. | correctness + adversarial + reliability | 0.75 | ✅ |
| 8 | `history-chart-stream.service.spec.ts:91` | The backfill→live **buffer-seeding seam is never tested** — every test uses an empty backfill or unsubscribes before live, so the seed loop (`:81-84`) could be deleted and all tests pass. The PR's core contract (history seeds the window) is unproven. | testing | 0.88 | — |
| 9 | `history-chart-stream.service.ts:87` | The backfill **rejection/`catch` path** (→ `HISTORY_UNAVAILABLE`) is untested; specs only use `mockResolvedValue`. This is the real-deployment sad path. | testing | 0.90 | — |
| 10 | `history-chart-stream.service.ts:94` | **Disposal/teardown untested** — no test unsubscribes, so a live-subscription leak (delta pushing into a destroyed chart) would go uncaught. | testing | 0.85 | — |
| 11 | `history-chart-stream.service.ts:112` | The **timer resampling path never fires in tests** (no fake timers; the spec stubs `subscribePath` with a plain `Subject`, not the real `BehaviorSubject`). The mechanism that keeps the chart advancing between deltas is unverified — and this is exactly what hides #7. | testing + correctness + adversarial | 0.80 | — |
| 12 | `widget-data-chart.component.ts:111` | The widget integration (engine branching, `localStorage` override precedence, `historyUnavailable` signal) has **zero component tests**. | testing | 0.80 | — |

### P3 — Low

| # | File:line | Issue | Reviewer(s) | Conf |
|---|-----------|-------|-------------|------|
| 13 | `chart-perf.util.ts:1` | 120 lines of A/B instrumentation ship in the product bundle (runtime-gated → no-op in normal use, so no runtime cost), and its documented on-widget overlay is unbuilt (`snapshot()`/`IChartPerfSnapshot` are half-dead). Remove or build-time-gate before promoting #64. | maintainability | 0.75 |
| 14 | `history-chart-stream.service.ts:66` | "History **disabled**" (a user setting) is conflated with "no **provider**": both surface the "a history provider is required" message, and the flag is read once at subscribe so a late server-version resolution / runtime flip isn't re-evaluated (fresh-load race → stuck unavailable until rebuild). | reliability + maintainability + correctness | 0.70 |
| 15 | `dataset-stream.service.ts:222` | ✅ **Fixed in review** — doc referenced `getDatasetObservable()`, deleted in this PR; updated to `getDatasetBatchThenLiveObservable()`. | maintainability | 0.90 |
| 16 | `history-chart-stream.service.ts:113` | ✅ **Fixed in review** — `_tick` underscore-prefixed param renamed to `tick` per the repo convention (CLAUDE.md: don't underscore unused params). | project-standards | 0.60 |
| 17 | `chart-stats.util.spec.ts:34` | Circular-domain test asserts range only (`[0,2π)`), not the exact wrapped value; the `signed` domain and 3+-angle / empty-buffer edges are untested. | testing | 0.75 |

- **Vestigial `ema`/`dema` average options** (`widget-data-chart.component.ts:877`) resolve to no data in *both* engines; the new path inherits the dead config surface. Route: fold into the #64 cleanup or file a small follow-up. *(Not introduced here.)*
- The `:min`/`:max` columns requested by `fetchBackfill` but never read by the mapper are **already captured in #87** ("Minor, same code path") — not re-filed.

### Coverage

- Nothing suppressed (all findings ≥ 0.60 confidence).
- Cross-reviewer agreement boosted #5, #7, #11, #14 (independent reviewers converged).
- Version-bump policy confirmed **not applicable** to skip (no `VERSION`/`debian/changelog`, ships no `.deb`); no `any`/eslint-disable introduced; zoneless/signals/OnPush patterns correct.

### Verdict — Ready to merge as opt-in scaffolding; **not** ready to promote to default (#64)

The engine is safely default-off and green on lint/build/tests, so it can land as the reviewable base the cluster references. But #1, #3, #5, #6, #7 (correctness/reliability) and #2, #4, #13 (maintainability debt) should be resolved, and the test gaps (#8–#12) closed, **before #64 flips the default**. Findings #1, #3, #5, #6, #7 are new and distinct from #85/#86/#87 — candidates for tracking as additional #64 blockers.

<sub>🤖 Review synthesized from 6 independent persona reviewers via Claude Code.</sub>

mairas:
## Review follow-up — fixed here vs. filed

Per the routing decision, the cheap/local findings are fixed on this branch; the deeper ones are filed as #64 blockers.

**Fixed in this PR:**
- #15 — stale `getDatasetObservable()` doc reference → `getDatasetBatchThenLiveObservable()`.
- #16 — `_tick` unused param renamed to `tick` (repo convention).
- #4 — path normalization aligned to the recorder's `/^(vessels\.)?self\./` (removes the `vessels.self.`-prefix divergence).
- #17 — circular stats tests strengthened: exact wrapped values, a 0/360-wrap aggregation case, and the `signed` domain.
- #8, #9, #10 — added tests for the backfill→live buffer-seeding seam, the backfill-rejection path (→ `HISTORY_UNAVAILABLE`), and subscription teardown/disposal.

**Filed as #64 blockers** (deeper / entangled / design):
- #130 (transient-failure black hole), #131 (clock-domain + resampling), #132 (seam value semantics), #133 (share domain + circular stats — subsumes the angular-default correctness gap), #134 (component tests, perf-util removal, disabled-vs-unavailable messaging).

Lint, prod build, and the chart specs (now 23 tests across the three chart files) are green locally; CI runs the full suite.

Issue #131, opened 2026-07-02
History-chart engine: server-timestamped points on a client-clock realtime axis (clock skew + frozen-hold resampling)


mairas:
## Defect

Two related time-domain defects found during PR #127 review.

**1. Clock skew blanks the chart.** Both backfill and live points are stamped with **server** time (mapped History row time / `u.data.timestamp`), but the chart.js `realtime` axis scrolls and prunes on the **client** `Date.now()`. A marine tablet without NTP (SK server GPS-synced) offsets every point out of the visible window, so a successful backfill + live stream renders a permanently **blank** chart. The recorder is immune because it stamps `Date.now()` client-side.

**2. Frozen-hold resampling.** `startLive` uses `merge(path$.pipe(take(1)), timer(…)+withLatestFrom(path$))` over a `BehaviorSubject`. The first sample is emitted twice (duplicate point, double-weighted window stat), and held resamples reuse the value's frozen server timestamp — so when a sensor goes silent the trace freezes at an old x and is then pruned, blanking the chart during a dropout instead of holding the last value across to "now".

## Where

- `history-chart-stream.service.ts` — `startLive` (the `merge`/`timer`/`withLatestFrom` pipeline and the `timestamp = … instanceof Date ? … : Date.now()` stamping); `fetchBackfill` mapped point timestamps.

## Direction (not prescriptive)

Pick one clock domain for both the axis and the data — translate server timestamps into the client domain on ingest (offset measured at backfill time), or stamp held resamples with the tick time — and de-duplicate the first live value so `take(1)` + the first timer tick don't both emit it. The backfill and live tail must agree on the clock so the seam is continuous.

## Testing

Add a fake-timer regression test. The current spec stubs `subscribePath` with a plain `Subject` (not the real `BehaviorSubject`) and never advances the timer, so the resampling path — exactly where these bugs live — is unexercised.

## Scope

History engine only, default-OFF. **Blocks #64 promotion.**

PR #138, opened 2026-07-02
refactor(charts): unify angle-domain + circular stats across both chart engines (#133)

### LINKED ISSUE #133: History-chart engine: share the recorder's angle-domain resolution and circular stats instead of duplicating/omitting them
## Context

Found during PR #127 review. The window derivation was genuinely extracted into `chart-window.util` (both engines consume it), but two related pieces were **not** — the "both engines share stats" design is only half-realized.

## Defect

**1. Circular stats are a copy, not a shared source.** `circularMeanRad` / `circularMinMaxRad` / `normalize*` now live in **three** places — `chart-stats.util.ts` (new), `dataset-stream.service.ts` (recorder, still its own), and `history-to-chart-mapper.service.ts`. Only the History engine uses the new util; the recorder was never repointed. Angle math must be edited in three places to stay consistent, and can silently drift.

**2. Angle-domain resolution is omitted.** The recorder auto-detects the circular domain from the path's base unit (`resolveAngleDomain(path, unit, override)` returns `'signed'`/`'direction'` when `unit === 'rad'`). The History path derives the domain **only** from the explicit `datachartAngleRange` override and hardcodes `baseUnit: ''`, so any angular path *without* an explicit angle-range set is treated as **scalar** — its client-side live avg/min/max use arithmetic (not circular) mean. Distinct from #87 (that is the server-side `:avg`); this breaks the purely **client-side** live stats.

## Where

- `chart-stats.util.ts` vs `dataset-stream.service.ts` (`resolveAngleDomain`, `circularMeanRad`, `circularMinMaxRad`) vs `history-to-chart-mapper.service.ts`.
- `widget-data-chart.component.ts` — the history branch derives `domain` from `datachartAngleRange` only and sets `baseUnit: ''`.

Extract one shared angle-domain resolver + circular-stats util that both engines consume (delete the duplicates), and resolve the History path's domain from the path's base unit with the override as a tiebreak — mirroring the recorder.

> PR #127 already aligned the `self.` / `vessels.self.` path-normalization divergence between the two engines; the full stats/domain dedup belongs here.

## Scope

Squarely part of the #64 consolidation. **Blocks #64 promotion.**

mairas:
## What

Consolidates the duplicated angle math + domain resolution flagged in the #127 review (issue #133). One shared circular-stats source (`chart-stats.util`) and one shared angle-domain resolver (new `angle-domain.util`); the recorder, the mapper, and the History engine all consume them. **Foundational** for the rest of the #64 blocker chain (#131/#87/#132/#130/#85/#134 build on these shared utilities).

## Why

Circular math (`circularMean` / `circularMinMax` / normalize) existed in **three** places — `chart-stats.util`, the recorder, and the mapper — with two subtly different signed-normalizer formulas. Angle-domain resolution existed **twice** and incompatibly: the recorder auto-detects from the path's base unit + a signed-angle allowlist, but the History engine derived the domain from the explicit angle-range override only and hardcoded `baseUnit: ''`. So an angular (`rad`) path charted **without** an explicit override was treated as **scalar**, and its client-side live avg/min/max were computed with linear math — wrong across the 0/360° wrap.

## Changes

- **Recorder** (`DatasetStreamService`): delete the private circular helpers + `resolveAngleDomain` + allowlist; `updateDataset` delegates to `computeWindowStats`. Behaviour preserved for scalar and circular domains.
- **Mapper** (`HistoryToChartMapperService`): drop its private circular copies and the redundant `unit` option — `domain` is now authoritative.
- **History engine**: resolve the domain from the path's real base unit (`DataService.getPathUnitType`) via the shared resolver. The widget now passes only the raw override, not a synthesized `unit`/`domain`.
- **Signed normalizer** standardized on the `atan2` form `(-π, π]`; exactly 180° now maps to `+π` on both engines (was `-π` on the recorder).

## Behaviour preservation

`updateDataset` → `computeWindowStats` is equivalent for the guaranteed inputs (buffer length ≥ 1, smoothing period ≥ 1). The full recorder spec passes locally, including the seeded-history recompute and circular-domain cases. The only intended change is the ±π boundary above, pinned with a test.

## Testing

`npm run lint` ✓ · `npm run build:prod` ✓ · 47 tests across the 6 affected specs pass locally (recorder behaviour-preservation suite, mapper, engine, the new `angle-domain.util` spec, a `#6`-style regression proving a rad path with no override now aggregates circularly, and the ±π boundary). CI runs the full matrix.

Closes #133.

### DIFF
--- a/src/app/core/components/widget-history-chart-dialog/widget-history-chart-dialog.component.ts
+++ b/src/app/core/components/widget-history-chart-dialog/widget-history-chart-dialog.component.ts
@@ -207,7 +207,6 @@ export class WidgetHistoryChartDialogComponent implements OnInit, AfterViewInit,
       }
 
       const datapoints = this.historyMapper.mapValuesToChartDatapoints(response, {
-        unit: 'number',
         domain: 'scalar'
       });
 
--- a/src/app/core/services/dataset-stream.service.spec.ts
+++ b/src/app/core/services/dataset-stream.service.spec.ts
@@ -312,7 +312,7 @@ describe('DatasetStreamService', () => {
         };
 
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        const datapoints = (service as any).convertHistoryToDatapoints(response, 'number', 'scalar');
+        const datapoints = (service as any).convertHistoryToDatapoints(response, 'scalar');
 
         expect(datapoints.length).toBe(3);
         expect(datapoints[0].data.lastAverage).toBeNull();
@@ -345,7 +345,7 @@ describe('DatasetStreamService', () => {
         };
 
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        const datapoints = (service as any).convertHistoryToDatapoints(response, 'number', 'scalar');
+        const datapoints = (service as any).convertHistoryToDatapoints(response, 'scalar');
 
         expect(datapoints.length).toBe(2);
         expect(datapoints[0].data.value).toBe(0.5);
@@ -370,7 +370,7 @@ describe('DatasetStreamService', () => {
         };
 
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        const datapoints = (service as any).convertHistoryToDatapoints(response, 'rad', 'direction');
+        const datapoints = (service as any).convertHistoryToDatapoints(response, 'direction');
 
         expect(datapoints.length).toBe(3);
         expect(datapoints[0].data.lastAverage).toBeNull();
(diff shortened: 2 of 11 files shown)


mairas:
## Code review — 4 independent personas (correctness, adversarial, maintainability, testing)

Fresh-context Opus reviewers over `main...HEAD`. Focus for a behavior-preserving refactor: does the shared code truly reproduce the deleted duplicated code? All findings below have been **addressed in the branch** (force-pushed).

### P1 — fixed

| # | Finding | Reviewers | Fix |
|---|---------|-----------|-----|
| 1 | **Override silently dropped when metadata isn't loaded.** `resolveAngleDomain` checked `baseUnit !== 'rad'` before the override, so a rad path whose producer is idle this session (`getPathUnitType` → null — common when viewing *history*) reverted an explicit `signed`/`direction` chart to **linear** stats (~180° error on wind angle). Not behaviour-preserving vs the old History engine. | correctness + adversarial (converged) | `resolveAngleDomain` now honours an explicit override when the unit is rad **or unknown**, and stays scalar only for a *positively* non-rad unit. Added specs for null-unit + override, non-rad + override, and unknown + no-override. |

### P2/P3 — fixed

| # | Finding | Reviewer | Fix |
|---|---------|----------|-----|
| 2 | Recorder spec proved the *scalar* `updateDataset` delegation but never the *circular* one — a dropped domain arg (the #6 bug, recorder side) would pass every test. | testing | Added a recorder test asserting `updateDataset(ds, 'direction')` computes the circular mean (~0°) over a wrap-straddling buffer. |
| 3 | Mapper had no `signed`-domain test after the branch selector (`unit` → `domain`) and normalizer (`mod` → `atan2`) changed. | testing | Added a mapper `signed`-domain test crossing ±π (asserts +π at the boundary). |
| 4 | `SIGNED_ANGLE_PATHS` exported with no external consumer. | maintainability | Made it module-private. |
| 5 | Dead `lastAngleAverage` field on `IDatasetServiceDatapoint`. | maintainability | Removed. |

### Verified non-issues (residual risk, no change)

- **±π boundary flip** (recorder's old `mod` gave −π, shared `atan2` gives +π): a 5M-sample scan found the *only* divergence is at exact odd multiples of π; worst non-π difference is 4.9e-15. +π is correct for the documented `(-π, π]`, both engines now agree, and it's cosmetic/measure-zero for continuous data. Pinned by a test.
- **`updateDataset` → `computeWindowStats` equivalence** holds for every reachable input (buffer ≥ 1, smoothingPeriod ≥ 1, both floored/guaranteed); empty-buffer and `sp < 1` are unreachable.
- **SMA slice**, `value` field, and the mapper's `circularMeanRad` vs old inline `atan2` accumulation are bit-identical.

### Noted (not blocking)

- Domain resolution now reads live `getPathUnitType` at subscription while the recorder captures `baseUnit` at dataset-create; same path, so they agree in practice, but they could momentarily differ if metadata arrives between the two. Pre-existing timing property; the override fix removes the worst consequence.

### Verdict — ready once CI is green

The refactor is behaviour-preserving for every reachable input; the one genuine regression (override + missing metadata) is fixed with coverage.

<sub>🤖 Review synthesized from 4 independent persona reviewers via Claude Code.</sub>

PR #139, opened 2026-07-02
fix(charts): render the History live tail on the client clock with a staleness gap (#131)

### LINKED ISSUE #131: History-chart engine: server-timestamped points on a client-clock realtime axis (clock skew + frozen-hold resampling)

mairas:
## What

Fixes #131: the History engine's live tail rendered on the wrong clock, so under clock skew the whole chart blanked, and a frozen sensor froze the trace until it was pruned.

> Stacked on #133 (base branch `feat/133-shared-angle-domain-stats`). GitHub will retarget to `main` when #133 merges.

## Why

Backfill and live points were stamped with the SK **server** timestamp, but chart.js's `realtime` axis scrolls and prunes on the **client** `Date.now()`. On a marine tablet without NTP, every server-stamped point falls outside the visible window → blank chart (the recorder is immune; it stamps `Date.now()`). A silent sensor's held resamples reused the frozen server timestamp, so the trace stopped advancing in x and was pruned.

## Changes

- **`startLive`**: stamp each emission with the client clock at emit time, so held resamples advance in x and skew no longer hides the series.
- Share one path subscription (`shareReplay`) across the freshness tracker, the immediate first value, and the resampler.
- **Staleness gap**: after 3 missed sample intervals with no fresh delta, break the trace once (emit a gap marker) so a dropout reads as a gap, then stop advancing until data resumes.
- **`fetchBackfill`**: shift server-time history timestamps into the client clock via the response `range.to` offset, so backfill lines up with the live tail.

## Design note

Per the decision on the silent-sensor question, this ships **gap-after-staleness** (not flat-line hold). The threshold is `3 × sampleTime` (a single module constant, tunable) — a heuristic that assumes the sensor updates near the chart cadence; a genuinely slow sensor on a fine chart could show gaps, so the constant may want tuning against real hardware.

## Testing

`npm run lint` ✓ · `npm run build:prod` ✓ · 12 engine specs pass, incl. the new server→client offset test and a fake-timer test proving the held value advances in x then breaks into a gap once stale.

Closes #131.

### DIFF
--- a/src/app/core/services/history-chart-stream.service.spec.ts
+++ b/src/app/core/services/history-chart-stream.service.spec.ts
@@ -99,12 +99,14 @@ describe('HistoryChartStreamService', () => {
     await Promise.resolve();
     expect(emissions[0]).toEqual([]); // empty batch first
 
-    // First live value routes through take(1) — no timer needed.
-    path$.next({ data: { value: 5, timestamp: new Date(1000) }, state: 'normal' });
+    // First live value routes through take(1) — no timer needed. Fresh source timestamp (now).
+    const before = Date.now();
+    path$.next({ data: { value: 5, timestamp: new Date() }, state: 'normal' });
 
     expect(emissions.length).toBe(2);
     const point = emissions[1] as IDatasetServiceDatapoint;
-    expect(point.timestamp).toBe(1000);
+    // #131: stamped with the client clock at emit time, not the value's raw source timestamp.
+    expect(point.timestamp).toBeGreaterThanOrEqual(before);
     expect(point.data.value).toBe(5);
     expect(point.data.lastAverage).toBe(5);
     expect(point.data.lastMinimum).toBe(5);
@@ -124,7 +126,7 @@ describe('HistoryChartStreamService', () => {
     await Promise.resolve();
 
     // The live value aggregates over the seeded buffer [10, 20], not a fresh empty one.
-    path$.next({ data: { value: 30, timestamp: new Date(3000) }, state: 'normal' });
+    path$.next({ data: { value: 30, timestamp: new Date() }, state: 'normal' });
 
     const point = emissions[1] as IDatasetServiceDatapoint;
     expect(point.data.value).toBe(30);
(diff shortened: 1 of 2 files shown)


mairas:
## Code review — 3 personas (correctness, reliability, adversarial) + fixes applied

All three converged on the staleness gap being keyed to the **chart cadence** (`3 × sampleTime`), not the **source rate** — so a slow-but-alive sensor false-gaps and a day-window masks real dropouts. The adversarial pass added three more. All fixed by reworking staleness around the **value's own source-timestamp age** (force-pushed).

| # | Finding | Reviewers | Fix |
|---|---------|-----------|-----|
| 1 | **Staleness keyed to render cadence, not source rate** (P1). `3 × sampleTime` false-gaps healthy slow sensors on short windows; masks dropouts on long windows. | correctness + reliability + adversarial (converged) | Staleness now = the value's source-timestamp age exceeding `max(floor, 3 × observed source interval)`, with a generous bootstrap default until the interval is learned. Decoupled from `sampleTime`. |
| 2 | **Stale cached seed drawn as fresh** (P2). `subscribePath` replays the last cached value on subscribe; a long-dead sensor drew a fake "live" point at *now*. | adversarial | Same source-age check: the replayed value's old timestamp → immediate gap, not a fresh point. |
| 3 | **Missing `range.to` → offset 0 → skew reintroduced** (P2). | adversarial | Offset now falls back to the newest sample's server timestamp when `range.to` is absent/unparseable. |
| 4 | **Seam dropout drawn as a connecting line** (P2). Data ending before *now* connected straight into resumed live data. | adversarial | `startLive` breaks the trace at the seam when the newest backfill point is well before *now*. |
| 5 | Wall-clock step (NTP) can invert the seam / spurious-gap (P3). | adversarial | **Noted, not fixed** — rare, self-heals as backfill ages out; a `performance.now()` monotonic clock is a larger change better weighed with #85/#130. |

### Verified sound (no change)

Offset sign + NaN fallback, `take(1)` first-value freshness, `sub.unsubscribe()` tearing down both children (no leak via `shareReplay refCount`), and the NaN gap marker being guarded downstream by `Number.isFinite` (title/annotations stable; chart.js breaks the line).

### New tests

Source stops → one gap; **slow-but-alive (5s interval, 500ms cadence) → zero gaps** (the P1 regression); **stale cached seed → gap not a fresh point**; plus the existing offset test.

### Noted for follow-up (not this PR)

The backfill HTTP request isn't cancelled on teardown (pre-existing; the `disposed` guard prevents corruption, it just wastes a request). Better addressed with the fetch-path changes in #85/#130.

<sub>🤖 Review synthesized from 3 independent persona reviewers; fixes applied and re-verified.</sub>

PR #119, opened 2026-07-02
Cap gauge needle animation to a short fixed window (replicates part of mxtommy/Kip#1101)

mairas:
## Motivation

The ng-compass and ng-radial gauges derived `animationDuration` from the path sample time minus a small constant (`sampleTime - 50` / `sampleTime - 25`), so the needle animation nearly filled each sample interval (~95% duty cycle). A gauge fed changing data repainted its needle layer almost continuously through its own rAF loop, pinning gauge-heavy dashboards on low-power marine hardware. Upstream mxtommy/Kip#1101 caps the animation to a short fixed window so each gauge idles most of the interval.

## Approach

Cherry-picked from mxtommy/Kip#1101 with `git cherry-pick -x`, preserving upstream authorship. Note: the upstream PR is still **open**; if it changes before merge there, reconcile per the issue notes.

- `5c790b22` — adds `src/app/core/utils/gauge-animation.util.ts` (`GAUGE_ANIMATION_MS = 120`, clamped to never exceed the sample interval, 120 ms fallback for non-finite/<= 0 sample times) plus a spec, and swaps the call sites in `widget-gauge-ng-compass.component.ts` and `widget-gauge-ng-radial.component.ts`. Applied cleanly with no conflicts.

## Deliberate divergence from a faithful replica

- **ng-linear also capped** (separate fork-authored commit): `widget-gauge-ng-linear.component.ts` carries the identical `sampleTime - 25` pattern that upstream left unfixed, so the same `gaugeAnimationDurationMs()` cap is applied there as well.

## Verification

- `npx ng test --include='src/app/core/utils/gauge-animation.util.spec.ts'` — 1 test file, 3/3 tests passed
- `npm run lint` — all files pass linting
- `npm run build:prod` — bundle generation complete (only the pre-existing `piexifjs` non-ESM warning)

Closes #117

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #125, opened 2026-07-02
Throttle + value-dedupe own-ship updates (replicates part of mxtommy/Kip#1101)

mairas:
## Motivation

Part of adopting the AIS radar freeze-fix series from mxtommy/Kip#1101 (still **open** upstream; reconcile if it changes before merge). Upstream measured a −63% reduction in handler wait time on the AIS radar from cutting per-fix churn.

`applyOwnShipUpdate` set the `_ownShip` signal with a fresh object on every `self.navigation.*` delta, unthrottled. The radar's render effect guards re-renders by `ownShipRef` reference equality, which a fresh object per delta always defeats — so a streaming compass/GPS drove a full O(targets) re-render on every fix even when nothing changed.

## Approach

Cherry-picked from mxtommy/Kip#1101 with `git cherry-pick -x`, preserving upstream authorship and commit order:

- `b75a2315` — Add failing test for own-ship update throttling (RED)
- `b63859d4` — Throttle + dedupe own-ship updates (GREEN)

Own-ship updates now accumulate into a `pendingOwnShip` buffer and fire `ownShipDirty$`, throttled (250 ms, leading+trailing, `takeUntilDestroyed`) into `flushOwnShip()`, which sets the `_ownShip` signal only when `ownShipChanged()` detects a field-level difference (headingTrue, courseOverGroundTrue, speedOverGround, position lat/lon). Unchanged fixes keep the same object reference, so the radar's reference-equality guard short-circuits.

## Adaptations

None. Both commits auto-merged cleanly onto the eviction branch (this matches upstream's own commit order); the spec's new describe block landed at end of file after the fork-added eviction tests, and the service hunks applied at their intended locations. No fork-authored commits were needed.

## Stacking

Stacked on the target-eviction PR #120 (both change `ais-processing.service.ts`); base is `replicate/pr-1101-ais-target-eviction`. Retarget to `main` after #120 merges, and merge this PR only after it.

## Verification

- `npx ng test --include='src/app/core/services/ais-processing.service.spec.ts'` — 1 file, **17/17 tests passed** (10 dispatch + 5 eviction + 2 new throttle tests)
- `npm run lint` — all files pass
- `npm run build:prod` — succeeds (pre-existing `piexifjs` CommonJS warning only)

Closes #118

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #140, opened 2026-07-02
Adopt the Kip#1101 freeze-measurement harness, ported to Skip

### LINKED ISSUE #129: Adopt Kip#1101's perf-harness freeze-measurement tooling
Follow-up to the mxtommy/Kip#1101 adoption series (#114–#118). That PR is roughly half measurement tooling: a self-contained `perf-harness/` directory we deliberately excluded from the five fix replicas (#119–#122, #125). This issue tracks adopting it as its own change.

## Motivation

Skip has no performance or freeze measurement tooling at all — no benchmark directory, no browser-driving deps, unit tests only. We just adopted five freeze fixes entirely on upstream's measurements; we cannot produce before/after numbers on our own build, and future perf-sensitive fork work (chart/history engine, widget rendering) has no regression net. The harness's CPU-throttled headless-Chromium setup (10× throttle emulating low-power hardware) maps directly onto Skip's Pi-class HaLOS deployment target.

## What upstream provides

- `probe.js` — injected via Playwright before page load (app code stays clean): longest task / total blocking time, 16 ms timer lateness (queued-handler wait, the "can't exit fullscreen" proxy), rAF frame gaps, heap slope.
- Mock Signal K server (same-origin, configurable delta rates, path counts, AIS count/churn).
- `run.mjs` + `lib/build-serve.mjs` — builds any git ref's production bundle in an isolated worktree; named scenarios (`resize-storm`, `delta-storm-30x10`, `reconnect-backlog`, `ais-radar-150`, `gauges-16`, `ais-growth-churn`) run K times, reported as median/p95 with raw runs kept.
- Deterministic screenshot + click-verification tooling.
- Own `package.json`; never touches the app's dependencies.

## Adoption plan

Cherry-pick the harness commits from the already-fetched `upstream-pr-1101` branch — `caa13a41` (harness) and `e631654a` (screenshot/click tooling) — then port in fork-authored commits:

- [ ] Drop upstream's `results/` baselines and reports (theirs, not ours)
- [ ] Serving path: harness assumes Kip's root serving; Skip serves at `/@halos-org/skip/` (baseHref, manifest scope)
- [ ] Auth: harness bootstraps Kip into anonymous local mode via localStorage config; Skip is session-cookie SSO-only, so the mock server must fake `GET /skServer/loginStatus` (and whatever else `AuthenticationService`/`SsoRedirectService` probe at startup)
- [ ] Config bootstrap: port `lib/kip-config.mjs` dashboard/config builders to Skip's config namespace and current config version (#84/#105 moved these)
- [ ] Capture Skip baselines: pre-fix main (`11f3fbd0`) vs main after #119–#122/#125 merge — `run.mjs --branch <ref>` builds from any ref, so the pre-fix baseline can be captured retroactively
- [ ] Decide CI stance: local-only tool at first; a CI freeze-budget gate is a separate later decision

## Verification

- Harness runs end-to-end against a Skip production build: all scenarios complete, results JSON written with sane values
- The captured before/after baselines show the adopted fixes' effect (directionally consistent with upstream's numbers)
- No changes to the app's `package.json` or runtime code

## Notes

Do this after the five fix PRs land, so the first real use is measuring them. Upstream #1101 is still open; if the harness changes before upstream merge, reconcile. `verify-click.mjs`'s original role (radar hit-testing) is already covered in Skip by the unit spec added in PR #122 — keep the tool anyway, it generalizes.

mairas:
## Motivation

Skip adopted five freeze fixes from mxtommy/Kip#1101 (#119, #120, #121, #122, #135) entirely on upstream's measurements: we had no way to produce before/after numbers on our own build, and future perf-sensitive work (chart/history engine, widget rendering) had no regression net. Kip#1101's other half is a self-contained measurement harness; #129 tracks adopting it. Its 10x CPU-throttled headless-Chromium setup maps directly onto Skip's Pi-class HaLOS target. First real use: measuring the five fixes we already merged (baselines below).

## Approach

Three tiers, by how much of the upstream code survives contact with Skip:

**Cherry-picked verbatim** (`-x`, upstream authorship preserved):
- `caa13a41` — harness core: probe, mock Signal K server, scenario runner, cross-ref worktree builder.
- `e631654a` — deterministic AIS-radar screenshot tooling.

**Adopted with attribution** (fork commits citing the upstream SHA, because the source commits carry material we must not take):
- `report.mjs` + the `npm ci` → `npm install` build change from `fb391574`, *excluding* upstream's `results/` baselines — those measure Kip builds and are meaningless for Skip.
- `verify-click.mjs` from `dd5b59f2` — the rest of that commit is app code Skip already replicated in #122.

**Fork-ported** (upstream's approach is dead on Skip):
- Upstream bootstraps Kip into anonymous localStorage-config mode. Skip is session-cookie SSO-only and persists config server-side, so the mock now serves the whole session/config surface: `GET /skServer/loginStatus` (logged-in admin), `applicationData` GET/POST (the scenario's full `IConfig` document; boot JSON-Patch/autosave absorbed with 200 so no error snackbar lands in the measurement), the `/plugins` surface reporting the history-series plugin disabled (suppresses the reconcile POST ~750 ms after boot), and `server.version` 2.24.0 (Skip gates widget history on >= 2.22.1; upstream's 2.3.0 silently fails that semver check). Only `skip.connectionConfig` is seeded in localStorage — the sole key Skip reads at boot.
- Serving base `/@halos-org/skip/` (Skip's baseHref) in all five sites; widget factories fixed for Skip's schema (radial gauge `displayScale`; upstream's top-level `minValue`/`maxValue` are dead and silently render a 0–100 scale).
- Determinism hardening: the gstatic reachability probe is answered locally (fulfilled 204 — aborting it triggers retry cadence inside the window), boot *asserts* the exact rendered widget count (a mis-seeded config still boots a plausible empty dashboard), and scenario rate changes restart live WS timers (Skip opens its socket once at APP_INITIALIZER, so upstream's connect-time rate snapshot streamed everything at boot-default 10 Hz).
- Chunked runs accumulate into `results/<label>.json` per scenario, so long suites survive command time ceilings and mid-suite crashes.
- `report.mjs` additionally prints each label's median probe-window length per scenario: the window self-extends while the throttled main thread is busy, so raw window sums (blocking time) drift with it — see the ais-growth-churn honesty note below.

`perf-harness/` is self-contained: own `package.json`, no changes to the app's dependencies or any `src/` runtime code.

## Baselines: 11f3fbd0 (pre-fixes) vs main 2e358f17

Full suite, 4 repeats, 10x CPU throttle, Chrome 149. Medians; full tables in `perf-harness/results/report-pre-freeze-fixes-vs-main-freeze-fixes.md`, raw runs in the results JSONs. The after-ref also contains #127 (history-chart prototype), see diagnostics.

| Scenario | Blocking time (ms) | Max handler wait (ms) | Verdict |
|---|--:|--:|---|
| resize-storm | 1900 → 112 (**-94%**) | 324 → 134 (-59%) | The headline: #121's coalescing kills the fullscreen enter/exit freeze. Longest task 177 → 73 ms, dropped frames 37 → 21. Windows are tight (±2%), so this survives normalization. |
| gauges-16 | 1 → 0 | 48 → 39 (-19%) | Small, consistent improvement; Skip's pre-fix gauge cost was already near-zero at this throttle, so #119 has little left to show here. |
| ais-growth-churn | 27231 → 25163 | 8696 → 6447 | **Null-to-weak.** The raw -8% blocking is a window-length artifact and the -26% max-wait sits inside same-code variance; only p95 wait (-24%) has directional support. See diagnostics. Churn remains heavy — an open cost. |
| ais-radar-150 | 7826 → 8398 | 1406 → 1511 | **Null result.** See diagnostics — apparent deltas are same-build noise. |
| idle-numeric-24 | 0 → 0 | 4 → 7 | Null, as expected. |
| delta-storm-30x10 | 0 → 0 | 9 → 4 | Null — no delta-path fix was among the five replicas; matches upstream's finding that ingestion isn't the freeze. |
| reconnect-backlog | 0 → 4 | 58 → 61 | Null (single ~50 ms parse task on both refs). |

**Honesty notes / diagnostics** (committed as `results/diag-*.json`):
- ais-growth-churn first looked directionally better (-8% blocking, -26% max wait). Neither claim survives scrutiny. Blocking is a raw sum over the probe window, and the pre-fix churn windows ran long (median 33.8 s vs 31.1 s; the window self-extends while the main thread is busy) — normalized to a common 30 s window the medians are 24809 vs 24833 (+0.1%), i.e. no change. The -26% max-wait sits inside same-code run-to-run variance: `diag-pre127` (63cf071e, effectively identical code to main for this scenario) has churn max-wait median 3539 vs main's 6447, a same-code swing larger than the claimed improvement. Only p95 wait (6658 → 5091, -24%) reproduces directionally. `report.mjs` now prints median window length per scenario so this artifact class is visible in every future report.
- ais-radar-150 first looked like a regression (+47% longest task, +30% heap). Two diagnostics say otherwise: `diag-pre127` (63cf071e = the five fixes without #127) shows the same numbers as main, ruling out #127; `diag-pre-variance` (the *same pre-fix build* measured again) spans longest-task 610–1071 ms and blocking 7.8–9.7 s — the entire pre/post delta sits inside same-build run-to-run spread. The baseline pre-fix run was a low draw. Verdict: no measurable change on this scenario, high variance; #122/#135 target rotation-only and own-ship-driven frames, and this scenario re-renders 150 moving targets at 4 Hz regardless.
- Heap-growth medians on the AIS scenarios (9 → 18 MB churn) are noise-dominated: same-ref raw spreads run 5–21 MB (no forced GC; usedJSHeapSize includes uncollected garbage). No leak signal either way.

## Verification

- `node smoke.mjs` — probe detects a synthetic 400 ms block.
- Full 7-scenario suite ran twice (both refs) with zero boot-assertion failures; every rep rendered the exact expected widget count; delta counts match scenario rate semantics (16 deltas at 2 Hz / 8 s).
- Boot-path diagnostic: zero 404s, zero SPA-fallback hits on `/signalk|/plugins|/skServer`, zero console errors; radar screenshot and click-verification tools pass.
- `npm run lint` passes (flat config only matches `**/*.ts`/`**/*.html`; harness `.mjs` is out of scope by design); `node --check` on all harness modules.
- App build untouched: no `src/` or app `package.json` changes anywhere in the branch.

## Documentation

- `CLAUDE.md` gains a "Performance and freeze measurement (perf-harness/)" section (the auto-loaded surface): what the harness is, when to reach for it, the three commands, the traps (per-label result accumulation, probe-window drift, heap noise, AIS variance, boot-assertion semantics, nested worktrees, `CHROME_BIN`), and the committed baseline labels with their refs.
- `perf-harness/README.md` is the full operator guide: architecture and mock-auth surface, scenario table, metric interpretation (which metric answers which question, which claims the data cannot support), an add-a-scenario recipe, and a troubleshooting table (boot assertion, degraded boot, HTML fallthrough, report ENOENT, port in use, Chrome path).

Closes #129

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -19,6 +19,30 @@ Skip — an Angular 21 (zoneless, signals, new control flow) Signal K marine mul
 
 Node: the app builds on Node 20+; the kip-plugin's built-in history provider needs Node ≥22.5 (`node:sqlite`). CI runs the matrix 20/22/24.
 
+## Performance and freeze measurement (perf-harness/)
+
+Self-contained freeze/jank harness (own `package.json`, never touches app deps): builds the **real production bundle**, drives it in headless Chromium at **10× CPU throttle** (Pi-class HaLOS target) against a **mock Signal K server** on one origin. Use it for perf-sensitive changes to widgets, rendering, or the data pipeline, and for before/after numbers when replicating upstream perf work. Full operator docs: `perf-harness/README.md`.
+
+The three commands (from `perf-harness/`, after `npm install` there):
+
+```bash
+node run.mjs --public ../public --label dev --scenarios resize-storm --repeats 2  # quick, prebuilt ../public
+node run.mjs --branch main --label main       # full 7-scenario suite against any git ref
+node report.mjs --a pre-freeze-fixes --b main-freeze-fixes   # compare two labels
+```
+
+Traps that cost real time:
+
+- **Results accumulate per label**: scenarios merge into any existing `results/<label>.json`. Reusing a label for a different ref keeps the old ref's scenario entries under the new `branch` field — delete the file first.
+- **Blocking time and delta counts are raw sums over the probe window**, and the window self-extends while the throttled main thread is busy. Check the report's window-ms context row before claiming a regression from a small delta.
+- **Heap metrics are noise-dominated** (no forced GC; same-ref spreads of 5–21 MB observed). No leak claims from them without a sustained slope across repeats.
+- **AIS scenarios have large run-to-run variance** (same-build blocking spread 7.8–9.7 s observed) — never conclude from a single rep; compare against a same-build re-run (`results/diag-*.json` are committed examples).
+- **A boot-assertion failure** (`boot check failed: N widget-host2 rendered, expected M`) means the seeded config is wrong for that build (version or widget-schema drift) — not a flake. A mis-seeded config boots a plausible empty dashboard; the assert exists to catch exactly that.
+- `--branch` builds register **nested git worktrees** under `perf-harness/worktrees/`; `git worktree prune` alone does not deregister them — `git worktree remove` each one, or delete `perf-harness/worktrees/` and then prune.
+- Chrome path defaults to the macOS app bundle; set `CHROME_BIN` on other platforms.
+
+The mock serves Skip's full session/config surface (`loginStatus`, `applicationData`, `/plugins`) because Skip is session-SSO-only with server-side config — a localStorage-only bootstrap boots a degraded app; see the README. Committed baselines: `pre-freeze-fixes` (11f3fbd0, before the #119/#120/#121/#122/#135 freeze fixes) and `main-freeze-fixes` (2e358f17).
+
 ## Testing reality (read before touching specs)
 
 - The runner is **vitest** (`vitest.config.ts`, `environment: jsdom`) driven by `@angular/build:unit-test`. `src/test.ts` is the setup file and **monkey-patches `TestBed.configureTestingModule`** to inject a large set of global stubs/providers (SettingsService, AuthenticationService, SignalKConnectionService, ConnectionStateMachine, MatDialog refs, the widget host directives, etc.). Specs override these by listing their own local providers (local wins; globals are prepended).
--- a/perf-harness/.gitignore
+++ b/perf-harness/.gitignore
@@ -1,0 +1,3 @@
+node_modules/
+worktrees/
+results/*.raw.json
--- a/perf-harness/.gitignore
+++ b/perf-harness/.gitignore
@@ -1,3 +1,4 @@
 node_modules/
 worktrees/
 results/*.raw.json
+results/shots/
--- a/perf-harness/.gitignore
+++ b/perf-harness/.gitignore
@@ -1,4 +1,5 @@
 node_modules/
 worktrees/
 results/*.raw.json
+results-*.log
 results/shots/
--- a/perf-harness/README.md
+++ b/perf-harness/README.md
@@ -1,0 +1,57 @@
+# KIP freeze-audit measurement harness
+
+Reproducible, honest before/after measurement of the main-thread **freezes** users
+report ("randomly unresponsive; can't exit fullscreen"). Self-contained — its deps
+live in `perf-harness/package.json` and never touch the app's `package.json`.
+
+## What it measures (the symptom, not proxies)
+
+The same probe (`probe.js`) is injected into every build under test via Playwright
+`addInitScript`, so measurement code is identical across branches:
+
+| Metric | Meaning |
+|---|---|
+| `longTaskMaxMs` / `blockingTimeMs` | Longest single main-thread task, and total blocking time (Σ max(0, dur−50)). The browser's own "the main thread was stuck" signal. |
+| `taskLatencyMaxMs` / `p95` | A 16 ms self-rescheduling timer's lateness = **how long a queued handler (e.g. the exit-fullscreen keydown) waits**. The honest "frozen" proxy. |
+| `frameMaxGapMs` / `framesDropped` | rAF inter-frame gaps → dropped frames. |
+| `heapGrowthMB` / `heapSlopeKBps` | `usedJSHeapSize` slope over time (unbounded-growth findings). |
+
+## How it works
+
+`run.mjs` builds a branch's **production** bundle in an isolated git worktree
+(`lib/build-serve.mjs`), serves it plus a **mock Signal K server** on one origin
+(`lib/server.mjs` — no CORS), injects the probe + a localStorage config that puts
+KIP in anonymous local mode (`lib/kip-config.mjs`), throttles CPU via CDP to emulate
+low-power marine hardware, runs each scenario K times, and writes
+`results/<label>.json` (raw + median/p95).
+
+## Scenarios → audit findings (`scenarios.mjs`)
+
+| Scenario | Reproduces |
+|---|---|
+| `resize-storm` | Rank 1 — shared ResizeObserver reallocates every canvas in one task (the fullscreen enter/exit storm). |
+| `delta-storm-30x10` / `reconnect-backlog` | Rank 2 — delta ingestion / reconnect snapshot fan-out. |
+| `ais-radar-150` | Ranks 4/5 — AIS radar full re-render loops. |
+| `gauges-16` | Rank 7 — ng-canvas-gauge animation duty cycle. |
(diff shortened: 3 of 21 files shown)


mairas:
## Review synthesis (3 independent lenses)

Three fresh-context reviewers examined the branch: **port fidelity** (every ported/adopted file diffed against upstream-pr-1101/cited SHAs, divergences checked for deliberateness and correctness against Skip's app source), **measurement honesty** (re-ran 3 scenarios on both committed builds, deliberately mis-seeded a config to test the boot assertion, audited network egress, cross-checked prose against committed JSONs), and **repo hygiene** (diff scope, lint/test integrity, README command accuracy, stray artifacts, commit hygiene). Verdicts: PASS / publishable-with-one-correction / PASS.

### Findings and resolution

**Major — ais-growth-churn improvement claims not supported by the branch's own data** (measurement honesty). Verified against the committed JSONs and confirmed:
- The published -8% blocking delta is a measurement-window artifact: blocking is a raw sum over a probe window that self-extends under main-thread load, and the pre-fix churn windows ran 31.9–35.0 s vs 30.2–32.0 s post. Normalized to 30 s, medians are 24809 vs 24833 (+0.1%).
- The -26% max-wait sits inside same-code variance: committed `diag-pre127` (effectively identical code to main for this scenario) has churn max-wait median 3539 vs main's 6447.
- Only p95 wait (-24%) has directional support.

**Fixed** in b7666ab3 + PR body edit: the churn row is downgraded to **null-to-weak** with a full honesty note (window medians, normalization, diag-pre127 variance), and `report.mjs` now prints each label's median probe-window length per scenario so window drift is visible next to the raw sums — the committed baseline report is regenerated (churn context row now shows 33771 vs 31073 ms; resize-storm windows are tight at ±2%, so the -94% headline is unaffected and survives normalization).

**Minor — blockingTimeMs un-normalized over a load-dependent window; report.mjs never surfaced windowMs** (measurement honesty). Same root cause as the major; fixed in b7666ab3 via the lighter of the two remedies the reviewer proposed (surface median windowMs rather than convert blocking to a rate — a rate would silently change every published table's semantics vs upstream Kip#1101 reports). Documented as a deliberate divergence in the README.

**Minor — README provenance comment "pre-#119..#125" wrong** (repo hygiene). Confirmed via `git log --merges 11f3fbd0..2e358f17`: the merged replicas are #119/#120/#121/#122/#135. Fixed by amending the README commit (now 1702a6ea).

**Nit — README divergences section omitted the WS rateHz timer-restart divergence** (port fidelity). Confirmed: documented in the commit body and PR but missing from the README list it belongs on. Fixed in 1702a6ea (one bullet).

**Nit — nested git worktrees under `perf-harness/worktrees/` stay registered in the parent clone until `git worktree prune`** (repo hygiene). The persistence is intentional (build reuse, matches upstream), but the registration behavior is now noted in the README's Run section (1702a6ea); the review-session worktrees themselves were pruned from the parent clone.

**Dismissed — report.mjs default `--a master` cannot exist on Skip** (port fidelity, nit). Left as-is, as the reviewer itself suggested: all documented usage passes explicit `--a`/`--b`, changing the default doesn't remove the failure mode (any typo'd label ENOENTs identically), and the windowMs change already keeps the report.mjs diff against upstream `fb391574` minimal — no reason to widen it for a cosmetic default.

### Re-verification after fixes

- `node smoke.mjs` → SMOKE OK; `idle-numeric-24` end-to-end against the built app → boot assertion passes (widgets=24), deltas=16 (2 Hz × 8 s), sane metrics; regenerated report byte-consistent with the committed JSONs.
- `npm run lint` → "All files pass linting".
- `git diff main...HEAD` → 20 files, all under `perf-harness/`; cherry-picked upstream commits untouched (README fix amended into the fork-authored docs commit; history force-pushed accordingly).

PR #143, opened 2026-07-03
fix(ais): keep indexes consistent on at-birth eviction

### LINKED ISSUE #123: AIS: track evicted at birth by cap enforcement leaves dangling contextIndex entry
Found during review of PR #120 (replication of upstream mxtommy/Kip#1101). The behavior is inherited byte-identical from upstream commit ca1df062, so PR #120 deliberately did not diverge to fix it.

**Defect**

In `src/app/core/services/ais-processing.service.ts`, `resolveTrack` calls `createTrack` (which does `tracks.set` + `enforceTargetCap`) and only afterwards does `contextIndex.set(update.context, created.id)`.

If the tracks map is at the cap (`maxTargets`, default 500) and a new context arrives whose first update timestamp is strictly older than every retained track's `lastUpdateAt` (e.g. a buffered/replayed delta), `enforceTargetCap`'s min-find evicts the just-created track inside `createTrack`. `removeTrack`'s contextIndex sweep finds no entry for it yet, and `resolveTrack` then inserts a contextIndex entry pointing at a track no longer in the tracks map.

**Impact**

Small. Requires older-than-all-retained timestamps while at cap. The stale entry is overwritten if the same context ever sends another update (`tracks.get(staleId)` returns undefined, so a fresh track is created and the entry replaced); it leaks permanently only for contexts that never speak again. One Map entry per such context.

**Possible fix**

Register the context in `contextIndex` before cap enforcement runs (or run cap enforcement after `resolveTrack` completes), so `removeTrack`'s sweep can clean it up. Since this code is a faithful cherry-pick from upstream Kip, consider reporting/fixing upstream first and cherry-picking the fix, to keep the fork's divergence minimal.

mairas:
## Motivation

At the target cap (`maxTargets`, default 500), a new AIS context whose first update timestamp is strictly older than every retained track's `lastUpdateAt` (e.g. a buffered/replayed delta) gets evicted by `enforceTargetCap` inside `createTrack` — but `resolveTrack` registered the `contextIndex` entry only *afterwards*, leaving a dangling entry pointing at a track no longer in the `tracks` map. It leaks permanently for contexts that never speak again.

Auditing all creation paths surfaced a second dangle on the same scenario: `resolveTrack` returned the evicted (dead) track object, so when the stillborn context's first update was an `mmsi` path, `applyMmsi` registered the dead id in `mmsiIndex` too.

## Approach

Two coordinated changes in `ais-processing.service.ts`:

- `createTrack` now registers the `contextIndex` entry itself, **before** `enforceTargetCap` — so if the cap evicts the newborn, `removeTrack`'s sweep cleans the entry. All index registration (tracks, contextIndex, mmsiIndex) is now co-located ahead of cap enforcement.
- `resolveTrack` returns `null` when the created track did not survive the cap, so `applyAisUpdate` drops the update instead of mutating a dead object or letting `applyMmsi` index it.

The invariant now holds on every path: **every id stored in `contextIndex`/`mmsiIndex` refers to a live entry in `tracks`.** Behavior for surviving tracks is unchanged; a stillborn track's update was already effectively dropped pre-fix (the track was gone from the map), so returning `null` only removes the dangling side effects.

**Deliberate divergence from upstream Kip:** this eviction code was cherry-picked byte-identical from upstream (mxtommy/Kip#1101, commit ca1df062, still open upstream), which has the same defect. Skip declares rebaseability a non-goal, so we fix it here; upstream should eventually receive an equivalent fix, and future cherry-picks from #1101 must reconcile against this change knowingly.

## Verification

- **RED proof:** the two new specs (`leaves no dangling contextIndex entry for a track evicted at birth`, `leaves no dangling mmsiIndex entry when the first update of a track evicted at birth is an mmsi`) fail on pre-fix code at the dangling-index assertions (`expected true to be false` at the `contextIndex.has(stillborn)` / `mmsiIndex.has('300000099')` checks) — commit 8ddd8767, 2 failed | 17 passed.
- After the fix: `npx ng test --include='src/app/core/services/ais-processing.service.spec.ts'` — 19/19 passed.
- `npx ng test --include='src/app/widgets/widget-ais-radar/widget-ais-radar.component.spec.ts'` — 5/5 passed (radar consumes the `targets()` signal).
- `npm run lint` — clean.
- `npm run build:prod` — succeeds.

Closes #123

### DIFF
--- a/src/app/core/services/ais-processing.service.spec.ts
+++ b/src/app/core/services/ais-processing.service.spec.ts
@@ -282,6 +282,48 @@ describe('AisProcessingService target eviction (bounds unbounded growth)', () =>
     expect(internals().contextIndex.has(contexts[3])).toBe(true);
     expect(internals().contextIndex.has(contexts[4])).toBe(true);
   });
+
+  /**
+   * A track evicted at birth: the map is at the cap and the newcomer's first
+   * timestamp is older than every retained track's lastUpdateAt (e.g. a
+   * buffered/replayed delta), so enforceTargetCap removes the just-created
+   * track inside createTrack. No index may keep referencing it afterwards.
+   */
+  function fillToCap(cap: number): void {
+    internals().maxTargets = cap;
+    for (let i = 0; i < cap; i++) {
+      pushAt(`vessels.urn:mrn:imo:mmsi:${300000001 + i}.navigation.position.latitude`, 10 + i, EVENT_TS + (i + 1) * 1000);
+    }
+    expect(internals().tracks.size).toBe(cap);
+  }
+
+  it('leaves no dangling contextIndex entry for a track evicted at birth', () => {
+    fillToCap(3);
+
+    const stillborn = 'vessels.urn:mrn:imo:mmsi:300000099';
+    pushAt(`${stillborn}.navigation.position.latitude`, 20, EVENT_TS);
+
+    expect(internals().tracks.size).toBe(3);
+    expect(internals().contextIndex.has(stillborn)).toBe(false);
+    for (const [context, id] of internals().contextIndex.entries()) {
+      expect(internals().tracks.has(id), `contextIndex entry for ${context} points at a missing track`).toBe(true);
+    }
+  });
+
+  it('leaves no dangling mmsiIndex entry when the first update of a track evicted at birth is an mmsi', () => {
+    fillToCap(3);
+
+    const stillborn = 'vessels.urn:mrn:imo:mmsi:300000099';
+    pushAt(`${stillborn}.mmsi`, '300000099', EVENT_TS);
+
+    expect(internals().tracks.size).toBe(3);
+    expect(internals().mmsiIndex.has('300000099')).toBe(false);
+    for (const [mmsi, ids] of internals().mmsiIndex.entries()) {
+      for (const id of ids) {
+        expect(internals().tracks.has(id), `mmsiIndex entry for ${mmsi} points at a missing track`).toBe(true);
+      }
+    }
+  });
 });
 
 /**
--- a/src/app/core/services/ais-processing.service.ts
+++ b/src/app/core/services/ais-processing.service.ts
@@ -561,8 +561,10 @@ export class AisProcessingService {
     if (existing) return existing;
 
     const created = this.createTrack(update.context, update.type, timestampMs, undefined);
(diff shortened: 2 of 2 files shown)


mairas:
## Review synthesis

Two independent reviewer passes were run against this PR; both approve with zero findings.

### Lenses

1. **Index-consistency invariant audit** — every mutation path of `tracks`/`contextIndex`/`mmsiIndex` post-fix (creation via context and mmsi, at-birth eviction, TTL sweep, cap eviction, `status: remove`, stale-entry self-heal), plus eviction-order/victim-selection preservation and RED/GREEN verification.
2. **Testing** — RED/GREEN discipline, mutation-kill coverage, assertion quality (invariant-based, not implementation-order-based), vacuous-pass guards.

### Findings

None. Highlights from the reviews:

- The invariant "every id in `contextIndex`/`mmsiIndex` refers to a live entry in `tracks`" holds on all mutation paths: index registration is co-located ahead of `enforceTargetCap` in `createTrack`, so `removeTrack`'s sweep cleans a stillborn's entries, and `resolveTrack`'s null return prevents the second dangle (in `applyMmsi`) that the issue's suggested reorder alone would have left.
- RED reproduced at test-only commit 8ddd8767 (exactly the 2 new dangling-index assertions fail, 17/19 pass); fix commit 4eb6f358 passes 19/19.
- Mutation testing gives a clean kill matrix: reverting only the null-return fails exactly the `mmsiIndex` test; reverting only registration-before-enforcement fails exactly the `contextIndex` test. Each coordinated change is independently covered.
- Victim-selection tests pass identically at the pre-fix commit, confirming eviction order semantics are unchanged (`enforceTargetCap` reads only `this.tracks`, whose insertion order the diff does not alter).

### Fixed / Dismissed

Nothing to fix or dismiss — both reports returned empty findings lists.

### Re-verification (post-review, local worktree)

- `npx ng test --include='src/app/core/services/ais-processing.service.spec.ts'`: 19/19 passed
- `npm run lint`: clean
- `npm run build:prod`: success (pre-existing piexifjs CommonJS warning only)
- CI: all 13 checks green on head 4eb6f358

Note: as stated in the PR body, this fix deliberately diverges from upstream Kip PR #1101 (still open upstream); Skip does not target rebaseability, so future upstream cherry-picks should reconcile knowingly.

PR #144, opened 2026-07-03
fix(ais-radar): refresh own-ship before settled early return

### LINKED ISSUE #124: AIS radar: settled-rotation early return can briefly freeze own-ship icon/vector
Found during review of #122 (replication of upstream Kip PR mxtommy/Kip#1101). The code is inherited byte-identical from upstream commit dd5b59f2; this is an upstream defect, tracked here so it isn't lost.

## Behavior

`render()` in `widget-ais-radar.component.ts` early-returns when `!dataChanged && rotationSettled && hasRenderedOnce`. Own-ship heading is deliberately excluded from `RenderSignature` (that is the point of the optimization), but the early return also skips the O(1) own-ship re-orient and `renderOwnShipVector` call. Two narrow staleness windows result:

1. **North-up after a north-settled course-up stint:** in north-up, `targetRotation` is pinned to 0 and `smoothRotation` is never called, so a `viewRotationSmoothed` lingering within 0.5° of 0 (from a prior course-up stint at heading ~north) keeps `rotationSettled` true. A heading/COG change with bit-identical GPS position and unchanged targets ref then hits the early return, freezing the own-ship icon and COG-vector angle.
2. **SOG-only change in settled course-up:** SOG is not in the signature, so the own-ship vector length stays stale.

Both self-heal on the next update that touches any signature field (GPS position, targets ref, theme, ...) — typically within ~1s on live data.

## Why not fixed in #122

#122 is a faithful replication; diverging on this file would create conflict burden for future upstream cherry-picks. The regression vs the fork base (610d35da kept `ownShipRef` identity in the signature) is minor and self-healing.

## Suggested resolution

Report upstream to mxtommy/Kip and pick up the fix via the normal replication flow. If it bites in practice before then, a fork-authored follow-up could move the own-ship re-orient + `renderOwnShipVector` above the early return (they are O(1) per frame).

mairas:
## Motivation

`render()` early-returns when `!dataChanged && rotationSettled && hasRenderedOnce`. Own-ship heading/COG/SOG are deliberately excluded from `RenderSignature` (that is the point of the settled-frame optimization), but the early return also skipped the O(1) own-ship re-orient and `renderOwnShipVector` call. Two staleness windows (#124):

1. **North-up after a north-settled course-up stint**: north-up pins `targetRotation` to 0 and never calls `smoothRotation`, so a `viewRotationSmoothed` lingering within 0.5° of 0 keeps `rotationSettled` true. A heading/COG change with bit-identical GPS position and unchanged targets ref then froze the own-ship icon and COG-vector angle.
2. **SOG-only change in settled course-up**: SOG is not in the signature, so the own-ship vector length stayed stale.

Both self-healed on the next signature-field change, but on live data that is a visible ~1 s freeze of the own-ship glyph.

## Approach

Move the cheap per-frame block — viewBox/scale/rotation-group transforms, own-ship icon re-orient, and `renderOwnShipVector` — above the early return. All of it is O(1) and already ran on every non-settled frame; nothing O(targets) moves (rings, target build/vectors/icons/selection stay behind the return, still gated on `dataChanged`).

Design notes:

- `smoothRotation` now also runs on settled frames, so the smoothed rotation converges on sub-threshold heading drift instead of freezing, and `lastViewRotation` stays in step with the drawn rotation-group transform (hit-testing consistency).
- Init-order safe: the top-of-function guard still covers `renderState`/`svg`/`root`/`rotationGroup`; `ownShipLayer` access is optional-chained and `renderOwnShipVector` guards it too. On the first data frame the pre-rebuild `g.radar-ownship` selection is empty (no-op) and `renderRings()` applies the same rotation when it creates the group.
- On an early-returned frame the skipped `remainingRotation > 0.5` reschedule cannot trigger anyway: settled means the pre-smooth delta is ≤ 0.5°, and the post-smooth remainder only shrinks.

**Deliberate divergence from upstream Kip PR #1101** (mxtommy/Kip, still open): this file was inherited byte-identical from upstream commit dd5b59f2, and this fix diverges from it by design — Skip declares rebaseability a non-goal. Future upstream cherry-picks in this area must reconcile against this change, and upstream should eventually receive the same fix (report via the normal replication flow).

## Verification

TDD, RED first: `test(ais-radar)` commit adds three tests to the TestBed-free spec (prototype-backed context, spied d3 layers and O(targets) methods, real geometry/signature/smoothing helpers). On pre-fix code the two staleness tests fail with

```
AssertionError: expected "vi.fn()" to be called with arguments: [ 'transform', 'rotate(90)' ] — Number of calls: 0
AssertionError: expected "vi.fn()" to be called 1 times, but got 0 times
```

while the third test pins the preserved perf contract (no `buildTargets`/`renderRings`/`renderTargets`/`scheduleRender` on settled frames) and passes both before and after.

Post-fix:

- `npx ng test --include='src/app/widgets/widget-ais-radar/widget-ais-radar.component.spec.ts'` — 8/8 pass
- `npx ng test --include='src/app/core/services/ais-processing.service.spec.ts'` — 17/17 pass
- `npm run lint` — clean
- `npm run build:prod` — succeeds (pre-existing piexifjs CommonJS warning only)

Closes #124

### DIFF
--- a/src/app/widgets/widget-ais-radar/widget-ais-radar.component.spec.ts
+++ b/src/app/widgets/widget-ais-radar/widget-ais-radar.component.spec.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi, type Mock } from 'vitest';
 import { WidgetAisRadarComponent } from './widget-ais-radar.component';
 
 /**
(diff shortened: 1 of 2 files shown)


mairas:
## Review synthesis

Three independent reviewer lenses were run against this PR:

1. **Correctness** — do the changes close both #124 staleness windows; init-order safety; effect liveness for heading-only/SOG-only updates. **APPROVE.** Both windows verified closed empirically (RED/GREEN swap against the parent commit); the render effect provably fires for heading/SOG-only changes via `flushOwnShip`'s field compare; nothing O(targets) moved above the settled return.
2. **Perf-adversary** — guarding the #1101 settled-frame early-return optimization. **APPROVE.** Storm path is instruction-identical to pre-fix; settled frames gain only O(1) work at data rate (<= ~10 Hz, never 60 fps); the reschedule is unreachable on settled frames.
3. **Testing** — RED/GREEN verification, behavior-vs-implementation pinning, determinism. **APPROVE.** TDD contract verified end-to-end: test commit fails exactly the two staleness tests pre-fix, all pass at head; assertions pin behavior at the d3 boundary.

### Findings and resolution

| # | Severity | Finding | Resolution |
|---|----------|---------|------------|
| 1 | minor (perf) | viewBox and root-scale writes were moved above the settled early return although both derive solely from width/height, which is in the render signature (`sizeKey`) — settled frames rewrote identical attribute values at own-ship data rate, and same-value `setAttribute` on the SVG root is not guaranteed to be a no-op. | **Fixed** in 76e8a96 (fix commit amended): both writes moved back below the early return. Any size change implies `dataChanged`, so they still run exactly when they can change. Only the rotation-dependent writes and the own-ship refresh stay above the return — that is the #124 fix. Side benefit: keeps the settled-return placement closer to upstream Kip #1101 for future cherry-pick reconciliation. |
| 2 | nit (testing) | Test 1 (north-up heading staleness) lacked a `buildTargets`-not-called assertion, so adding heading/COG to `RenderSignature` — silently defeating the streaming-compass optimization the code comment documents — would go undetected (test 3's heading is constant, so it would not catch it either). | **Fixed** in 2125571 (test commit amended so RED stays RED): added `expect(harness.buildTargets).not.toHaveBeenCalled()` to test 1, pinning that the scenario stays a settled frame. |
| 3 | nit (correctness) | Own-ship vector vs icon z-order within `ownShipLayer` is now history-dependent: on the first data frame the vector is appended before the icon group (icon draws above vector); a vector exit/re-enter flips it back. | **Dismissed.** Purely cosmetic with no functional effect: `raiseOwnshipAndVector()` already raises the whole layer above targets, the overlap zone is a few pixels where the dashed vector root meets the glyph, and both stackings are legible (icon-above-vector arguably reads better, per the reviewer). Adding per-frame raise/lower calls inside the hot render path to pin a cosmetic ordering is not warranted. |

### Verification at amended head 76e8a96

- RED preserved: at amended test commit 2125571 (pre-fix component), 2 failed / 6 passed — exactly the two #124 staleness tests fail with the original assertions; the added guard assertion holds on both sides.
- GREEN: `npx ng test --include='src/app/widgets/widget-ais-radar/widget-ais-radar.component.spec.ts'` — 8/8 pass.
- `npm run lint` — clean. `npm run build:prod` — succeeds.
- Branch force-pushed with `--force-with-lease`; CI re-runs on the new head.

PR #145, opened 2026-07-03
fix(charts): fall back to the live tail on a transient backfill failure (#130)

### LINKED ISSUE #130: History-chart engine: transient backfill failure permanently disables the chart (no retry, no live fallback)
## Defect

Found during PR #127 review. `fetchBackfill` returns `null` for **both** "no history provider" and any **transient** failure — `HistoryApiClientService.getValues` catches every network/5xx/timeout error and returns `null`, indistinguishable from the no-provider case. On `null`, `getBackfillThenLive` emits `HISTORY_UNAVAILABLE` and `complete()`s **without ever starting the live delta tail**.

So a brief SK-server restart or network blip *during the one-shot backfill* leaves the chart stuck on "history data unavailable" with no live data and no retry — until an unrelated rebuild (config/theme change or page reload). The recorder keeps showing live data through the same blip.

Distinct from #85 (reconnect re-backfill): this fires on the *initial* backfill, and there is no live-only fallback at all.

## Where

- `history-chart-stream.service.ts` — the `if (batch === null) { subscriber.next(HISTORY_UNAVAILABLE); subscriber.complete(); return; }` branch returns before `startLive`.
- `history-api-client.service.ts` — `getValues` `catch { … return null; }` conflates transient failure with no-provider.

Distinguish "no provider" from "request failed" (typed error / separate signal). On a transient failure, either retry the backfill with bounded backoff or fall through to the live delta tail, rather than completing. At minimum, don't `complete()` on a transient error.

## Scope

History engine only, behind the default-OFF `chartEngine` toggle. **Blocks #64 promotion.**

mairas:
## What

Fixes #130: a transient failure during the History engine's one-shot backfill left the chart permanently stuck on "history unavailable" — no retry, no live fallback — until an unrelated rebuild.

## Why

`HistoryApiClientService.getValues` caught **every** error and returned `null`, indistinguishable from the genuine no-provider case. `getBackfillThenLive` then emitted `HISTORY_UNAVAILABLE` and `complete()`d without ever starting the live delta tail. So a passing server restart or network blip during that single backfill request killed the chart, while the recorder rode straight through the same blip.

## Changes

- **`getValues` now distinguishes the two conditions:** it returns `null` only for a genuine no-provider (no service URL, or a 404/501), and throws a typed **`HistoryRequestError`** for a transient failure (network error, timeout, or 5xx). A bounded 30 s `timeout` is added so a server that accepts the connection but never responds can't hang the backfill forever.
- **`getBackfillThenLive`** shows `HISTORY_UNAVAILABLE` only on the `null` (no-provider) result. On a `HistoryRequestError` it **rides the live delta tail** with no backfill seed, so live data keeps rendering through the blip; a reconnect re-backfill (#85) will fill the gap. An unexpected (non-request) error still degrades to empty, now with a log line.
- The two other `getValues` callers (recorder backfill, history-chart dialog) swallow the new throw to `null`, preserving their existing graceful no-data behaviour.

## Reviewer catch worth highlighting

An independent reliability review found that the transient live-fallback, started with `offsetMs = 0`, would **false-gap a healthy source under server-behind-client clock skew** — precisely the just-rebooted marine-Pi-without-NTP case that triggers the fallback. The live tail stamps the chart axis with the client clock (fine), but its staleness check compared `Date.now()` against the raw server timestamp, reading the skew as age. Fixed by **deriving the offset from the first live sample** when no backfill anchored it, with a reproducing test (60 s skew + two deltas) that the previous zero-skew test couldn't catch.

## Scope

History engine only, behind the default-OFF `chartEngine` / `skip.chartEngine` flag. **Blocks #64 promotion.**

## Testing

`npm run lint` ✓ · `npm run build:prod` ✓ · full `ng test` suite **820 pass** locally, incl. the new transient-fallback, clock-skew-offset, and `getValues` 501/5xx/network-error contract tests. CI runs the full matrix.

Closes #130.

### DIFF
--- a/src/app/core/components/widget-history-chart-dialog/widget-history-chart-dialog.component.ts
+++ b/src/app/core/components/widget-history-chart-dialog/widget-history-chart-dialog.component.ts
@@ -195,12 +195,14 @@ export class WidgetHistoryChartDialogComponent implements OnInit, AfterViewInit,
     const convertUnitTo = this.resolveConvertUnitTo(rawPath);
 
     for (const candidate of requestCandidates) {
+      // A transient failure (network/5xx/timeout) now throws; treat it like an empty result so this
+      // candidate is skipped and the next one (or an empty chart) is tried.
       const response = await this.historyApiClient.getValues({
         paths: candidate.paths,
         context: candidate.context,
         duration,
         resolution: resolutionSeconds
-      });
+      }).catch(() => null);
 
       if (!response?.data?.length) {
         continue;
--- a/src/app/core/services/dataset-stream.service.ts
+++ b/src/app/core/services/dataset-stream.service.ts
@@ -315,8 +315,9 @@ export class DatasetStreamService implements OnDestroy {
         resolution: historyResolutionSeconds
       };
 
-      // Fetch history data
-      const response = await this.history.getValues(query);
+      // Fetch history data. A transient failure (network/5xx/timeout) now throws; treat it like
+      // "no data" so the seed is skipped and the live recorder carries on.
+      const response = await this.history.getValues(query).catch(() => null);
 
       if (!response || !response.data || response.data.length === 0) {
         console.log(`[DatasetStreamService] No history data available for ${configuration.path}`);
--- a/src/app/core/services/history-api-client.service.spec.ts
+++ b/src/app/core/services/history-api-client.service.spec.ts
@@ -3,7 +3,16 @@ import { beforeEach, describe, expect, it, afterEach } from 'vitest';
 import { HttpClientTestingModule, HttpTestingController } from '@angular/common/http/testing';
 import { BehaviorSubject } from 'rxjs';
 import { SignalKConnectionService, IEndpointStatus } from './signalk-connection.service';
-import { HistoryApiClientService } from './history-api-client.service';
+import { HistoryApiClientService, HistoryRequestError } from './history-api-client.service';
+
+const CONNECTED_ENDPOINT: IEndpointStatus = {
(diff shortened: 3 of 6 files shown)


mairas:
## Independent reliability review (recording it here — should have been posted before merge)

An independent reliability pass on this diff drove the fixes that landed in it:

- **[medium — fixed]** The transient live-fallback started the tail with `offsetMs = 0`. On a server whose clock trails the client (a just-rebooted marine Pi without NTP — exactly the situation that triggers the fallback), the staleness check reads the skew as *age* and **false-gaps a healthy source**, defeating the whole point of the fix. Fixed by deriving the offset from the first live sample when no backfill anchored it, with a reproducing test (60 s skew + two deltas) that the original zero-skew single-delta test could not catch.
- **[low — fixed]** Unexpected (non-`HistoryRequestError`) backfill errors degraded to `HISTORY_UNAVAILABLE` silently — added a `console.error` so a real mapper/logic bug isn't indistinguishable from a legitimate no-provider.
- **[low — by design]** 400/403 are classified as transient (ride the live tail) rather than no-provider. Deliberate: only 404/501 mean "no provider"; any other failure rides live rather than claiming permanent absence. Documented in the `getValues` comment.
- Also added a bounded 30 s request timeout so a server that accepts the connection but never responds can't wedge the backfill forever (the commit message had listed "timeout" among handled failures before the code actually delivered it).

Teardown race, unhandled-rejection coverage across all `getValues` callers, and test quality were checked and clean.

🤖 review by Claude Code

Issue #76, opened 2026-07-01
Tracking: codebase-review backlog — wave sequencing & decision gates


mairas:
Sequencing tracker for the codebase-review backlog. **Regrounded 2026-07-13** against the current clone (main). Supersedes the 2026-07-02/03 grounding: the original review waves and every decision gate are now settled, so this reground records the settled decisions, marks what the History-engine landing (#64) mooted, and folds in the ~30 issues filed since (#148–#247) into current clusters.

Core rule unchanged: **decisions that delete code come before fixes inside that code.** `rebase-risk` is informational only (staying rebaseable on upstream is a non-goal).

## Settled decisions (durable record)

- **#46 auth** — cookie-only; token mode + device tokens deleted. Mooted #18/#22.
- **#66 chartplotter mode** — removed; URL normalization re-homed into route config.
- **#84 config namespace** — cold-turkey `kip`→`skip` on both storage tiers (shipped as #105).
- **#17 settings.service rewrite** — signals-first over a single persistence path; shipped (PRs #160/#166/#168/#172). Spawned #170 and #159; unblocked #79.
- **#64 trend charting** — DECIDED & SHIPPED: the SK History-API engine is promoted and the **client-side recorder is retired**. `dataset-stream.service.ts` / `DatasetService` are gone; the Data Chart widget is now History-backfill + live-tail with a graceful "history unavailable" empty state (no recorder fallback). Spawned #85; the fixed-120 density (#86) and circular-avg (#87) bugs were fixed and closed.
- **#68 desktop swipe** — resolved (mouse-movement reveal of the actions shelf; closed).
- **#6 strictNullChecks** — direction committed; ratchet plan with #23 as the landed pilot slice.

## Mootness check (from the #64 landing)

- **#13** (HF-07 — history batch delivery reaching into `ReplaySubject._buffer`) — **MOOT & CLOSED in this reground.** The batched `_buffer` read lived in the recorder that #64 retired; no `._buffer` access remains anywhere in `src`.
[...]

PR #146, opened 2026-07-03
Fix shared-config mutation, loose typing, and WS frame validation (review batch)

### LINKED ISSUE #23: [TS-02] Loose typing masks latent bugs: undefined pushed to stream and wrong return types
**Severity:** medium | **Confidence:** confirmed | **Area:** type-safety | **Source:** codebase review 2026-06-30, finding `TS-02`

## Problem

Three locations in `src/app/core/services/data.service.ts` contain type assignments that the TypeScript compiler silently accepts only because strict null checks are disabled. In `timeoutPathObservable` (lines 729–746), `let timeoutValue: IPathUpdate;` is declared but assigned only inside an `if` block that guards a subset of `pathType` values; the subsequent unconditional `pathRegister.pathDataUpdate$.next(timeoutValue)` call will push `undefined` into the widget data stream for any unrecognised `pathType`. The helper `typeFromUnits` (lines 45–54) is annotated as returning `string` but executes `return undefined` when `!units`, so its declared return type is a lie. A third assignment at line 398 sets `let pathType: string` to `undefined` via a ternary, which is again accepted only under non-strict typing.

## Impact

Under normal operation callers tend to pass valid `pathType` values, so the `undefined` push in `timeoutPathObservable` is latent rather than constantly triggered — but it is a real defect path, not merely a style issue. Widget data streams receiving `undefined` instead of a typed `IPathUpdate` can cause silent display errors or unexpected widget state in a marine navigation context where data reliability matters. The mismatched return type of `typeFromUnits` means callers that depend on always receiving a `string` may behave incorrectly when `units` is falsy. These defects are invisible to the compiler today and will surface as build errors the moment `strictNullChecks` is enabled (as planned in `TS-01`), making them a prerequisite fix for that hardening work.

## Affected code

- `src/app/core/services/data.service.ts:729-746`
- `src/app/core/services/data.service.ts:45-54`
- `src/app/core/services/data.service.ts:398`

## Suggested direction

Fix the three defects individually before or alongside enabling strict flags. For `timeoutPathObservable`, either give `timeoutValue` a safe default initialiser or guard the `.next` call so it is only invoked when `timeoutValue` has been assigned. For `typeFromUnits`, change the return type annotation to `string | undefined` (or a narrower union such as `'Date' | 'number' | undefined`) to match what the function actually returns. For line 398, change the type annotation of `pathType` to `string | undefined` to reflect the ternary's possible output. These corrections are the natural first dividends of the `TS-01` strict-flag initiative.

## Verification

Verified against the codebase: all three cited locations match the described defects, and `tsconfig.json` has no `strict` or `strictNullChecks` flag set (confidence: confirmed).

### LINKED ISSUE #24: [TS-03] WebSocket boundary casts Signal K frames without runtime validation
**Severity:** medium | **Confidence:** confirmed | **Area:** type-safety | **Source:** codebase review 2026-06-30, finding `TS-03`

## Problem

The Signal K WebSocket connection in `signalk-delta.service.ts` is typed as `WebSocketSubject<object>` (line 62). Raw frames from the socket are passed directly into `processWebsocketMessage(message: ISignalKDeltaMessage)` (line 221) with no runtime guard in between. Because every field of `ISignalKDeltaMessage` is optional, TypeScript's structural type system accepts an `object` value as a valid argument — this is a compile-time coincidence, not a validation step. The downstream parsers then unconditionally trust the shape: `parseUpdates` calls `item.path.startsWith(...)` without checking that `path` is a non-null string, and `parseSkMeta` dereferences `metadata.value.properties[key]` where `properties` is typed `object` with no null guard. Malformed or hostile delta frames therefore produce uncaught runtime exceptions rather than explicit, recoverable failures.

## Impact

Any deviation from the expected Signal K delta shape — a buggy server plugin, a protocol version mismatch, or a network interception attack — can throw an unhandled `TypeError` deep inside the service. In a marine instrument panel context this means silent widget failures or a blank display at sea with no actionable error surfaced to the user. The project already ships `@signalk/server-api` as a dependency but does not use its validated delta types at this boundary, so stronger typing is available at no extra cost.

## Affected code

- `src/app/core/services/signalk-delta.service.ts:62` — socket typed `WebSocketSubject<object>`
- `src/app/core/services/signalk-delta.service.ts:221` — raw frame passed to `processWebsocketMessage` with no guard
- `src/app/core/services/signalk-delta.service.ts:333-357` — `processWebsocketMessage(message: ISignalKDeltaMessage)` entry point
- `src/app/core/services/signalk-delta.service.ts:479-495` — `parseSkMeta` dereferences `metadata.value.properties[key]` without null check
- `src/app/core/interfaces/signalk-interfaces.ts:67-103` — `ISignalKDeltaMessage` has every field optional, making it structurally equivalent to `object`
- `src/app/core/interfaces/signalk-interfaces.ts:186` — `properties` typed as bare `object` inside `ISkMetadata`

## Suggested direction

Add a type guard at the WebSocket boundary before dispatching to `processWebsocketMessage`. The discriminator fields already present on valid deltas — `updates`, `requestId`, `self`, `errorMessage` — are sufficient to narrow the type without a heavy schema library. At minimum, guard that `item.path` is a non-null string before calling `.startsWith`. Consider adopting the delta types from `@signalk/server-api` (already a declared dependency) in place of the hand-rolled `ISignalKDeltaMessage` to benefit from any upstream validation improvements. The change is self-contained and does not require restructuring the service.

## Verification

Verified against the codebase at commit HEAD of the SKip fork (confidence: confirmed). All cited file paths and line references were confirmed to exist and match the described code.

### LINKED ISSUE #26: [TS-05] Stringly-typed dispatch and magic-number status handling in storage/requests services
**Severity:** medium | **Confidence:** confirmed | **Area:** type-safety | **Source:** codebase review 2026-06-30, finding `TS-05`

## Problem

`StorageService.patchConfig` accepts `ObjType` as a free-form `string` and dispatches on it via a switch whose cases are magic string literals such as `'IAppConfig'`, `'IThemeConfig'`, `'Array<IUnitDefaults>'`, and `'Array<IDatasetDef>'`. The switch's default branch only emits a `console.warn`, leaving the `document` variable undefined with no compile-time guard — meaning a typo at any call site silently no-ops the save and is never caught by the compiler. In `SignalkRequestsService`, `deltaStatusCodes` is an object literal keyed by status-code numbers with no type annotation or index signature, and `updateRequest` branches on a hand-written disjunction of magic numbers (`statusCode == 200 || == 202 || == 400 || ...`) repeated across several `if` blocks. Both constructs are completely invisible to TypeScript's type checker because the project's `tsconfig.json` sets neither `strict` nor `noImplicitAny`.

## Impact

Silent mis-dispatch in `patchConfig` means a mistyped config-type string produces no error, no save, and no user feedback — user settings are silently discarded. In a marine instrument panel context this could mean preferences, unit defaults, or dataset definitions are lost without any indication. The magic-number status-code pattern in `SignalkRequestsService` is brittle: adding or renaming a status code requires changes in multiple disconnected places, and omitting one is undetectable at compile time.

## Affected code

- `src/app/core/services/storage.service.ts:359-453`
- `src/app/core/services/signalk-requests.service.ts:12-22`
- `src/app/core/services/signalk-requests.service.ts:186-188`

## Suggested direction

Type the `ObjType` parameter of `patchConfig` as a string-literal union (or enum) so the switch becomes exhaustiveness-checked by TypeScript. Give `deltaStatusCodes` an explicit `Record<number, string>` annotation to eliminate the implicit-any numeric index. Replace the hand-written magic-number disjunction in `updateRequest` with a `Set`-based or enum-based lookup so that every status code is defined in exactly one place and compiler-checked. These are small, targeted changes that close a class of silent runtime failures.

## Verification

Verified against the codebase: both source files exist at the cited paths and the described constructs are present (confidence: confirmed).

### LINKED ISSUE #100: app-initNetwork mutates the shared DefaultConnectionConfig singleton by reference (sibling of #35)
Surfaced while reviewing #35 (PCS-05).

## Defect

`app-initNetwork.service.ts` (~line 369, `loadLocalStorageConfig`) does the same mutate-the-shared-singleton-by-reference anti-pattern #35 fixed in `settings.service`:

```
this.config = DefaultConnectionConfig;
this.config.signalKUrl = window.location.origin;
// ...and later mutates isRemoteControl / instanceName / configVersion on the same object
```

On a fresh boot with no stored connection config, this mutates the shared `DefaultConnectionConfig` module singleton in place (`signalKUrl` null → origin, etc.). The `Readonly<IConnectionConfig>` annotation added in #35 does **not** catch it — the reference is laundered through the mutable `this.config: IConnectionConfig` alias, so TS allows the write.

Blast radius is limited today because `settings.service.getDefaultConnectionConfig()` now clones and overwrites `signalKUrl`/`kipUUID`, but it is the identical latent bug class.

## Fix

`cloneDeep` the constant before assigning to `this.config`, mirroring `settings.service.getDefaultConnectionConfig()`.

Ref: `src/app/core/services/app-initNetwork.service.ts:369`.

mairas:
Batches the four independent items from the #76 "Actionable now" block into one PR (per maintainer preference, to limit review churn). Each fix is its own atomic commit; none share files.

## What's in here

**`fix(services): clone default connection config on boot`** — closes #100
`loadLocalStorageConfig` assigned the shared `DefaultConnectionConfig` module singleton by reference and then mutated it in place (`signalKUrl`, version upgrades, v13 migration fields). Now `cloneDeep`s the constant first, mirroring `settings.service.getDefaultConnectionConfig()`. Regression test pins that the singleton's `signalKUrl` stays `null` (asserting the known blank value so the test fails under the bug regardless of spec execution order). Note: unlike settings.service, this path does not regenerate `kipUUID` after cloning — the constant already gets a fresh UUID at module evaluation each page load, so both paths yield a per-boot UUID; left as-is deliberately.

**`fix(core): guard timeout emit, fix loose types`** — closes #23 (pilot slice for the #6 ratchet)
- `timeoutPathObservable` no longer pushes `undefined` into the widget data stream for unrecognised path types (the downstream `widget-streams.directive` conversion map dereferences `x.data.value` and would crash). Recognised types emit exactly the same payload as before.
- `typeFromUnits` return type corrected to `string | undefined`; both callers verified safe (one guarded by a truthy-units check, the other feeds `ISkPathData.type` whose `undefined` state the code already branches on).
- `pathType` local typed `string | undefined` to match its ternary.
- 4 new behavioral tests; the unrecognised-path-type test was verified red before the fix.

**`fix(delta): validate SK frames at WS boundary`** — closes #24
- Hand-rolled `isDeltaFrame` type guard on the four discriminators (`updates`, `requestId`, `self`, `errorMessage`); frames failing it are logged and dropped — never thrown on.
- `parseUpdates` drops items whose `path` isn't a string (keeping siblings), and the `meta`/`values` checks are now `Array.isArray` so null entries can't throw.
- `parseSkMeta` drops null/undefined `value` and treats `properties: null` as the single-emission branch instead of throwing in `Object.keys`.
- 9 new malformed-frame specs, verified red first; all 19 pre-existing characterization tests unchanged. Adopting `@signalk/server-api` delta types in place of the all-optional `ISignalKDeltaMessage` remains a possible follow-up (deliberately out of scope here).

**`refactor(core): typed config and status dispatch`** — closes #26
- `patchConfig`'s `ObjType` is now the exported string-literal union `TConfigObjectType` (exactly the 8 literals the switch dispatches on; all call sites verified in-union). The default branch stays as runtime defense and gains a `never` exhaustiveness check. The stale docstring list (which disagreed with the switch) now points at the union.
- `deltaStatusCodes` annotated `Record<number, string>`; the six-way magic-number disjunction replaced with a single `handledStatusCodes` Set. Behavior unchanged — 500/502/504 still route to the unknown-code branch, now pinned by a new characterization test.
- `patchConfig`'s `value: any` is out of scope (existing eslint-disable retained).

## Verification

- `npm run lint` clean; full suite green locally on the assembled branch: 134 files / 841 tests (baseline on main: 821).
- Every behavioral fix had its test verified red before the fix (TDD).
- `tsc --noEmit` on the app and spec tsconfigs passes with the narrowed `patchConfig` signature (lint alone does not typecheck).

Pre-existing defects surfaced while implementing (multi-source timeout-reset gap, de-facto-optional interface fields, delta `$source`/`timestamp` validation gaps) are being filed as separate issues rather than folded in here.

### DIFF
--- a/src/app/core/services/app-initNetwork.service.spec.ts
+++ b/src/app/core/services/app-initNetwork.service.spec.ts
@@ -16,6 +16,7 @@ import { InternetReachabilityService } from './internet-reachability.service';
 import { DatasetStreamService } from './dataset-stream.service';
 import { Injector } from '@angular/core';
 import { ensureLocalStorage } from '../../../test-helpers/local-storage.test-helper';
+import { DefaultConnectionConfig } from '../../../default-config/config.blank.const';
 
 // jsdom has no FontFace; preloadFonts() constructs one during the end-to-end initNetworkServices runs.
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -265,6 +266,21 @@ describe('AppNetworkInitService', () => {
 
     it('should be created', () => {
         expect(service).toBeTruthy();
+    });
+
+    describe('loadLocalStorageConfig default seeding (issue #100)', () => {
+        it('clones DefaultConnectionConfig instead of mutating the shared singleton', () => {
+            localStorage.removeItem('skip.connectionConfig');
+
+            (service as unknown as { loadLocalStorageConfig: () => void }).loadLocalStorageConfig();
+
+            const config = (service as unknown as { config: IConnectionConfig }).config;
+            expect(config).not.toBe(DefaultConnectionConfig);
+            expect(config.signalKUrl).toBe(window.location.origin);
+            // The blank singleton stays pristine: earlier end-to-end runs in this file also seed the
+            // default config, so assert the known blank value rather than a captured "before" value.
+            expect(DefaultConnectionConfig.signalKUrl).toBeNull();
+        });
     });
 
     it('should use connection retry window when no timeout is provided', async () => {
--- a/src/app/core/services/app-initNetwork.service.ts
+++ b/src/app/core/services/app-initNetwork.service.ts
@@ -13,6 +13,7 @@ import { SignalKConnectionService } from "./signalk-connection.service";
 import { AuthenticationService, ILoginStatus } from './authentication.service';
 import { SsoRedirectService } from './sso-redirect.service';
 import { DefaultConnectionConfig } from '../../../default-config/config.blank.const';
+import { cloneDeep } from 'lodash-es';
 import { BehaviorSubject, Observable, Subscription } from 'rxjs';
 import { DataService } from './data.service';
 import { SignalKDeltaService } from './signalk-delta.service';
@@ -367,7 +368,7 @@ export class AppNetworkInitService implements OnDestroy {
     this.config = JSON.parse(localStorage.getItem(CONNECTION_CONFIG_KEY));
 
     if (!this.config) {
-      this.config = DefaultConnectionConfig;
+      this.config = cloneDeep(DefaultConnectionConfig);
       this.config.signalKUrl = window.location.origin;
       console.log(`[AppInit Network Service] Connection Configuration not found. Creating configuration using Auto-Discovery URL: ${this.config.signalKUrl}`);
       this.setLocalStorageConfig();
--- a/src/app/core/services/data.service.spec.ts
+++ b/src/app/core/services/data.service.spec.ts
@@ -292,4 +292,73 @@ describe('DataService', () => {
     subA.unsubscribe();
     subB.unsubscribe();
   });
+
+  it('resets value, timestamp and state on timeout for a recognised pathType', () => {
+    let latest: IPathUpdate | undefined;
+    service
+      .subscribePath('self.environment.wind.speedApparent', 'default')
+      .subscribe(update => (latest = update));
+
+    dataPathUpdates$.next({
+      context: 'self',
+      path: 'environment.wind.speedApparent',
+      source: 'test-source',
+      timestamp: '2026-01-01T00:00:01.000Z',
+      value: 5.5,
+    });
+    expect(latest!.data.value).toBe(5.5);
+
+    service.timeoutPathObservable('self.environment.wind.speedApparent', 'number');
+
+    expect(latest!.data.value).toBeNull();
+    expect(latest!.data.timestamp).toBeNull();
+    expect(latest!.state).toBe(States.Normal);
+  });
+
+  it('does not emit on timeout for an unrecognised pathType', () => {
+    const updates: IPathUpdate[] = [];
+    service
+      .subscribePath('self.navigation.position', 'default')
+      .subscribe(update => updates.push(update));
+
+    const countBefore = updates.length;
+    service.timeoutPathObservable('self.navigation.position', 'object');
+
+    expect(updates.length).toBe(countBefore);
+    expect(updates.every(update => update !== undefined)).toBe(true);
+  });
+
+  it('derives path type from meta units when meta precedes the value', () => {
+    metadataUpdates$.next({
+      context: 'self',
+      path: 'electrical.batteries.10.current',
+      meta: { description: 'Battery current', units: 'A', properties: {} },
+    });
+    metadataUpdates$.next({
+      context: 'self',
+      path: 'navigation.datetime',
+      meta: { description: 'Time', units: 'RFC 3339 (UTC)', properties: {} },
+    });
+    metadataUpdates$.next({
+      context: 'self',
+      path: 'design.aisShipType',
+      meta: { description: 'Ship type', units: undefined, properties: {} },
+    });
+
+    expect(service.getPathObject('self.electrical.batteries.10.current')!.type).toBe('number');
+    expect(service.getPathObject('self.navigation.datetime')!.type).toBe('Date');
+    expect(service.getPathObject('self.design.aisShipType')!.type).toBeUndefined();
+  });
+
+  it('leaves path type undefined when the first received value is null', () => {
+    dataPathUpdates$.next({
+      context: 'self',
+      path: 'navigation.anchor.position',
+      source: 'test-source',
+      timestamp: '2026-01-01T00:00:01.000Z',
+      value: null,
+    });
+
+    expect(service.getPathObject('self.navigation.anchor.position')!.type).toBeUndefined();
+  });
 });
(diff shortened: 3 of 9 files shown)


mairas:

Eight independent reviewer personas (correctness, maintainability, testing, project-standards, adversarial, security, reliability, api-contract) reviewed this diff in fresh contexts; findings were deduplicated and each surviving finding was then challenged by an independent adversarial verifier. Seven personas returned no findings; two flagged the same gap, which the verifier confirmed.

### Confirmed finding (minor — test coverage)

**WS boundary guard: request/error frame passthrough is untested.** `isDeltaFrame`'s `requestId` and `errorMessage` clauses are never exercised *through* `handleStreamFrame` — the new passthrough tests cover only `updates` and `self` frames, and the existing requestId/errorMessage specs call `processWebsocketMessage` directly, bypassing the guard. If either clause regressed, valid PUT request-response frames would be silently dropped at the boundary (no confirmation/error toast on a command) with the suite staying green — a silent-failure path this fix explicitly exists to prevent.

**Resolution:** adding two passthrough tests mirroring the update/self ones, amended into the `fix(delta)` commit on this branch.

No correctness, security, reliability, or API-contract findings survived verification.

mairas:
Review finding addressed: two boundary-passthrough tests added (request-response frame reaching the requests stream; errorMessage frame accepted and logged rather than dropped), amended into the `fix(delta)` commit (`c72f59a6`). Suite green locally: 134 files / 843 tests, lint clean.

PR #147, opened 2026-07-03
Characterization tests for the four smoke-stubbed core services

### LINKED ISSUE #27: [TST-01] ~47% of spec files are smoke-only stubs including several non-trivial services
**Severity:** medium | **Confidence:** confirmed | **Area:** testing | **Source:** codebase review 2026-06-30, finding `TST-01`

## Problem

48 of 102 spec files (approximately 47%) in the codebase contain a single `it('should be created', ...)` assertion and nothing else. These stubs only verify that Angular's dependency-injection testbed can construct the class — they make no behavioral assertions. While this is acceptable for trivial dialog components, the set also includes several substantial services with hundreds of lines of non-trivial logic. Because all 102 spec files are counted in the spec-to-source ratio, the apparent coverage of 102 specs against 197 source files materially overstates actual behavioral test coverage.

## Impact

Non-trivial services such as the dashboard service, notifications service, and remote-dashboards service carry complex stateful logic that is entirely untested beyond instantiation. Bugs in these services would not be caught by the existing test suite. The inflated spec count makes the project's test health appear stronger than it is, which can mislead contributors prioritizing test work. This is a pre-existing upstream gap that the fork inherits.

## Affected code

- `src/app/core/services/signalk-requests.service.spec.ts` (covers a 251-line service)
- `src/app/core/services/notifications.service.spec.ts` (covers a 423-line service)
- `src/app/core/services/dashboard.service.spec.ts` (covers a 447-line service)
- `src/app/core/services/remote-dashboards.service.spec.ts` (covers a 229-line service)
- `src/app/core/services/app-service.spec.ts` (covers a 273-line service)

## Suggested direction

Track smoke-only stubs separately from specs with real behavioral assertions so that reported coverage reflects actual testing depth. For the five substantial services identified, replace the single-assertion stubs with behavior-focused tests that exercise meaningful logic paths — prioritizing `dashboard.service`, `notifications.service`, and `remote-dashboards.service` given their size. Alternatively, delete the boilerplate stubs for services not yet ready for real testing so they no longer inflate the count.

## Verification

Verified against the codebase (confidence: confirmed): `find src -name "*.spec.ts" | wc -l` = 102, source files = 197. 48 spec files contain exactly one `it` block whose description matches "should be created" or "should create" with no other assertions. All five named service spec files confirmed present and match the described stub pattern.

mairas:
Closes #27 as scoped by the #76 tracker: real behavioral specs for the four substantial services that only had `should be created` smoke stubs (signalk-requests was dropped from scope — covered by #8). One commit per service; each touches only that service's spec file. No app code is changed.

## What's pinned

**`dashboard.service` — 33 tests** (blank-dashboard seeding, route-param-driven activation, bounds checks, add/update/delete/duplicate, `updateConfiguration` clone semantics, previous/next cycling, widget actions, clipboard sanitization, persistence effect). Pins several surprising behaviors as-is, notably the inverted previous/next direction semantics and delete() not remapping the active index — defect issues follow separately.

**`notifications.service` — 26 tests** (alarm aggregation from deltas, max-severity selection, `notifications.security` filtering, update/delete/dedupe transitions, severity/mute gating, alarm sound track selection, config enable/disable stream handling, reset behavior, Signal K PUT helpers).

**`remote-dashboards.service` — 24 tests** (construction-time remote-path clears with exact payloads, enable/disable publication, display-name changes, remote screen-change requests incl. coercion and rejection paths, publish helpers).

**`app-service` — 31 tests** (theme color publication, brightness CSS custom properties, day/night/red-night toggles, auto night mode transitions, browser/OS detection branches). jsdom returns `''` for `getComputedStyle` custom properties, so theme-color tests pin structure/identity; inline-style properties round-trip and are asserted by value.

## Approach notes

- Characterization discipline: tests document what the code **does today**, including oddities, with a brief comment where the pinned behavior is surprising. Anything defect-shaped was recorded and is being filed as GitHub issues referencing the pinning tests — the specs themselves stay faithful to current behavior.
- All mocking is via spec-local providers overriding the global `src/test.ts` stubs (local wins); `src/test.ts` is untouched.
- Effects are flushed with `TestBed.tick()` following the existing `widget-windsteer` spec pattern; `DataService` mocks emit synchronously (BehaviorSubject-backed) to match the real `subscribePath` contract several effects implicitly rely on.

## Verification

`npm run lint` clean; full suite green locally: 134 files / 936 tests (baseline on main: 821). All 114 new tests pass through the `@angular/build:unit-test` builder locally.

One pinned oddity is deliberately not getting an issue: modern Opera reporting as "Chrome" in the about info (UA token order) — cosmetic, not worth tracking.

### DIFF
--- a/src/app/core/services/app-service.spec.ts
+++ b/src/app/core/services/app-service.spec.ts
@@ -1,16 +1,312 @@
 import { TestBed } from '@angular/core/testing';
-import { beforeEach, describe, expect, it } from 'vitest';
+import { MatIconRegistry } from '@angular/material/icon';
+import { BehaviorSubject, Observable } from 'rxjs';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import packageInfo from '../../../../package.json';
+import { States } from '../interfaces/signalk-interfaces';
 import { AppService } from './app-service';
+import { DataService, IPathUpdate } from './data.service';
+import { SettingsService } from './settings.service';
+
+class SettingsServiceMock {
+  public themeName$ = new BehaviorSubject<string>('');
+  public autoNightMode$ = new BehaviorSubject<boolean>(false);
+  public redNightMode$ = new BehaviorSubject<boolean>(false);
+  public nightModeBrightness = 0.27;
+
+  getThemeNameAsO(): Observable<string> { return this.themeName$.asObservable(); }
+  getAutoNightModeAsO(): Observable<boolean> { return this.autoNightMode$.asObservable(); }
+  getRedNightModeAsO(): Observable<boolean> { return this.redNightMode$.asObservable(); }
+  getNightModeBrightness(): number { return this.nightModeBrightness; }
+}
+
+const modeUpdate = (value: string): IPathUpdate => ({
+  data: { value, timestamp: null },
+  state: States.Normal
+});
 
 describe('AppService', () => {
-  let service: AppService;
+  let settings: SettingsServiceMock;
+  let envMode$: BehaviorSubject<IPathUpdate>;
 
   beforeEach(() => {
-    TestBed.configureTestingModule({});
-    service = TestBed.inject(AppService);
+    document.body.className = '';
+    document.body.removeAttribute('style');
+    settings = new SettingsServiceMock();
+    envMode$ = new BehaviorSubject<IPathUpdate>(modeUpdate('day'));
+    TestBed.configureTestingModule({
+      providers: [
+        { provide: SettingsService, useValue: settings },
+        { provide: DataService, useValue: { subscribePath: () => envMode$.asObservable() } as Partial<DataService> }
+      ]
+    });
   });
 
-  it('should be created', () => {
-    expect(service).toBeTruthy();
+  afterEach(() => {
+    Reflect.deleteProperty(window.navigator, 'userAgent');
+    Reflect.deleteProperty(window.navigator, 'platform');
+  });
+
+  function setNavigator(userAgent: string, platform = ''): void {
+    Object.defineProperty(window.navigator, 'userAgent', { configurable: true, value: userAgent });
+    Object.defineProperty(window.navigator, 'platform', { configurable: true, value: platform });
+  }
+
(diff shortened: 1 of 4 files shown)


mairas:

Five independent reviewer personas (correctness, maintainability, testing, project-standards, adversarial) reviewed this diff in fresh contexts; findings were deduplicated and each surviving finding was then challenged by an independent adversarial verifier. Two findings survived, both minor, both about fixture/coverage fidelity rather than the pinned behaviors themselves.

### Confirmed findings (minor — test fidelity)

1. **Notification dedupe: message-only and method-only branches are unpinned.** `processNotificationDeltaMsg` fires an update when state, message, OR method differs; the spec pins only state-change and all-unchanged. The verifier confirmed by mutation testing: deleting either the message comparison or the method `isEqual` comparison leaves all 26 tests green — so those branches aren't characterized.
   **Resolution:** two additional cases (message-only change, method-only change), amended into the notifications commit.

2. **Dashboard router stub mismodels the production route tree.** The app route is top-level `dashboard/:id`, so in production the param is found only by walking `root.firstChild`; the stub puts `:id` directly on the root snapshot with `firstChild: null`, so the tree-walk branch of `getRouteParam` — the one that actually runs in the app — is never exercised. A root-only lookup would pass every test while breaking production deep-links.
   **Resolution:** stub reworked to carry the param on a `firstChild` snapshot (root paramMap empty), amended into the dashboard commit.

No findings against the pinned behaviors themselves survived — the deliberately pinned oddities are tracked as #148–#153.

mairas:
Both review findings addressed, amended into their originating commits:
- `test(notifications)` (`339fa257`): message-only and method-only change cases added — the mutations the verifier used to prove the gap now fail the suite.
- `test(dashboard)` (`65450339`): router stub reworked to carry `:id` on a `firstChild` snapshot with an empty root paramMap, matching the production route tree, so `getRouteParam`'s tree-walk is exercised by all six route tests.

Suite green locally: 134 files / 938 tests, lint clean.

PR #157, opened 2026-07-03
refactor(charts): remove widgetHistoryDisabled setting + perf instrumentation (#134)

### LINKED ISSUE #134: History-chart engine: pre-promotion polish (component tests, remove perf instrumentation, disabled-vs-unavailable messaging)
## Context

Lower-severity items found during PR #127 review, to clear before #64 flips the History engine to default.

## Items

- [ ] **Component test coverage.** `widget-data-chart.component.ts` gained real branching in this PR — engine selection, the `localStorage` override precedence (`skip.chartEngine` > per-widget config > `recorder`), and the `historyUnavailable` signal — with **zero** component tests. Add coverage (CI is the source of truth for `localStorage`-touching specs; local jsdom lacks a usable `localStorage`).

- [ ] **Remove the A/B perf instrumentation.** `chart-perf.util.ts` (120 lines) ships in the product bundle, runtime-gated to a no-op (`skip.chartPerf`); its documented on-widget overlay was never built, so `snapshot()`/`IChartPerfSnapshot` are half-dead. It exists only to quantify the recorder-vs-History comparison — remove it (and the flag key) or move it behind a build-time dev flag once the A/B decision is made.

- [ ] **Fix disabled-vs-unavailable messaging.** `getWidgetHistoryDisabled()` (a user setting) maps to the same `HISTORY_UNAVAILABLE` sentinel and "a history provider is required" message as a genuinely absent provider; the flag is also read once at subscribe, so a late server-version resolution / runtime flip isn't re-evaluated (cold-start race → stuck unavailable until rebuild). Distinguish the two conditions and react to the flag observable (`getWidgetHistoryDisabledAsO()`).

## Scope

mairas:
## What this does

Pre-promotion polish for the chart pipeline, in three parts.

### 1. Remove the A/B chart-perf instrumentation

The runtime-gated no-op perf probe (`skip.chartPerf`) shipped in the bundle but its on-widget overlay was never built. This deletes the util, drops the `CHART_PERF_FLAG_KEY` constant, and unthreads the probe from its sole consumer (`HistoryChartStreamService`). No other stream behaviour changes.

### 2. Remove the `widgetHistoryDisabled` setting entirely

Product decision: the "Disable widget historical charts" setting is gone. **History is available whenever the Signal K server supports it** — there is no user-facing off switch.

Nothing is lost by removing it. When history is genuinely unavailable — no History API provider, *including a server too old to have the History API* — `HistoryApiClientService.getValues` returns `null`, the stream emits `HISTORY_UNAVAILABLE`, and the widget already renders "History data unavailable". The old server-unsupported case is therefore fully covered by the existing no-provider path; the disabled flag had no remaining job.

Removed across the stack:

- `HistoryChartStreamService`: the disabled guard at the top of `getBackfillThenLive`, plus the now-unused `SettingsService` injection/import. `HISTORY_UNAVAILABLE` is now documented as meaning only "no history provider" (the one production behaviour change: the stream no longer knows about a disabled setting).
- `SettingsService`: the `widgetHistoryDisabled` subject, its getters/setters, and the `pushSettings` server-version gating block — which also lets the now-unused `compare` (compare-versions) and `SignalKConnectionService` dependencies go.
- `IAppConfig.widgetHistoryDisabled` (old saved configs carrying the key are harmless — the extra JSON key is ignored, no migration needed).
- Display-options toggle (`display.component.html`/`.ts`). The provider-selection radios and the `isHistoryApiSupported()` wrapper are untouched (separate item).
- `widget-host2`: the history-popup early-return. The two-finger-tap / right-click history popup now always opens (subject to the remaining static-dashboard + overlay guards); on a server without history it simply shows no data.

### 3. Add widget-data-chart component tests

New `widget-data-chart.component.spec.ts` covers the real branching #134 targeted (minus the removed disabled state): recorder-vs-history engine dispatch, the `localStorage['skip.chartEngine']` override beating per-widget config, and the `HISTORY_UNAVAILABLE` empty state. Chart.js and its plugins are mocked self-contained (jsdom can't instantiate a real chart); the override is driven with `vi.stubGlobal` so it's deterministic.

## Verification

- `npm run lint` — clean
- `npx ng test` — 959 tests passed (135 files)
- `npm run build:prod` — succeeds

Closes #134 — the disabled-vs-unavailable messaging item is resolved by removing the setting: there is now a single "History data unavailable" state, shown whenever the server can't serve history.


mairas:
## Review — reworked branch

Reviewed: perf-instrumentation removal + full `widgetHistoryDisabled` removal. **Clean; ready to merge.**

**Why this was reworked (context):** the original #134 approach — distinguish disabled-vs-unavailable + a cold-start recovery effect — was found **inert** by an independent correctness pass. The "late server-version resolution" race it targeted can't occur: `serverVersion$` is set on the endpoint response (`signalk-connection.service.ts:306`), and `pushSettings()` (which computes the flag) runs only after the storage bootstrap that depends on that same connection — so the version is always resolved before the flag is computed. And `SettingsService` never re-emits the flag on a late version anyway. Combined with the product decision that history is always available when the server supports it, the setting was removed entirely.

**Verified on the removal:**
- Grep-clean: zero remaining references to `widgetHistoryDisabled` / `historyDisabled` / `chart-perf` / `CHART_PERF` under `src`.
- The orphaned injects came out correctly and compile-clean: `SignalKConnectionService` + `compare` from `settings.service.ts` (only the removed history block used them), and the `settings` inject from `widget-host2.component.ts` (only the removed popup guard used it).
- Removing the optional `IAppConfig.widgetHistoryDisabled` needs no migration — old configs load fine and the field is simply dropped on next save.
- Unsupported-server behaviour is preserved without the flag: no History API → `getValues` returns `null` → `HISTORY_UNAVAILABLE` → the existing "History data unavailable" empty state. `HistoryChartStreamService` JSDoc updated so the sentinel documents "no provider" only.
- The two-finger-tap / right-click history popup now always opens (subject to the remaining static-dashboard + overlay guards); on a server without history it shows no data, consistent with the new model.

**Tests:** the new `widget-data-chart.component.spec.ts` is meaningful — mounts the real component and asserts the correct engine service is dispatched per `chartEngine` (recorder / default / history), the `skip.chartEngine` localStorage override precedence, and the unavailable empty state.

**CI:** all checks green; `npm run lint` / full `ng test` / `build:prod` pass locally too.

🤖 review by Claude Code

PR #161, opened 2026-07-03
fix(charts): back History backfill Value series with :last (closes #87, #132)

### LINKED ISSUE #87: History-chart backfill plots linear (:avg) means for circular/direction paths — wrong at the 0/360° wrap

### LINKED ISSUE #132: History-chart engine: 'Value' series is server-average in backfill but raw instantaneous when live (seam discontinuity)
## Defect

Found during PR #127 review. The `Value` series (`dataset[0]`) is plotted from the History `:avg` column during backfill (`HistoryToChartMapperService` maps `avg` → value) but from the **raw instantaneous** delta once the live tail takes over (`Number(u.data.value)`). The same visual trace switches statistical meaning at the join — a smooth averaged line abruptly becomes a noisy instantaneous one — and is inconsistent with the recorder, whose `Value` series is always instantaneous.

The min/max annotation drifts the same way: whole-window `:avg`-derived on the last backfill point, then rolling-buffer-derived once live.

## Where

- `history-chart-stream.service.ts` — `fetchBackfill` (maps the `:avg` column into `value`) vs `startLive` (`value = Number(u.data.value)`).

Choose one contract for `dataset[0]` and make both phases honor it — e.g. request/plot a per-bucket instantaneous-equivalent for backfill, or plot the same aggregate on both sides — so the seam is continuous.

## Scope

mairas:
## Why

The History chart engine backfilled its **Value** series with `:avg` — the arithmetic mean of each resolution bucket. For circular/direction paths that mean is wrong at the 0/360° wrap (#87), and because the live tail plots raw instantaneous samples, the backfill→live handoff showed a visible seam (#132).

This is the first step of #64 (promote the History engine): the promoted engine must be seam-free and angle-safe before it becomes the default.

## What

Back the Value series with the raw per-bucket **`:last`** sample instead of `:avg`:

- **Client** — `history-chart-stream` backfill now requests `:sma` + `:last` (dropping the dead `:min`/`:max` columns the mapper never read; it derives min/max/avg from the value column). `history-to-chart-mapper` prefers a `:last` column and keeps `:avg`/`:average` as the fallback, so the other callers (history-chart dialog, the still-present recorder seed) are untouched.
- **kip-plugin** — a client-only change would have been inert: the bundled provider coerced any unknown method to `:avg` (`THistoryMethod` had no `last`). So `:last` is implemented end-to-end in the provider — method type, request normalization/parsing, the `applyMethod` passthrough, `aggregateBucket` (returns the bucket's final raw sample), and the OpenAPI enum.

`:last` cannot invert at the 0/360° wrap and matches the client-stamped live tail, so both #87 and #132 are resolved by the same change.

### Behavior note for reviewers

The reference lines (`lastAverage`/`lastMinimum`/`lastMaximum`) are computed from the Value column, so sourcing it from `:last` (raw samples) rather than `:avg` (bucket means) shifts their backfill values. That's intentional — it makes backfill consistent with the live tail, which already computes stats over raw samples.

## Verification

- `npm run test:plugin` — green, incl. a new test that `:last` returns the last raw sample per bucket (not the mean).
- Mapper spec — `:last` preferred over `:avg`; the `:avg`/`:average` fallback tests still pass.
- Stream spec — backfill requests `:last` + `:sma:`, never `:avg`/`:min`/`:max`.
- `npm run lint` — clean.

Closes #87
Closes #132

Part of #64.

### DIFF
--- a/kip-plugin/src/index.ts
+++ b/kip-plugin/src/index.ts
@@ -679,7 +679,7 @@ const start = (server: ServerAPI): Plugin => {
     if (raw === 'average') {
       return 'avg';
     }
-    if (raw === 'min' || raw === 'max' || raw === 'sma' || raw === 'ema' || raw === 'avg') {
+    if (raw === 'min' || raw === 'max' || raw === 'sma' || raw === 'ema' || raw === 'avg' || raw === 'last') {
       return raw;
     }
     return 'avg';
--- a/kip-plugin/src/kip-series-contract.ts
+++ b/kip-plugin/src/kip-series-contract.ts
@@ -1,4 +1,4 @@
-export type THistoryMethod = 'min' | 'max' | 'avg' | 'sma' | 'ema';
+export type THistoryMethod = 'min' | 'max' | 'avg' | 'sma' | 'ema' | 'last';
 
 export type TElectricalFamilyKey = 'batteries' | 'solar' | 'chargers' | 'inverters' | 'alternators' | 'ac';
 export type TElectricalExpansionMode = 'bms-battery-tree' | 'solar-tree' | 'charger-tree' | 'inverter-tree' | 'alternator-tree' | 'ac-tree';
--- a/kip-plugin/src/openApi.json
+++ b/kip-plugin/src/openApi.json
@@ -187,7 +187,8 @@
                 "max",
                 "avg",
                 "sma",
-                "ema"
+                "ema",
+                "last"
               ]
             }
           },
--- a/kip-plugin/src/sqlite-history-storage.service.ts
+++ b/kip-plugin/src/sqlite-history-storage.service.ts
@@ -1049,7 +1049,7 @@ export class SqliteHistoryStorageService {
   private parseMethod(value?: string): THistoryMethod | undefined {
     if (!value) return undefined;
     const normalized = value.toLowerCase();
-    if (normalized === 'min' || normalized === 'max' || normalized === 'avg' || normalized === 'sma' || normalized === 'ema') {
+    if (normalized === 'min' || normalized === 'max' || normalized === 'avg' || normalized === 'sma' || normalized === 'ema' || normalized === 'last') {
       return normalized;
     }
     return undefined;
@@ -1159,7 +1159,7 @@ export class SqliteHistoryStorageService {
     if (rows.length === 0) return [];
 
     const method = request.method ?? 'avg';
-    if (method === 'min' || method === 'max' || method === 'avg') {
+    if (method === 'min' || method === 'max' || method === 'avg' || method === 'last') {
       return rows.map(entry => ({ timestamp: Number(entry.ts_ms), value: Number(entry.value) }));
     }
 
@@ -1233,6 +1233,12 @@ export class SqliteHistoryStorageService {
       return Math.max(...values);
     }
 
+    if (method === 'last') {
+      // Relies on bucket lists being timestamp-ascending, which holds because selectRowsForPaths
+      // fetches `ORDER BY path ASC, ts_ms ASC` and downsampleIfNeeded preserves that order.
+      return values[values.length - 1];
+    }
+
     const sum = values.reduce((acc, value) => acc + value, 0);
     return sum / values.length;
   }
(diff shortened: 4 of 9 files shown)


mairas:
## Review synthesis (correctness + adversarial personas)

**Correctness — clean, zero diff defects.** Traced end-to-end: `:last` returns each bucket's final raw sample (`ORDER BY … ts_ms ASC` → `applyMethod` passthrough → per-bucket `aggregateBucket`), handles empty/single/non-finite/`resolution<=0` cases, the mapper's `valueIndex` preference and fallback are correct, and dropping `:min`/`:max` is safe (the mapper derives min/max from the value column, never from API columns).

**Adversarial — bundled path sound (61 plugin tests green); two findings + residual risks.** Addressed:

| Finding | Action |
|---|---|
| **SMA overlay still arithmetic** — now visibly disagrees with the corrected `:last` Value line at the 0/360° wrap on angle charts | Filed **#162** and **gated the default engine flip (PR5)** on it. Not fixed here: it's a separate code path (the smoothing computation), and the History engine is default-OFF, so no production user sees it yet. |
| **Mapper single-column fallback** could plot a lone `:sma` column as the Value series if a non-kip provider drops the `:last` label | Fixed: the fallback now guards on `smaIndex < 0`, so a lone `:sma` column is never treated as the raw value (+ test). |
| **Ordering contract** — `aggregateBucket('last')` depends on ascending row order from an `ORDER BY` ~220 lines away | Documented the ascending-order dependency in-place. |

Added hardening tests: `:sma`+`:last` column independence (the exact backfill shape), raw passthrough at zero resolution, and skipping a trailing non-finite sample.

**Known limitation (tracked as a #64 follow-up):** `:last` correctness is a provider contract. A non-bundled provider that silently coerces `:last` to an average (and still labels the column `last`) reintroduces the wrap bug on that stack — the client can't detect it. Providers that error on `:last` degrade gracefully (live-tail fallback or `HISTORY_UNAVAILABLE`). The bundled kip-plugin implements `:last` for real.

PR #164, opened 2026-07-03
feat(charts): stream the numeric mini-chart from the History engine

mairas:
## Why

The numeric widget's background sparkline (`MinichartComponent`) was the last **recorder-only** consumer besides windtrends — it read a pre-recorded dataset by `datasetUUID` from `DatasetStreamService` and had no History-engine path. Deleting the recorder (a later #64 PR) would silently blank it. This migrates it first so the recorder can go without a regression.

## What

- **`MinichartComponent`** now streams from `HistoryChartStreamService.getBackfillThenLive` instead of `dsService.getDatasetBatchThenLiveObservable(datasetUUID)`. Its window is the same fixed 12s (`minute` / `0.2`) the recorder used, derived through the shared `resolveWindowMs` + `deriveDataSourceInfo` — the recorder derived sampling the exact same way (`dataset-stream.service.ts:136-137`), so cadence, buffer size and rendering are unchanged (12000ms → 100ms / 120pts / smoothing 30).
- **`widget-numeric`** drops its orchestrator dataset sync (`syncNumericMiniChartDataset` / `removeDatasetIfExists`) and the `datasetUUID` wiring — the sparkline no longer needs a user-managed dataset. The now-misnamed `manageDatasetAndChart` becomes `updateMiniChartVisibility`.
- **No-provider** emissions leave the sparkline empty instead of crashing (`isHistoryUnavailable` guard).

Mirrors the already-reviewed history branch in `widget-data-chart`.

## Verification

- New `minichart.component.spec.ts` (the component's first tests): asserts the derived History params (path/source/window/sampleTime/points/smoothing), the no-path guard, clean `HISTORY_UNAVAILABLE` handling, and batch-then-live plotting. `widget-numeric`'s change is a pure removal, covered by the compile and the mini-chart's data-path spec.

Part of #64.

### DIFF
--- a/src/app/widgets/minichart/minichart.component.spec.ts
+++ b/src/app/widgets/minichart/minichart.component.spec.ts
@@ -1,0 +1,121 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+// Chart.js and its plugins cannot instantiate under jsdom. Mock them self-contained (no
+// importOriginal, no outer references) so the vi.mock hoisting does not trip a TDZ. The mock
+// Chart carries a truthy `ctx` so the streaming callback's ctx guard passes.
+vi.mock('chart.js', () => {
+  class MockChart {
+    public static register(): void { /* noop */ }
+    public ctx = {};
+    public data: { datasets: { data: unknown[] }[] };
+    public options: unknown;
+    constructor(_ctx: unknown, config: { data: { datasets: { data: unknown[] }[] }; options: unknown }) {
+      this.data = config.data;
+      this.options = config.options;
+    }
+    public update(): void { /* noop */ }
+    public destroy(): void { /* noop */ }
+  }
+  return {
+    Chart: MockChart,
+    TimeScale: {}, LinearScale: {}, LineController: {}, PointElement: {},
+    LineElement: {}, Filler: {}, CategoryScale: {}
+  };
+});
+vi.mock('chartjs-adapter-date-fns', () => ({}));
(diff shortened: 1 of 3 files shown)


mairas:

**Clean — zero defects; the migration is faithful.** Traced old vs new across field fidelity (only `timeScaleFormat` was consumed, always `'minute'` from the orchestrator, now reproduced exactly), emission handling (`isHistoryUnavailable` guard correctly placed; removed `shift()` blocks were already dead), widget-numeric visibility (identical `showMiniChart` signal; template `@if` drives teardown; no dataset leak since the old dataset was `serialize:false`), the theme-effect guard (`if (this.chart)` is strictly safer, can't NPE), and the `startChart` gate (same net outcome, no other callers).

**One intended behavior change (recorded for awareness, not a defect):** the mini-chart now **backfills ~12s on mount**. The old recorder started blank because it gated history-seeding on `sampleTime ≥ 1000ms` and the mini-chart derives 100ms — so it never seeded and filled from live deltas only. `getBackfillThenLive` always backfills, so each visible numeric mini-chart now issues one bounded History-API request at load and shows history immediately. Consistent with the History engine's design (one aggregation query per config change, no polling).

**Testing:** the smoke spec covers the derived params, the no-path guard, clean `HISTORY_UNAVAILABLE`, and batch-then-live plotting. Deeper edge cases (theme-guard re-theme, `showMiniChart` toggle-off teardown, `GAP_POINT` staleness breaks) ride on `HistoryChartStreamService`'s own tests.

PR #165, opened 2026-07-03
feat(charts): stream the wind-trends chart from two History feeds

mairas:
## Why

`widget-windtrends-chart` was the last recorder-only consumer. It read two pre-recorded datasets (`-twd` direction, `-tws` speed) by UUID from `DatasetStreamService`, with no History-engine path — so deleting the recorder would blank it.

## What

Re-point it onto **two independent `HistoryChartStreamService.getBackfillThenLive` streams**, one per path, composed in the widget:

- **No new engine capability.** The two wind series (`self.environment.wind.directionTrue`, `self.environment.wind.speedTrue`) are separate paths, so two ordinary single-path streams — each with its own backfill + live tail, sharing the realtime time axis — cover it. `HistoryChartStreamService` stays single-purpose.
- **Window/sampling** derive from the widget's configured `timeScale` (period 30) via the shared `resolveWindowMs` + `deriveDataSourceInfo` — the same values the recorder used (the orchestrator created both datasets with `timeScale`/period 30).
- **TWD's circular domain** is auto-resolved from the path's unit (no explicit override), matching the recorder's behavior.
- Drops the orchestrator dataset sync (`syncWindTrendsDatasets`) and the `-twd`/`-tws` UUID lookups. A no-provider emission on either stream leaves that series empty (`isHistoryUnavailable` guard). The manual `maxDataPoints` trim is preserved.

Mirrors the minichart migration (#164) with two streams.

## Verification

- New `widget-windtrends-chart.component.spec.ts` (the widget's first tests): asserts two streams open with the direction/speed paths and derived window params (`Last 30 Minutes` → 1,800,000ms / 3600 / 500 / 125), clean `HISTORY_UNAVAILABLE` on either stream, and batch plotting for both series.

Part of #64.

### DIFF
--- a/src/app/widgets/widget-windtrends-chart/widget-windtrends-chart.component.spec.ts
+++ b/src/app/widgets/widget-windtrends-chart/widget-windtrends-chart.component.spec.ts
@@ -1,0 +1,126 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+// Chart.js cannot instantiate under jsdom. Mock it self-contained (no importOriginal / outer refs)
+// so vi.mock hoisting does not trip a TDZ. The mock Chart carries a truthy `ctx`.
+vi.mock('chart.js', () => {
+  class MockChart {
+    public static register(): void { /* noop */ }
+    public ctx = {};
+    public data: { datasets: { data: unknown[] }[] };
+    public options: { plugins?: Record<string, unknown>; scales?: Record<string, unknown> } = {};
+    constructor(_ctx: unknown, config: { data: { datasets: { data: unknown[] }[] }; options: unknown }) {
+      this.data = config.data;
+      this.options = (config.options ?? {}) as typeof this.options;
+    }
+    public update(): void { /* noop */ }
+    public destroy(): void { /* noop */ }
+  }
+  return {
+    Chart: MockChart,
+    TimeScale: {}, LinearScale: {}, LineController: {}, PointElement: {},
+    LineElement: {}, Filler: {}, Title: {}, SubTitle: {}
+  };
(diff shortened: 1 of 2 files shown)


mairas:

**Clean — zero defects; the migration is faithful.** Traced all six areas: every `datasetConfig.timeScaleFormat` → `this.timeScale` read site is reached only after `startWidget` sets it (no null read); `needsRebuild` reproduces the old rebuild/first-run logic; the `startWidget` guard is equivalent; both subscriptions place the `isHistoryUnavailable` guard before the branch and keep the manual `maxDataPoints` trim byte-for-byte; TWD's `'direction'` domain resolves identically (the recorder set **no** angle override for these datasets either — verified against `syncWindTrendsDatasets` + its spec), and `unwrapAngles`/`transformRows` are untouched; `ngOnDestroy` cleanup is intact.

**Correction to the PR description:** this PR removes the widget's **call** to `syncWindTrendsDatasets`, not the method. `WidgetDatasetOrchestratorService.syncWindTrendsDatasets` is now dead code — it's removed with the whole orchestrator in PR6. (`DashboardHistorySeriesSyncService.mapWindTrendsWidget` still reconciles the two wind paths independently, so history-series sync is unaffected.)

**Note (matches old behavior, not a regression):** the backfill-array branch pushes the full History batch without a trim, relying on the server honoring the requested resolution — identical to the old recorder's array branch.

**Testing gaps (accepted at smoke-spec level):** the timeScale-change rebuild path, the live-branch trim, and `unwrapAngles` value-correctness on history-sourced points are not asserted; angle handling rides on `unwrapAngles` being untouched + the shared engine's own tests.

PR #166, opened 2026-07-03
Retire useSharedConfig; fix v11 upgrade loop and stale export (#17 PR-2)

mairas:
PR-2 of the #17 settings.service rewrite (plan v2.1 in the issue body): retire the `useSharedConfig` flag. The flag has been frozen `false` since #82 removed its toggle UI, nothing in settings.service branches on it, and its six surviving peripheral read sites hid two live defects — both fixed here with tests written red-first against the defective behavior.

## The two defect fixes

- **v11 upgrade boot loop**: a same-origin install holding a v11 server profile transformed the config in memory but wrote the result *only to localStorage* before reloading — the server profile stayed v11, so the upgrade re-triggered on every boot, forever. The upgrade now runs the server pipeline for every install: `listConfigs(11)` → back up each slot to 11.99 → `upgradeConfig` → `setConfig`, so all v11 slots upgrade, not just the active one. Anti-loop test asserts the v12 result lands via `setConfig`.
- **Stale config export**: backup/export read localStorage (where nothing has been written since #82) while all writes go to the server. `getActiveConfig` now always builds from the server-backed config; the orphaned `getLocalConfigFrom*` helpers are deleted. Red-run proof: the old path exported a planted stale blob.

## The retirement itself

- Field removed from `IConnectionConfig`, `DefaultConnectionConfig`, settings.service (hydration, write-through, serialization). Stray keys in stored connection blobs are ignored on load — no connection-version bump needed.
- The 404 classification is un-gated: a genuine storage 404 now classifies as `missing-shared-config` and reaches the recovery toast (previously flag-false installs degraded to `unknown` with no recovery path).
- `migrateRemoteControlToDevice` collapses exactly to its pinned-false paths (the flag-true "defer the lift" branch was unreachable).
- Dead legacy-upgrade machinery removed (first commit): `probeLegacyUpgradeAvailability` probes fileVersion-9 configs under the `skip` appid, but v9/v10 configs only ever existed under upstream KIP's `kip` appid — it could never return true, making the "Upgrade" recovery-toast variant dead UI. The live "Create default" variant is untouched. `upgradeDashboardWidgets` went with its sole (deleted) caller.
- Forced updates: the four flag-referencing specs, the `src/test.ts` stub literal, and the perf-harness config seeder.

Deliberately kept: `v10-config-interface.ts` still models the historical stored shape (it describes old data, not current behavior); the v10 localStorage upgrade arm survives per the plan's decision 5.

## Verification

`npm run lint` clean; full suite 135 files / 960 tests green (main baseline 958, −1 collapsed shared/local matrix duplicate, +3 corrective tests); `npm run build:dev` clean. All three corrective tests verified red against the pre-change code.

Refs #17 (PR-2 of 4; plan and sequencing in the issue body).

### DIFF
--- a/perf-harness/lib/kip-config.mjs
+++ b/perf-harness/lib/kip-config.mjs
@@ -42,7 +42,7 @@ export function connectionConfig(subscribeAll = false) {
     configVersion: 13, kipUUID: '00000000-0000-4000-8000-000000000001',
     signalKUrl: '__ORIGIN__', // replaced with the served origin at inject time
     proxyEnabled: false, signalKSubscribeAll: subscribeAll,
-    useSharedConfig: true, sharedConfigName: 'default',
+    sharedConfigName: 'default',
     isRemoteControl: false, instanceName: '',
   };
 }
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -216,28 +216,17 @@ export class AppComponent implements OnInit, AfterViewInit, OnDestroy {
 
       this.missingConfigPromptShown = true;
       const cfgName = issue.sharedConfigName || 'default';
-      const canUpgradeLegacy = issue.legacyUpgradeAvailable === true;
-      const actionLabel = canUpgradeLegacy ? 'Upgrade' : 'Create';
-      const message = canUpgradeLegacy
-        ? `Server is reachable, but no current version shared configuration '${cfgName}' exists for this user. A legacy configuration was found and can be upgraded. Upgrade now?`
-        : `Server is reachable, but no shared configuration '${cfgName}' exists for this user. Create a default configuration now?`;
       const ref = this.toast.show(
-        message,
+        `Server is reachable, but no shared configuration '${cfgName}' exists for this user. Create a default configuration now?`,
         0,
         true,
         'warn',
-        actionLabel
+        'Create'
       );
 
       ref.onAction()
         .pipe(takeUntilDestroyed(this._destroyRef))
-        .subscribe(() => {
-          if (canUpgradeLegacy) {
-            this.upgrade.runUpgrade();
-          } else {
-            this.settings.resetSettings();
-          }
-        });
+        .subscribe(() => this.settings.resetSettings());
     });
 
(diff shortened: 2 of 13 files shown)


mairas:

Six reviewer personas (correctness, maintainability, testing, project-standards, adversarial, reliability) reviewed the diff independently; deduplicated findings were each challenged by an adversarial verifier briefed that flag-true states are unreachable in real deployments. Five findings survived — none critical, none refuted. All are now addressed, amended into the retirement commit (`4b3c2c2a`).

### Behavior findings (fixed, both proven red-first)

1. **404 un-gating over-matched** *(minor — correctness + adversarial)*: the un-gated `status === 404` catch classified *any* bootstrap 404 as `missing-shared-config` — including a 404 from `GET /signalk/` discovery (rethrown through `handleError`), which would show the misleading "Create a default configuration?" toast for a problem `resetSettings()` cannot fix. **Fix:** the classification now lives in an inner catch on the config fetch itself; all other 404s fall through to `unknown` (the pre-PR behavior). Pinned by a new discovery-404 test, verified red.
2. **v11 upgrade could wedge the blocking overlay** *(minor — reliability)*: a `listConfigs(11)` rejection reached the outer catch, which only pushed an error — `upgrading` stayed true and the full-screen spinner (no dismiss affordance) wedged the session. Newly reachable, since the server pipeline is now the live path. **Fix:** the outer catch clears `upgrading`; deliberately no reload there (a persistent failure would otherwise boot-loop — the server still holds v11, so the upgrade retries next boot). Pinned by a new failure-path test, verified red.

### Test/comment findings (fixed)

3. **Stale call-site comment** *(minor, found by two personas)*: the `migrateRemoteControlToDevice` call site still described the deleted "skipped on degraded boot, retries later" deferral — updated to the new profile → legacy-blob → defaults semantics.
4. **Anti-loop test didn't pin reload-once** *(nit)*: added a fake-timer test asserting `reloadApp` fires exactly once after the v11 pipeline.

Suite after fixes: lint clean, 135 files / 963 tests green (963 = main's 958 − 1 collapsed matrix duplicate + 6 new tests).

PR #167, opened 2026-07-03
fix(charts): derive backfill SMA circularly, matching the live tail (#162)

### LINKED ISSUE #162: History backfill SMA overlay uses arithmetic mean — wrong at 0/360° for angle paths (needs circular smoothing)
## Problem

The History chart engine's **backfill SMA overlay** is an arithmetic moving average, which is wrong at the 0/360° wrap for direction/angle paths — the same bug #87 fixed for the Value column, but on the smoothing line.

- Backfill SMA comes from the bundled provider's arithmetic window mean (`kip-plugin/src/sqlite-history-storage.service.ts:1166-1177`, then arithmetic bucket aggregation at `:1240-1241`). The client mapper only *wraps* that value (`history-to-chart-mapper.service.ts:109`) — wrapping cannot undo an arithmetic average, so near a 0/360 crossing the SMA line dips toward ~180°.
- The **live tail** computes SMA circularly (`chart-stats.util.ts` `circularMeanRad`). So on an angle chart the smoothing line is arithmetic-and-wrong in the backfill half and circular-and-correct in the live half: a wrong dip plus a step discontinuity at the backfill→live seam.

## Why it surfaced now

PR #161 (closes #87/#132) switched the **Value** line to `:last` (raw per-bucket sample), which is angle-correct. That makes the still-arithmetic SMA overlay *visibly disagree* with the corrected Value line at every wrap crossing — previously both dipped to ~180° together, so the error was hidden.

## Fix direction

Compute the backfill SMA client-side from the raw `:last` samples using the same domain-aware smoothing the live tail uses (`chart-stats.util`), and stop requesting the server-side arithmetic `:sma` column. That unifies backfill and live on one stats path and removes the arithmetic-SMA provider dependency entirely — consistent with #64's shared-stats-util goal.

## Gate

**Blocks the default engine flip (#64 PR5).** The History engine is default-OFF today, so this artifact reaches no production user yet — but it must be fixed before `chartEngine: 'history'` becomes the default, or direction/heading/wind-angle trend charts ship a wrong, discontinuous smoothing line.

Related: #87, #132, #64.

mairas:
## Why

PR #161 fixed the History backfill **Value** line (`:last`, angle-correct), but left the **SMA overlay** on the server's arithmetic `:sma` column. On direction/heading/wind-angle charts that mean is wrong at the 0/360° wrap and discontinuous with the live tail's circular SMA — and #161 made the mismatch conspicuous by fixing the Value line beside it. This is the gate on promoting the History engine to default (PR5).

## What

Derive the backfill SMA **client-side** instead of requesting it from the server:

- `fetchBackfill` now requests only `:last` (drops `:sma:N`).
- Each backfill point's `sma` is computed via the shared `computeWindowStats(values.slice(0, i+1), smoothingPeriod, domain)` — the **exact same function the live tail uses** over its buffer. So the backfill SMA is circular for angle domains and identical to what the live tail produces over the same samples: wrap-correct and seam-free.
- **Scalar paths unchanged** — `computeWindowStats` returns an arithmetic mean for `scalar`, so non-angle charts get the same SMA values, now computed client-side.

No server or `kip-plugin` change — the provider still supports `:sma`; the chart path just no longer asks for it.

## Verification

- Updated stream spec: backfill requests `:last` only (no `:sma`/`:avg`/`:min`/`:max`); a new test proves the direction-path SMA is the **circular** mean (~0.0175 rad for samples straddling the wrap), not the arithmetic mean (~2.11 rad) the server `:sma` would have produced.

Closes #162. Unblocks PR5 (the default-engine flip).

### DIFF
--- a/src/app/core/services/history-chart-stream.service.spec.ts
+++ b/src/app/core/services/history-chart-stream.service.spec.ts
@@ -58,15 +58,48 @@ describe('HistoryChartStreamService', () => {
     expect(Array.isArray(first)).toBe(true);
     expect((first as IDatasetServiceDatapoint[]).map(p => p.data.value)).toEqual([1, 3]);
     expect(history.getValues).toHaveBeenCalledTimes(1);
-    // The Value series is the raw per-bucket `:last` sample (angle-safe, seam-free with the live
-    // tail), plus the `:sma` overlay. The dead `:min`/`:max` columns are no longer requested.
+    // Only the raw per-bucket `:last` sample is requested; the SMA overlay is derived client-side, so
+    // no `:sma`/`:avg`/`:min`/`:max` aggregate columns are asked of the server (#162).
     const query = history.getValues.mock.calls[0][0];
     expect(query.paths).toContain(':last');
-    expect(query.paths).toContain(':sma:');
+    expect(query.paths).not.toContain(':sma');
     expect(query.paths).not.toContain(':avg');
     expect(query.paths).not.toContain(':min');
     expect(query.paths).not.toContain(':max');
     expect(query.from).toBeTruthy();
+  });
+
+  it('derives the backfill SMA circularly for direction paths (no server :sma)', async () => {
+    history.getValues.mockResolvedValue({ context: 'vessels.self', range: {}, values: [], data: [] });
+    // Direction samples straddling the 0/360° wrap: ~355°, ~5°, ~3°.
+    mapper.mapValuesToChartDatapoints.mockReturnValue([
+      { timestamp: 1000, data: { value: 6.19591884457987 } },
+      { timestamp: 2000, data: { value: 0.08726646259971647 } },
+      { timestamp: 3000, data: { value: 0.05235987755982989 } }
+    ]);
+    const params: IHistoryChartStreamParams = { ...PARAMS, angleDomainOverride: 'direction', smoothingPeriod: 3 };
+    const first = await firstValueFrom(make().getBackfillThenLive(params));
+    const points = first as IDatasetServiceDatapoint[];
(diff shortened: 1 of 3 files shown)


mairas:
## Review synthesis (adversarial verification)

**No correctness defects.** The circular SMA is verified: `windowSma` over the trailing window gives the exact circular mean (`0.01749591…` for the wrap-straddling test), circular-mean is wrap-invariant (so the mapper's pre-wrapping + the util's re-wrap is a genuine no-op), and the backfill SMA is continuous with the live tail by construction — both now call the same `windowSma` over the same trailing window of the same value buffer.

Two substantive items were raised and handled:

- **Performance (medium) — fixed in this PR.** The first pass called `computeWindowStats(values.slice(0,i+1), …)` per point, which recomputes whole-prefix min/max/avg (a sort + full-prefix trig) that the caller discards — O(n²·log n), a real freeze risk at load when several trend widgets backfill on the Pi-class target. Replaced with `windowSma` over only the trailing smoothing window: O(n·smoothingPeriod), no sort, no wasted aggregates. `computeWindowStats` now uses `windowSma` too, so the live tail and backfill stay identical.
- **Scalar semantics (low) — intended, noted.** The old server `:sma` was a moving average over N *raw* samples (then bucket-averaged); the client SMA is over N *bucket* (`:last`) values. They differ only when the source samples faster than the chart resolution. This is the correct direction: `smoothingPeriod` is defined as "trailing points" (buckets), and the live tail already smooths over buckets — so the old server behavior was the inconsistent one, and this removes a latent scalar seam jump.

**Tests added:** the direction-path circular SMA (vs the ~2.11 rad arithmetic result), and a scalar trailing-window SMA that also exercises the early clamp (point 0 → 1-element window).

PR #169, opened 2026-07-03
feat(charts): make History the default engine, drop the recorder branch

mairas:
## Why

All recorder-only consumers are migrated (minichart #164, windtrends #165) and the History backfill is now angle-correct end to end (`:last` #161, circular SMA #162). So the History engine can become the default, and `widget-data-chart`'s opt-in A/B branch — the last recorder code path — can go.

## What

In `widget-data-chart`:

- **Flip** `DEFAULT_CONFIG.chartEngine` to `'history'`.
- **Remove** the `chartEngine` A/B `computed` and its `skip.chartEngine` localStorage override — the widget no longer chooses an engine.
- **Delete** the recorder branches in `rebuildForDataset` (the `syncDataChartDataset` + `getDatasetConfig`/`getDataSourceInfo` path) and `startStreaming` (the `getDatasetBatchThenLiveObservable` tail). Both now unconditionally use the History path.
- **Drop** the `DatasetStreamService` and `WidgetDatasetOrchestratorService` injections/imports (no longer referenced here). Renamed the now-misleading `dsServiceSub` field to `streamSub`.

Deliberately **left for PR7** (the config migration): the vestigial `chartEngine` config field, the `CHART_ENGINE_OVERRIDE_KEY` constant in `config-storage.const.ts`, and clearing any stale `skip.chartEngine` localStorage value.

After this, `widget-data-chart` no longer references the recorder — **PR6 can delete `DatasetStreamService` + the orchestrator.**

## Depends on

Must merge **after #167** (the circular-SMA fix, #162) — already merged — so the now-default engine's smoothing line is angle-correct.

## Verification

- Spec updated: the four engine-dispatch/override tests collapse to one ("streams from the History engine — the only engine"); the `HISTORY_UNAVAILABLE` empty-state test stays. Recorder mocks/providers removed.

Part of #64.

### DIFF
--- a/src/app/widgets/widget-data-chart/widget-data-chart.component.spec.ts
+++ b/src/app/widgets/widget-data-chart/widget-data-chart.component.spec.ts
@@ -30,8 +30,6 @@ import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { EMPTY, Subject } from 'rxjs';
 import { WidgetDataChartComponent } from './widget-data-chart.component';
 import { IWidgetSvcConfig } from '../../core/interfaces/widgets-interface';
-import { DatasetStreamService } from '../../core/services/dataset-stream.service';
-import { WidgetDatasetOrchestratorService } from '../../core/services/widget-dataset-orchestrator.service';
 import { HistoryChartStreamService, HISTORY_UNAVAILABLE } from '../../core/services/history-chart-stream.service';
 import { WidgetRuntimeDirective } from '../../core/directives/widget-runtime.directive';
 import { UnitsService } from '../../core/services/units.service';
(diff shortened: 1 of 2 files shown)


mairas:

**Clean — zero defects; the flip is faithful.** Traced: the retained History branch populates both `datasetConfig` (synthesized inline) and `dataSourceInfo` synchronously before every downstream read (`setChartOptions`/`startStreaming`/`createDatasets`), so no code now reads an undefined config; the dropped `if (!this.datasetConfig) return` guard was recorder-async-specific and unreachable on the History path; `pathSignature` dropping `cfg.chartEngine` only removes a now-constant term; the `dsServiceSub`→`streamSub` rename is consistent across all 5 sites; and no `chartEngine()`/`dsService`/`datasetLifecycle`/`CHART_ENGINE_OVERRIDE_KEY` references remain.

**Deferred to PR7 (as planned):** `CHART_ENGINE_OVERRIDE_KEY` in `config-storage.const.ts` is now an unused export, folded into the same `chartEngine`-field cleanup + config migration.

**Test added** for the migration-critical behavior the review flagged: a saved `chartEngine:'recorder'` config now routes to the History engine (the field is ignored, not dispatched on).

Issue #170, opened 2026-07-03
pushSettings persist-on-missing writes serialize mid-hydration, reverting later-hydrated fields


mairas:
Surfaced by the #17 PR-4 characterization pass (settings.service.spec.ts, hydration describe).

## Defect

`pushSettings()` fires its persist-on-missing bootstrap writes (missing `autoNightMode` / `redNightMode` / `nightModeBrightness`) **mid-hydration**: each write serializes the full `IAppConfig` from the reactive holders at that moment, so fields hydrated *after* the night-mode checks — notably `browserTabTitle` — ride along with their pre-hydration defaults.

Concrete loss: a stored profile that lacks `autoNightMode` but carries a custom `browserTabTitle` gets a bootstrap patch that writes `browserTabTitle: 'SKip'` back to the server — the custom title is silently reverted on the next reload.

Only sparse configs are exposed (persist-on-missing fires only when a night-mode field is absent), which is why this has gone unnoticed.

## Fix direction

Hydrate all fields into the reactive holders first, then fire any persist-on-missing writes (or coalesce them into one write built from fully-hydrated state). Deliberately **not** folded into the #17 PR-4 rewrite, which is bound to bit-for-bit hydration behavior preservation; fix as a follow-up once PR-4's single write path has landed (the fix becomes a one-line ordering change there). The characterization suite pins only order-safe blob content, so the fix will not invalidate it.

PR #160, opened 2026-07-03
Drop dead storage-routing code from settings.service (#17 PR-1)

mairas:
PR-1 of the #17 settings.service rewrite (plan v2.1 in the issue body): the behavior-preserving dead-code sweep. Pure deletion — 259 lines removed, 58 added (all from unwrapping live branches).

## What's removed

- The `useServerStorage = true` constant and all 12 `if`/`else` shells around it: live server paths kept verbatim (including the `storageServiceReady$` guards), dead localStorage else-branches deleted.
- The three unreachable save helpers (`saveAppConfigToLocalStorage`, `saveThemeConfigToLocalStorage`, `saveLDashboardsConfigToLocalStorage`). `saveConnectionConfigToLocalStorage` stays — device-scope persistence is localStorage by design (4 live call sites).
- `replaceConfig` — its sole caller was `loadDemoConfig`'s dead else branch (found during the sweep, folded in).
- Vestigial `widgets` array + `getWidgets()` (always returned `[]`, zero callers).
- `StorageService.patchGlobal` (zero production callers) with its 3 spec assertions — the only spec edits in this PR; they tested the removed method.
- widget-anchor-alarm's injected-but-never-used `SettingsService`; main.ts's redundant `SettingsService` provider entry (`providedIn: 'root'`). `UnitsService`/`AppNetworkInitService` deliberately untouched — bare `@Injectable()`, main.ts is their sole registration.

Note: the plan's `getWidgetHistoryDisabledAsO` item was already satisfied by #157, which removed the whole field between plan-writing and this PR.

## Verification

- `npm run lint` clean; full suite 135 files / 956 tests green — exactly 3 fewer than main's 959, matching the removed patchGlobal tests; every other spec unmodified.
- `npm run build:dev` completes (catches DI/template breakage the specs might miss, e.g. the main.ts provider removal).

Refs #17 (PR-1 of 4 + fast-follow; sequencing and full plan in the issue body).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #79, opened 2026-07-01
Migrate settings.service consumers to signals; retire the *AsO() observable bridge


mairas:
Fast-follow to #17 (settings-store rewrite), non-gating. Decided per #17 open-question A1: the `*AsO()` observable getters are a **transitional bridge**, not a permanent API.

After #17's core rewrite lands the internal signals plus `toObservable(signal)` bridges, migrate the reactive consumers off the bridge:
- The ~7 consumers that do `toSignal(getXAsO())` (app.component, menu-actions, app-service ×3, remote-dashboards ×2) read the readonly signal directly — deletes the signal→observable→signal round-trip.
- The two `.subscribe()` side-effect consumers (`units.service`, `notifications.service`) convert to `effect()`.
- Remove each `*AsO()` method once it has no remaining references; update the `SettingsServiceStub` in `src/test.ts` accordingly.

Each edit is individually CI-verifiable. Do not start until #17's core PR has merged. Tracked in #76.

mairas:
Unblocked: #17 landed (PR #172). Scope is now exactly: convert the two remaining `.subscribe()` consumers (units.service `getDefaultUnitsAsO`, notifications.service `getNotificationServiceConfigAsO`) to `effect()` over the signal accessors, then delete those two bridges and their sync-replaying backing subjects from settings.service. The real-service bridge tests in settings.service.spec.ts pin the current sync-emission contract and get deleted with the bridges.

Issue #159, opened 2026-07-03
src/test.ts global app-class stubs are DI-inert under the vitest builder


mairas:
Discovered during the #17 ground-truth pass (2026-07-03).

## Problem

The global app-class stubs registered in `src/test.ts` (`SettingsServiceStub`, `AuthenticationServiceStub`, `SignalKConnectionServiceStub`, `ConnectionStateMachineStub`, `UnitsService` stub, widget directive stubs) are **DI-inert** under the current `@angular/build:unit-test` setup: specs receive the *real* services.

Mechanism: `src/test.ts` is loaded by vitest as a `setupFiles` entry (vite-transformed), while spec and app code go through the Angular builder bundle — so the class used as a DI token in `test.ts` and the class injected by components are different module instances. Only framework-token providers (`MAT_DIALOG_DATA`, `MatDialogRef`, `ActivatedRoute`, zoneless CD, `HttpClientTesting`, icon registration) actually take effect.

## Evidence

- Verbose runs show the real `SettingsService` constructor log in `app.component.spec` and `menu-actions.component.spec`.
- `AppComponent` calls `getBrowserTabTitleAsO()` at construction (`app.component.ts:70`) — a method `SettingsServiceStub` does not have — and the spec passes.
- The stub has silently drifted (also missing `getConfigVersion`, `getThemeConfig`, `getActiveProfileName`, `resetSettings`, …) with no spec failing.

## Consequences

1. The repo CLAUDE.md "Testing reality" guidance ("the most common spec failure is stub drift — fix the stub, not app code") describes a mechanism that no longer operates for app tokens and should be rewritten.
2. Specs currently exercise real services more than intended — arguably good (the #17 review found this catches the `requireSync` NG0601 hazard), but it is accidental, and construction-time resilience of real services is now load-bearing for ~30 spec files.
3. Restoring stub effectiveness naively (e.g. compiling `test.ts` through the builder via `--setup-files`) would break specs until the drifted stubs are extended — verified: `app.component.spec` fails with `getBrowserTabTitleAsO is not a function`.

## Suggested direction

Decide deliberately: either (a) delete the inert app-token stubs and embrace real-services-plus-spec-local-mocks as the documented model, or (b) restore global stub effectiveness and pay down the stub drift. Update the CLAUDE.md testing section to match reality either way. Verify the same inertness holds in CI before changing anything.

PR #172, opened 2026-07-03
Rewrite settings.service: signals-first core over a single write path (#17 PR-4)

mairas:
PR-4 of the #17 settings.service rewrite — the core: signals-first internals behind a single persistence path, closing RED-05 (duplicated per-setter persistence), ANG-02 (BehaviorSubjects → signals), and the PCS-08 remainder. Plan v2.1 in the issue body; PRs #160/#166/#168 prepared the ground.

## Structure: characterize first, then rewrite against the oracle

- **`test(settings): characterize persistence core`** — 24 new tests pinning current behavior *before* any rewrite: configVersion preservation on every write path, resetSettings (including today's silent no-op when storage isn't ready), the persist-on-missing hydration writes, the two untested granular patch mappings, the migrate-fallback identity read, and the `getDefault*` localStorage side effects. Three pins are explicitly flagged as Phase-B-mutable; everything else survives the rewrite byte-identical.
- **`refactor(settings): signals-first core, one write path`** — the equivalence-oracle commit: all ten BehaviorSubjects become signals with readonly public accessors; every app-scope setter funnels through two private helpers (`saveAppConfig` for the whole blob with version preservation, `saveConfigSection` for granular writes) — the single write path that makes the split-brain bug class unrepresentable. Six `*AsO()` getters deleted and their eight `toSignal` consumer sites migrated to direct signal reads (menu-actions, app-service ×3, app.component, remote-dashboards ×2, toast.service → `computed`). The two `.subscribe()` consumers (units, notifications) keep their `*AsO()` bridges, backed by synchronously-replaying sources fed from the same write path — real-service tests pin the synchronous first emission (bare `toObservable` would emit async-first and break `notifications._notificationConfig`). The characterization suite passes unmodified against this commit.
- **`fix(settings): loud reset failure, pure defaults`** — the three corrective changes, each proven red-first against the rewritten core: `resetSettings` with storage not ready now surfaces a MatSnackBar error instead of silently doing nothing; the dead localStorage side-effect writes are stripped from `getDefaultAppConfig`/`getDefaultThemeConfig`/`getDefaultDashboardsConfig` (the connection-config default write stays — live device scope, pinned); `getDefaultThemeConfig` returns a fresh clone instead of aliasing the shared const.
- **`refactor(storage): drop dead widget/layout patches`** — the `IWidgetConfig`/`ILayoutConfig` patch cases and union members go (zero callers; the `never` exhaustiveness check compiles over the narrowed union).

## Public-surface delta

Added: readonly signal accessors for the nine reactive settings. Removed: the six migrated `*AsO()` getters (and their `src/test.ts` stub members). Unchanged: all sync `getX()`/`setX()` (now thin signal reads), `KipUUID`, `configUpgrade`, `proxyEnabled`/`signalKSubscribeAll` (still externally read), the two retained bridges, and the whole device/connection path.

Not in this PR: #170 (the pre-existing mid-hydration serialization quirk found while characterizing — hydration here is bit-for-bit); #79 remains the fast-follow that converts the two `.subscribe()` consumers to `effect()` and retires the last bridges.

## Verification

Lint clean; `tsc --noEmit` passes on both app and spec tsconfigs (the narrowed union and deleted getters compile everywhere); full suite 137 files / 996 tests green after every commit; `npm run build:dev` clean. Red-run evidence for the corrective commit: exactly the three flagged pins failed against the rewritten core, then went green with the fixes.

Refs #17 (PR-4 of 4 — closes out the rewrite; #79 and #170 are the tracked follow-ups).


mairas:

Eight reviewer personas (correctness, maintainability, testing, project-standards, adversarial, api-contract, reliability, security) reviewed the diff independently; deduplicated findings were adversarially verified, with signal-semantics claims checked against actual Angular behavior rather than folklore. Notably, the correctness reviewer's deep-check for signal-vs-subject semantic drift (reference-equality dedupe, emission timing) across all migrated consumers came back clean, as did reliability's failure-path traces and api-contract's zero-stragglers sweep of the six deleted getters.

**Two findings survived — the same gap, found from two angles** (adversarial + api-contract): the global `SettingsServiceStub` in `src/test.ts` lost its deleted `*AsO()` members but never gained the nine new signal accessors, so it stopped mirroring the real public surface. Currently masked by the stub's DI-inertness (#159), but that made the suite green only accidentally — and this PR extended the drift.

**Resolution (amended into the rewrite commit, `61e0534e`):** the stub now exposes all nine readonly signal accessors with the same defaults as the real service, and its test-utility setters drive the signals alongside the legacy subjects, so the double stays faithful whichever way #159 is eventually resolved.

**Rejected by verification** (examples): a third report of the same stub gap was refuted as stated because it assumed the consumer specs inject the stub — a probe proved they receive the real service; a `saveConfigSection` docstring nit didn't survive scrutiny.

Suite after the fix: lint clean, 137 files / 996 tests green, `tsc --noEmit` passing on app and spec configs.

PR #173, opened 2026-07-03
feat(config): migrate app config to v13, retiring the recorder's dataset model

mairas:
## What

PR7a of the #64 recorder-retirement epic (the migration half; the dataset-config UI removal is PR7b). Retires the client-side recorder's now-dead config footprint and migrates stored configs **v12 → v13**.

- **v12→v13 migration** in `ConfigurationUpgradeService` + its `app.component` trigger arm: drops the app-level `dataSets` registry and strips residual per-widget `datasetUUID`/`chartEngine`, stamping v13. `MIGRATION_OUTPUT_VERSION` stays pinned at 12 (legacy transforms keep stamping 12; the new step stamps 13 itself, so a future bump re-raises the upgrade flag rather than relabeling).
- Bump `LATEST_APP_CONFIG_VERSION` 12→13 and the dashboard schema `configVersion`.
- Remove `IAppConfig.dataSets` and all its plumbing (settings field/read/write, `getDataSets`/`saveDataSets`, the storage `Array<IDatasetDef>` patch case, the demo/blank defaults, the test stub) — nothing consumed it after the recorder was deleted.
- Remove the vestigial `chartEngine` field and the dead `CHART_ENGINE_OVERRIDE_KEY`.
- Delete the legacy dataset-migration code (`extractAppDatasets`/`migrateDatasetsToDataCharts`): it only fed the pre-fork Kip named-dataset model, unreachable under Skip's own storage namespace (`SERVER_CONFIG_APPID='skip'` + `skip.*` localStorage) and not a supported input.

`IDatasetServiceDatasetConfig` is kept — still the chart-widget config type.

## Verification against reality (the handoff plan was materially stale)

A cold verification pass corrected several plan assumptions before implementation: there's no generic upgrade loop (a hardcoded dispatcher + an `app.component` trigger that only handled v11 — a v12 bump would otherwise strand configs on the upgrade gate); `settings.service` has no own version constant; there is **no** StorageService version-write-suppression to guard against; the schema has no `app`/`dataSets`/`chartEngine` keys to drop; and `buildAppStorageObject` re-serializes `dataSets` on every write (the repopulation trap — handled by removing the field).

## Migration chain

v11 configs chain to v13 across two boots (v11→v12 stamps 12 → reload → v12→v13 stamps 13). Import (`profile.service.importProfile`) still rejects non-current versions by design; in-place stored profiles auto-migrate.

## Tests

- `npm run lint` — clean. `npm run build:dev` — green.
- New v12→v13 migration tests (TDD: drop registry, strip vestige, stamp v13, reload-once, skip non-v12).
- Green locally: configuration-upgrade (10), settings (22), profile (32), config-component (10), storage (33), widget-data-chart (2), app.component (1). Version-gate seed bumps applied to profile/config-component/settings specs.

Part of #64.

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -103,7 +103,7 @@ export class AppComponent implements OnInit, AfterViewInit, OnDestroy {
       if (this.settings.configUpgrade()) {
         const liveVersion = this.settings.getConfigVersion();
 
-        if (liveVersion === 11) {
+        if (liveVersion === 11 || liveVersion === 12) {
           this.upgrade.runUpgrade(liveVersion);
         }
 
--- a/src/app/core/components/options/configuration/config.component.spec.ts
+++ b/src/app/core/components/options/configuration/config.component.spec.ts
@@ -151,7 +151,7 @@ describe('SettingsConfigComponent', () => {
       getThemeConfig: ReturnType<typeof vi.fn>;
       loadConfigFromLocalStorage: ReturnType<typeof vi.fn>;
     };
-    const serverApp = { configVersion: 12, browserTabTitle: 'Server' };
+    const serverApp = { configVersion: 13, browserTabTitle: 'Server' };
     settings.getAppConfig.mockReturnValue(serverApp);
     settings.getDashboardConfig.mockReturnValue([{ id: 'dash-1' }]);
     settings.getThemeConfig.mockReturnValue({ themeName: 'dark' });
--- a/src/app/core/constants/config-storage.const.ts
+++ b/src/app/core/constants/config-storage.const.ts
@@ -28,6 +28,3 @@ export const SSO_REDIRECT_BUDGET_KEY = 'skip.ssoRedirectAttempts';
 
 /** Gesture-diagnostics debug flag key. */
 export const GESTURES_DEBUG_KEY = 'skip.gesturesDebug';
-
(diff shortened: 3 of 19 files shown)


mairas:
## Rebased onto #172 + independent review

**Reconciled with #172** (`refactor/17-core-rewrite`, the signals-first settings/storage persistence rewrite, which landed while this was in flight). #172 kept the `dataSets` registry and added persistence-characterization tests for it; this PR removes the registry. Resolution: adopted #172's structure (signal accessors, `applyUnitDefaults`/`applyNotificationConfig`, one write path, dropped `IWidgetConfig`/`ILayoutConfig`) and additionally stripped every `dataSets` touchpoint — the `pushSettings` read, `buildAppStorageObject` field, `getDataSets`/`saveDataSets`, the `Array<IDatasetDef>` patch type + case, and #172's new dataSets characterization tests in the settings and storage specs. The migration logic (configuration-upgrade + trigger) was untouched by the rebase.

**Independent adversarial review: 0 findings** across three lenses (migration correctness & reachability, removal completeness / the repopulation trap, legacy-migration regression), each with per-finding adversarial verification. The v11→v12→v13 chain, the trigger wiring, and the registry removal all verified clean.

Re-verified post-rebase: `build:dev` green, `lint` clean, and specs green — settings (44), storage (34), configuration-upgrade (10), app.component (1), plus profile/config-component/widget-data-chart from before.

Issue #183, opened 2026-07-04
Navigation UX redesign: auto-hiding toolbar, natural-scroll nav, unified widget gestures


mairas:
## Navigation UX redesign (tracking)

Replace Skip's gesture-driven, undiscoverable navigation with a discoverable, robust model built on natural inputs, keeping edit mode. Full plan lives in the clone at `_local/plans/2026-07-04-001-feat-navigation-ux-redesign-plan.md` (not committed); this epic and its sub-issues carry the executable detail.

### The model

- One **auto-hiding top toolbar** (overlay, no permanent chrome) replaces both swipe-summoned sidenavs. Revealed by vertical scroll-up / swipe-up / wheel-up, or by tapping the transient edge-peek; cued by that edge-peek on any tap or mouse-move; hidden on scroll-down, tap-away, or idle. Auto-hides in edit mode too (never covers the top row being edited).
- A **persistent alarm badge** — the single sanctioned exception to "no permanent chrome" — shows active alarms independent of the toolbar (safety).
- **Page navigation** via horizontal two-finger scroll / swipe / wheel-`deltaX`, plus a persistent page-icon control in the toolbar (each page's assigned icon; tap to jump; no prev/next).
- Widgets get exactly **two gestures**: long-press (locked → open history plot directly, non-core/unadvertised) and single-tap (edit → one action popup). Interactive widgets keep single-tap for their own control in locked mode.
- **Edit mode retained** as the accidental-change guard.
- **Notifications downgraded**; **terminology + route** shift to "page" (`/dashboard/:id` → `/page/:id` with redirect; whole surface stays the "dashboard").

Retired: vertical-swipe page nav, double-tap, right-click, two-finger tap, click-drag mouse "swipe", top-edge pull-down, all permanent nav chrome (except the alarm badge).

### Key decisions

- Two input sources → shared intents: a dedicated `wheel` handler (`deltaX`→page, `deltaY`→toolbar) + repointed touch-swipe axes; both call one `DashboardService` nav API.
- Reveal/peek handler is **shell-level** (not the grid's `editor`-gated `GestureDirective`), suppressed while `isDragging`; edge-peek is tappable to expand.
- Horizontal `wheel` must be non-passive + `preventDefault` (blocks browser back/forward on macOS).
- Vertical handler yields to scrollable widget content before revealing.
- Persistent alarm badge = repurposed `NotificationOverlayService`, visible only while an alarm is active.
[...]

PR #200, opened 2026-07-04
PR-4: dashboards editor single-tap action menu (dashboard editor gesture alignment)

### LINKED ISSUE #188: PR-4: Align Settings → Dashboards editor to the single-tap popup model
**PR-4 — dashboards-editor alignment (separable; may land last or be deferred).**

### Unit 9b: Dashboards-editor gesture alignment
The Settings → Dashboards panel (`dashboards-editor.component.html`) reuses the retired double-tap/long-press vocabulary. Align it to the new single-tap popup model for consistency (R6/R19/R20).

**Files:** Modify `src/app/core/components/dashboards-editor/dashboards-editor.component.{ts,html}`; Test `…spec.ts`.

**Tests:** single-tap → popup for rename/duplicate/delete (no double-tap/long-press); delete still confirms.

Separable: can ship after the core redesign or be dropped if scope tightens, but leaving retired gestures alive here while removing them everywhere else is inconsistent.

mairas:
PR-4 of the navigation UX redesign (#183) — the last sub-issue. The Settings → Dashboards editor was the one surface still using the retired gesture vocabulary: double-tap a page tile to edit, long-press for a Duplicate/Delete bottom sheet. This brings it in line with the rest of the redesign.

## What changes

- **Single tap on a page tile → one responsive action menu** (Edit / Duplicate / Delete) at the tap point, reusing the same `ActionMenuComponent` the widget grid uses (mat-menu on desktop, bottom drawer under 600px).
- **Delete now confirms** via a dialog before removing the page and its widgets (previously the bottom sheet deleted immediately).
- The bespoke `DashboardsBottomSheetComponent`, the press/drag-arbitration state it needed (`_sheetOpen`, `suppressDrag`, `cancelPointerSequence`, …), and two dead fields (`pageTitle`, `DomSanitizer`) are removed.

## Tap vs. drag

The tiles are `cdkDrag` reorderable and now also carry `kipGestures (tap)`. Tap-vs-drag is left entirely to the gesture directive's movement threshold: a real reorder moves past the tap slop (20px touch / 8px mouse) and emits no tap, so it never opens the menu. There is deliberately **no** `isDragging` guard on the tap handler — gating on the shared drag signal would swallow a legitimate tap whose minor travel already tripped cdkDrag's lower 5px start threshold (common on a rocking marine display), which review flagged as a real defect.

## Review

Five independent persona reviewers (correctness, adversarial, maintainability, testing, project-standards). One genuine defect (the tap-swallow above) fixed; the `private`-visibility and test-coverage findings applied; no standards violations. The positional `_menuIndex` capture and the app-global `isDragging` signal were noted as residual risks — both safe today (the menu is modal; `dashboards()` has no live async mutation) and out of scope to re-architect here.

Verification: `lint` clean, `build:prod` green, editor spec 10/10 locally (menu contents, single-tap + keyboard open, edit/duplicate apply-on-save, delete-confirms-and-only-on-confirm, stale-index guard). CI runs the full matrix.

Closes #188. Refs #183.


mairas:
## Review synthesis

Five independent persona reviewers (correctness, adversarial, maintainability, testing, project-standards), each in a fresh context.

### Confirmed and fixed

| # | Finding | Severity | Fix |
|---|---------|----------|-----|
| 1 | **Jittery tap swallowed.** A tap with 6–20px travel trips cdkDrag's 5px start threshold (`isDragging` → true) but still emits a `tap` (within the 20px tapSlop); the old `onTileTap` `isDragging` guard then discarded that legitimate tap, so the menu failed to open. Common on a rocking marine display. (adversarial + correctness) | P3 | Removed the guard. Tap-vs-drag is arbitrated by the gesture directive's movement threshold alone — a real reorder moves past tapSlop and emits no tap. This also fixes the post-drop tap-swallow correctness raised and removes the app-global-signal dependency. |
| 2 | `editDashboard` / `duplicateDashboard` only called internally after the refactor — should be `private` (consistency with `confirmDelete`). (maintainability) | P3 | Made both `private`. |
| 3 | Test gaps: save/apply path never exercised (dialog mock always emitted null), keyboard `onTileKey` untested, stale-index guard untested, and 4 tests leaked real CDK overlays by not spying `open`. (testing ×3, standards) | P2 | Spec expanded to 10 tests; spy `ActionMenuComponent.open` in `beforeEach` + `fixture.destroy()`; added save-apply, keyboard, and stale-index cases. |

### Verified sound (adversarial traced 6 break scenarios)

Real reorder never pops the menu (moves past tapSlop → no tap); deleting the **last** page recreates a blank "Page 1" with `activeDashboard` reset; double-tap is inert (`enableDoubleTap=false`); the mat-menu/bottom-drawer breakpoint + `position:fixed` anchor keep the menu on-screen; **delete cannot proceed without confirmation** (only the Confirm button closes `true`). No standards violations (uses `unknown` not `any`, serving path untouched, OnPush retained, no stub drift).

### Residual risks (safe today, documented — not acted on)

- `_menuIndex` is a positional index captured at menu-open and re-read at action time. Safe because both menu variants are modal (backdrop blocks tile interaction) and `dashboards()` has no live async mutation path. If a future feature adds live sync (multi-tab storage events, remote push, profile switch), switch to capturing `dashboard.id`. `drop()` reorder coverage remains a pre-existing gap.

**Verdict: ready.** lint clean, build green, spec 10/10 locally; CI runs the full matrix.

Issue #156, opened 2026-07-03
signalk-requests: 500/502/504 described but routed to unknown-status toast; 200-message typo


mairas:
Surfaced while fixing #26 (PR #146), which pinned the current behavior with a characterization test rather than changing it.

1. `signalk-requests.service.ts` — `deltaStatusCodes` carries descriptions for 500/502/504, but `updateRequest` routes those codes to the **"Unknown Request Status Code"** error toast (which then prints the very description it claims not to know). Either handle them as known codes or drop the unused descriptions.
2. User-visible message typo for statusCode 200: `"The request was successfully."` — asserted verbatim by existing characterization specs; update spec and string together.

Issue #155, opened 2026-07-03
data.service: timeout reset only reaches the first registration for a path


mairas:
Surfaced while fixing #23 (PR #146).

## Defect

`data.service.ts` `timeoutPathObservable` matches registrations with `this._pathRegister.find(item => item.path == path)` — loose `==` and first-match-by-path-only. When the same path is registered under multiple `$source`s, only the first-registered subject receives the timeout reset; other-source registrations are never reset.

The #23 fix deliberately did not change this (scope). The fix likely needs to iterate all registrations for the path (and use `===`).

Issue #150, opened 2026-07-03
dashboard.service: blank-dashboard seeding aliases DefaultDashboard's configuration array


mairas:
Surfaced by the #27 characterization pass (PR #147, `dashboard.service.spec.ts`). Same defect class as #35 and #100 (mutating a shared module-level default).

## Defect

`dashboard.service.ts:46-50` — blank-dashboard seeding spreads `DefaultDashboard` entries **shallowly**, so the new dashboard's `configuration` array is the same object as the module-level `DefaultDashboard` constant. Any in-place mutation of that dashboard's widgets corrupts the shared default for every later blank creation.

## Fix

Deep-clone the default (the `cloneDeep` pattern used by `settings.service.getDefaultConnectionConfig()` and, since PR #146, `app-initNetwork`). PR #147 pins the current aliasing behavior — update that assertion intentionally.

Issue #149, opened 2026-07-03
dashboard.service: delete() leaves activeDashboard pointing at a shifted index


mairas:
Surfaced by the #27 characterization pass (PR #147, `dashboard.service.spec.ts`).

## Defect

`dashboard.service.ts:182-191` — `delete()` never remaps `activeDashboard` when a dashboard at a lower index is removed; only overflow past the end is clamped. The active index then silently points at a *different* dashboard. Callers must navigate afterwards to stay consistent, and nothing enforces that.

## Minor siblings (same file, fold into the fix or split)

- Lines 101-130: a URL id param like `'1.5'` passes `Number()` parsing and the bounds check (`1.5 >= 0 && 1.5 < length`), so `activeDashboard` can be set to a fractional index.
- Lines 232, 236: `console.error` prefix reads `"Dashboard Service]"` — missing the opening bracket.

Pinned by PR #147 tests; update those assertions intentionally when fixing.

PR #201, opened 2026-07-04
fix: characterization-pass follow-up defects (#149, #150, #155, #156)

### LINKED ISSUE #35: [PCS-05] Default-config singletons are mutated by reference instead of cloned

### LINKED ISSUE #184: PR-0: Fix page-nav semantics (misnamed next/previous + inconsistent bindings)
**PR-0 — lands first, standalone. De-risks the rest of the redesign.**

## Unit 8a: Page-nav semantics fix

**Goal:** Correct the misnamed/inconsistent next/previous page navigation and update all direct callers in lockstep.

**Diagnosis (verified against code):** the methods are **misnamed, not inverted relative to each other** — `navigateToNextDashboard()`/`nextDashboard()` both *decrement* the index (retreat), `navigateToPreviousDashboard()`/`previousDashboard()` both *increment* (advance); this direction is **pinned** by `dashboard.service.spec.ts`. The user-visible inconsistency lives in `dashboard.component`: swipe-up binds `previousDashboard()` (advance) while `Ctrl+↑` binds `nextDashboard()` (retreat). Direct callers outside the shell: `widget-iframe`, `widget-freeboardsk`, `widget-anchor-alarm`. `RemoteDashboardsService.navigateTo(idx)` is absolute/direction-agnostic.

**Requirements:** R13 (correctness prerequisite).

**Files:**
- Modify: `src/app/core/services/dashboard.service.ts` (rename/correct so `next` advances / `previous` retreats)
- Modify (callers, lockstep): `src/app/widgets/widget-iframe/*`, `src/app/widgets/widget-freeboardsk/*`, `src/app/widgets/widget-anchor-alarm/*`, `src/app/core/components/dashboard/dashboard.component.*` (swipe + hotkey bindings); audit `src/app/core/services/remote-dashboards.service.ts`
- Test (re-pin): `src/app/core/services/dashboard.service.spec.ts`, `src/app/core/components/widget-iframe/widget-iframe.component.spec.ts`

**Approach:** fix naming/semantics once; preserve current clamp/wrap; make swipe and hotkey directions agree; re-pin the direction assertions to the corrected semantics.

**Test scenarios:**
- Happy path: `next` advances (index+1 within clamp/wrap), `previous` retreats; re-pinned specs assert corrected direction.
- Edge case: first/last boundary matches prior clamp/wrap, identical across callers.
- Integration: the three iframe widgets and `dashboard.component` bindings all navigate consistently for the same intent.

**Verification:** full unit suite green (previously-pinned direction tests now assert corrected semantics).

mairas:
## What

Fixes four correctness defects surfaced by the #27 characterization pass (PR #147/#146). Each fix intentionally re-pins the characterization test that previously asserted the buggy behavior.

- **#149 — `dashboard.service.delete()` index bookkeeping.** Deleting a lower-indexed dashboard left `activeDashboard` pointing at a *different* dashboard (only end-overflow was clamped). Now the active index follows its dashboard down and clamps. Folds in two same-file siblings: `setActiveDashboardIndex` accepted fractional indexes (a URL id like `1.5` passed the bounds check), and two `console.error` prefixes in `duplicate()` read `"Dashboard Service]"` with no opening bracket.
- **#150 — blank-seed array aliasing.** Constructor blank-dashboard seeding shallow-spread `DefaultDashboard`, so the new dashboard's `configuration` array was the *same object* as the module-level constant; in-place widget edits corrupted the shared default. Now deep-cloned (same defect class as the already-fixed #35/#100).
- **#155 — `data.service.timeoutPathObservable` first-match only.** Used `_pathRegister.find(item => item.path == path)` (loose `==`, first match); when a path was registered under multiple `$source`s only the first reset. Now iterates all registrations via `_pathRegisterByPath.get(path)`.
- **#156 — `signalk-requests` 500/502/504 routing + 200 wording.** 500/502/504 had descriptions but were routed to the "Unknown Request Status Code" toast (which then printed the description it claimed not to know). Now handled as known server errors surfaced as an error toast with their description. Also fixes the 200 message typo (`"successfully"` → `"successful"`).

## Not included: #148

#148 (prev/next navigation inverted) was requested but is **already fixed on `main`** by `c5b858c4` (2026-07-04), which corrected both cycling methods and their `navigateTo*` siblings, did the caller audit, and re-pinned the direction assertions. That commit recorded `Closes #184`, leaving #148 open as a stale duplicate — it should just be closed.

## Verification

- `npm run lint` — clean.
- `ng test` on the three affected specs — 61/61 pass. Each updated/new assertion is a genuine regression catch (fails on the pre-fix code).
- `npm run build:dev` — clean.

CI runs the full matrix (lint, vitest 20/22/24, build 20/22/24).

### DIFF
--- a/src/app/core/services/dashboard.service.spec.ts
+++ b/src/app/core/services/dashboard.service.spec.ts
@@ -167,6 +167,13 @@ describe('DashboardService', () => {
       expect(service.activeDashboard()).toBe(1);
       expect(consoleError).toHaveBeenCalledTimes(2);
     });
+
+    it('rejects a fractional index instead of activating a non-integer dashboard', () => {
+      service.setActiveDashboardIndex(1);
+      service.setActiveDashboardIndex(1.5);
+      expect(service.activeDashboard()).toBe(1);
+      expect(consoleError).toHaveBeenCalledTimes(1);
+    });
   });
 
   describe('add and update', () => {
@@ -196,13 +203,41 @@ describe('DashboardService', () => {
   });
 
   describe('delete', () => {
-    it('removes the dashboard without remapping the active index', () => {
+    it('shifts the active index down so it follows the same dashboard when a lower one is removed', () => {
       setup();
       service.setActiveDashboardIndex(1);
       service.delete(0);
       expect(service.dashboards().map(d => d.id)).toEqual(['d-1', 'd-2']);
-      // Pins today's bookkeeping: the index is not shifted down, so it now points at the former third dashboard.
+      // The active dashboard (d-1) shifts from index 1 to index 0; the active index follows it.
+      expect(service.activeDashboard()).toBe(0);
+      expect(service.dashboards()[service.activeDashboard()!].id).toBe('d-1');
+    });
+
+    it('leaves the active index unchanged when a higher-indexed dashboard is removed', () => {
+      setup();
+      service.setActiveDashboardIndex(1);
+      service.delete(2);
       expect(service.activeDashboard()).toBe(1);
+      expect(service.dashboards()[service.activeDashboard()!].id).toBe('d-1');
+    });
+
+    it('keeps the active index on the dashboard that slides up when the active, non-last one is deleted', () => {
+      setup();
+      service.setActiveDashboardIndex(1);
+      service.delete(1);
+      expect(service.dashboards().map(d => d.id)).toEqual(['d-0', 'd-2']);
+      expect(service.activeDashboard()).toBe(1);
+      expect(service.dashboards()[service.activeDashboard()!].id).toBe('d-2');
+    });
+
+    it('ignores an out-of-range index without touching the active dashboard', () => {
+      setup();
+      service.setActiveDashboardIndex(0);
+      service.delete(-1);
+      service.delete(5);
+      expect(service.dashboards().map(d => d.id)).toEqual(['d-0', 'd-1', 'd-2']);
+      expect(service.activeDashboard()).toBe(0);
+      expect(consoleError).toHaveBeenCalledTimes(2);
     });
 
     it('clamps the active index when it falls past the end', () => {
--- a/src/app/core/services/dashboard.service.spec.ts
+++ b/src/app/core/services/dashboard.service.spec.ts
@@ -92,8 +92,10 @@ describe('DashboardService', () => {
       expect(dashboards[0].name).toBe('Page 1');
       expect(dashboards[0].id).toMatch(UUID_PATTERN);
       expect(widgetsOf(dashboards[0])[0].input.widgetProperties.type).toBe('widget-tutorial');
-      // Pins today's aliasing: the blank dashboard shares its configuration array with the DefaultDashboard constant.
-      expect(dashboards[0].configuration).toBe(DefaultDashboard[0].configuration);
+      // The blank dashboard deep-clones the DefaultDashboard constant so later in-place edits cannot corrupt the shared default.
+      expect(dashboards[0].configuration).not.toBe(DefaultDashboard[0].configuration);
+      expect(widgetsOf(dashboards[0])[0]).not.toBe(DefaultDashboard[0].configuration![0]);
+      expect(dashboards[0].configuration).toEqual(DefaultDashboard[0].configuration);
       expect(consoleWarn).toHaveBeenCalled();
     });
 
--- a/src/app/core/services/dashboard.service.ts
+++ b/src/app/core/services/dashboard.service.ts
@@ -122,7 +122,7 @@ export class DashboardService {
     if (itemIndex === this.activeDashboard()) return;
     // No change if the same dashboard is selected to prevent unnecessary cascading updates
 
-    if (itemIndex >= 0 && itemIndex < this.dashboards().length) {
+    if (Number.isInteger(itemIndex) && itemIndex >= 0 && itemIndex < this.dashboards().length) {
       this.activeDashboard.set(itemIndex);
     } else {
       console.error(`[Dashboard Service] Invalid dashboard ID: ${itemIndex}`);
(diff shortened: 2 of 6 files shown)


mairas:

**Scope:** 6 files, core services (`dashboard`, `data`, `signalk-requests`) + specs, diffed against `main`.
**Intent:** four correctness fixes from the #27 characterization pass (#149, #150, #155, #156), each re-pinning the test that formerly asserted the buggy behavior.
**Reviewers (6):** correctness, testing, maintainability, project-standards (always-on) + reliability (timeout/error paths) + adversarial (delete index arithmetic).

Findings are confidence-gated (≥0.60) and deduped; cross-reviewer agreement noted. **All in-scope findings were addressed in the pushed commits**; two pre-existing/out-of-scope items were filed as issues.

### P2 — High

| # | File | Issue | Reviewer(s) | Conf | Resolution |
|---|------|-------|-------------|------|------------|
| 1 | dashboard.service.ts:196 | `delete()` with an out-of-range/negative `itemIndex` shifted (or, for a negative index at active 0, set **-1**) the active index for a delete that removed nothing — a regression vs the old safe no-op; -1 then crashes `loadDashboard`. | adversarial (correctness concur) | 0.68 | **Fixed** — bounds guard at top of `delete()` matching `duplicate()`/`navigateTo()`; test added. |
| 2 | dashboard.service.spec.ts | The `<` boundary that *is* the #149 fix was unpinned — no test deletes the **active, non-last** dashboard, so a `<`→`<=` regression survived the suite. | testing | 0.82 | **Fixed** — added active-non-last delete test (+ out-of-range test). |

### P2 — High (pre-existing, out of scope → filed)

| # | File | Issue | Reviewer(s) | Conf | Resolution |
|---|------|-------|-------------|------|------------|
| 3 | dashboard.component.ts:235 | Deep-link/reload to an invalid `/page/:id` (out-of-range, or fractional now that #149 rejects non-integers) leaves `activeDashboard` null; `loadDashboard` has no null guard → `undefined.configuration` TypeError / blank page. Fractional already crashed pre-#149 via `dashboards()[1.5]`. | adversarial | 0.66 | **Filed #205** — pre-existing; component-level, outside this PR's 3-service scope. |

### P3 — Moderate

| # | File | Issue | Reviewer(s) | Conf | Resolution |
|---|------|-------|-------------|------|------------|
| 4 | data.service.ts:718 | #155 now resets **all** source registrations for a path. Correct when the whole path is silent, but on redundant-sensor setups one source's timeout nulls a still-live sibling (+ default), flashing it blank and reverting alarm state until the next delta. | correctness + reliability | ~0.80 | **Filed #206** — the all-reset is #155's requested fix and correct for the common case; source-scoping is the follow-up. |
| 5 | data.service.spec.ts:341 | Test comment narrated the old implementation ("the old first-match lookup…") — violates the repo's comment policy (comments state WHY, never what changed). | maintainability + project-standards | ~0.80 | **Fixed** — trimmed to the invariant. |
| 6 | dashboard.service.spec.ts:96 | #150 clone test asserted only `configuration` array identity; a shallow `[...arr]` clone sharing nested widget objects would still pass. | testing | 0.85 | **Fixed** — added nested-widget identity assertion. |
| 7 | signalk-requests.service.ts:139 | Server-error branch: only 500-with-message was tested; 502/504 and the no-message path (`detail = description`) were uncovered. | testing (reliability/standards/correctness concur) | ~0.80 | **Fixed** — parameterized 500/502/504 + added a no-message case. |

### Not actioned (advisory)

- `serverErrorStatusCodes` repeats 500/502/504 already in `deltaStatusCodes`/`handledStatusCodes` — a three-way sync cost (maintainability, 0.60). Acceptable for three codes; reviewer agreed no change needed now.

### Coverage

- Suppressed: findings below 0.60 confidence.
- Residual risk (pre-existing, not filed): the `try/catch` in `updateRequest` calls `requestStatus$.error(err)` on failure, which permanently terminates the shared request-status Subject for the session. `splice` on a found index effectively can't throw, so trigger probability is very low — noted, not filed.
- Verified by review: the #155 timeout key matches `subscribePath`'s key (same `normalizedPath`); `_pathRegisterByPath` arrays are app-lifetime (no concurrent-mutation risk); the #156 server-error branch falls through to dispatch+splice with no pending-request leak; 401 stays untoasted.

### Verdict

**Ready to merge** once CI is green. All P2/P3 in-scope findings are fixed in the pushed commits (66/66 targeted specs pass, lint + dev build clean); the two pre-existing items are tracked in #205 and #206.

<sub>Review by 6 independent persona agents; synthesis deduped and confidence-gated.</sub>

PR #207, opened 2026-07-04
feat(widget-config): always offer "Any" data source option

### LINKED ISSUE #203: Widget config: replace disappearing "default" data source with an always-present "Any" option
## Summary

In widget configuration, the **Data Source** dropdown for a path behaves inconsistently depending on how many sources currently publish that path:

- **One source** → the only option is a literal `default`, and the field is set to `default`.
- **Two or more sources** → `default` disappears from the list entirely; only the concrete source keys are offered. An existing `default` selection is silently reset, forcing the user to re-pick a concrete source.

This makes `default` look like a per-path artifact that comes and goes, when it is actually the most useful option — the server-merged, priority-following value. Proposal: always present it as a stable **Any** entry at the top of the list, with the concrete sources listed underneath.

## Why `default` already means "any source"

`default` is not a placeholder for "the one source we have". At the data layer, `source === 'default'` reads `pathValue` — the Signal K server's merged / priority-selected value:

- `data.service.ts:258` — `currentValue = source === 'default' ? dataPath.pathValue : dataPath.sources?.[source]?.sourceValue ?? null`
- `data.service.ts:364-382` — documents that with multiple sources and SK priorities configured, `default` follows whichever source the server currently favors, giving automatic failover (e.g. two GPS units, one drops out, the value keeps flowing).

So "accept data from any source, let the server decide" is exactly `default`'s runtime behavior — and it is *most* valuable when a path has several sources, which is precisely when the current UI removes it.

## Current behavior (where)

Two config components carry the same logic:

1. `src/app/widget-config/path-control-config/path-control-config.component.ts:219-234`
   - 1 source → `availableSources = ['default']`
   - \>1 source → `availableSources = Object.keys(pathObject.sources)` (no `default`); a saved `default` value is `reset()`.
   - Template `path-control-config.component.html:83-87` renders each option as the raw string, so the label shown is literally `default`.

2. `src/app/widget-config/dataset-chart-options/dataset-chart-options.component.ts:141-155` (data chart widget)
   - Same 1-vs-many split; `default` is likewise absent from the >1 branch.
   - Template `dataset-chart-options.component.html:50-57`.

## Problems

- **Confusing label.** `default` reads as "the default choice", not "the merged/any-source value". Users don't know what selecting it does.
- **It vanishes when it matters most.** The moment a second source appears for a path, the "any source / failover" option is removed, and users are pushed to pin a single concrete source — losing automatic failover.
- **Silent reset (path-control-config only).** Re-opening a widget's config after a path gains a second source clears a previously-saved `default` selection, so the field shows empty/invalid until re-selected.

## Proposed change

- Always include an **Any** option as the first entry in the Data Source list, regardless of source count. It maps to the existing stored value `'default'` (display label "Any", stored value unchanged).
- List the concrete sources below it (grouped/ordered as today).
- Stop dropping `default` and stop resetting a saved `default` selection when a path has multiple sources.
- Apply to both components above so widget paths and data-chart series behave the same.

Design direction only — no code prescribed here.

## Back-compat / migration

None required. The stored config value stays `'default'`; "Any" is only the display label for it. Existing widgets configured with `default` keep working and now render a stable, meaningful option instead of one that may disappear.

## Verification

- Path with a single source: Data Source shows **Any** (selected), no literal `default`.
- Path with multiple sources: Data Source shows **Any** at top plus each concrete source; selecting **Any** persists and yields server-priority/failover behavior at runtime.
- Re-opening a widget whose saved source is `default` after a second source has appeared: selection is preserved (not reset), shown as **Any**.
- Same for the data-chart widget's Source field.

mairas:
## Why

In widget configuration, the **Data Source** dropdown behaved inconsistently depending on how many sources currently published a path:

- **One source** → the only option was a literal `default`.
- **Two or more sources** → `default` vanished entirely; only concrete source keys were offered, and a saved `default` selection was silently reset (in `path-control-config`), forcing the user to re-pick a concrete source.

That hid `default` exactly when it matters most. At the data layer, `source === 'default'` reads the server's merged, priority-selected value (`data.service.ts` `subscribePath`) — it follows Signal K source priority and automatically fails over between sources (e.g. two GPS units, one drops, the value keeps flowing). "Accept data from any source" is its actual behavior, and that's most useful with multiple sources.

Closes #203.

## What

- Always list **Any** (the stored `'default'` value) first, with concrete sources below, regardless of source count.
- A freshly selected path defaults to **Any**; an existing selection (concrete or `default`) is preserved — no more silent reset when a path gains a second source.
- Applies to both source dropdowns: widget path (`path-control-config`) and data-chart series (`dataset-chart-options`).
- The dropdown *label* maps `'default'` → "Any"; the **stored config value stays `'default'`**, so no migration and full back-compat with existing widgets.

## Verification

- `npm run lint` — clean.
- `npm run build:prod` — succeeds (pre-existing `piexifjs` CommonJS warning only).
- Component specs (both files), covering single-source, multi-source (Any preserved — the #203 regression), preserved concrete selection, and fresh-path default:
  - `path-control-config.component.spec.ts` — 5 passed
  - `dataset-chart-options.component.spec.ts` — 4 passed


mairas:
## Code review — feat(widget-config): always offer "Any" data source option

**Scope:** `feat/203-any-data-source` vs `main` (6 files, ~40 impl lines + templates + specs). **Intent:** always offer "Any" (stored value `'default'`) first in the Data Source dropdown regardless of source count; preserve existing selection; fresh path defaults to "Any". Applies to `path-control-config` and `dataset-chart-options`.

**Reviewers:** correctness, testing, maintainability, project-standards (4 always-on; no conditional personas — small diff, no auth/API/reliability/data-mutation surface).

### P2 — should fix before merge

| 1 | `dataset-chart-options.component.ts:138` | **Switching path keeps a stale source instead of "Any".** `changePath` calls `setPathSources` without resetting `datachartSource` first (only the null-path branch resets). The new `setPathSources` sets `'default'` only when the value is falsy, so a concrete source picked for the *previous* path (e.g. `gps.0`) survives when the user selects a new path that lacks it. The `required` control passes with the stale string, the select renders blank, and the saved chart subscribes to a non-existent source → null data. The old code reset to `'default'` for single-source target paths, so this is a regression. `path-control-config` avoids it via its `setValues` flag; the chart component has no equivalent. | correctness, maintainability | 0.85 | fix-here |
| 2 | `path-control-config.component.spec.ts:17` | **`let pathObject: any` + eslint-disable** suppresses `@typescript-eslint/no-explicit-any` (an error per CLAUDE.md). The sibling spec changed in this same PR types its fixture as `Pick<ISkPathData, 'sources'>` with no suppression. Avoidable and inconsistent. | project-standards | 0.72 | fix-here |
| 3 | `path-control-config.component.ts:225` | **Empty-source load branch untested.** The new guard `if (setValues \|\| !sourceControl.value)` has two triggers, but only `setValues=true` is exercised. The `!value` disjunct (non-fresh load with empty saved source, reachable from `ngOnInit`) rewrites `''`→`'default'` — a behavior change with no test. The chart spec covers the equivalent case; path-control doesn't. | testing | 0.82 | fix-here |

### P3 — discretionary

| 4 | `path-control-config.component.ts:223` | **Source ordering diverges:** chart sorts the concrete keys (`.sort()`), path-control does not. Same conceptual dropdown, two orderings. | maintainability | 0.72 | fix-here (add `.sort()`) |
| 5 | both `.html` | **"Any" label mapping has no DOM test.** The headline user-visible change (`{{source === 'default' ? 'Any' : source}}`) is asserted nowhere; specs check the internal source arrays. MatSelect option rendering is awkward in this jsdom harness — noting rather than forcing a brittle test. | testing | 0.85 | advisory |
| 6 | both `.spec.ts` | **Tests call private methods via `as unknown as` casts** and assert internal arrays (`availableSources`/`pathSources()`) rather than driving the public triggers (`ngOnInit` valueChanges, `changePath`). Form-control-value assertions are fine; the coupling is a pragmatic tradeoff given the harness. | testing | 0.72 | advisory |

### Coverage
- Residual risk (both components): a saved concrete source that no longer exists in `pathObject.sources` is preserved as the form value but omitted from the options, leaving a blank-but-required select. Behavior matches old code for the multi-source case; not newly introduced except via finding #1.

### Verdict: **Ready with fixes** — address #1 (regression), #2 (lint policy), #3 (coverage), and #4 (consistency). #5/#6 are advisory.

mairas:
## Review findings addressed

Folded into the branch (amended, force-pushed) rather than a follow-up commit, since each fix revises code the same commit introduced.

| # | Finding | Resolution |
|---|---------|------------|
| 1 | P2 — chart `changePath` kept a stale concrete source when switching paths | `changePath` now `reset()`s `datachartSource` before `setPathSources`, so a fresh path falls through to "Any". The `ngOnInit` load path still calls `setPathSources` directly and keeps the saved selection. `dataset-chart-options.component.ts:137` |
| 2 | P2 — `let pathObject: any` + eslint-disable in spec | Typed the fixture as `Partial<ISkPathData>` with a shared `src()` helper, matching the sibling spec; removed the suppression. `path-control-config.component.spec.ts` |
| 3 | P2 — empty-source load branch untested | Added `defaults an empty saved source to "Any" on load` (`setValues=false`, empty value → `'default'`). `path-control-config.component.spec.ts` |
| 4 | P3 — source ordering diverged between the two components | path-control now `.sort()`s the concrete keys, matching the chart component. `path-control-config.component.ts:223` |

Also added a regression test for #1 that drives the **public** `changePath` (not the private method), covering the testing reviewer's coupling note (#6) for the highest-value case: switch to a path lacking the previously selected source → value becomes "Any".

**Advisory, not changed:**
- #5 (DOM label test): MatSelect option text renders only when the overlay panel opens, which is awkward and brittle in this jsdom harness. The `'default' → 'Any'` mapping is a one-line template ternary; not worth a fragile test.
- #6 (private-method coupling): kept for the remaining unit assertions — driving the debounced path `valueChanges` for every case is heavier and more brittle than the harness supports. The new #1 test exercises the public path.

**Verification:** `npm run lint` clean · `npm run build:prod` succeeds · `path-control-config.component.spec.ts` 6 passed · `dataset-chart-options.component.spec.ts` 5 passed.

Issue #217, opened 2026-07-07
fix: Skip's SSO redirect breaks when framed without a session


mairas:
Found during the #213 panel spike.

When Skip is embedded in an iframe (the Freeboard panel) **without** a Signal K session, its boot logic redirects the frame to the SSO login endpoint (`https://<host>/sso?flow=openid_connect…`). That endpoint sets `frame-ancestors 'none'`, so the browser blocks the navigation and the panel renders broken instead of authenticating.

The authenticated same-origin happy path does **not** trigger this — the shared session cookie is present, so Skip never redirects (verified working). It bites on session expiry mid-use, or a cross-origin embed.

Skip should detect it's running framed and avoid a top-frame SSO redirect — e.g. surface an in-panel 'sign in' affordance, or break the redirect out to `window.top`. `SsoRedirectService` owns the sign-in redirect.

Issue #216, opened 2026-07-07
feat: chromeless/embedded Skip route for the Freeboard panel


mairas:
Follow-up to #213 (v1 panel ships pointing at Skip's full shell).

The Freeboard plotter-extension panel currently loads `/@halos-org/skip/` — Skip's full app shell. Skip's responsive phone-width layout fits the drawer, but its own toolbar/navigation is redundant chrome inside a host-owned panel (Freeboard already provides the panel title + close).

Add a compact/embedded render mode — a route or query flag (e.g. `/@halos-org/skip/#/embed` or `?chrome=off`) that shows a single dashboard without Skip's app chrome — and point the panel plugin's `SKIP_URL` (plugin/index.js) at it.

mairas:
## Implementation plan (independent design pass, 2026-07-15)

### Architecture — one new service + reuse existing gates
**`EmbedModeService`** (root, dependency-free, ChromeVisibilityService-shaped): reads `new URLSearchParams(window.location.search)` **once at construction**, exposes `embed` + `profile` signals. Force-constructed by injecting it into `AppNetworkInitService` (the first blocking `provideAppInitializer`).

**Key insight — boot-latching is free.** The flags live in the **pre-hash** query (`/@halos-org/skip/?embed=1&profile=day#/page/0`). Because the app uses `withHashLocation`, every `router.navigate()` only touches the fragment, so `window.location.search` is structurally untouched by in-app navigation. "Boot-latched" = "read `window.location.search` once" — no `queryParamsHandling` plumbing anywhere.

### The behaviors (each grafts onto an existing gate — minimal new surface)
- **Chrome off** — DOM-*unmount* (not CSS-hide): `@if (dashboardVisible() && !embed())` on the toolbar (`app.component.html:19`). Removes toolbar + page-icon strip + all its buttons.
- **Read-only / edit-gestures off** — single choke point: pin `isDashboardStatic` **true** at `DashboardService.setStaticDashboard/toggleStaticDashboard` under `embed()`. That one signal already gates gridstack drag/resize, widget tap→config, group edit, and all three unlock entry points (Ctrl+Shift+E, toolbar, empty-state). Plus one template gap: gate the "Unlock and Customize" empty-state button with `!embed()`.
- **Kept, zero change** — alarm badge (CDK overlay, already static-gated) + swipe/wheel page-nav (on the shell div, already static-gated).
[...]

PR #178, opened 2026-07-03
Remove bundled kip-plugin history provider and its Display config UI

mairas:
Retires Skip's bundled server-side history provider and the Settings UI that configured it. History is now served entirely by an external Signal K History API provider (InfluxDB via `signalk-to-influxdb2`, or `signalk-parquet`); Skip is a webapp only, with no server-side plugin.

Closes #176 and #177.

## Why
- The plugin's home-baked `node:sqlite` history capture is retired; the marine stack's history provider is external.
- After the image-asset routes were removed earlier, `kip-plugin/` was a pure history provider — so retiring that role means removing it wholesale.
- The "Widget Historical Data" section of Settings → Display existed only to enable/select the plugin as the History API provider. With the plugin gone, it configured a provider that no longer exists.

## What changed
**#176 — plugin removal**
- Delete `kip-plugin/` entirely.
- Relocate the shared `kip-series-contract` into `src/app/core/contracts/` — it was re-exported from the plugin, and the **kept** client History-API consumers (`KipSeriesApiClientService`, `DashboardHistorySeriesSyncService`, `widget-host2`, `widget-history-chart-dialog`) still need it. Content is identical; import paths are unchanged.
- Strip plugin wiring from `package.json` (`main`, `plugin/**` in `files`, the `signalk-node-server-plugin` keyword, `signalk-plugin-enabled-by-default`, and the `build:plugin`/`test:plugin`/`build:all` scripts) and drop the `plugin-node-test` / `plugin-build` CI jobs.
- Clean plugin-only entries from `.gitignore` / `.npmignore`, and drop the Node ≥22.5 (`node:sqlite`) note and `build:all` from `CLAUDE.md` / `README.md`. Changelog entry added.

**#177 — Display UI removal**
- Remove the "Widget Historical Data" expansion panel and the now-unused `mat-radio` import, `SignalKConnectionService` injection, and `compare-versions` import.
- Remove `providerMode`, `isKipHistoryProviderSelectable`, `isHistoryApiSupported`, `KIP_DATA_PLUGIN_ID`, and `getKipPluginConfig`/`setKipPluginConfig`. `applyAndSaveSettings` no longer writes plugin config; it persists the local display settings and reports success. `PluginConfigClientService` is kept for the Automatic Night Mode (derived-data) flow.

[...]

PR #218, opened 2026-07-07
feat: expose Skip as a Freeboard-SK plotter-extension panel

### LINKED ISSUE #213: feat: expose Skip as a Freeboard-SK plotter-extension panel
## Summary

Expose Skip inside Freeboard-SK as a **panel** using the Freeboard-SK [plotter extension provider](https://github.com/SignalK/freeboard-sk/blob/master/docs/api/plotter_extension_provider_plugins.md) mechanism, so a user working in the Freeboard chartplotter can open Skip's instruments in a drawer without leaving Freeboard.

This is the inverse of what Skip already does: `widget-freeboardsk` embeds Freeboard-SK as an iframe inside a Skip dashboard. Here Freeboard is the host and Skip is the contributor.

## Direction / how the mechanism works

The mechanism has a **host** (the chartplotter — Freeboard-SK) and **contributors** (Signal K server plugins that inject UI into the host). Contributions are declared in a manifest registered as a read-only `plotterExtensions` resource; iframe assets are served by the plugin at `/plotterext/<id>`. Discovery is by enumerating `plotterExtensions` resources over REST — there is no webapp-only path, so a small server-side plugin is required (see prerequisite below).

A **panel** is an iframe of arbitrary content where the host owns the chrome (title bar, close button). The plugin supplies only content. Skip is already a full webapp that connects to Signal K directly over its same-origin session, so it needs nothing from the host to render or get data.

## Prerequisite: companion server plugin

Skip is webapp-only today — the plugin component was removed in #176/#177. The `plotterExtensions` resource and the `/plotterext/<id>` asset route can only be created by a Signal K server plugin. So this issue depends on reintroducing a thin companion plugin whose only jobs are:

- register a `plotterExtensions` resource provider returning Skip's manifest
- (optionally) serve panel assets; for the panel the manifest `url` can instead point at Skip's existing same-origin mount `/@halos-org/skip/`

This prerequisite is shared with the widget issue (see below); build it once.

## Approach

1. Companion plugin registers a manifest with:
   - one `panels.iframe` contribution whose `url` targets a compact Skip surface (a single dashboard route, not the full app shell)
   - optionally a `buttons` contribution so Freeboard shows a toolbar button that opens the panel (`action.type: openPanel` / `togglePanel`)
2. Panel iframe loads Skip same-origin; the session cookie authenticates it, exactly like `widget-freeboardsk` does in reverse. Skip connects to Signal K on its own — the plotterext bus is **not** required for data.
3. `connectExtension()` handshake is **optional**, added later for niceties only: host-driven open/close, `keepAlive` lifecycle, and syncing night mode / display units from the host. Skip already has the postMessage plumbing (`widget-freeboardsk`) to build on.

## Scope decisions

- **Which Skip surface** goes in the drawer is a product decision, not a code one. A full MFD shell is awkward in a narrow panel; pick one dashboard or a slim instrument strip. Likely needs a dedicated route/layout that renders without Skip's own chrome.
- Panel sizing/orientation: Freeboard drawers are narrow; the chosen surface must be legible there.

## Open questions / risks

- **Mechanism maturity.** `apiVersion: '1'`; verify `signalk-plotterext-bus` is actually published and that host support ships in the Freeboard-SK version we bundle before investing.
- Confirm Freeboard accepts a same-origin `url` pointing at an existing webapp mount (`/@halos-org/skip/`) rather than an asset under `/plotterext/`. The doc warns against `/plugins/*`; an existing webapp mount should be fine but needs checking.
- Confirm the `plotterExtensions` resource type doesn't collide with anything, given the recent first-wins provider conflict we hit on charts (resources-provider vs charts-plugin).

## Estimate

Small, once the companion plugin exists — days, not weeks. ~90% of the "glance at Skip from Freeboard" value for ~10% of the widget cost.

Related: widget variant — #214

mairas:
## What

Bundles a small Signal K server plugin in the Skip package that registers a read-only `plotterExtensions` resource. A supporting Freeboard-SK (`3.0.0-beta`+) reads it and shows a **"Skip" toolbar button** that opens Skip in a side panel — Skip and Freeboard side by side, without leaving the chartplotter.

## Why this shape

Freeboard's [plotter-extension mechanism](https://github.com/SignalK/freeboard-sk/blob/master/docs/api/plotter_extension_provider_plugins.md) is delivered by a server plugin (the resource is only creatable via `registerResourceProvider`; there's no webapp-only path). A **panel** is just an iframe of arbitrary content with host-owned chrome, and Skip already connects to Signal K over its same-origin session — so the panel needs no bundle, no `signalk-plotterext-bus`, and no asset route. The manifest URL is server-relative, so it points straight at Skip's own serving path.

Folded into the webapp package (rather than a standalone plugin) so it's version-locked to Skip, always points at the correct `/@halos-org/skip/` path, and ships/deploys with Skip — no second package to wire into an image story Skip doesn't have yet. The plugin removed in #178 was Kip's heavyweight server-side *history provider*; this is an unrelated read-only UI registration.

## Changes

- `plugin/index.js` — the plugin: registers `plotterExtensions` with one panel (`/@halos-org/skip/`, `keepAlive`) + a toggle button.
- `package.json` — `signalk-node-server-plugin` keyword, `main`, and `plugin/**` in `files`.
- CHANGELOG entry under Unreleased.

## Verification

Proven end-to-end on halpi.hurma (Freeboard `3.0.0-beta.0`, SK `2.27.0`): SK loads Skip as a plugin, the resource serves the manifest, Freeboard renders the "Skip" button, and clicking it opens the panel with live Skip (authenticated same-origin path confirmed working). CI-relevant checks run locally green (`build:prod`; lint/betterer untouched — plugin is non-`src` JS).

## Scope / follow-ups

v1 points the panel at Skip's existing responsive shell (its phone-width layout fits the drawer). Tracked separately:
- #216 — a chromeless/embedded Skip route as the panel target.
- #217 — Skip's SSO redirect breaks when framed without a session (happy path unaffected).

## Deploy note

The Skip package now ships `plugin/`, so deployment must copy it (not just `public/` + `package.json`). The plugin must be **enabled** on the Signal K server to activate.

Closes #213

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 ## New Features
+* Freeboard-SK panel: Skip registers itself as a Freeboard-SK plotter extension, so a supporting Freeboard-SK (3.0.0-beta or later) shows a toolbar button that opens Skip in a side panel. Ships as a Signal K server plugin bundled in the Skip package, enabled by default.
 * Same-origin Signal K sign-in (SSO): when KIP is served by a Signal K server on the same origin, it authenticates with the server's session cookie instead of its own username/password form. With an OIDC/SSO server, KIP joins the existing session and redirects to the server login when needed — no second sign-in, and OIDC-provisioned users (who have no server-local password) can use KIP. Cross-origin standalone use (a PWA pointed at a remote server) keeps the existing token sign-in.
 ## Improvements
 * The Connectivity settings show your Signal K session identity in same-origin mode (including a read-only-session indicator) instead of a credential form.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -74,10 +74,10 @@ The mock serves Skip's full session/config surface (`loginStatus`, `applicationD
 
 ## Fork-specific gotchas
 
-- **The package name is the serving path.** `angular.json` `baseHref`, the `dev` serve-path, and `src/manifest.json` `id`/`scope`/`start_url` must all stay `/@halos-org/skip/`. Branding lives in `brand/` (vector master, repo-only), `src/assets/` (favicon/icon set), `src/manifest.json`, and `signalk.displayName` in `package.json`.
+- **The package name is the serving path.** `angular.json` `baseHref`, the `dev` serve-path, `src/manifest.json` `id`/`scope`/`start_url`, and `SKIP_URL` in `plugin/index.js` must all stay `/@halos-org/skip/`. Branding lives in `brand/` (vector master, repo-only), `src/assets/` (favicon/icon set), `src/manifest.json`, and `signalk.displayName` in `package.json`.
 - **`gh` defaults PRs to the `upstream` remote.** Always pass `--repo halos-org/skip` to `gh pr create` / `gh pr merge`. (The clone is set up with `gh repo set-default halos-org/skip` and a disabled upstream push URL, but pass it explicitly anyway.)
 - **Upstream's own vitest CI is red**; a rebase onto `mxtommy/kip` can re-introduce test failures. The fork's fixes are small and live in `src/test.ts` stubs plus a few specs — re-apply them rather than disabling tests.
-- Not published to npm (publishing would list it in the Signal K app store). Deploy by placing the built `public/` output where the SK server serves the webapp.
+- Not published to npm (publishing would list it in the Signal K app store). Deploy by placing the package where the SK server serves it: the built `public/` output (the webapp) **and** `plugin/` (the bundled Freeboard-SK panel plugin, referenced by `main`) alongside `package.json`. A `public/`-only copy leaves `main` dangling and the plugin fails to load.
 
 ## Inherited upstream docs — treat skeptically
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@halos-org/skip",
   "version": "4.8.0",
+  "main": "plugin/index.js",
   "publishConfig": {
     "access": "public"
   },
@@ -32,6 +33,7 @@
   },
   "keywords": [
     "signalk-webapp",
+    "signalk-node-server-plugin",
     "signalk-category-instruments",
     "signalk-category-notifications",
     "signalk-category-ais",
@@ -46,8 +48,10 @@
     "appIcon": "assets/icon-72x72.png",
     "displayName": "Skip"
   },
+  "signalk-plugin-enabled-by-default": true,
   "files": [
     "public/**",
+    "plugin/**",
     "README.md",
     "LICENSE",
     "CHANGELOG.md",
--- a/plugin/index.js
+++ b/plugin/index.js
@@ -1,0 +1,79 @@
+// Signal K server plugin bundled with the Skip webapp package.
+//
+// It registers Skip as a Freeboard-SK "plotter extension" panel: a read-only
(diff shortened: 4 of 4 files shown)


mairas:

**Scope:** `plugin/index.js` (new), `package.json`, `CHANGELOG.md`, `CLAUDE.md` vs `main`.
**Intent:** Fold a Signal K server plugin into `@halos-org/skip` that registers a read-only `plotterExtensions` resource, so Freeboard-SK shows a "Skip" panel button.
**Team (7):** correctness, testing, maintainability, project-standards (always-on) + api-contract (manifest is a host contract), security (new server-reachable resource), adversarial (new file ≥50 LOC, external-integration surface).

Findings were verified before acting — two of the highest-severity ones did not survive verification.

### Dismissed after verification

| # | Finding | Reviewer | Why dismissed |
|---|---------|----------|---------------|
| 1 | Webapp disappears from `/skServer/webapps` when the plugin is disabled (P2, conf 0.80) | adversarial | **Refuted empirically.** Disabled the plugin on the deployed server (resource → 404, confirming off); `@halos-org/skip` still lists. The source trace was against a different signalk-server version than the one deployed (2.27/2.28). |
| 2 | `start()` re-registers the provider on restart → double registration (P2, conf 0.62) | maintainability | **Refuted.** correctness + adversarial both traced SK's `onStopHandler` → `resourcesApi.unRegister(plugin.id)` runs before `stop()`; re-enable cleanly re-registers (Map.set). The `running` gate is a correct safety net. |

### Fixed in this PR

| # | Finding | Reviewer | Resolution |
|---|---------|----------|------------|
| 3 | Feature inert until an operator manually enables the plugin | correctness, adversarial | Added `signalk-plugin-enabled-by-default: true`. **Verified**: from a fresh state (no config) the plugin auto-enables, the resource registers, and the webapp still lists. |
| 4 | Documented deploy copies only `public/` + `package.json`, so `main: plugin/index.js` dangles → `MODULE_NOT_FOUND` (P2, conf 0.82) | adversarial, project-standards | Updated the `CLAUDE.md` deploy note to require shipping `plugin/`; the private operator recipe is updated too. Repo `files` already included `plugin/**`. |
| 5 | Manifest `version` hardcoded `'1'`, never tracks releases (P3, conf 0.65) | api-contract | Sourced `version` from `package.json` (now `4.8.0`); `apiVersion` stays `'1'`. |

### Deferred → #221 (P3, intentional for v1)

- `requires: []` omits capabilities the panel uses; safe on the beta host, may need declaring before Freeboard-SK GA (changing it now risks hiding the extension).
- No unit tests for the plugin's three contracts (running gate, read-only rejections, id match); `plugin/` is outside the Angular vitest scope.
- `slot: 'primary'` rendered into a secondary toolbar group — placement is host-version-dependent.

### Coverage

- security: no findings — read-only static manifest, no untrusted-input sink, no fs/net/eval.
- Residual risk carried: the manifest contract is validated only by manual end-to-end testing against Freeboard-SK `3.0.0-beta.0`; no automated schema check guards upstream drift (folded into #221).
- Suppressed below-threshold findings: a few; nothing P0/P1.

### Verdict

**Ready to merge** once CI is green. All confirmed findings are fixed and verified end-to-end on halpi (fresh-install auto-enable + button renders); the two P2s were refuted with evidence; the remaining P3s are tracked in #221.

PR #219, opened 2026-07-08
fix(windsteer): base laylines and wind sectors on true wind

mairas:
## Why

The windsteer widget's close-hauled ("layline") lines were centered on the **apparent** wind angle, and the wind-shift sectors were built from **apparent** wind stored **boat-relative**. Both are the wrong reference:

- Close-hauled/beating angle is a true-wind quantity. Drawing the lines at `AWA ± laylineAngle` puts them off the **A** indicator instead of the **T** indicator, shifted toward the bow by `TWA − AWA` (10–20° when beating). When you're actually sailing the layline, your bow doesn't touch it.
- A wind-shift sector should track the **true wind direction** (compass), so only real shifts move it. Stored boat-relative, it smeared under heading changes.

This is upstream KIP behavior (`mxtommy/kip`); the sibling **racesteer** widget already uses true wind for its laylines. Fixing it here is a deliberate divergence.

## What

- **Laylines** now center on the boat-relative true wind (`TWA ± laylineAngle`) and recompute on the TWA stream. The compass/simple-mode branch collapses into a shared `toDialLocal()` helper, so simple mode tracks the wind too (it previously pinned to ±angle off the bow).
- **Hide, don't fake.** Laylines are hidden when the *configured* true-wind path produces no value, instead of drawing a bogus apparent-derived line. Gated by a new `trueWindActive` signal — keyed off whatever path the widget is configured with, nothing hardcoded.
- **Sectors** are fed from the true wind **direction** (compass frame) and converted to boat-relative at render through the same `toDialLocal()` helper, keeping laylines and sectors in one frame. Stale sectors are cleared when true wind drops out.

## Verification

- New unit tests encode the fix: close-hauled lines center on true wind (not apparent), hide when true wind is unavailable, and sectors position from the compass true-wind direction. `6/6` svg + `5/5` widget specs pass.
- `lint`, `build:prod`, and the `strictNullChecks` betterer ratchet all clean (ratchet tightened by one).
- Deployed to a live device (halpi.hurma) and confirmed visually: laylines render centered on the true wind, and hide when the configured path is absent.

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -225,15 +225,15 @@ exports[`strictNullChecks`] = {
       [42, 57, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
       [43, 57, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"]
     ],
-    "src/app/widgets/svg-windsteer/svg-windsteer.component.ts:3669162095": [
+    "src/app/widgets/svg-windsteer/svg-windsteer.component.ts:3559272095": [
       [28, 59, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
-      [39, 50, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
-      [44, 46, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
-      [45, 47, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
-      [46, 51, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
-      [48, 57, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
+      [40, 50, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
+      [45, 46, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
+      [46, 47, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
+      [47, 51, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
       [49, 57, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
-      [50, 57, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"]
+      [50, 57, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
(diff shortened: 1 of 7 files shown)


mairas:

**Mode:** interactive · **Scope:** `fix/windsteer-laylines-true-wind` vs `main`
**Intent:** move windsteer close-hauled lines + wind-shift sectors from apparent-wind to true-wind reference; hide laylines when the configured true-wind path is absent.
**Reviewers (5, parallel, independent contexts):** correctness, testing, maintainability, project-standards, adversarial. Adversarial added for subtle reference-frame geometry (~85 non-test lines changed).

All findings below were fixed in-branch and folded into the two commits (force-pushed).

### P1 — Critical

| # | File | Issue | Reviewers | Status |
|---|------|-------|-----------|--------|
| 1 | svg-windsteer.component.ts:522 | Simple (bow-fixed) mode rendered wind sectors at the absolute compass bearing, displaced from the wind by the heading — `computeSectorPath` used `compass.newValue` for the frame conversion, which the dial forces to 0 in simple mode. | correctness, adversarial, testing | **Fixed** — convert using the real `compassHeading` input; compass mode unchanged. Now covered by a simple-mode geometry test. |

### P2 — Moderate

| # | File | Issue | Reviewers | Status |
|---|------|-------|-----------|--------|
| 2 | svg-windsteer.component.ts:388 | The new sector-clear branch set the paths to `''` but didn't cancel the in-flight sector animation, so a running frame could overwrite the cleared path and leave a stale sector on screen — defeating the branch's purpose. | adversarial | **Fixed** — cancel `port/stbdSectorAnimId` and reset `windSectorsInitialized`, mirroring the disable branch. |

### P3 — Low

| # | File | Issue | Reviewer | Status |
|---|------|-------|----------|--------|
| 3 | svg-windsteer.component.ts:264 | Stale comment claimed laylines recompute in the AWA effect (they moved to the TWA effect). | maintainability | **Fixed** — removed. |
| 4 | svg-windsteer.component.svg:559 | Child layline `[style.display]` re-checked `trueWindActive()` — a strict subset of the parent group's gate, so redundant. | maintainability | **Fixed** — dropped the child bindings. |

### Pre-existing (filed as follow-up)

| File | Issue | Routing |
|------|-------|---------|
| svg-windsteer.component.svg | `#PortLayline` binds `closeHauledLineStbdPath()` and vice-versa (ids mislabeled vs bound paths); polar→cartesian projection duplicated 4×. | **#220** — cosmetic/refactor, out of scope for this fix. |

### Advisory / residual risk (no code change)

- **`trueWindActive` staleness:** it flips false only on an explicit null delta. If the source goes silently stale or the path is reconfigured to empty, laylines keep the last TWA. This matches the widget's overall behavior (`enableTimeout: false`, no staleness signal anywhere); the "never arrives" case is handled. Not a new regression.
- **windsteer vs racesteer divergence:** windsteer sectors now derive from true-wind direction; racesteer's still derive from apparent wind. A fix to one won't port cleanly to the other.

### Coverage

- Tests added for: simple-mode laylines + sectors, sector clear-on-null, and the widget-level feed switch (`trueWindActive` toggle, apparent no longer feeds sectors). svg 9/9, widget 7/7.
- Suppressed: findings below 0.60 confidence. `project-standards`: clean (no `any`, betterer ratchet regenerated 574→573, lint green).

### Verdict: Ready to merge (pending CI)

All P1/P2/P3 findings resolved in-branch; pre-existing items tracked in #220.

Issue #225, opened 2026-07-09
Unit 3: Animated page transition (two-phase slide)


mairas:
Part of #194 (page transition animation). Unit 3 of 3.

## Goal

Consume the direction, run the two-phase slide with an off-screen content swap, manage the guard flag, and fall back to an instant swap for reduced-motion and non-nav-driven changes.

**Requirements:** R1, R2, R3, R4, R7, R8, R9, R10

**Dependencies:** Unit 1 (`.page-slide` wrapper), Unit 2 (direction + flag).

## Files

- Modify: `src/app/core/components/dashboard/dashboard.component.ts`
- Modify: `src/app/core/components/dashboard/dashboard.component.scss` (if easing/transform helpers need declaring)
- Test: `src/app/core/components/dashboard/dashboard.component.spec.ts`

## Approach

- Add a `viewChild` for the `.page-slide` element.
- Restructure the `activeDashboard` effect to call a new `runPageChange(idx)` (untracked), replacing the direct `loadDashboard` call.
- `runPageChange`: consume `pendingPageDirection`; if null or `prefers-reduced-motion` → `loadDashboard(idx)` (existing behavior) and return. Otherwise `beginPageTransition()`, then (try/finally) exit phase → `loadDashboard(idx)` off-screen → reposition wrapper to the opposite edge without transition → enter phase, then `endPageTransition()`.
- Direction sign: `next` = exit to −100%, enter from +100%; `prev` mirrored.
- Durations/easing: 180 ms per phase, exit `ease-in`, enter `ease-out`, 100% translate (tune on the Pi-class throttle during verification).
- Wrap the per-phase animation in an overridable `animatePhase(...)` method (Web Animations API, transform-only) so specs can stub it — mirror the `now()`-override pattern in `scroll-nav.directive.ts`.
- `prefers-reduced-motion` via `matchMedia('(prefers-reduced-motion: reduce)')`, guarded for absence like `app-help.component.ts`.

**Execution note:** implement `runPageChange` test-first against a stubbed `animatePhase` — real Web Animations `.finished` does not resolve in jsdom.

## Test scenarios

- Happy path: pending `next` + motion allowed → `animatePhase(exit)`, then `loadDashboard`, then `animatePhase(enter)`, in order; `isPageTransitioning` true during, false after.
- Edge (direction sign): `prev` produces the mirrored translate targets vs. `next`.
- Reduced-motion: with reduce active → `loadDashboard` immediately, `animatePhase` never called, flag never left set.
[...]

Issue #224, opened 2026-07-09
Unit 2: Service travel-direction + non-interruptible transition guard


mairas:
Part of #194 (page transition animation). Unit 2 of 3.

## Goal

Capture wrap-aware travel direction and expose the non-interruptible transition flag + guard on `DashboardService`, without consuming them yet (inert on its own — the flag stays false and the direction is set-but-unread until Unit 3).

**Requirements:** R5 (one trigger for all entry points), R6 (source-provided, wrap-aware direction), R9 (non-interruptible guard)

**Dependencies:** none (can land before or after Unit 1).

## Files

- Modify: `src/app/core/services/dashboard.service.ts`
- Test: `src/app/core/services/dashboard.service.spec.ts`

## Approach

- Add `isPageTransitioning` (signal, default false) and `pendingPageDirection` (`'next' | 'prev' | null`), plus `begin/endPageTransition()` helpers that flip the flag.
- In `navigateToNextDashboard` / `navigateToPreviousDashboard` / `navigateTo`: return early if `isPageTransitioning()`; compute the target as today; **only when the target differs from the current active index**, record `pendingPageDirection` — `'next'`/`'prev'` for the dedicated methods (including wrap), `sign(index − active)` for `navigateTo` — before navigating.
- Do not change routing behavior otherwise. The component (Unit 3) writes the flag; here it only stays false, so the guard is inert.

**Note:** every user-facing nav funnels through these three methods (swipe/wheel/hotkey via `app.component.pageNav`, toolbar via `page-nav-control`, remote via `remote-dashboards.service`, and the freeboardsk/anchor-alarm/iframe widgets), so this is the single choke point.

## Test scenarios

- Happy path: next from a middle page records `next`; previous records `prev`.
- Edge (wrap): next from the last page records `next` (not `prev`); previous from page 0 records `prev`.
- Edge (jump): `navigateTo(higher)` records `next`; `navigateTo(lower)` records `prev`.
- Edge (no-op): `navigateTo(current)` records no direction and starts no transition; next/previous with a single dashboard (target === active) records no direction.
- Guard: with `isPageTransitioning` true, all three methods return without calling `router.navigate`.
- `endPageTransition` clears the flag.

## Verification

- Spec passes in CI; direction and guard behavior covered; no routing regressions.

Issue #223, opened 2026-07-09
Unit 1: Slide wrapper scaffold for page transitions


mairas:
Part of #194 (page transition animation). Unit 1 of 3.

## Goal

Introduce the `.page-slide` wrapper the animation will drive, with no behavioral or visual change. Lands the DOM structure Unit 3 animates; inert on its own.

**Requirements:** R4 (structural groundwork), R7

**Dependencies:** none.

## Files

- Modify: `src/app/core/components/dashboard/dashboard.component.html`
- Modify: `src/app/core/components/dashboard/dashboard.component.scss`

## Approach

- Wrap the dashboard surface (the `<gridstack>` and the empty-state overlay) in a single `.page-slide` container: `position: relative`, full width/height, `transform: none` at rest so it is inert.
- Keep the container as the positioning context so the absolute empty-state (`inset: 0`) and edit-mode buttons anchor to it and will travel with the page.
- No transform logic yet.

**Patterns to follow:** existing `:host` flex layout and `.grid-stack { width/height: 100% }` in `dashboard.component.scss`.

## Test scenarios

- Test expectation: none — structural wrapper only; verify no regression manually/visually.

## Verification

- App renders and behaves identically: grid sizes correctly (cellHeight from host height), empty-state overlay centered, edit mode (add/drag/resize via `getCellFromPixel`) unaffected, `npm run lint` clean.

PR #226, opened 2026-07-09
feat(dashboard): animate page transitions with a horizontal slide

### LINKED ISSUE #194: feat(nav): animate page transitions by sliding the page contents horizontally
Part of the navigation UX redesign (epic #183). Follow-up from PR-1 (#193) testing.

## Overview

Page (dashboard) switching is an instant swap. Add a horizontal slide so the direction of travel is felt. Chosen model: a **single-element push** — the one reused `DashboardComponent` / one in-place gridstack slides out in the direction of travel, swaps contents while off-screen, and slides the new page in from the opposite edge. Both animated phases move the same way, with an immediate handoff at the (invisible) midpoint. Transform-only, no `@angular/animations`, `prefers-reduced-motion` → instant swap, transitions non-interruptible.

Choreography (`next` = travel left; `prev` mirrors):

```
 t0   |==P0==|            P0 centered
 t1        P0==|          P0 sliding left, exiting left edge
 t2      ·····            empty background (crossover frame, no dwell)
 t3            |==P1      P1 entering from right, still moving left
 t4   |==P1==|            P1 centered
```

## Problem Frame

`page/:id` maps to a single `DashboardComponent` with no custom `RouteReuseStrategy`, so `page/0 → page/1` reuses the component; its `activeDashboard()` effect calls `grid.load(...)` and mutates one gridstack in place. There is only ever one grid in the DOM, so a true two-page filmstrip is out of scope — the slide is a single element pushed out and in, triggered off the `activeDashboard` change (which uniformly covers swipe, wheel, toolbar tab, hotkey, remote, and widget-driven nav). Angular's built-in router-outlet animations do not fire on a param-only change, so hooking `activeDashboard` is both necessary and better.

## Requirements

- R1. Swapped page content slides horizontally in the direction of travel; out one edge, in from the opposite edge, both moving the same way.
- R2. `next` travels left; `prev` mirrors.
- R3. Single-element choreography, immediate handoff, no dwell on the empty background.
- R4. Only the page content translates; the fixed toolbar stays above and does not move.
- R5. Triggered on the active-dashboard change so every entry point animates identically.
- R6. Travel direction comes from the navigation source (wrap-aware), not index comparison.
- R7. Transform-only, cheap on the Pi; no new animation dependency.
- R8. Respect `prefers-reduced-motion` → instant swap.
- R9. Non-interruptible: page-nav input during an in-flight transition is ignored.
- R10. Incoming content loads while off-screen; no partially-loaded page is ever visible mid-slide.

## Scope Boundaries

- Not a two-page filmstrip; both pages are never on screen at once; the midpoint is an empty background frame.
- No finger-following / drag-to-track; commit-triggered only.
- No change to `RouteReuseStrategy` or router-outlet animation wiring.
- Dashboard-to-dashboard nav only; settings/options/remote/help/data routes unaffected.

## Key Technical Decisions

- **Direction captured at the nav-method call site, consumed by the effect.** The three `DashboardService` methods know the wrap-aware direction; they record a one-shot `pendingPageDirection`. The component effect reads-and-clears it. `pendingPageDirection === null` cleanly means "not a guarded nav" (boot, deep link, `delete()`) → instant swap. Direction is armed **only when the target index differs** from the current, so a same-target no-op leaves no stale direction.
- **Non-interruptible guard = an `isPageTransitioning` flag on `DashboardService`,** checked by the three nav methods (the single choke point every entry point funnels through). The component sets it true at slide start / false at end via `begin/endPageTransition`, try/finally so it always clears. Coexists with the unchanged 500 ms wheel inertia cooldown (separate concern).
- **Single `.page-slide` wrapper inside `DashboardComponent`;** the host's existing `overflow: hidden` clips it; the fixed toolbar is untouched. Transform-only, compositor-friendly, no layout thrash.
- **Web Animations API for phase sequencing** (`.finished` promises): exit `0 → ∓100%`, off-screen `loadDashboard` + reposition to `±100%`, enter `±100% → 0`. Per-phase animate wrapped in an overridable method for testability (jsdom's `element.animate` doesn't resolve `.finished`).
- **Durations/easing:** 180 ms/phase (~360 ms total), exit `ease-in`, enter `ease-out`, 100% translate — tune on the Pi-class throttle.

## Open Questions

**Resolved during planning:** durations/easing (180 ms/phase, ease-in/ease-out); wrapper placement + off-screen load gating (`.page-slide` + `runPageChange` midpoint swap); guard mechanism vs. wheel cooldown (service flag + unchanged cooldown).

**Deferred to implementation:** WAAPI `fill`/`commitStyles` handling; whether a 1-frame rAF settle is needed before the enter phase on heavy pages (decide on-device); optional watchdog to auto-clear the flag if a router navigation is ever cancelled/errored (low risk — `/page/:id` has no guards).

## Implementation units

- [ ] #223 — Unit 1: Slide wrapper scaffold (inert DOM groundwork)
- [ ] #224 — Unit 2: Service travel-direction + non-interruptible transition guard
- [ ] #225 — Unit 3: Animated page transition (two-phase slide, reduced-motion fallback)

## Risks

- Heavy incoming page renders during the enter phase → jank on the Pi: load happens off-screen at the midpoint (R10); add a 1-frame rAF settle before enter if needed.
- WAAPI `.finished` unreliable in jsdom: behavior tested against an overridable `animatePhase` stub; timing verified on device/perf-harness.
- Adding the wrapper shifting gridstack coordinate math: wrapper is full-size at `translateX(0)` at rest; Unit 1 verifies edit-mode add/drag (`getCellFromPixel`) before any transform.

---

<details><summary>Original ask (pre-plan)</summary>

Animate the transition by **sliding the page contents horizontally** in the direction of travel (next = current slides out left, incoming enters from the right; previous = mirror). The motion should read as moving along a horizontal strip of pages, matching the horizontal nav gesture.

Pointers: nav goes through `DashboardService`; direction derived from next-vs-previous with wrap-around; respect `prefers-reduced-motion`; keep it cheap (transform-based, no layout thrash); the auto-hiding toolbar is `position: fixed` and should sit above the sliding content.

</details>

mairas:
## Summary

Page (dashboard) switching was an instant swap — no spatial continuity, so the direction of travel wasn't felt. This animates the switch as a **horizontal slide** in the direction of travel, cheap enough for the Pi-class target.

Part of the navigation UX redesign (epic halos-org/skip#183). Closes halos-org/skip#194, halos-org/skip#223, halos-org/skip#224, halos-org/skip#225.

## Approach

A **single-element push**, not a two-page filmstrip: there is only ever one reused `DashboardComponent` / one in-place gridstack, so the slide moves that single wrapper. For `next` the current page exits left, the incoming page loads while off-screen, then enters from the right — both phases moving the same way, with an immediate handoff at the (invisible) midpoint. `prev` mirrors.

```
 t0   |==P0==|         P0 centered
 t1        P0==|       P0 sliding left, exiting left edge
 t2      ·····         empty background (crossover frame, no dwell)
 t3            |==P1   P1 entering from right, still moving left
 t4   |==P1==|         P1 centered
```

The work splits cleanly across the service (which knows *where* nav is going) and the dashboard host (which owns *how* the page renders):

- **`DashboardService`** — `navigateTo` / `navigateToNext` / `navigateToPrevious` now record a one-shot, wrap-aware travel direction (consumed once by the host) and return early while a transition is in flight. Every nav entry point (swipe, wheel, toolbar tab, hotkey, remote, widget buttons) funnels through these three methods, so this is a single choke point for both direction capture and the non-interruptible guard.
- **`DashboardComponent`** — a `.page-slide` wrapper around the grid; the `activeDashboard` effect now runs `runPageChange`, which consumes the direction and drives a two-phase, transform-only Web Animations slide, loading the incoming page off-screen at the midpoint. When the change carries no direction (initial boot, deep link, browser history, delete) or `prefers-reduced-motion` is set, it falls back to the previous instant swap. The per-phase animate call is isolated behind an overridable method so specs can drive the sequence without real Web Animations.

Direction comes from the nav source rather than comparing indices, so wrap-around at the first/last page slides the correct way. Animation is transform-only (no layout thrash), the fixed toolbar stays above and does not move, and the host's existing `overflow: hidden` clips the sliding wrapper with no new CSS.

## Commits

1. `feat(dashboard): wrap page content in .page-slide for transitions` — inert wrapper scaffold.
2. `feat(dashboard): capture page travel direction and add transition guard` — service direction + non-interruptible flag.
3. `feat(dashboard): slide page contents on navigation` — the two-phase animated slide + reduced-motion fallback.

## Testing

- `npm run lint` — clean.
- `npm run build:dev` — compiles.
- `npm run betterer` — strictNullChecks baseline regenerated, unchanged at 573 (no new issues; the two edited files' hash keys updated per the repo's ratchet workflow).
- `dashboard.service.spec.ts` — 44 pass (7 new): direction incl. wrap, direct-jump sign, same-target no-op, single-dashboard no-op, one-shot consume, and the in-flight guard rejecting re-entrant nav.
- `dashboard.component.spec.ts` — 18 pass (5 new): phase/load ordering for next and prev, instant swap when no direction is pending, instant swap under reduced-motion, and flag cleanup on a failing phase.

**Still pending:** runtime visual verification against a live/mock Signal K backend — confirming on-device that the slide reads correctly (direction incl. wrap, no flash at the off-screen jump, toolbar stationary, timing) and that reduced-motion swaps instantly. The unit tests cover sequencing and decision logic, not the rendered animation.

No additional operational monitoring required — this is a client-side UI animation with no server, data, or persistence impact. Validation is visual: after deploy, switch pages by swipe/wheel/toolbar-tab/hotkey and confirm the slide direction (including wrap at the ends), that the toolbar stays put, and that an instant swap occurs under `prefers-reduced-motion`. Rollback is a straight revert of this branch if the animation misbehaves on the Pi target.

---

🤖 Generated with Claude Opus 4.8 (1M context, extended thinking) via [Claude Code](https://claude.com/claude-code)


mairas:
## Code review — page transition animation

**Scope:** `git diff bc03d72..HEAD` (3 commits, 7 files). **Mode:** interactive, PR-first.
**Reviewers:** correctness, testing, maintainability, project-standards (always-on) + reliability, adversarial (async transition sequence + state mutation across every nav entry point).
**Verdict: Not ready to merge** — one P2 concurrency defect (three independent reviewers converged), one P2 wedge risk, and one P2 test gap the plan explicitly required. Happy-path feel is fine; the defects are edge cases.

### P2 — Moderate

|---|------|-------|-----------|------|-------|
| 1 | `dashboard.component.ts` `runPageChange` | **Grid desyncs from URL when a non-guarded `activeDashboard` change lands mid-slide.** The `isPageTransitioning` guard only gates the three `navigate*` methods. Browser **back/forward** (and any future non-`navigate*` mutation) reaches `activeDashboard` via `NavigationEnd → setActiveDashboardIndex`, so the effect re-enters `runPageChange` while the first run is still awaiting its exit phase. The second run takes the instant path (direction already consumed → `null`) and loads immediately; then the first run's deferred `loadDashboard(capturedId)` executes ~180 ms later and **wins**. Final state: URL + page-nav highlight point at the back-target, grid shows the original swipe target — persists until the next nav. | correctness, reliability, adversarial | 0.75 | manual |
| 2 | `dashboard.component.ts` `runPageChange` finally / `ngOnDestroy` | **No teardown/timeout backstop — a non-settling slide wedges all navigation.** `isPageTransitioning` is a root-singleton cleared only by `runPageChange`'s `finally`, which runs only if both `animatePhase` promises settle. If `animation.finished` never settles, or the component is destroyed mid-slide, the flag can stay `true` forever; every `navigate*` then early-returns → nav dead until reload. `ngOnDestroy` neither cancels the in-flight animation nor clears the flag. (In practice current Chromium resolves `finished` after the duration even when detached, so the common case self-heals — but there's no guarantee.) | reliability | 0.60 | manual |
| 3 | `dashboard.component.spec.ts` | **The flag↔guard integration scenario the plan required is missing.** The mock's `begin/endPageTransition` are no-op stubs that never flip a real signal, and `order[]` records only `animate`/`load` — so a regression that moved `beginPageTransition()` *after* the first `await` (opening a re-entrancy window during phase 1, exactly what the flag prevents) would still pass every assertion. Nothing proves the flag is observable to the service guard while a slide is genuinely in flight. | testing | 0.85 | manual |

### P3 — Low

|---|------|-------|-----------|------|-------|
| 4 | `dashboard.service.ts` `navigate*` | **Guard set-timing gap.** The flag is checked synchronously in `navigate*` but only *set* later (in the component effect after the router round-trip). A second nav issued in that sub-tick window (Ctrl+Arrow auto-repeat, remote/programmatic burst) isn't blocked and recomputes from the stale active index. Closing #1 by arming the flag synchronously at commit also closes this. | adversarial, correctness | 0.58 | manual |
| 5 | `dashboard.service.ts` `_pendingPageDirection` | **Stale direction mis-animates a later unrelated change.** Direction is set before `router.navigate` resolves but consumed only when the effect fires. If that navigation is cancelled/redirected (or never yields a distinct `activeDashboard` change), the direction lingers and the next non-animating change (back, deep link) consumes it → a backward nav can slide forward. Low likelihood today (`/page/:id` has no route guards). Fix: tie the direction to its target index / clear on `NavigationCancel|Error`. | adversarial, correctness, maintainability | 0.52 | manual |
| 6 | `dashboard.component.ts:124` | **`void runPageChange(...)` → unhandled rejection if `loadDashboard` throws.** The instant-swap branch sits outside the `try/finally`, and `loadDashboard` dereferences `dashboards()[id].configuration` with no bounds/null guard. Reachability is low (indices are valid in normal operation), but a failed load surfaces as a discarded unhandled rejection with no toast rather than the synchronous effect error it was before. `loadDashboard`'s missing error handling is pre-existing. Fix: guard invalid/null index + `.catch` on the void call. | correctness, reliability | 0.50 | manual (pre-existing component) |
| 7 | `dashboard.component.spec.ts` | **`!slide` fallback branch untested.** `instrument()` always mocks `_pageSlide()` to an element, so the "wrapper not ready, direction pending" path (one-shot direction consumed, no animation, no `beginPageTransition`) is never exercised. | testing | 0.80 | manual |
| 8 | `dashboard.component.spec.ts:393` | **"phase fails" test models an impossible rejection.** The real `animatePhase` wraps `.finished` in `.catch(() => undefined)` and never rejects; the test forces a rejection that can't occur. The realistic mid-slide failure is `loadDashboard` throwing between phases (already stubbed) — retarget the test there. | testing | 0.68 | advisory |
| 9 | `dashboard.component.ts:270` | **`±100` off-screen distance left inline** while its siblings (`PAGE_SLIDE_PHASE_MS`, easings) are named constants. Extract `PAGE_SLIDE_OFFSCREEN_PCT = 100`. | maintainability | 0.62 | gated_auto |
| 10 | `dashboard.service.ts:25` | **`PageTransitionDirection` duplicates `PageNavDirection`** (`'next' \| 'prev'`, in `scroll-nav.directive.ts`). Defensible (importing the directive type into a service is worse coupling), but make it deliberate — a one-line comment or a shared neutral module. | maintainability | 0.60 | advisory |

### Requirements completeness (plan R1–R10)

All met in the happy path. R9 (non-interruptible) is **partially** met — the guard holds once the slide is animating, but findings #1/#4 show it doesn't cover non-`navigate*` `activeDashboard` changes or the pre-animation window. R1/R7/R10's real rendered choreography is verified only at call-ordering level (unavoidable in jsdom) — hence the on-device pass now underway.

### Coverage / notes

- **Positives:** `project-standards` clean — the version-bump policy correctly does **not** apply (skip produces no `.deb`: no `VERSION`/`debian/`/`apps/metadata.yaml`, CI has no version-bump-check job), so no bump was needed and none was made. betterer regen consistent (573→573; only `dashboard.component.ts`'s hash rekeyed, issue count 36→36). CSS checks came back clean: `:host { overflow:hidden }` clips the wrapper, `translateX(100%)` fully clears (element = viewport width), the `−100%↔+100%` jump is between awaits with no intervening paint (no flash), and the fixed toolbar isn't reparented by the transform.
- **Untested (deferred to on-device, now running):** the real `animatePhase` WAAPI path (`fill:'forwards'` + `setSlideOffset` + `animation.cancel()`), real `prefers-reduced-motion` `matchMedia`, and the `effect → runPageChange` wiring.
- **Suppressed:** none below the 0.60 gate that were P2+.

### Suggested fix order (before merge)

1. **#1 + #4** together: at the exit-phase midpoint, load the *current* `activeDashboard()` (not the captured id) and/or use a per-run token to abort a superseded run; arm the flag synchronously at nav commit. This closes the desync and the set-timing gap.
2. **#2:** clear the flag + cancel the animation in `ngOnDestroy`; optionally race each phase against a max-duration timeout.
3. **#3:** add the flag↔guard integration test with a deferred `animatePhase`.
4. Fold in **#5–#8** (cheap correctness/test hardening); **#9–#10** at discretion.

mairas:
## Addressed review findings

Two commits (`374402e2` fix, `0e31b791` doc) resolve the P2s and the cheap P3s from the [review](https://github.com/halos-org/skip/pull/226#issuecomment-4928766672).

**Core redesign of `runPageChange`:** it now takes no captured index — it always loads the **current** `activeDashboard()` at the slide midpoint, ignores re-entrant calls while a slide is animating (`_slideInFlight`), and settles on the latest target after the enter phase.

| # | P | Status | How |
|---|---|--------|-----|
| 1 | P2 | **Fixed** | Midpoint loads current `activeDashboard()`, not a captured id; `_slideInFlight` makes a re-entrant `runPageChange` (browser back/forward mid-slide) a no-op; a post-enter check settles on any change that landed during the slide. Grid can no longer diverge from the URL. |
| 2 | P2 | **Fixed** | `ngOnDestroy` sets `_destroyed`, cancels the in-flight animation, and calls `endPageTransition()`; the slide bails after each phase if destroyed. A torn-down component can't wedge the root-scoped flag. |
| 3 | P2 | **Fixed** | New test wires `begin/endPageTransition` to a real signal and gates `animatePhase` on a deferred promise, asserting the flag is held true for the whole slide and cleared after — pinning the flag-armed-before-first-yield timing the guard depends on. |
| 6 | P3 | **Fixed** | `loadDashboard` now takes `number \| null` and early-returns on null/out-of-range, so a stray change no longer throws into a discarded promise. Side benefit: strictNullChecks **573 → 571** (two pre-existing call sites became null-safe). |
| 7 | P3 | **Fixed** | Added a test for the wrapper-not-ready instant path (direction consumed exactly once, no animation, no `beginPageTransition`). |
| 8 | P3 | **Fixed** | Retargeted the failure test to `loadDashboard` throwing between phases (the real, reachable mid-slide failure) instead of an `animatePhase` rejection the production code can't emit. |
| 9 | P3 | **Fixed** | Extracted `PAGE_SLIDE_OFFSCREEN_PCT`. |
| 10 | P3 | **Fixed** | Comment on `PageTransitionDirection` noting the intentional mirror of `PageNavDirection`. |
| 4 | P3 | **Deferred** | Closing the pre-animation set-timing window needs the flag armed synchronously in the service `navigate*` methods, which widens the wedge surface (a committed nav that yields no `activeDashboard` change would strand the flag). The router coalesces rapid nav to the correct final page, so impact is a possible wrong-direction animation on a `next`-then-`prev` burst — cosmetic. Not worth the added surface. |
| 5 | P3 | **Deferred (largely closed by invariant)** | With the target-differs guard, a `navigate*` always yields the `activeDashboard` change that consumes its direction on the very next tick, and a fresh `navigate*` is blocked mid-slide — so a stale direction can't linger to mis-animate a later change in practice (Skip's `/page/:id` route has no guards that would cancel a navigation). |

## Verification

- `npm run lint` clean; `npm run build:prod` clean.
- `dashboard.component.spec.ts` — **23 pass** (10 for the transition, incl. the new re-entrancy, settle-on-latest, flag-held, wrapper-not-ready, and destroy-clears-flag cases).
- `dashboard.service.spec.ts` — **44 pass** (unchanged; service behavior untouched).
- `npm run betterer` — **571** (improved).
- Redeployed to `halpi.hurma` for continued on-device testing.

PR #231, opened 2026-07-10
fix(gauge-steel): clear resize timer on destroy (port Kip #1126)

mairas:
Ports upstream **mxtommy/Kip#1126** ("Fix minor gauge steel memory leak", squash `bf6bb874`).

## Problem
`GaugeSteelComponent.onResized` arms a 120 ms debounce timer (`resizeTimer`) but `ngOnDestroy` never cleared it. A resize landing just before destruction leaves the callback scheduled to run against a torn-down gauge instance.

## Fix
Clear the pending `resizeTimer` at the top of `ngOnDestroy`. Single hunk in `src/app/widgets/gauge-steel/gauge-steel.component.ts`; applied cleanly from upstream (line offset only). No betterer/lint impact.

### DIFF
--- a/src/app/widgets/gauge-steel/gauge-steel.component.spec.ts
+++ b/src/app/widgets/gauge-steel/gauge-steel.component.spec.ts
@@ -50,4 +50,26 @@ describe('GaugeSteelComponent', () => {
     // NaN section that does not draw; the upper is converted (identity mock -> 11.5).
     expect(sectionSpy).toHaveBeenCalledWith(10, 11.5, 'red');
   });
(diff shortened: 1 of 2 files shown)


mairas:

**Finding (minor, testing):** the resize-timer cleanup shipped without a test.
**Addressed:** added a fake-timer spec — `onResized` arms the 120 ms debounce, `ngOnDestroy` clears it, and the debounced rebuild does not fire after destroy. The fix itself is correct (timer cleared before the gauge/canvas teardown).

PR #232, opened 2026-07-10
fix(ais-radar): stabilize target dialog lifecycle (port Kip #1097)

mairas:
Ports upstream **mxtommy/Kip#1097** ("Reapply fix AIS panel lifecycle and ExpressionChangedAfterItHasBeenCheckedError", squash `8ad825ca`).

## Changes
- `dialog-ais-target.component.ts`: start the clock in `ngOnInit` instead of the constructor, avoiding `ExpressionChangedAfterItHasBeenCheckedError` under OnPush change detection.
- `widget-ais-radar.component.ts`: pass a `structuredClone` of the AIS track when opening the target dialog, so subsequent live stream updates don't mutate the dialog's snapshot.

Applied cleanly from upstream (line offsets only). `.betterer.results` regenerated (issue count unchanged at 571; the `widget-ais-radar` entry was rekeyed by the content change). Production build passes.

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -240,10 +240,10 @@ exports[`strictNullChecks`] = {
       [23, 41, 4, "Argument of type \'null\' is not assignable to parameter of type \'string\'.", "2087897566"],
       [24, 42, 4, "Argument of type \'null\' is not assignable to parameter of type \'string\'.", "2087897566"]
     ],
-    "src/app/widgets/widget-ais-radar/dialog-ais-target/dialog-ais-target.component.ts:1637448081": [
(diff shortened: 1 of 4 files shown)


mairas:

**Finding (significant):** the open-time `structuredClone` froze the dialog, which is designed to update live — position, CPA, TCPA and the "Received X ago" counter stopped updating while the still-ticking clock inflated the frozen age (a live, transmitting target would read as going stale).
**Addressed:** the dialog now re-reads the live track by id from `AisProcessingService` each render, keeping the snapshot only as a fallback once the target is evicted — restoring live data while staying mutation-safe. Added a live-read + snapshot-fallback spec.

PR #233, opened 2026-07-10
fix(autopilot): guard V1 autopilotMode path assignment (port Kip #1096)

mairas:
Ports upstream **mxtommy/Kip#1096** ("Fix", squash `adfa758c`).

## Problem
`startV1Subscriptions()` wrote `this.runtime.options().paths['autopilotMode'].path = …` with no guards. Skip was more exposed than upstream's pre-fix code — it lacked even the `?.` after `options()` — so a missing config or absent `autopilotMode` entry threw.

## Fix
Read `runtime.options()` once and guard both `cfg?.paths` and `paths['autopilotMode']` before assigning. Manual port (skip's line had diverged from upstream). Bonus: betterer strictNullChecks improved by 2 (571 → 569). Lint clean.


mairas:

**Finding (minor, testing):** the V1 `autopilotMode` path guard shipped with no test for its three branches.
**Addressed:** added specs for no-`paths`, `paths` without `autopilotMode` (both must not throw), and the present case rewriting the path to the V1 legacy path. The guard itself is correct.

mairas:
Note: while this was open, main independently gained an equivalent guard (commit `0e98bd49`, a strictNullChecks fix that also handles the missing-`autopilotMode` case with a warning). Rebased onto that — the guard code is now main's, so this PR reduces to the **test coverage** for it (which main's guard otherwise lacks). The three specs pass against main's implementation.

PR #234, opened 2026-07-10
fix(horizon): correct frame visibility logic (port Kip #1128)

mairas:
Ports upstream **mxtommy/Kip#1128** ("correct frame visibility logic for WidgetHorizonComponent", squash `8c7b2ad6`).

## Problem
The Horizon widget config has a **"Show Frame"** checkbox bound directly to the `noFrameVisible` field (`root-modal-widget-config.component.html:720`, no inversion). `buildOptions` computed `frameVisible: !(cfg.gauge?.noFrameVisible ?? false)`, so checking "Show Frame" set the field `true` and then *hid* the frame — the opposite of what the label promises.

## Fix
Drop the negation: `frameVisible: (cfg.gauge?.noFrameVisible ?? false)`, matching the checkbox. One-line manual edit (skip's surrounding `frameDesign` line differs from upstream, so the patch was applied by hand). Betterer/lint clean.

### DIFF
--- a/src/app/widgets/widget-horizon/widget-horizon.component.spec.ts
+++ b/src/app/widgets/widget-horizon/widget-horizon.component.spec.ts
@@ -1,0 +1,53 @@
+import { signal } from '@angular/core';
+import { TestBed } from '@angular/core/testing';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { WidgetHorizonComponent } from './widget-horizon.component';
+import { WidgetRuntimeDirective } from '../../core/directives/widget-runtime.directive';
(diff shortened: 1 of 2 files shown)


mairas:
## Code review — correctness, testing, maintainability

**Finding (real bug):** the fix was incomplete. `frameVisibleView` (line 117, drives the wrapper padding) still negated `noFrameVisible` while `buildOptions().frameVisible` no longer did, so the two consumers of the same field had opposite polarity and the padding ran inverse to the frame — frame drawn → 0px padding, no frame → 10px.
**Addressed:** dropped the negation on line 117 so both consumers agree, and added a spec pinning the invariant (frame option and padding-driver track together for `noFrameVisible` true/false).

PR #238, opened 2026-07-10
feat(windsteer): gate COG/waypoint/drift on compass mode (port Kip #1099)

mairas:
Ports the behavioral part of upstream **mxtommy/Kip#1099** ("streamline Windsteer", squash `377fae9c`).

## What it does
The COG, waypoint, drift and current indicators are only meaningful when the wind steer has a compass reference. This change:
- **SVG** (`svg-windsteer.component.svg`): gates the `wptIndicator`, `setIndicator`, `layerCurrent` and `cogIndicator` display bindings on `&& compassModeEnabled()`, so they hide when compass mode is off.
- **Config** (`root-modal-widget-config.component.ts`): adds `setupWindsteerControlState()`, which disables the COG/waypoint/drift toggles when compass mode is off and re-enables them when it's on (reacting to the `compassModeEnabled` control).
- **Null guard** (`widget-windsteer.component.ts`): the wind-sector cutoff falls back to 5 s if `windSectorWindowSeconds` is unset.

## Port notes
- SVG and widget-windsteer hunks applied cleanly; the config method was hand-inserted (skip's `ngOnInit` had diverged by line offset).
- **Deliberately omitted** the upstream `svg-windsteer.component.ts` hunk: it's incidental input type-widening (`input<number>` → `input<number | undefined>`) plus `?? 0` defaults that skip's existing `updateWindSectors` early-return guard already covers. Porting it would inflate the strictNullChecks betterer ratchet for no behavior change.
- betterer strictNullChecks improved (the `?? 5` guard fixed one issue); lint and production build pass.


mairas:
## Code review — correctness, adversarial, testing, maintainability

**Findings (test gaps — the shipped code is correct):** the SVG compass-mode gating and the config toggle-disabling had no tests. **Nits:** the `?? 5` fallback duplicated the widget's own default, the sync closure named one control but gated three, and there was a stray blank line.
**Addressed:** added specs for the SVG gating (indicators hide when compass off, show when on) and the config method (COG/waypoint/drift disabled when compass off, re-enabled on toggle); extracted the 5 s default into a single constant, renamed the closure, removed the blank line.

One integration risk was checked and cleared: disabling the config toggles does not strip them from persisted config, because `submitConfig()` uses `getRawValue()`.

PR #244, opened 2026-07-13
Add deploy-halos target for durable Signal K install

mairas:
## Why

Skip was deployed by copying `public/` + `plugin/` + `package.json` straight into the Signal K server's `node_modules`. An undeclared package there is *extraneous*, so the next `npm install` in the SK data dir (any plugin add/update, or the influxdb prestart install) **prunes** it — Skip kept silently vanishing from the device after unrelated plugin work.

## What

Adds `./run deploy-halos <target>`:

- Builds the production bundle **locally** (never on the device — HaLOS test-device policy).
- Installs the package under `system-plugins/skip/` on the SK data volume, symlinks it into `node_modules` the way npm links a `file:` dep, and declares `@halos-org/skip: file:system-plugins/skip` in the SK `package.json` — so it survives prunes and restarts like an app-store install.
- `local` runs every step on the device; a hostname deploys over ssh (via `bash -s`, since the device login shell is fish).
- Verifies the result through Traefik on `:4430` with the device CA (never the container's incidental `:3000` port).

## Docs

Updates the fork's deploy note to point at `deploy-halos` and warn against the bare `node_modules` copy. Trivial fly-by in the same file: drops the undefined "standard CI triad" phrasing — the sentence already states the mechanism (publishing a GitHub release triggers `release.yml` to `npm publish`).


mairas:
## Code review — `deploy-halos` target

**Scope:** `run` (+120, the new `deploy-halos` function), `CLAUDE.md` (1 doc line). Untracked `perf-harness/mock-serve.mjs` excluded (not staged).
**Intent:** build Skip locally, install onto a HaLOS device's Signal K as a durable `file:` dependency that survives `npm` prune; `local` runs on-box, a hostname deploys over ssh.
**Reviewers (7):** correctness, testing, maintainability, project-standards (always) + security, reliability, adversarial. Model: opus, independent contexts.

Two convergences worth calling out: **correctness + adversarial** independently hit the migration bug (P1), and **correctness + reliability** independently hit the verify-block abort (P2). **security** and **project-standards** both returned no findings — the nested-heredoc design passes derived values as positional args (no injection), TLS is genuinely verified (`--cacert`, no `-k`), and the docstring convention / test-device policy / Traefik-`:4430` policy all check out.

---

### P1 — High

| # | Location | Issue | Reviewers | Conf |
|---|----------|-------|-----------|------|
| 1 | `run:113` `ln -sfn` | **Migration silently serves stale code.** On any device still holding the old bare-copy `node_modules/@halos-org/skip` (a real directory — the exact state this feature exists to replace), `ln -sfn ../../system-plugins/skip <dir>` does **not** replace the directory and does **not** error: GNU `ln` descends into it and creates a nested `.../skip/skip` symlink, exit 0. Node keeps resolving to the stale copy, the new build is never served, and the verify block still reports success (the stale dir is a valid package). Empirically confirmed on coreutils 9.7. The one upgrade path the command targets is the one it breaks, silently. | correctness, adversarial | 0.90 |

**Fix:** clear the link location before linking, e.g. `rm -rf "${data}/node_modules/${pkg}"` immediately before the `ln -sfn` block (handles both a stale real dir and an old symlink). Add a post-condition assert that the path is a symlink resolving into `system-plugins/`.

### P2 — Moderate

| # | Location | Issue | Reviewers | Conf |
|---|----------|-------|-----------|------|
| 2 | `run:141` verify | **Verify aborts *after* the restart on the port/domain lookups.** Under the REMOTE block's `set -o errexit -o pipefail`, `port=$(grep '^signalk-server=' /etc/halos/port-registry 2>/dev/null \| cut …)` and `domain=$(hostname -d 2>/dev/null)` are bare assignments: a missing/renamed registry key or a non-zero `hostname -d` aborts the whole command — *after* `docker restart` already applied the deploy. The `${port:-4430}` / `${domain:+…}` fallbacks that were clearly meant to tolerate this never execute. The author already guarded `code=$(… \|\| true)` two lines down but not these. | correctness, reliability | 0.90 |
| 3 | `run:159` | **Verify never fails the command, yet prints "verified above".** Every verify failure (SK not healthy in budget, webapp not listed, false-negative) downgrades to a printed `WARN`, then the function unconditionally echoes `Deployed … (verified above, via Traefik on port 4430).` and returns 0. A CI/automation caller — or the operator skimming — reads success even when SK never came back or the deploy is finding #1's stale copy. | reliability, testing | 0.82 |

**Fix (2):** append `|| true` to the `port=` and `domain=` assignments (matching the existing `code=` guard). **Fix (3):** track a verify status and either exit non-zero or drop the "verified" wording when a check didn't pass.

### P3 — Low

| # | Location | Issue | Reviewers | Conf |
|---|----------|-------|-----------|------|
| 4 | `run:147` | Verify curls set no `--max-time`/`--connect-timeout`, so the "75s" wait budget is illusory against a hung Traefik/SK backend (connection accepted, response stalls). `25×3s` may also be too short for a cold restart on a Pi. | reliability | 0.78 |
| 5 | `run:150` | `curl … \| grep -qF` under `pipefail`: `grep -q` closes the pipe on match, and if `/skServer/webapps` exceeds the pipe buffer (~64KB, plausible on a device with many webapps) curl dies with SIGPIPE → pipeline non-zero → false `WARN` on a served webapp. Capture to a var, then grep. | correctness, reliability | 0.62 |
| 6 | `run:91` | Pre-deploy backup tarballs accumulate in `$HOME` unbounded (no retention) on a storage-constrained device; same-second re-deploys collide on the timestamp filename and overwrite the prior backup. | maintainability, reliability, adversarial | 0.70 |
| 7 | `run:87,118` | No cheap regression gate: `bash -n` does **not** parse the quoted heredoc bodies (PREP/REMOTE/PY), so the most edit-prone code is unchecked. The device-independent pieces — the python `package.json` registration, the scope-aware symlink depth (`../../` vs `../`), the base-URL derivation — are unit-testable off-device but have zero coverage. A lint/`py_compile` gate + a small unit test would be proportionate. | testing | 0.82 |
| 8 | `run:57,84` | Docstring says "the build runs here, **never on the device**", but `deploy-halos local` runs `build` on the box it's invoked on — i.e. the device. Fine when that box is the dev/primary host, but the "never on the device" wording is self-contradictory for `local` mode; reword. | adversarial | 0.6 |
| 9 | `run:46,71,119` | Maintainability: hardcoded HaLOS paths with no shared source of truth; 3-level heredoc nesting (bash→bash→python) is the least legible, most-likely-edited region; ~115-line target bundles 7 responsibilities (build/backup/sync/link/register/restart/verify) — an outlier vs the file's one-line npm wrappers. | maintainability | 0.6–0.68 |

---

### Coverage & residual risks (non-blocking)

- **Non-transactional deploy, no auto-rollback.** `rsync --delete` → symlink → package.json rewrite → `docker restart` run in sequence and the restart precedes verify; a mid-sequence failure or dropped ssh leaves partial state, recoverable only by manually restoring the tarball.
- **Fresh-install path** (no existing `${data}/package.json`) makes the python `json.load` raise → aborts *after* the filesystem mutations, leaving files+symlink but the dep unregistered and no restart (adversarial, 0.6).
- **Non-atomic `package.json` rewrite** races with SK's own npm installs (app-store plugin add, influxdb prestart) — a concurrent writer's entry can be silently dropped (adversarial, 0.6). Narrow window.
- **`$target` foot-guns:** a value named `local`/`localhost` can't be reached over ssh (sentinel collision); a leading-dash `$target` is parsed as an ssh/rsync option (operator-only, not exploitable as shipped) — security flagged both as sub-threshold residuals, not vulnerabilities.
- **Verify FQDN fragility:** `https://$(hostname)${domain:+.$domain}:…` can mis-derive (double-domain, or bare hostname) and fail cert-SAN matching → spurious `WARN` on a good deploy. Verify-only impact.
- **Suppressed:** security & project-standards returned 0 findings; their observations are the residuals above. No findings dropped below the 0.60 gate.

---

### Verdict: **Ready with fixes** — 3 must-fix, all small

1. **P1 #1** (`rm -rf` before `ln`) — blocks merge: without it, migrating a device from the old method silently serves stale code and reports success.
2. **P2 #2** (`|| true` on port/domain) and **P2 #3** (verify should signal failure, not claim "verified") — the verify block currently both false-fails and false-succeeds depending on the path.

P3 #4–#8 are quick hardening (curl `--max-time`, SIGPIPE-safe grep, backup retention, a `bash -n`/`py_compile` gate + tiny unit test, docstring wording). P3 #9 is discretionary. Everything is contained to the new function; the docs change is fine as-is.

mairas:
## Review follow-up — findings triaged, fixes applied

Re-triaged against the reality of this project: a **single, owner-controlled device**, interactive use, no CI consuming this target. The reviewers (esp. adversarial) scoped some findings for a "many HaLOS installs / multiple developers" world that doesn't apply here.

### P1 (ln stale-directory) — deprioritized, not fixed
The trigger is **device state**, not multiple developers: `ln -sfn` only no-ops when `node_modules/@halos-org/skip` is a *real directory*. That state doesn't exist on the target (it's a symlink), and the only way it reappears on a controlled box is installing Skip from the SK app store first. Low-likelihood, self-evident if it ever happens, and the guard (`rm -rf` before `ln`) is complexity this environment doesn't need. Simplicity preferred over the corner case.

### Applied (folded into the feat commit)
- **Verify no longer over-claims.** Final line dropped the "(verified above)" assertion — the `OK`/`WARN` + HTTP code speak for themselves; the tool no longer says "verified" on a `WARN`. (was P2 #3)
- **Verify can't abort the run.** Removed the `port=$(grep … port-registry)` lookup entirely and hardcoded `:4430` (matches the "always Traefik 4430" policy — *fewer* lines, and no `set -e` abort after the restart); guarded `hostname -d` with `|| true`. (was P2 #2)
- **`--max-time 10`** on all three verify curls, so a stalled backend can't hang the deploy indefinitely. (was P3 #4)
- **Docstring fixed** — "the build runs here, never on the device" was false for `deploy-halos local`; now says the build runs on the machine you invoke it from. (was P3 #8)

### Intentionally not done (over-engineering for this context)
SIGPIPE-safe grep (needs a >64 KB webapps response — not a real body size here), backup-tarball retention logic, a `bash -n`/`py_compile` CI gate + unit tests, and the naming/nesting refactors. All add machinery for negligible return on a solo one-device ops target.

Security and project-standards were clean. Merging.

PR #248, opened 2026-07-13
fix(dashboard): send invalid /page/:id deep links to page 0

### LINKED ISSUE #205: Deep-link / invalid /page/:id crashes the dashboard (loadDashboard has no null guard)
## Problem

`DashboardComponent.loadDashboard(dashboardId)` dereferences the dashboard without a null/undefined guard, and the effect that drives it can be called with an invalid index. A deep-link (or reload) to a `/page/:id` that isn't a valid dashboard index leaves `activeDashboard` at `null`, and once the gridstack grid is ready the component crashes / renders blank.

```
// dashboard.component.ts
effect(() => {
  const activeIdx = this.dashboard.activeDashboard();   // number | null
  untracked(() => { this.loadDashboard(activeIdx); });  // called with null / invalid
});

private loadDashboard(dashboardId: number): void {
  const dashboard = cloneDeep(this.dashboard.dashboards()[dashboardId]); // dashboards()[null|9|1.5] === undefined
  const _gridstack = this._gridstack();
  if (!_gridstack?.grid) { queueMicrotask(() => this.loadDashboard(dashboardId)); return; }
  _gridstack.grid.load(dashboard.configuration);        // undefined.configuration -> TypeError
  ...
}
```

Two ways in:

1. **Out-of-range id** — boot to `/page/9` on a 3-dashboard install. `applyDashboardParam('9', true)` parses `9` (not NaN), so neither `defaultToZero` branch fires; `setActiveDashboardIndex(9)` rejects the out-of-range index and `activeDashboard` stays `null`. `/page/abc` is safe (NaN → default 0); only the numeric-but-invalid case leaks through.
2. **Fractional id** — `/page/1.5`. `dashboard.service.ts` now rejects non-integer indices (from #149), so `activeDashboard` stays `null`; before #149 it set `activeDashboard` to `1.5`, which was also broken (`dashboards()[1.5]` is `undefined`). Either way the page crashes — #149 only changed the route in.

## Scope / root cause

- `DashboardService.applyDashboardParam` (`dashboard.service.ts`) only falls back to index 0 for `null`/empty/`NaN`, not for a finite-but-invalid (out-of-range / non-integer) id.
- `DashboardComponent.loadDashboard` (`dashboard.component.ts`) has no guard for an out-of-range/`null` `dashboardId`.

## Suggested direction

- `loadDashboard`: bail (or clamp to a valid index) when `dashboards()[dashboardId]` is `undefined`, instead of dereferencing it.
- `applyDashboardParam`: treat any id that `setActiveDashboardIndex` would reject the same as `NaN` — fall back to 0 on the `defaultToZero` path so a stale/invalid deep link lands on page 0.

## Context

Pre-existing; surfaced by an adversarial review pass on PR #201 (which fixed the sibling `delete()` out-of-range regression but deliberately left this component-level crash out of scope). Not introduced by #149.

mairas:
## What

An invalid `/page/:id` deep link — out-of-range (e.g. `/page/9` after dashboards were removed) or fractional (`/page/1.5`) — left the dashboard **blank and stuck** on that URL instead of falling back to page 0.

## Why

`applyDashboardParam` only fell back to page 0 for `null`/empty/`NaN` ids. A finite-but-invalid id parses as a real number, so it was handed to `setActiveDashboardIndex`, which correctly rejects it — but that left `activeDashboard` at `null`, and the component renders an empty dashboard.

This is the remaining half of #205. The component-level crash guard (`loadDashboard` dereferencing `undefined`) already landed in `374402e2`; this fixes the leftover blank-page behaviour so a stale deep link lands on page 0.

## How

- `setActiveDashboardIndex` now returns whether it applied the index.
- `applyDashboardParam` falls back to page 0 on the initial (default-to-zero) path when the finite id is rejected — treating out-of-range/fractional ids the same as `NaN`.
- Later in-app navigations keep the current page on a bad id, unchanged (the `console.error` diagnostic is preserved on that path).

## Tests

Added two `dashboard.service.spec.ts` cases — first `NavigationEnd` with `/page/9` and `/page/1.5` → page 0. Existing route / `setActiveDashboardIndex` tests are unchanged.

## Verification note

Not run locally: the on-boat host this was authored on has no `node_modules`, and the HaLOS test-device policy forbids installing/building there. Relying on CI (Node 24: lint + betterer:ci + tests) to verify. `dashboard.service.ts` is not in `.betterer.results`, so the ratchet is unaffected.

Closes #205

### DIFF
--- a/src/app/core/services/dashboard.service.spec.ts
+++ b/src/app/core/services/dashboard.service.spec.ts
@@ -133,6 +133,23 @@ describe('DashboardService', () => {
       expect(service.activeDashboard()).toBe(0);
     });
 
+    it('defaults to 0 when the first NavigationEnd param is a finite but out-of-range index', () => {
+      setup();
+      router.setIdParam('9');
+      router.emitNavigationEnd();
+      expect(service.activeDashboard()).toBe(0);
+      // The rejected id is still logged once; the fallback to page 0 is silent.
+      expect(consoleError).toHaveBeenCalledTimes(1);
+    });
+
+    it('defaults to 0 when the first NavigationEnd param is a fractional index', () => {
+      setup();
+      router.setIdParam('1.5');
+      router.emitNavigationEnd();
+      expect(service.activeDashboard()).toBe(0);
+      expect(consoleError).toHaveBeenCalledTimes(1);
+    });
+
     it('follows id param changes on later navigations', () => {
       setup();
       router.emitNavigationEnd();
--- a/src/app/core/services/dashboard.service.ts
+++ b/src/app/core/services/dashboard.service.ts
@@ -123,23 +123,30 @@ export class DashboardService {
       return;
     }
 
-    this.setActiveDashboardIndex(parsed);
+    // A finite but out-of-range or non-integer id (e.g. a stale /page/9 deep link)
+    // is rejected by setActiveDashboardIndex; on the initial (default-to-zero) path,
+    // land on page 0 rather than leaving the dashboard unset and rendering blank.
+    if (!this.setActiveDashboardIndex(parsed) && defaultToZero) {
+      this.setActiveDashboardIndex(0);
+    }
   }
 
   /**
    * Sets the active dashboard index in the service.
    * This only updates the internal state and does NOT trigger navigation or URL changes.
    * @param itemIndex The index of the dashboard to activate.
+   * @returns true if the index was valid and applied (or already active); false if rejected.
    */
-  public setActiveDashboardIndex(itemIndex: number): void {
-    if (itemIndex === this.activeDashboard()) return;
+  public setActiveDashboardIndex(itemIndex: number): boolean {
     // No change if the same dashboard is selected to prevent unnecessary cascading updates
+    if (itemIndex === this.activeDashboard()) return true;
 
     if (Number.isInteger(itemIndex) && itemIndex >= 0 && itemIndex < this.dashboards().length) {
       this.activeDashboard.set(itemIndex);
-    } else {
-      console.error(`[Dashboard Service] Invalid dashboard ID: ${itemIndex}`);
+      return true;
     }
+    console.error(`[Dashboard Service] Invalid dashboard ID: ${itemIndex}`);
+    return false;
   }
 
   /**


mairas:
## Code review — #205 fix

**Scope:** `dashboard.service.ts` + `dashboard.service.spec.ts` (1 commit vs `main`). `perf-harness/mock-serve.mjs` is untracked and excluded.
**Intent:** an invalid `/page/:id` deep link (out-of-range / fractional) must land on page 0 on the initial route path instead of leaving `activeDashboard` null (blank), without changing later-navigation behavior or dropping the invalid-id diagnostic.
**Mode:** interactive. **Reviewers (5):** correctness, testing, maintainability, project-standards (always-on) + api-contract (public `setActiveDashboardIndex` return type changed `void → boolean`).

The core fix is **confirmed correct**: correctness traced every branch and found no defect; api-contract confirmed `void → boolean` is backward-compatible (sole caller is `applyDashboardParam`, no external callers, no stub drift); project-standards confirmed no `any`, betterer baseline unaffected (`dashboard.service.ts` isn't in `.betterer.results`), and the commit is conventional-compliant.

### P2 — Moderate

| 1 | dashboard.service.spec.ts:136 | New fallback tests asserted only `activeDashboard()===0`, not that the invalid-id `console.error` still fires — but preserving that diagnostic is part of the intent, so a refactor that dropped the log on the fallback path would stay green. | testing, correctness | 0.82 | safe_auto → **fixed in this review** |

### P3 — Low

| 2 | dashboard.service.ts:71 | The immediate-init path (`applyDashboardParam(param, false)` at construction) with an invalid id is untested. Theoretical ordering: if initial router navigation completes *before* the service constructs, immediate-init won't fall back (`defaultToZero=false`) and `navEnd$`'s `take(1)` never re-fires → blank page. Both reviewers hedged; depends on injection timing. | testing, correctness | 0.70 | residual / possible follow-up |
| 3 | dashboard.service.ts:114 | The fallback-to-0 rule is spelled out in three guard branches (null/empty, NaN, finite-invalid). A future change to the fallback target must touch all three. Current early-return form is also readable — judgment call. | maintainability | 0.60 | advisory |

### Pre-existing (not introduced here)

| # | File | Issue | Route |
|---|------|-------|-------|
| 4 | dashboard.service.ts:144 | The index range-check (`isInteger && 0<=i<length`) is hand-written across `setActiveDashboardIndex`, `delete`, `duplicate`, and `navigateTo`. Notably **this fix does the right thing** — it reuses the new boolean return rather than adding a 5th copy. | optional follow-up issue; does **not** block merge |

### Applied fixes

- Added `expect(consoleError).toHaveBeenCalledTimes(1)` to both new tests — asserts the rejected id is logged once and the fallback to page 0 is silent. Folded into the fix commit.

### Coverage

- Suppressed: 0 findings below threshold. All 5 reviewers returned.
- Testing gaps (not blocking): no deep-link test for a negative id (`/page/-1`) or the exact boundary `id === length` — both are covered at the `setActiveDashboardIndex` unit level and share the same rejection path as the tested `/page/9`. The fallback tests don't assert the page swap is non-animated (`consumePendingPageDirection` stays null).
- Verification note: not run locally (this on-boat host has no `node_modules`; HaLOS policy forbids installing/building here). CI (Node 24: lint + betterer:ci + tests) is the gate.

---

**Verdict: Ready to merge** once CI is green. The one actionable gap (P2) is fixed in this review; the P3s are advisory/residual and the pre-existing DRY smell is explicitly not widened by this change. Finding #2 is worth a quick follow-up check of `DashboardService` injection timing but is pre-existing and out of #205's stated scope.

PR #252, opened 2026-07-13
fix(data): scope path timeout reset to the timed-out source

### LINKED ISSUE #206: timeoutPathObservable resets all sources for a path, nulling healthy sibling sources
## Problem

`DataService.timeoutPathObservable(path, pathType)` resets **every** registration for a path (PR #201 / #155 changed it from a first-match lookup to iterating `_pathRegisterByPath.get(path)`). That is correct when the whole path goes silent, but the caller only knows that **one source's** stream timed out — it passes no source:

```
// widget-streams.directive.ts
this.data.timeoutPathObservable(normalizedPath, pathType);   // no source

// data.service.ts
const registrations = this._pathRegisterByPath.get(path);
for (const registration of registrations) {
  registration.pathDataUpdate$.next({ data: { value: null, timestamp: null }, state: States.Normal });
}
```

On a vessel with redundant sensors (e.g. two GPS or two wind sources) shown by separate widgets bound to different `$source`s, when the dead source's timeout fires it also nulls the still-live sibling's registration (and the `default` bucket, which every source's delta updates). The live widget flashes blank and any active alarm/warn state momentarily reverts to `Normal`, recovering only on the next delta — a visible flicker exactly during the partial-failure condition redundant sensors exist to smooth over.

The single-source case (one widget on `default`) is unaffected and behaves as before.

## Suggested direction

Thread the source through so the timeout resets only the registration(s) it applies to:

- Pass the timed-out `source` from `WidgetStreamsDirective` into `timeoutPathObservable`.
- Reset only the registration whose `source` matches, plus the `default` bucket **only** when the timed-out source is `default` (default silence implies all sources silent).

## Context

Follow-up to #155 (PR #201). #155's stated fix was "iterate all registrations for the path," which is what shipped and is correct for the common whole-path-silent case; this issue tracks the source-scoping precision refinement. Surfaced by correctness + reliability review of PR #201.

mairas:
## What

`DataService.timeoutPathObservable` nulled **every** registration for a path when any one source's stream timed out. On a vessel with redundant sensors (two GPS, two wind) shown by separate widgets bound to different `$source`s, a dead source's timeout also cleared the still-live sibling and the `default` bucket — a visible blank flicker (and a momentary alarm/warn → Normal revert) exactly during the partial failure redundant sensors exist to smooth over.

Closes #206.

## Why it was like this

#155 (PR #201) fixed a real bug where a first-match-by-path lookup reset only *one* of a path's registrations. It over-corrected to "reset all". #206 refines that: reset only the registration for the source that actually timed out.

## How

- `timeoutPathObservable(path, pathType)` → `timeoutPathObservable(path, source, pathType)`; it resets only the registration whose `source` matches.
- `WidgetStreamsDirective` passes the stream's effective source (the same `pathCfg.source?.trim() || 'default'` it already subscribes with) into the timeout call.

Correctness is preserved for a genuinely-silent path: each `(path, source)` stream has its own timeout, so when the whole path goes quiet every widget's stream self-times-out and clears its own registration — the `default` bucket included (default is fed by every source, so it only times out once all sources are silent).

## Tests

- `data.service.spec.ts`: the old *"resets every source registration… not just the first"* test (which pinned the #201 all-at-once behavior) is rewritten to *"resets only the timed-out source registration, leaving sibling sources live"* — two sources + default, time out one, assert the other two are untouched. The default-timeout and unrecognised-pathType tests are updated to the new signature.
- `widget-streams.directive.spec.ts`: both fake-DataService stubs updated to the 3-arg signature; the timeout-call assertion now checks the source (`default`) is passed; the #1069 TTL stub nulls the timed-out source's subject.

## Verification note

Not run locally (this on-boat host has no `node_modules`; HaLOS policy forbids installing/building here). Relying on CI (Node 24: lint + betterer:ci + tests). Neither changed source file is in `.betterer.results`, so the ratchet is unaffected.

### DIFF
--- a/src/app/core/directives/widget-streams.directive.spec.ts
+++ b/src/app/core/directives/widget-streams.directive.spec.ts
@@ -14,6 +14,7 @@ class FakeDataService {
     subjects = new Map<string, Subject<IPathUpdate>>();
     timeoutCalls: {
         path: string;
+        source: string;
         pathType: string;
     }[] = [];
 
@@ -27,8 +28,8 @@ class FakeDataService {
         return this.subjects.get(key)!.asObservable();
     }
 
-    timeoutPathObservable(path: string, pathType: string): void {
-        this.timeoutCalls.push({ path, pathType });
+    timeoutPathObservable(path: string, source: string, pathType: string): void {
+        this.timeoutCalls.push({ path, source, pathType });
     }
 }
 
@@ -365,7 +366,22 @@ describe('WidgetStreamsDirective', () => {
         // Do not emit anything; advance virtual time beyond 20ms to trigger timeout
         await vi.advanceTimersByTimeAsync(30);
         expect(dataSvc.timeoutCalls.length).toBe(1);
-        expect(dataSvc.timeoutCalls[0]).toEqual({ path: 'env.to', pathType: 'string' });
+        expect(dataSvc.timeoutCalls[0]).toEqual({ path: 'env.to', source: 'default', pathType: 'string' });
+    });
+
+    it('forwards a configured non-default source into timeoutPathObservable', async () => {
+        vi.useFakeTimers();
+        vi.spyOn(console, 'log');
+        const cfg = makeCfg({ path: 'env.to', source: 'n2k-1', pathType: 'string', sampleTime: 100, displayName: 'Test', enableTimeout: true, dataTimeout: 0.02 });
+        directive.setStreamsConfig(cfg);
+
+        const hits: string[] = [];
+        directive.observe('p', u => hits.push(String(u?.data?.value)));
+
+        // The stream's effective source must reach the reset so a source-bound widget
+        // clears its own registration, not the default bucket (#206).
+        await vi.advanceTimersByTimeAsync(30);
+        expect(dataSvc.timeoutCalls[0]).toEqual({ path: 'env.to', source: 'n2k-1', pathType: 'string' });
     });
 
     it('applies convertUnitTo to numeric values (initial + sampled)', async () => {
@@ -486,7 +502,7 @@ describe('WidgetStreamsDirective', () => {
  */
 class TtlFakeDataService {
     subjects = new Map<string, BehaviorSubject<IPathUpdate>>();
-    timeoutCalls: { path: string; pathType: string }[] = [];
+    timeoutCalls: { path: string; source: string; pathType: string }[] = [];
 
     private keyFor(path: string, source?: string): string {
         return `${path}|${source?.trim() || 'default'}`;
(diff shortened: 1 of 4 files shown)


mairas:
## Code review — #206 source-scoped timeout reset

**Scope:** `data.service.ts` + `widget-streams.directive.ts` + their specs (1 commit vs `main`).
**Intent:** scope `timeoutPathObservable` to the timed-out source so a dead source doesn't null live sibling widgets on other sources of the same path. Preserve whole-path-silent clearing and the invalid-source guard.
**Reviewers (7):** correctness, testing, maintainability, project-standards (always-on) + api-contract (signature change), reliability (timeout/failure logic), adversarial (reverses a tested behavior).

The fix is **confirmed sound for its stated scope**: correctness, reliability, and adversarial all traced that each `(path, source)` stream self-times-out, so a genuinely silent path still clears every widget while a single dead source no longer clobbers live siblings — the exact #206 fix. api-contract confirmed the signature change is atomic (sole prod caller + all specs updated, no stub drift, `source` correctly middle-positioned). project-standards clean (no `any`, betterer baseline unaffected, conventional commit). The rewritten sibling-isolation test is a genuine regression guard (fails on the old null-all, passes on the fix).

### P2 — Moderate

| 1 | data.service.ts:732 | **`enableTimeout:false` widgets no longer cleared by a sibling's timeout.** They never self-time-out, so a widget on a dead source stays frozen on stale data where the old null-all incidentally cleared it. `enableTimeout:false` is a shipped default (boolean-switch, autopilot). Arguably *more* correct (respects the opt-out) but a silent reversal with a safety angle. | adversarial | 0.72 | **follow-up filed: #254** |

### P3 — Low

| 2 | widget-streams.directive.spec.ts | No end-to-end test threading a **non-default** source through the directive to `timeoutPathObservable` — a regression hardcoding `'default'` would pass every test. | testing, correctness, api-contract, adversarial (4×) | 0.85 | safe_auto → **fixed in this review** |
| 3 | data.service.ts:733 | **Shallow reset:** nulls the derived `pathDataUpdate$` but leaves upstream `_pathData$` stale; a later `_pathState$` emission via `combineLatest` can resurface the stale value. Pre-existing; narrower blast radius now. | correctness, reliability | 0.65 | pre-existing → noted in #254 |
| 4 | widget-streams.directive.ts:101 | `source?.trim() || 'default'` normalization now repeated at 3 sites; a drift between them would desync subscribe vs timeout source. Extract `normalizeSource`. | maintainability | 0.68 | advisory (pre-existing) |
| 5 | widget-streams.directive.ts:157 | Unbounded per-cycle `console.log` during a sustained outage. | reliability | 0.62 | advisory (pre-existing) |

### Applied fixes

- Added `forwards a configured non-default source into timeoutPathObservable` to the directive spec (finding #2) — locks the effective-source wiring the fix depends on. Folded into the commit.

### Coverage

- Suppressed: 0. All 7 reviewers returned.
- Not run locally (no `node_modules` on the on-boat host; HaLOS policy forbids installing/building). CI (Node 24: lint + betterer:ci + tests) is the gate; neither source file is in `.betterer.results`.

---

**Verdict: mergeable once CI is green.** The change correctly implements #206 and is sound for self-timing-out widgets. The one substantive new finding (P2, `enableTimeout:false` reversal) is a product-intent decision, not a defect in this fix — tracked in **#254**. Pre-existing P3s (shallow reset, log noise, normalization DRY) are noted, non-blocking.

Issue #243, opened 2026-07-10
Port Kip #1129 (+#1130): bound DataService memory growth (refcounted release + removePathsForContext)


mairas:
## Summary

Upstream Kip PR https://github.com/mxtommy/Kip/pull/1129 fixes unbounded memory growth in `DataService` (surfaced while investigating upstream #770 — the "app goes unresponsive after a while" freeze). skip has the same leak and cannot take the upstream patch, so it needs a manual port.

### What upstream #1129 does

1. **Refcounts shared path registrations.** `subscribePath(path, source)` dedupes callers onto one registration; the PR adds a `refCount` to `IPathRegistration` and changes `unsubscribePath` from `unsubscribePath(path)` to `unsubscribePath(path, source)`, tearing the registration down only when the last caller releases it. This makes it safe for widgets to release their base registration on destroy without killing another widget still sharing the same `(path, source)`.
2. **Adds `removePathsForContext(context)`** to `DataService` — prunes `_skData`, the full-tree/meta caches, and pending notification state for every path under a context, so an evicted AIS target's path data actually leaves memory. It is a no-op for the `self` context.
3. **Makes callers release their registrations:** `widget-streams.directive` tracks `baseIdentities` per path and calls `unsubscribePath` on reconfigure/remove/destroy; `remote-control.component` releases the bound display's two path registrations on display switch and on destroy; `dataset-stream.service` releases on `stop()`; `ais-processing.service` calls `removePathsForContext` when a track is evicted (and relaxes the TTL sweep cadence).

### Why skip cannot cherry-pick this

skip's `DataService` and its callers diverged from Kip in exactly this area:

- **skip has no `unsubscribePath` at all.** It was deliberately removed in commit `cc8a1c9c` ("refactor(data): drop unsafe unused unsubscribePath", Refs #31) because the old single-arg version tore down shared registrations without refcounting. So skip does not have the method to modify — it has to (re)introduce a *refcounted* one from scratch, and then wire up every caller.
- **`IPathRegistration` (`src/app/core/services/data.service.ts:95`) has no `refCount`.** `subscribePath` (`:235`) pushes to `_pathRegister` (`:289`) and `_pathRegisterByPath` with no removal path anywhere. `_skData` is never pruned per-context. The leak is present and unmitigated.
[...]

PR #256, opened 2026-07-13
feat(data): bound DataService memory growth (refcount foundation, #243 stage 1)

mairas:
**Stage 1 of 2 for #243** (bound DataService memory growth). Foundation only — the DataService teardown contract. No caller wires to it yet, so this PR changes **no runtime behavior**; it adds the API + refcount bookkeeping that stage 2 uses to release registrations. See the [plan on #243](https://github.com/halos-org/skip/issues/243#issuecomment-4958701522).

## What
- `IPathRegistration` gains `refCount` (init 1) and a stored `combinedSub`.
- `subscribePath` bumps `refCount` on a dedup hit; stores the internal `combineLatest` subscription.
- `unsubscribePath(path, source)` — decrements; only at `refCount 0` unsubscribes `combinedSub`, completes the four subjects, and removes the registration from both `_pathRegister` and `_pathRegisterByPath`. Matched by **path AND source**. Unknown path/source is a safe no-op.
- `removePathsForContext(context)` — prunes `_skData`, `_pendingPathStates`, and `_dataServiceMetaByPath` (+ derived caches / full-tree emit when active) for a context's paths. **Never touches registrations.** No-op for self.

## Why — and the reversal
This reintroduces the teardown contract removed in `cc8a1c9c` ("drop unsafe unused unsubscribePath", Refs #31). That commit removed the **unrefcounted, single-arg** version precisely because it could tear down a shared registration under a live co-subscriber. This adds back the **refcounted, path+source** version whose absence was the stated danger — the cc8a1c9c invariant ("co-subscribers cannot tear each other down") is preserved by refcounting, and its test still passes. The `combinedSub` + `Subscription` import cc8a1c9c dropped are restored so teardown at zero also stops the internal `combineLatest`.

## Tests
Ported the upstream `data.service.spec.ts` cases, adapted to skip's harness: refcounted teardown (release only completes at 0), source-scoped teardown (releasing one source leaves siblings live), unknown-path/source no-op, per-context removal (other contexts untouched), full-tree + meta cache pruning when active, prune-while-inactive-then-rebuild, and self no-op.

## Verification
Not run locally (no `node_modules` on the on-boat host; policy forbids installing/building). CI (Node 24: lint + betterer:ci + tests) is the gate; `data.service.ts` is not in `.betterer.results`. **The actual leak is not exercised by unit tests** — the real proof is a soak (watch `_pathRegister`/`_skData` stabilise during AIS churn), which lands with stage 2 and needs a dev-machine/live-boat soak, not CI.

Stage 2 (separate PR) wires the release sites: widget-streams, remote-control, history-chart-stream, and ais-processing (`removePathsForContext` on eviction).

Refs #243

### DIFF
--- a/src/app/core/services/data.service.spec.ts
+++ b/src/app/core/services/data.service.spec.ts
@@ -1,7 +1,7 @@
 import { TestBed } from '@angular/core/testing';
 import { Subject } from 'rxjs';
 import { beforeEach, describe, expect, it } from 'vitest';
-import { IMeta, IPathValueData } from '../interfaces/app-interfaces';
+import { IMeta, IPathValueData, IPathMetaData } from '../interfaces/app-interfaces';
 import { ISignalKDataValueUpdate, States } from '../interfaces/signalk-interfaces';
 import { DataService, IPathUpdate, IPathUpdateWithPath } from './data.service';
 import { SignalKDeltaService } from './signalk-delta.service';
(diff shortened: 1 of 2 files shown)


mairas:
## Code review — #243 stage 1 (DataService refcount foundation)

**Reviewers (7):** correctness, testing, maintainability, project-standards, api-contract, reliability, adversarial.

**No P0. The foundation is correct** — correctness and project-standards found no defects: refcount balance is safe (over-release past 0 no-ops because the entry is spliced), no `next`-after-`complete` (every emit path goes through `_pathRegisterByPath`, which is spliced on teardown), `removePathsForContext` branch guards are right, the trailing-dot prefix avoids sibling-context clobbering, and the self guard is correct. All 7 ported specs are genuine regression guards.

### Applied in this review
- **Deleted the write-only `_pathRegister` array** (flagged P2 by maintainability + correctness + reliability). It was never read — only pushed and, in this PR, spliced. `unsubscribePath` no longer does O(n) bookkeeping for dead state; teardown now touches only `_pathRegisterByPath`.
- **Corrected the `unsubscribePath` docstring** — the old "a caller that never subscribed is a safe no-op" was false when co-subscribers keep the registration alive (the spurious release still decrements). Now states the balanced-release contract explicitly.
- **Renamed `matchingPaths` → `matching`** (single `find()` result).
- **Added 3 tests:** re-subscribe-after-teardown (proves the registration is actually removed — since `next`-after-`complete` is silent, a complete-but-don't-splice bug would pass every other test), `removePathsForContext` no-match/no-spurious-emit, and a string-prefix sibling-context guard.

### P1 — carried to stage 2 (not a defect in this foundation)

| # | Issue | Reviewers | Conf |
|---|-------|-----------|------|
| 1 | **The refcount contract is discipline-dependent.** `refCount` is a bare counter with no per-acquire handle, so an over-release (or a release of a `(path, source)` a caller didn't acquire) tears down a shared registration under a live co-subscriber — the cc8a1c9c class of bug, gated but not eliminated. Inert now (no callers); it's the contract stage-2 wiring must honor. | adversarial, reliability, api-contract | 0.85 |

### P2 — stage-2 design risks

| # | Issue | Reviewers | Conf |
|---|-------|-----------|------|
| 2 | `getPathMetaObservable(path)` returns registration **[0]**`.pathMeta$`. When stage 2 releases the index-[0] ('default') registration while a sibling source survives, meta consumers (the metadata directive → zones) get `complete` and freeze on a still-live path. Emergent from teardown + the pre-existing index-[0] meta design. | adversarial | 0.66 |
| 3 | Stage-2 landmine in `WidgetStreamsDirective`: `subscribePath` is acquired per-`pathName` on base-key change, with no natural 1:1 release point — a release tied to pipeline re-subscription over-releases (→ #1); a release only on destroy under-releases bases dropped on base-key change (→ leak). Release must mirror per-`pathName`, releasing the OLD base before overwriting. | adversarial, api-contract | 0.7 |
| 4 | Stage-2 caller stubs will drift: `widget-streams` `FakeDataService`/`TtlFakeDataService` and the AIS `Partial<DataService>` mocks don't implement the new methods — whichever stage-2 caller lands must add them. | api-contract | 0.65 |

### Coverage / residual
- Not run locally (no `node_modules`; policy). CI (Node 24) is the gate; `data.service.ts` absent from `.betterer.results`.
- Noted dead-ish branch: the `_pendingPathStates` prune in `removePathsForContext` may be unreachable today (notification states are rewritten to `self.`, and self is excluded) — kept as it matches upstream and is defensive.

---

**Verdict: the foundation is mergeable once CI is green.** The P1/P2 items are all stage-2 caller-discipline / design concerns, not defects in this API (which has no callers). Two decisions to settle before stage 2 wires the release sites: **(a)** keep the upstream discipline-based contract vs. harden it with a disposable release handle; **(b)** decouple meta delivery from registration[0] so releasing a source doesn't freeze a live path's zones. These will be captured on #243.

PR #258, opened 2026-07-13
fix(data): guard delta emission against missing source/timestamp/path

### LINKED ISSUE #154: signalk-delta: validate $source/timestamp and metadata.path on emission (follow-up to #24)
Follow-up to #24 (PR #146), which scoped the WS-boundary guard to frame discriminators, `item.path`, and the `parseSkMeta` value/properties chain. Two adjacent gaps remain (no throw, so outside #24's recoverability requirement — these are correctness-of-emission gaps):

1. `parseUpdates` emits `update.$source` / `update.timestamp` without validating they exist: a discriminator-valid frame whose update lacks them emits `IPathValueData` with `undefined` source/timestamp into the data pipeline.
2. `parseSkMeta` emits `undefined.key`-style paths when `metadata.path` is missing.

Both should either drop-and-log (consistent with the #24 guard) or substitute explicit defaults — decide once for both.

mairas:
## Why
Follow-up to #24. Two emission-boundary gaps remained in `SignalKDeltaService`:
- `parseUpdates` emitted `IPathValueData` carrying `update.$source` / `update.timestamp` without checking they exist — a discriminator-valid frame lacking either pushed `undefined` source/timestamp into the data pipeline.
- `parseSkMeta` emitted an `undefined.key`-style path when `metadata.path` was missing.

## How
Two guards in the existing #24 drop-and-log style (`console.warn` + skip, no throw). Validation is **per-value / per-item**, not whole-frame: a sibling value or meta item with valid fields still emits. Notifications routing is untouched.

## Verification
- Malformed-frames spec suite extended: missing-`$source`, missing-timestamp, missing-meta-path drop cases + sibling-valid-still-emits (across updates and meta items); new specs fail against the old code.
- Local static gate: `lint` ✓, `betterer:ci` ✓ (183, unchanged), `tsc --noEmit` ✓. Unit specs run in CI.

Closes #154

### DIFF
--- a/src/app/core/services/signalk-delta.service.spec.ts
+++ b/src/app/core/services/signalk-delta.service.spec.ts
@@ -489,6 +489,84 @@ describe('SignalKDeltaService', () => {
           { context: CTX, path: 'navigation.speedOverGround', meta: value },
         ]);
       });
+
+      it('drops a value whose update is missing $source without throwing', () => {
+        const { service } = setup();
+        const out = collectData(service);
+        const warn = vi.spyOn(console, 'warn').mockImplementation(() => undefined);
+        const noSource = { timestamp: TS, values: [{ path: 'navigation.speedOverGround', value: 1 }] } as unknown as ISignalKUpdateMessage;
+        expect(() => parse(service, { context: CTX, updates: [noSource] })).not.toThrow();
+        expect(out).toEqual([]);
+        expect(warn).toHaveBeenCalled();
+        warn.mockRestore();
+      });
+
+      it('drops a value whose update is missing timestamp without throwing', () => {
+        const { service } = setup();
+        const out = collectData(service);
+        const warn = vi.spyOn(console, 'warn').mockImplementation(() => undefined);
+        const noTs = { $source: 'src.1', values: [{ path: 'navigation.speedOverGround', value: 1 }] } as unknown as ISignalKUpdateMessage;
+        expect(() => parse(service, { context: CTX, updates: [noTs] })).not.toThrow();
+        expect(out).toEqual([]);
+        expect(warn).toHaveBeenCalled();
+        warn.mockRestore();
+      });
+
+      it('still emits a sibling update with a valid $source and timestamp when another update is invalid', () => {
+        const { service } = setup();
+        const out = collectData(service);
+        const noSource = { timestamp: TS, values: [{ path: 'navigation.courseOverGround', value: 1 }] } as unknown as ISignalKUpdateMessage;
+        parse(service, { context: CTX, updates: [noSource, update([{ path: 'navigation.speedOverGround', value: 2 }])] });
+        expect(out).toEqual([
+          { context: CTX, path: 'navigation.speedOverGround', source: 'src.1', timestamp: TS, value: 2 },
+        ]);
+      });
+
+      it('warns once per update, not once per value, when a multi-value update lacks $source', () => {
+        const { service } = setup();
+        const out = collectData(service);
+        const notes: ISignalKDataValueUpdate[] = [];
+        service.subscribeNotificationsUpdates().subscribe(v => notes.push(v));
+        const warn = vi.spyOn(console, 'warn').mockImplementation(() => undefined);
+        const note = { path: 'notifications.mob', value: { state: 'alarm', message: 'MOB' } };
+        const noSource = { timestamp: TS, values: [
+          { path: 'navigation.speedOverGround', value: 1 },
+          { path: 'navigation.courseOverGround', value: 2 },
+          note,
+        ] } as unknown as ISignalKUpdateMessage;
+        parse(service, { context: CTX, updates: [noSource] });
+        expect(out).toEqual([]);
+        expect(warn).toHaveBeenCalledTimes(1);
+        expect(notes).toEqual([note]);
+        warn.mockRestore();
+      });
+
+      it('drops a meta item whose path is missing without throwing', () => {
+        const { service } = setup();
+        const metas: IMeta[] = [];
+        service.subscribeMetadataUpdates().subscribe(v => metas.push(v));
+        const warn = vi.spyOn(console, 'warn').mockImplementation(() => undefined);
+        const meta = [{ value: { units: 'm/s', description: 'Speed over ground' } } as unknown as ISignalKMeta];
+        expect(() => parse(service, { context: CTX, updates: [update([], { meta })] })).not.toThrow();
+        expect(metas).toEqual([]);
+        expect(warn).toHaveBeenCalled();
+        warn.mockRestore();
+      });
+
+      it('still emits a sibling meta item with a valid path when another lacks one', () => {
+        const { service } = setup();
+        const metas: IMeta[] = [];
+        service.subscribeMetadataUpdates().subscribe(v => metas.push(v));
+        const sogMeta = { units: 'm/s', description: 'Speed over ground' };
+        const meta = [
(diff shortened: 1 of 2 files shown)


mairas:
## Review synthesis

No changes were needed beyond the reviewed diff.

mairas:
## Review synthesis (standard multi-persona pass)

> **Supersedes** the earlier preliminary [_"Review synthesis"_ comment](https://github.com/halos-org/skip/pull/258#issuecomment-4960573487) on this PR — that was a single-persona automated pass, not the standard review. Use this one.

**Scope:** `fix/delta-emission-guards` — guard `SignalKDeltaService` emission (follow-up to #24, closes #154). `parseUpdates` now drops-and-logs a value whose `update.$source` / `update.timestamp` is missing; `parseSkMeta` drops-and-logs when `metadata.path` is missing. Validation is per-value/per-item, not whole-frame — a sibling valid value/meta still emits. No throw. Malformed-frames spec suite extended.

**Reviewers:** correctness, testing, maintainability, project-standards, reliability (5 independent persona passes, fresh context each).

**Diff verified:** the two new guards and the pre-existing commented block below are confirmed against `origin/fix/delta-emission-guards`.

---

### Findings — P3 (advisory, non-blocking)

| # | file:line | issue | reviewer(s) | conf. | route |
|---|-----------|-------|-------------|-------|-------|
| 1 | `src/app/core/services/signalk-delta.service.ts:382` | Update-level `$source`/`timestamp` guard sits **inside** `update.values.forEach`, in the else (non-notification) branch. Both fields are update-scoped (invariant across the loop), so a single malformed multi-value update re-evaluates the identical check and fires `console.warn` **once per value** — N log lines for one per-update defect. This service consumes the server's full delta firehose on a Pi-class HaLOS display; a producer chronically omitting `$source`/`timestamp` floods the console (amplified ×values-per-update) and adds main-thread work. Behaviour is otherwise correct: all affected values drop, siblings in other updates still emit. | correctness, maintainability, reliability | 0.70 | fix-here (optional) |

**Fix (agreed across the three personas):** hoist the missing-field check to compute a `const` once per `update` (above `update.values.forEach`), warn a single time when it fails, then inside the else branch just `if (missing) return;` without re-warning. **Do not** hoist above the notification branch — the current placement deliberately lets `notifications.*` values through even when `$source`/`timestamp` are absent (alarm/safety-relevant), and that exemption must be preserved. Small, behaviour-preserving; the specs' drop-and-sibling-emit assertions still hold.

> Three personas flagged this independently → treated as higher-confidence. Only the placement/log-cardinality is at issue; the drop semantics are correct.

### Findings — pre-existing

| # | file:line | issue | reviewer(s) | conf. | route |
|---|-----------|-------|-------------|-------|-------|
| 2 | `src/app/core/services/signalk-delta.service.ts:363` | Three-line commented-out `if (context != this._selfUrn) { return; }` block at the head of `parseUpdates`. Commented code is a maintenance liability — next reader can't tell if it's a disabled feature, a reminder, or abandoned. Pre-existing, but sits directly above the `parseUpdates` change under review. | maintainability | 0.85 | fix-here (trivial) — delete lines 363-365; if the self-context filter is real future intent, open an issue instead of leaving dead code in the parser |

---

### Residual risks

- **Missing-timestamp is now fatal to the value.** A value with a valid `$source` but no `timestamp` is discarded entirely rather than emitted with `timestamp: undefined`. The SK node server always stamps forwarded deltas, so real-boat data is unaffected; a non-conforming producer relayed through the server would lose values, not just timestamps. *(correctness, reliability)*
- **Object-`source`-only form is dropped.** The guard requires the `$source` string; the code never derives it from the spec's `source` object. Standard SK server output always includes `$source` on the WS firehose and the whole app buckets by `$source`, so real-world risk is low — but a direct connection to a producer emitting only `source` would silently lose all its values. *(correctness, reliability)*
- **Type, not emptiness.** Empty-string `$source` / `timestamp` / `path` (`''`) passes `typeof === 'string'` and still emits. Matches the "drop when *missing*" intent, but `''` is unscreened — an empty/dangling path could key a delta. *(correctness, maintainability)*
- **Notification exemption is deliberate but undocumented.** `notifications.*` items still emit without validating `$source`/`timestamp` (guard lives in the else branch). Safe today because the emitted notification item is `{path, value}` and never references those fields — but the guard's scope is narrower than the PR summary phrasing implies, and any future consumer reading source/timestamp off notification deltas would not get the guard. *(correctness, testing, maintainability, reliability)*
- **No try/catch around WS ingestion.** A synchronous throw inside `parseUpdates`/`parseSkMeta` propagates as an observable error, tears down the WS subscription, and stops the live stream until reconnect. These guards cover the known missing-field cases but don't eliminate the teardown surface against unforeseen malformations. Pre-existing architecture; a blanket try/catch around `handleStreamFrame`'s body would make any single malformed frame non-fatal. *(reliability)*
- **Betterer ratchet.** `signalk-delta.service.ts` has zero entries in `.betterer.results`, so this diff needs no regeneration. If the branch is later rebased/merged with main and touches any strictNullChecks-tracked file, `npm run betterer` must be re-run and the baseline committed, or `betterer:ci` fails with "unexpected changes". *(project-standards)*

### Testing gaps

- Drop specs assert only `warn.toHaveBeenCalled()` — they don't pin the message text (unlike the pre-existing boundary test at spec line 439 which asserts the exact string). A regression firing the *wrong* early-return guard would still pass. Optional hardening: `toHaveBeenCalledWith('[Delta Service] Dropping value update without a valid $source or timestamp:', expect.anything())` and the analogous meta message. *(testing, maintainability, project-standards)*
- No test for a **multi-value** update missing `$source`/`timestamp` — the per-value re-warn amplification (finding #1) is unverified, so neither the current behaviour nor a future log-once fix would be caught. Also would pin whether all siblings in the same malformed update drop. *(correctness, testing, maintainability, reliability)*
- No test for a `notifications.*` item alongside a data item in an update missing `$source`/`timestamp` — confirming the notification still forwards while the data value drops. Without it, a refactor hoisting the guard could silently break the deliberate exemption. *(correctness, testing, maintainability, reliability)*
- No test covers an empty-string `path`/`$source` on either guard — the type-only nature of the checks is unpinned. *(correctness, maintainability)*
- No test for an update carrying only the object `source` form (no `$source`) being dropped — would document the deliberate design decision and guard against a future change deriving `$source` from `source`. *(correctness, reliability)*
- No end-to-end test through the WS subscription (`connectWS` → `handleStreamFrame`); specs call `parse()` directly, so "a malformed frame never tears down the live stream" is unexercised. *(reliability)*
- **CI is the source of truth:** the six new malformed-frame specs were not executed here (live boat box, `ng`/`npm test` disallowed). CI (Node 24, `npm ci`) must confirm they pass before merge. *(project-standards)*

---

**Verdict: ready-to-merge.** No P0/P1/P2 after gating. The two surviving items are both P3/advisory: finding #1 (log-once hoist, agreed fix, behaviour-preserving) and finding #2 (delete the pre-existing commented block right above the change). Both are trivial and touch the exact code under review, so folding them into this PR is worthwhile — but neither blocks merge, and CI green on the new specs is the gate.

mairas:
**Review findings addressed** (both P3):
- Hoisted the `$source`/timestamp check to once-per-update so a malformed multi-value update warns once, not per value — `notifications.*` still route when those fields are absent.
- Deleted the dead commented-out self-context filter.
- New spec: a multi-value source-less update drops all its values, warns exactly once, and a notification in the same update still routes.

Local gate green (lint/betterer/tsc); CI re-running.

PR #259, opened 2026-07-13
fix(dashboard): integer-safe index validity guard across all sites

### LINKED ISSUE #251: Consolidate the duplicated dashboard-index range check in DashboardService
## Problem

The dashboard-index validity check — "an integer within `[0, dashboards().length)`" — is hand-written in four `DashboardService` methods (`src/app/core/services/dashboard.service.ts`), in three slightly different spellings:

- `setActiveDashboardIndex` — `Number.isInteger(itemIndex) && itemIndex >= 0 && itemIndex < this.dashboards().length`
- `delete` — `if (itemIndex < 0 || itemIndex >= this.dashboards().length)`
- `duplicate` — `if (itemIndex < 0 || itemIndex >= this.dashboards().length)`
- `navigateTo` — `if (itemIndex >= 0 && itemIndex < this.dashboards().length)`

A change to the bounds rule (e.g. an added dashboard-count constraint, or an off-by-one fix) has to be made in four places, and the spellings already differ — only `setActiveDashboardIndex` includes the `Number.isInteger` guard, so `delete`/`duplicate`/`navigateTo` would accept a fractional index.

## Suggested direction

Extract a single private predicate, e.g. `isValidDashboardIndex(index: number): boolean` (integer + in-range), and call it from all four sites. Decide deliberately whether the callers that currently omit the integer check should adopt it (they probably should).

## Context

Surfaced by the review of #248 (fall back to page 0 on an invalid `/page/:id`). That fix deliberately did **not** widen this smell — it reused `setActiveDashboardIndex`'s boolean return instead of adding a fifth copy of the predicate. This issue is the follow-up to consolidate the existing four. Low priority; pure maintainability, no behavior change intended beyond optionally tightening the non-integer cases.

mairas:
## Why
The dashboard-index validity check ("integer within `[0, dashboards().length)`") was hand-written in four `DashboardService` methods in three spellings — and only `setActiveDashboardIndex` had the `Number.isInteger` guard, so `delete` / `duplicate` / `navigateTo` accepted fractional indices. `duplicate(1.5)` reached `cloneDeep(undefined)` and threw.

## How
Extract one private predicate `isValidDashboardIndex(index)` (integer AND in range) and route all four sites through it. `delete` / `duplicate` / `navigateTo` now reject fractional/negative/out-of-range indices. `setActiveDashboardIndex` keeps its boolean-return contract (relied on by #248's page-0 fallback).

## Verification
- New specs assert fractional-index rejection at each of the four sites (the `duplicate(1.5)` case pins the latent crash); they fail against the old code.

Closes #251

### DIFF
--- a/src/app/core/services/dashboard.service.spec.ts
+++ b/src/app/core/services/dashboard.service.spec.ts
@@ -261,6 +261,15 @@ describe('DashboardService', () => {
       expect(consoleError).toHaveBeenCalledTimes(2);
     });
 
+    it('rejects a fractional index without removing anything', () => {
+      setup();
+      service.setActiveDashboardIndex(0);
+      service.delete(1.5);
+      expect(service.dashboards().map(d => d.id)).toEqual(['d-0', 'd-1', 'd-2']);
+      expect(service.activeDashboard()).toBe(0);
+      expect(consoleError).toHaveBeenCalledTimes(1);
+    });
+
     it('clamps the active index when it falls past the end', () => {
       setup();
       service.setActiveDashboardIndex(2);
@@ -288,6 +297,22 @@ describe('DashboardService', () => {
       expect(service.duplicate(-1, 'Copy', 'icon')).toBe(-1);
       expect(service.dashboards()).toHaveLength(3);
       expect(consoleError).toHaveBeenCalledTimes(2);
+    });
+
+    it('returns -1 for a fractional index instead of crashing on a missing dashboard', () => {
+      setup();
+      expect(service.duplicate(1.5, 'Copy', 'icon')).toBe(-1);
+      expect(service.dashboards()).toHaveLength(3);
+      expect(consoleError).toHaveBeenCalledTimes(1);
+    });
+
+    it('returns -1 for NaN without throwing on a missing dashboard', () => {
+      setup();
+      let result: number | undefined;
+      expect(() => { result = service.duplicate(NaN, 'Copy', 'icon'); }).not.toThrow();
+      expect(result).toBe(-1);
+      expect(service.dashboards()).toHaveLength(3);
+      expect(consoleError).toHaveBeenCalledTimes(1);
     });
 
     it('deep clones with fresh dashboard and widget UUIDs kept in sync', () => {
@@ -381,6 +406,13 @@ describe('DashboardService', () => {
       service.navigateTo(3);
       expect(router.navigate).toHaveBeenCalledTimes(1);
       expect(consoleError).toHaveBeenCalled();
+    });
+
+    it('navigateTo rejects a fractional index without navigating', () => {
+      service.navigateTo(1.5);
+      service.navigateTo(-1);
+      expect(router.navigate).not.toHaveBeenCalled();
+      expect(consoleError).toHaveBeenCalledTimes(2);
     });
 
     it('navigateToNextDashboard routes forward and navigateToPreviousDashboard backward, with wrap', () => {
--- a/src/app/core/services/dashboard.service.ts
+++ b/src/app/core/services/dashboard.service.ts
@@ -131,6 +131,11 @@ export class DashboardService {
     }
   }
 
+  /** True when index is an integer within the current dashboards bounds [0, length). */
+  private isValidDashboardIndex(index: number): boolean {
+    return Number.isInteger(index) && index >= 0 && index < this.dashboards().length;
+  }
+
   /**
    * Sets the active dashboard index in the service.
    * This only updates the internal state and does NOT trigger navigation or URL changes.
@@ -141,11 +146,11 @@ export class DashboardService {
     // No change if the same dashboard is selected to prevent unnecessary cascading updates
     if (itemIndex === this.activeDashboard()) return true;
 
-    if (Number.isInteger(itemIndex) && itemIndex >= 0 && itemIndex < this.dashboards().length) {
+    if (this.isValidDashboardIndex(itemIndex)) {
       this.activeDashboard.set(itemIndex);
       return true;
     }
-    console.error(`[Dashboard Service] Invalid dashboard ID: ${itemIndex}`);
+    console.error(`[Dashboard Service] Invalid dashboard index: ${itemIndex}`);
     return false;
   }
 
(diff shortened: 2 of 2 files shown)


mairas:
## Review synthesis

No changes were needed beyond the reviewed diff. The `duplicate(1.5)` spec pins the previously-latent `cloneDeep(undefined)` crash.

mairas:
## Review synthesis (standard multi-persona review)

> **This supersedes the earlier preliminary "Review synthesis" comment on this PR.** That one was a single-persona automated pass; this is the standard `review` skill output — four independent persona reviewers (correctness, testing, maintainability, project-standards), each in a fresh context, synthesized and deduplicated.

**Scope:** `src/app/core/services/dashboard.service.ts` (+ the `remote-dashboards.service.ts` caller path) and the accompanying specs.

**Intent:** Consolidate the dashboard-index range check into one private integer-safe predicate `isValidDashboardIndex`, used by `setActiveDashboardIndex`, `delete`, `duplicate`, and `navigateTo` (closes #251). `delete`/`duplicate`/`navigateTo` now reject fractional indices they previously accepted — pinning the latent `duplicate()` `cloneDeep(undefined)` crash. `setActiveDashboardIndex` keeps its boolean-return contract (#248).

**Reviewers:** correctness · testing · maintainability · project-standards.

---

### Findings

Nothing blocking. One advisory, cosmetic item survived gating (confidence gate ≥ 0.60; dedup by file/line/title; sorted severity then confidence).

#### P3 — advisory

| # | file:line | issue | reviewer(s) | confidence | route |
|---|---|---|---|---|---|
| 1 | `dashboard.service.ts:148/204/247/351` | Consolidated check, un-consolidated wording: three sites log `Invalid dashboard ID: ${itemIndex}` while `duplicate` logs `Invalid itemIndex: ${itemIndex}` (the odd one out). "dashboard ID" also mislabels an **index** — `Dashboard.id` is a separate UUID string (interface at line 12), so a future debugger searching for a dashboard whose `id` equals `1.5` finds nothing. | maintainability | 0.60 | fix-here (trivial, related) — optional |

Suggested fix: use one string across all four sites and name it an index, e.g. `` `[Dashboard Service] Invalid dashboard index: ${itemIndex}` ``. Pre-existing on `main`, but this PR is the natural place to unify it since it already touches all four call sites. Not a merge blocker.

*(Line numbers above are as they stand on `main`; they shift ~+5 once the predicate is added, matching the reviewer's cited 153/209/252/356.)*

---

### Residual risks (not blocking)

- **Duplicated inline guard in the remote path.** `remote-dashboards.service.ts:137` keeps its own range guard (`!isNaN && idx>=0 && idx<length`) that omits the integer check now centralized in the predicate. Not a bug — `navigateTo` re-validates — but a fractional remote value now silently no-ops (logs an error) instead of navigating. Candidate for the same consolidation; out of this diff's scope. *(correctness)*
- **Sibling index-taking methods stay unguarded.** `update()` (~line 197) and `updateConfiguration()` (~line 292) still accept unvalidated indices and silently no-op via array map on out-of-range/fractional input — no crash, just a silent no-op. Pre-existing and explicitly outside this PR's scope, but the new predicate makes the inconsistency more visible; a future reader may assume all index-taking methods are guarded. *(testing, maintainability — independent agreement)*
- **`setActiveDashboardIndex` early-return runs before validation** (`if (itemIndex === this.activeDashboard()) return true`), but `activeDashboard()` is always a valid in-range integer or `null` (delete clamps it), so nothing out-of-range/non-integer can slip through. Pre-existing, unaffected by the diff. *(correctness)*
- **Guard reads `this.dashboards().length` at call time** — correct for all current synchronous callers, but any future check-then-act across an async boundary could race between guard and mutation. *(maintainability)*
- **Betterer ratchet is content-hash keyed.** This diff needs no `.betterer.results` regeneration (no new tracked strict error). But if the branch is behind `main` and a later merge rehashes tracked files, `npm run betterer` must be run and committed or `betterer:ci` fails. Verify the branch is current with `main` before merge. *(project-standards)*

### Testing gaps (not blocking)

- **NaN / Infinity untested for `delete`/`duplicate`/`navigateTo`.** New tests cover only the fractional `1.5` case. `NaN`/`Infinity` take the same newly-rejecting path; `duplicate(NaN)` is the one that on `main` fell through the old bounds guard and crashed at `cloneDeep(undefined)`, so a `NaN` case for `duplicate` would most directly document the crash-prevention intent. Low realistic risk (route params are NaN-guarded upstream in `applyDashboardParam`; UI callers pass concrete list indices), but it is an untested branch of the exact crash class the PR set out to pin. *(correctness, testing — independent agreement)*
- **No happy-path regression assertion after the tightening.** No test asserts a valid integer index still reaches `duplicate`'s deep-clone/UUID-regeneration body, nor that a valid integer remote index still reaches `navigateTo` (the `remote-dashboards.service.ts` path) — i.e. that the stricter guard didn't block the normal flow. The existing "deep clones…" test uses index `0`, which the predicate accepts, so it passes regardless. Low risk given the unchanged happy-path body. *(correctness, maintainability)*
- **Predicate not unit-tested in isolation** — covered only transitively through the four public callers. Acceptable: it is private and every branch is exercised via the public methods. *(testing)*
- **Local `npm test` / `lint` / `betterer:ci` not run** (live-services box; jsdom local runs unreliable). Confirmation that the new specs pass and that `betterer:ci` reports no new strict issue relies on CI. *(project-standards)*

---

**Verdict: ready-to-merge.** No P0/P1/P2 after synthesis. The single surviving item is a pre-existing P3 advisory (log-message wording) that this PR *may* unify since it already touches all four sites, but need not. The intent is met: the consolidated integer-safe guard closes the fractional-index gap and pins the latent `duplicate()` crash, and `setActiveDashboardIndex` keeps its boolean contract. Recommended (non-blocking) follow-ups: unify the four log strings, and add a `NaN`-for-`duplicate` test to document the crash-prevention intent.

mairas:
**Review finding addressed** (P3): the four index-validity logs are unified to `Invalid dashboard index: ${itemIndex}` — the old "Invalid dashboard ID" mislabelled an index (`Dashboard.id` is a separate UUID) and `duplicate()` diverged with "Invalid itemIndex". Added a `duplicate(NaN)` no-throw/returns-−1 test. Local gate green; CI re-running.

PR #260, opened 2026-07-13
fix(app): apply auto night mode immediately when enabled late

### LINKED ISSUE #152: app-service: redNightMode branches identical; late-enabled auto night mode inert
Surfaced by the #27 characterization pass (PR #147, `app-service.spec.ts`).

## Defect 1: redNightMode effect's branches are identical

`app-service.ts:115-125` — the effect's `if`/`else` branches both call `toggleDayNightMode()`; the conditional is dead and one branch was presumably meant to differ (likely applying vs. not applying the red-night theme).

## Defect 2: enabling auto night mode late is inert

`app-service.ts:100-113` — the environment-mode effect reads `_useAutoNightMode()` only inside `untracked()`, so enabling the setting while the environment is already `night` does nothing until the mode value next *changes*. Signal coalescing also makes a day→night→day flicker within one tick invisible to it.

Pinned by the "does not apply the current night mode when auto night mode is enabled later" test; update intentionally when fixing.

mairas:
## Why
Surfaced by the #27 characterization pass. Two defects in `app-service.ts`:
1. The `redNightMode` effect's `if`/`else` branches were **identical** (both call `toggleDayNightMode()`) — a dead conditional.
2. The environment-mode effect read `_useAutoNightMode()` only inside `untracked()`, so enabling auto night mode while the environment was already `night` did nothing until the mode value next changed.

## How
- Collapse the dead `redNightMode` conditional to its single actual behavior (no invented red-night branch).
- The environment-mode effect now **tracks** the auto-night-mode setting, so toggling it on applies the current environment immediately.

## Verification
- The pinning spec ("does not apply the current night mode when auto night mode is enabled later") is flipped to assert it now **does** apply.

Closes #152

### DIFF
--- a/src/app/core/services/app-service.spec.ts
+++ b/src/app/core/services/app-service.spec.ts
@@ -201,9 +201,9 @@ describe('AppService', () => {
       expect(toggle).toHaveBeenCalledTimes(1);
     });
 
-    // The mode effect only reacts to mode changes; enabling the setting alone
-    // leaves an already-active night mode unapplied until the next transition.
-    it('does not apply the current night mode when auto night mode is enabled later', () => {
+    // Enabling the setting while the environment is already night applies the
+    // current mode immediately, without waiting for the next transition.
+    it('applies the current night mode when auto night mode is enabled later', () => {
       const service = createService();
       envMode$.next(modeUpdate('night'));
       TestBed.tick();
@@ -211,7 +211,9 @@ describe('AppService', () => {
 
       settings.autoNightMode.set(true);
       TestBed.tick();
-      expect(service.isNightMode()).toBe(false);
+      expect(service.isNightMode()).toBe(true);
+      expect(document.body.style.getPropertyValue('--kip-nightModeBrightness')).toBe('0.27');
+      expect(document.body.style.getPropertyValue('--kip-nightModeFilters')).toContain('sepia(0.5)');
 
       envMode$.next(modeUpdate('day'));
       TestBed.tick();
--- a/src/app/core/services/app-service.ts
+++ b/src/app/core/services/app-service.ts
@@ -77,6 +77,7 @@ export class AppService {
   private _environmentMode = toSignal(this._data.subscribePath(this.MODE_PATH, 'default'));
 
   private previousEnvironmentMode: string | null = null;
+  private previousUseAutoNightMode = false;
 
   public readonly appVersion = signal<string>(packageInfo.version);
   public readonly browserVersion = signal<string>('Unknown');
@@ -99,30 +100,30 @@ export class AppService {
 
     effect(() => {
       const environmentMode = this._environmentMode();
+      const useAutoNightMode = this._useAutoNightMode();
 
       untracked(() => {
         if (!environmentMode) return;
         const mode = environmentMode.data.value;
-        if (this.previousEnvironmentMode === mode) return; // No change in mode
+        const modeChanged = this.previousEnvironmentMode !== mode;
+        const autoNightModeJustEnabled = useAutoNightMode && !this.previousUseAutoNightMode;
 
         this.previousEnvironmentMode = mode;
+        this.previousUseAutoNightMode = useAutoNightMode;
 
-        if (this._useAutoNightMode()) {
-          this.isNightMode.set(mode === "night");
-          this.toggleDayNightMode();
-        }
+        if (!useAutoNightMode) return;
+        if (!modeChanged && !autoNightModeJustEnabled) return;
+
+        this.isNightMode.set(mode === "night");
+        this.toggleDayNightMode();
       });
     });
 
     effect(() => {
-      const redNightMode = this._redNightMode();
+      this._redNightMode();
 
       untracked(() => {
-        if (redNightMode) {
-          this.toggleDayNightMode();
-        } else {
-          this.toggleDayNightMode();
-        }
+        this.toggleDayNightMode();
       });
     });
 


mairas:
## Review synthesis

No changes were needed beyond the reviewed diff. The dead `redNightMode` branch was collapsed (not replaced with invented behavior).

mairas:
## Review synthesis — PR #260 (standard multi-persona)

> **This supersedes the earlier preliminary "review synthesis" comment on this PR.** That was a single-persona automated pass, not the standard review. Treat this comment as the authoritative synthesis.

**Scope:** `src/app/core/services/app-service.ts` (closes #152) plus its spec.
**Intent:** the environment-mode effect now *tracks* the auto-night-mode setting, so enabling it while the environment is already night applies the current mode immediately (previously inert inside `untracked()` until the next transition). The `redNightMode` effect's two identical if/else branches are collapsed (dead conditional removed, no invented behavior). Pinning specs flipped.
**Reviewers:** correctness, testing, maintainability, project-standards (four independent personas, fresh context each).

After synthesis (drop malformed → confidence gate → dedupe → severity sort): **2 findings survive, both P3.** No P0/P1/P2. One P3 (bare `this._redNightMode();` read, conf 0.55) was gated out below the 0.60 threshold — noted under *Gated out* below. No pre-existing findings.

### P3 — non-blocking

|---|---|---|---|---|---|
| 1 | `app-service.spec.ts:214` | Late-enable test asserts the internal `isNightMode()` boolean, not the applied night *styling*. A regression where `isNightMode.set(true)` runs but `toggleDayNightMode()` no longer applies night brightness/filters would still pass. CLAUDE.md calls out the night-mode case by name ("verify the user-visible effect, not a proxy"). Sibling transition test (line 175) already checks `--kip-nightModeBrightness '0.27'`; the late-enable path does not. | testing | 0.62 | fix-here (optional) |
| 2 | `app-service.ts:109` | `autoNightModeEnabled` names a rising-edge (`useAutoNightMode && !previousUseAutoNightMode`), not current state — sitting one line below the current-state `useAutoNightMode`. A future editor reading `if (!modeChanged && !autoNightModeEnabled) return;` could mistake it for the current-state flag and reintroduce the exact late-enable bug this PR fixes. Suggest renaming to `autoNightModeJustEnabled`. | maintainability | 0.62 | fix-here (optional) |

**Suggested fixes** (both trivial, both this-PR-if-convenient):
- #1: after line 214, add `expect(document.body.style.getPropertyValue('--kip-nightModeBrightness')).toBe('0.27');` so the test proves styling was applied on late-enable, not just that the flag flipped.
- #2: rename the const to `autoNightModeJustEnabled` (edge semantics legible without tracing back to `previousUseAutoNightMode`) and update its single use in the guard at line 115.

### Gated out (below confidence threshold — recorded, not actioned)

- `app-service.ts:123` (maintainability, P3, conf 0.55): the bare `this._redNightMode();` expression statement (return value discarded) exists only to register the signal dependency; to a reader unaware of signal tracking it looks like a deletable no-op, and deleting it would silently break red-night-mode reactivity. `app-service.spec.ts:225` guards that reactivity, so an accidental deletion would be caught by tests. A one-line why-comment would help but is below the reporting bar.

### Residual risks (no action required; recorded for merge awareness)

- **One-frame transient (correctness):** `display.component.applyAndSaveSettings` reads `app.isNightMode()` and may call `setBrightness(1)` synchronously *before* the now-tracked mode effect flushes; the effect then applies correct night brightness. Final state is correct — a one-frame full-brightness transient when enabling auto night mode while already night, not a lasting defect.
- **Undefined-first-emit safety (correctness / maintainability):** the fix depends on the environment-mode `toSignal` having a value when `autoNightMode` toggles. If the SK observable hasn't emitted (`environmentMode` undefined), the effect early-returns *before* writing `previousUseAutoNightMode`, so the edge flag stays unlatched. Harmless under the current `BehaviorSubject`-seeded mock (always has a value); the first real emit still applies via `modeChanged` (null → mode). A production `subscribePath` that emits no initial value could compute `autoNightModeEnabled` from a stale `false`. Not exercised by any test.
- **Signal `===` coalescing (correctness):** re-saving `autoNightMode` unchanged is an intended no-op (won't schedule the effect). The within-tick day→night→day flicker from #152 is out of this PR's scope — a known-unaddressed edge, not a regression.
- **Manual change-detection state (maintainability):** the effect now carries `previousEnvironmentMode` + `previousUseAutoNightMode` + an edge const inside an undocumented `untracked()` block; future edits must correctly order read-previous / compute-edge / write-previous / gate by hand. This diff's extension is minimal and correct, but the pattern is fragile.
- **Standards remit (project-standards):** `app-service.ts` is not in `.betterer.results` and the diff adds no `strictNullChecks` error, so no baseline regeneration is needed. Only the root CLAUDE.md governs (no AGENTS.md / dir-scoped standards). Correctness of the collapsed effect is outside the standards remit — covered by the correctness/testing personas above.

### Testing gaps (converged across personas)

- **Enable-while-day** (`autoNightModeEnabled` true, `mode !== 'night'`): the `isNightMode.set(false)` branch of the late-enable path is uncovered. Only the enable-while-night case is pinned. Low value but the sole untested combination of the new branch.
- **Disable path** (`useAutoNightMode` true→false, and disable-then-re-enable exercising `previousUseAutoNightMode` true→false→true): unverified. Behavior is unchanged from main (disabling does not revert active night styling), so risk is low.
- **DOM styling on late-enable:** no test asserts brightness/filters for the late-enable transition — only the boolean (see finding #1).
- **CI confirmation:** per CLAUDE.md, Settings-touching specs are unreliable in local jsdom and were not run on this host; confirm the CI unit suite is green before merge.

---

**Verdict: ready-to-merge.** No P0/P1/P2 after synthesis. The two surviving P3s are optional polish (a stronger spec assertion and a clearer const name) that are safe to land in this PR or skip; neither blocks merge. Land after the CI unit suite passes.

mairas:
**Review findings addressed** (both P3):
- The late-enable test now asserts the **applied effect** — the `--kip-nightModeBrightness` / `--kip-nightModeFilters` CSS vars `toggleDayNightMode()` sets — not just the `isNightMode` boolean proxy (verify-the-effect).
- Renamed the rising-edge local to `autoNightModeJustEnabled`.

Local gate green; CI re-running.

PR #261, opened 2026-07-13
fix(remote-dashboards): drop dead sub, republish screens only on rename

### LINKED ISSUE #153: remote-dashboards.service: dead subscription, redundant republish, shared screens array
Surfaced by the #27 characterization pass (PR #147, `remote-dashboards.service.spec.ts`). Three small items in one service — likely one cleanup PR.

1. **Dead subscription**: `remote-dashboards.service.ts:68` — `remoteScreenPosition = toSignal(subscribePath(ACTIVE_SCREEN_PATH, ...))` is created but never read anywhere. Dead code or an unfinished feature; decide and either use or delete.
2. **Redundant republish**: lines 80-103 — a display-name change while remote control is enabled re-sends the active dashboard index (`setScreenIndex`) in addition to republishing screens, because one effect handles both name changes and enable transitions through the same path.
3. **Shared array reference**: lines 200-203 — `setScreensOnRemote` shallow-copies the payload but shares the `screens` array by reference with the caller's object.

All pinned by PR #147 tests.

mairas:
## Why
Surfaced by the #27 characterization pass. Three issues in `RemoteDashboardsService`:
1. `remoteScreenPosition` (a `toSignal(subscribePath(...))`) was created but never read.
2. A display-name change while remote control was enabled **redundantly re-sent the active dashboard index** on top of republishing screens.
3. `setScreensOnRemote` shallow-copied the payload but shared the `screens` array by reference with the caller.

## How
- Delete the dead subscription and its now-unused `ACTIVE_SCREEN_PATH`.
- Split the toggle/rename effect: a name change republishes screens **only**. The rename effect skips its initial run so boot-time publishing stays owned by the toggle effect (no new redundant boot republish).
- Copy the `screens` array in `setScreensOnRemote`.

## Verification
- The three PR #147 characterization tests are flipped (subscribes to activeScreen only; rename republishes screens without the index; published payload shares no `screens` reference).

Closes #153

### DIFF
--- a/src/app/core/services/remote-dashboards.service.spec.ts
+++ b/src/app/core/services/remote-dashboards.service.spec.ts
@@ -14,7 +14,6 @@ const KIP_UUID = 'test-kip-uuid';
 const SET_DISPLAY_PATH = 'self.kip.remote.setDisplay';
 const SET_SCREEN_INDEX_PATH = 'self.kip.remote.setScreenIndex';
 const REQUEST_ACTIVE_SCREEN_PATH = 'self.kip.remote.requestActiveScreen';
-const OWN_SCREEN_INDEX_PATH = `self.displays.${KIP_UUID}.screenIndex`;
 const OWN_ACTIVE_SCREEN_PATH = `self.displays.${KIP_UUID}.activeScreen`;
 
 const DASHBOARDS: Dashboard[] = [
@@ -108,11 +107,10 @@ describe('RemoteDashboardsService', () => {
       expect(requests.putRequest).toHaveBeenNthCalledWith(3, REQUEST_ACTIVE_SCREEN_PATH, { displayId: KIP_UUID, screenIdx: null }, KIP_UUID);
     });
 
-    it('subscribes to its own display screenIndex and activeScreen paths', () => {
+    it('subscribes only to its own display activeScreen path', () => {
       createService();
 
       expect(data.subscribedPaths).toEqual([
-        { path: OWN_SCREEN_INDEX_PATH, source: 'default' },
         { path: OWN_ACTIVE_SCREEN_PATH, source: 'default' }
       ]);
     });
@@ -169,6 +167,22 @@ describe('RemoteDashboardsService', () => {
       expect(callsTo(SET_SCREEN_INDEX_PATH)).toHaveLength(0);
       expect(callsTo(SET_DISPLAY_PATH)).toHaveLength(1);
     });
+
+    it('publishes the screens on init when constructed with remote control already enabled', () => {
+      settings.isRemoteControl.set(true);
+
+      createService();
+
+      const published = callsTo(SET_DISPLAY_PATH)
+        .map(call => call[1] as { display: IScreensPayload | null })
+        .filter(payload => payload.display !== null);
+      expect(published.length).toBeGreaterThan(0);
+      expect(published[published.length - 1].display?.screens).toEqual([
+        { id: 'dash-0', name: 'Navigation', icon: 'sailing' },
+        { id: 'dash-1', name: 'Engine', icon: 'engine' },
+        { id: 'dash-2', name: 'Anchor', icon: 'anchor' }
+      ]);
+    });
   });
 
   describe('disabling remote control', () => {
@@ -204,7 +218,7 @@ describe('RemoteDashboardsService', () => {
       expect((displayCalls[0][1] as { display: IScreensPayload }).display.displayName).toBe('Nav Station');
     });
 
-    it('re-sends the active dashboard index on a name change', () => {
+    it('republishes screens only, without re-sending the active dashboard index', () => {
       dashboard.activeDashboard.set(2);
       createService();
       enableRemoteControl();
@@ -213,9 +227,8 @@ describe('RemoteDashboardsService', () => {
       settings.instanceName.set('Nav Station');
       TestBed.tick();
 
-      expect(callsTo(SET_SCREEN_INDEX_PATH)).toEqual([
-        [SET_SCREEN_INDEX_PATH, { displayId: KIP_UUID, screenIdx: 2 }, KIP_UUID]
-      ]);
+      expect(callsTo(SET_DISPLAY_PATH)).toHaveLength(1);
+      expect(callsTo(SET_SCREEN_INDEX_PATH)).toHaveLength(0);
     });
 
     it('ignores name changes while disabled', () => {
@@ -372,7 +385,7 @@ describe('RemoteDashboardsService', () => {
       expect(requests.putRequest).toHaveBeenNthCalledWith(2, REQUEST_ACTIVE_SCREEN_PATH, { displayId: 'other-display', screenIdx: 7 }, KIP_UUID);
     });
 
-    it('publishes a shallow copy of the screens payload', () => {
+    it('publishes a deep-enough copy of the screens payload, sharing no reference with the caller', () => {
       const service = createService();
       requests.putRequest.mockClear();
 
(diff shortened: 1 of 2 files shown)


mairas:
## Review synthesis

**Post-review hardening:** added a first-run guard on the new rename effect so boot-time publishing stays owned by the toggle effect — no redundant boot republish (which would have re-introduced exactly the class of bug this PR removes).

mairas:
## Review synthesis (standard multi-persona)

> **Supersedes** the earlier preliminary [Review synthesis comment](https://github.com/halos-org/skip/pull/261#issuecomment-4960574084) on this PR — that was a single-persona automated pass, not the standard review. Treat this comment as authoritative.

**Scope:** `RemoteDashboardsService` cleanup (closes #153). Two-file diff: `src/app/core/services/remote-dashboards.service.ts` + its spec.

**Intent:** delete the never-read `remoteScreenPosition` subscription and its dead `ACTIVE_SCREEN_PATH`; split the toggle/rename effect so a display-name change republishes screens only (no redundant `setScreenIndex`); a first-run guard on the new rename effect skips its initial run so boot publishing stays owned by the toggle effect; `setScreensOnRemote` copies the screens array (no shared reference). Three PR #147 characterization tests flipped.

**Reviewers (6, independent personas):** correctness, testing, maintainability, project-standards, reliability, adversarial.

---

### Findings (post-synthesis, after confidence gate + dedup)

#### P2 — should land before merge

|---|---|---|---|---|---|
| 1 | `remote-dashboards.service.ts:108` | The `nameEffectInitialized` first-run guard is the mechanism this PR introduces, but it is untested. Every spec constructs the service with `isRemoteControl=false` and enables later, so on the rename effect's first run `isRemoteControl()` is always false and the guard's `return` and the following `if (!isRemoteControl()) return` short-circuit identically. Deleting the guard changes zero test outcomes, so a regression that reintroduces the redundant boot-time screens publish ships silently. The guard's only real job surfaces on the reload path (`SettingsService` seeds `isRemoteControl` from stored config at `settings.service.ts:158`), which no spec exercises. | testing (echoed as a shared testing gap by **all six** personas) | 0.85 | **fix here** — add spec |

**Suggested fix (#1):** add a spec that calls `settings.isRemoteControl.set(true)` *before* `createService()`, with an instance name set, then asserts the boot sequence publishes exactly one non-null `SET_DISPLAY` payload (from the toggle effect) — i.e. `callsTo(SET_DISPLAY_PATH)` length 2 (one `display:null` clear + one payload), not 3. Without the guard the rename effect's first run adds a second payload publish and the assertion fails, proving the guard is load-bearing.

#### P3 — advisory

|---|---|---|---|---|---|
| 2 | `remote-dashboards.service.ts:104` | The new name-change effect + `nameEffectInitialized` boolean duplicate the pre-existing "Push dashboard configuration to remote" effect (`:118`), which already calls `shareScreens(getScreensPayload(dashboards))` under the same `isRemoteControl` guard and already embeds `displayName()` via `getScreensPayload`. It just doesn't *track* `displayName()`. Reading `this.displayName()` in that effect's tracked body covers name changes with no separate effect and no boot guard (it already runs once at boot). The chosen path leaves the service with two init booleans (`previousIsRemoteControl`, `nameEffectInitialized`) and three screens-publishing effects, forcing the next reader to trace an implicit "boot publishing is owned by the toggle effect" contract encoded only in a comment. | maintainability | 0.70 | consider (equivalent behavior, one fewer effect) |

Findings #1 and #2 are two sides of the same mechanism (the new rename effect + its guard): #1 says it is untested, #2 says the merge alternative would avoid needing the guard at all. If #2 is adopted, #1's guard-pinning test becomes moot; if the current shape is kept, add the #1 spec.

*No P0/P1 findings. `betterer` ratchet is not at risk — the changed src file has no entry in `.betterer.results` (project-standards).*

---

### Residual risks (advisory, no action required)

- **`setScreensOnRemote` throws on malformed payload:** `[...screensPayload.screens]` (`:~202/214`) would throw a `TypeError` if a non-null `IScreensPayload` with `undefined` screens were passed, where the old `{...screensPayload}` tolerated it. TypeScript makes `screens` required and every in-repo caller passes `null` or a real `getScreensPayload()` array, so no live path hits it — only reachable via a boundary-violating external call. (correctness, testing, maintainability, project-standards, adversarial)
- **Array copy is shallow:** the copy is one level deep; individual `DashboardListItem` elements stay shared with the caller. The flipped test "sharing no reference with the caller" only asserts array identity (`sent.display.screens !== payload.screens`), not element identity. Internal callers pass fresh objects from `getScreensPayload`, so internal flows are safe; the copy is shallower than the test name implies. (testing, maintainability, project-standards, adversarial)
- **Pre-existing boot double-publish (unchanged):** with remote control enabled at construction, the toggle effect and the dashboards effect both emit `setDisplay` on first flush, and the toggle + active-dashboard effects both emit `setScreenIndex`. Idempotent (same end state), predates this PR; the new guard correctly avoids adding a *third* `setDisplay`. (correctness, maintainability, adversarial)
- **Rename no longer resends active index:** rename-while-enabled now republishes the screens catalog (`setDisplay`) without the `setScreenIndex` the old merged effect sent. Safe **only** if the remote controller keeps its per-display active highlight as separate state across a `setDisplay`. That contract lives in a separate consumer not in this repo, so the PR's "redundant" claim is unverifiable here — worth confirming against a live controller. (adversarial)
- **Two effects vs one coalesced run:** if `isRemoteControl` and `instanceName` ever change in the same synchronous tick while running, two `setDisplay` publishes fire where the merged effect emitted one. No current path changes both in one tick post-boot — latent, not active. (adversarial)
- **First-run guard leans on Angular scheduling:** correctness of the boot-ownership assumption depends on Angular scheduling a newly-created effect's initial run independent of tracked-signal changes. Holds for current signal-effect semantics; a future scheduling change could invalidate it. (correctness)
- **Fire-and-forget SK PUTs (pre-existing):** all remote-display sync writes are best-effort with no timeout/retry; a lost `setDisplay` leaves the remote list stale until the next change republishes. Acceptable for a non-safety-critical display; outside this diff. (reliability)

### Testing gaps (advisory)

- No test constructs the service with `isRemoteControl` already true at boot (the reload path) — the exact case the first-run guard exists for. **(all six personas — dominant gap; this is the basis of finding #1)**
- No test asserts the rename effect does **not** re-fire when `isRemoteControl` toggles (it tracks only `displayName`), so a future change that added `isRemoteControl` tracking to it would go unnoticed. (correctness, testing)
- No test for the rename-while-disabled-then-enable sequence (that enabling republishes the *new* instance name via the toggle effect). (adversarial)
- No test pins the publish count for a simultaneous `isRemoteControl` + `instanceName` change. (maintainability, adversarial)
- Flipped copy test only checks array identity; element-sharing intent is unpinned, so a future deep-copy or re-alias would go uncaught. (testing, project-standards, adversarial)
- No characterization test drives a server-side rejection (403/500/504) through a real `setDisplay`/`setScreenIndex` publish at this service's level. (reliability)
- No test asserts that removing the `remoteScreenPosition` subscription dropped no needed side effect (coverage confirms the path is no longer subscribed, not that nothing depended on it). (reliability)

---

**Verdict: ready-with-fixes.** No P0/P1; the diff does what it intends and no correctness defect survived gating. Before merge, add the P2 guard-pinning spec (finding #1) so the load-bearing `nameEffectInitialized` guard can't be silently regressed — or adopt the maintainability merge (#2), which removes the need for the guard entirely and then only needs the display-name characterization specs re-run. Everything else is advisory.

<sub>Automated multi-persona synthesis. Not an approval or a change-request — no merge action taken.</sub>

mairas:
**Review findings addressed** (P2 + P3) via the reviewer’s simplification: dropped the separate rename effect **and** the `nameEffectInitialized` first-run guard entirely, and folded a tracked `displayName()` read into the existing config-push effect. That effect only calls `shareScreens` (never the active index), so #153’s "republish screens only on rename" holds — and there is no guard left untested (resolves the P2) and no duplicated effect (resolves the P3). Added a boot-enabled publish test. Local gate green; CI re-running.

Issue #241, opened 2026-07-10
Port Kip #1107: boolean switch/light OFF-state visibility + momentary-button immediate emit


mairas:
## Upstream

mxtommy/Kip PR: https://github.com/mxtommy/Kip/pull/1107 (merged, squash `f9d657cf`)

Upstream "Fix switch panel indicator state for better visibility" bundles two changes to the boolean switch-panel widgets:

1. **OFF-state visibility rework** (`svg-boolean-switch`, `svg-boolean-light`): the OFF indicator now uses a dedicated dimmer `offColor` (theme `*Dimmer`) instead of reusing `labelColor` / `var(--mat-sys-outline)`; the OFF label is dimmed (`opacity 0.6`) and its glow softened; the light gains layered radial specular/depth gradients plus a highlight ellipse for a 3D lit look; stroke widths move into the layout util (`switchStrokeWidth`, and `lightStrokeWidth` reduced 5→3).
2. **Momentary-button fix** (`svg-boolean-button`): removes the 200ms hold delay so the momentary button emits state immediately on press; swipe detection now sets `pressed = false` and clears timers directly, and the `holdTimeoutId` field is removed.

## Why this needs a manual port

The upstream patch does not apply — `git apply --check f9d657cf` fails on all present files, and it also targets `src/app/widgets/widget-boolean-switch/boolean-control-layout.util.ts`, **which does not exist in skip**.

At skip's fork point upstream's switch/light widgets used the same two-group SVG structure skip still ships (`<g id="disabled">` / `<g id="enabled">` with hardcoded coordinates and a `glow` filter). Upstream has since rebuilt these widgets around a single-layout model driven by `boolean-control-layout.util.ts` (`layout.*` bindings, `isEnabled ? … : …` ternaries, `dimensions()`). skip never merged that refactor, so the visual half of this PR must be reimplemented against skip's divergent templates rather than cherry-picked.

skip's `svg-boolean-button.component.ts` momentary logic, by contrast, is identical to upstream's pre-PR version, so change (2) is a small, self-contained edit — but it still needs to be hand-applied (surrounding strictNullChecks context defeats a clean apply) and it is a separate concern from the visual rework.

## Affected skip files

- `src/app/widgets/svg-boolean-switch/svg-boolean-switch.component.svg` — two-group structure, hardcoded coords, `fill:var(--mat-sys-outline)` for OFF track.
[...]

PR #262, opened 2026-07-13
fix(boolean-button): momentary button emits immediately on press (#241a)

mairas:
## Why
The momentary boolean button waited 200ms before its first emit, adding noticeable latency to a boat control (#241a, the momentary half of #241).

## How
Cut the pre-emit hold to **~75ms**: a deliberate press actuates in ~75ms (feels instant), while a swipe/scroll that crosses the 30px threshold within the window still cancels before any emit reaches the actuator — so an accidental scroll starting on the button does not fire it. The 100ms repeat while held, swipe-cancel, release teardown, and the null-safe `data()` guard are unchanged.

**Skip-specific choice** (per review): upstream Kip PR #1107 removed the delay entirely; on marine momentary actuators we keep a short guard window to avoid spurious actuation on touch scroll.

## Verification
- New spec: actuates after the hold not before; swipe-within-window cancels (no emit); quick-tap-in-window no-actuate; 100ms repeat while held; sub-threshold move doesn't cancel; release stops the interval; null `data()` no-op. Fake timers.
- Local static gate: `lint` ✓, `betterer:ci` ✓, `tsc --noEmit` ✓. Vitest runs in CI.

Scope is #241a only — OFF-state visuals (#241b) remain separate, so #241 stays open. Pre-existing `IDimensions` coupling surfaced in review filed as #263.

Part of #241 (#241a).

### DIFF
--- a/src/app/widgets/svg-boolean-button/svg-boolean-button.component.spec.ts
+++ b/src/app/widgets/svg-boolean-button/svg-boolean-button.component.spec.ts
@@ -1,0 +1,135 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { SvgBooleanButtonComponent } from './svg-boolean-button.component';
+import type { IDynamicControl } from '../../core/interfaces/widgets-interface';
+import type { IDimensions } from '../widget-boolean-switch/widget-boolean-switch.component';
+
+const controlMock: IDynamicControl = {
+  ctrlLabel: 'Bilge Pump',
+  type: '2',
+  pathID: 'test-path',
+  value: false,
+  color: 'contrast',
+  isNumeric: false
+};
+
+const dimensionsMock: IDimensions = { width: 180, height: 70 };
+
+function pointerAt(clientX: number, clientY: number): PointerEvent {
+  return { clientX, clientY } as unknown as PointerEvent;
+}
+
+describe('SvgBooleanButtonComponent momentary emit', () => {
+  let fixture: ComponentFixture<SvgBooleanButtonComponent>;
+  let component: SvgBooleanButtonComponent;
+  let emitted: IDynamicControl[];
+
+  const setup = (controlData: IDynamicControl | null = controlMock, render = true) => {
+    TestBed.configureTestingModule({ imports: [SvgBooleanButtonComponent] });
+    fixture = TestBed.createComponent(SvgBooleanButtonComponent);
+    component = fixture.componentInstance;
+    fixture.componentRef.setInput('controlData', controlData);
+    fixture.componentRef.setInput('theme', null);
+    fixture.componentRef.setInput('dimensions', dimensionsMock);
+    emitted = [];
+    component.toggleClick.subscribe((v) => emitted.push(v));
+    if (render) fixture.detectChanges();
+  };
+
+  beforeEach(() => TestBed.resetTestingModule());
+  afterEach(() => {
+    vi.useRealTimers();
+    TestBed.resetTestingModule();
+    vi.clearAllMocks();
+  });
+
+  it('actuates after the short hold, not before', () => {
+    setup();
+    vi.useFakeTimers();
+
+    component.handleClickDown(pointerAt(10, 10));
+    expect(emitted).toHaveLength(0);
+    vi.advanceTimersByTime(74);
+    expect(emitted).toHaveLength(0);
+
+    vi.advanceTimersByTime(1);
+    expect(emitted).toHaveLength(1);
+    expect(emitted[0].value).toBe(true);
+  });
+
+  it('does not actuate when a swipe cancels within the hold window', () => {
+    setup();
+    vi.useFakeTimers();
+
+    component.handleClickDown(pointerAt(10, 10));
+    // A scroll/swipe crossing the threshold before the hold elapses must suppress actuation entirely.
+    component.handlePointerMove(pointerAt(60, 12));
+    vi.advanceTimersByTime(500);
+
+    expect(emitted).toHaveLength(0);
+  });
+
+  it('does not actuate on a quick tap released within the hold window', () => {
+    setup();
+    vi.useFakeTimers();
+
+    component.handleClickDown(pointerAt(10, 10));
+    component.handleClickUp();
+    vi.advanceTimersByTime(500);
+
+    expect(emitted).toHaveLength(0);
+  });
+
+  it('repeats every 100ms while held', () => {
+    setup();
+    vi.useFakeTimers();
+
+    component.handleClickDown(pointerAt(10, 10));
+    vi.advanceTimersByTime(75);
+    expect(emitted).toHaveLength(1);
+
+    vi.advanceTimersByTime(100);
+    expect(emitted).toHaveLength(2);
+    vi.advanceTimersByTime(100);
+    expect(emitted).toHaveLength(3);
+  });
+
+  it('keeps the press when movement stays below the swipe threshold', () => {
+    setup();
+    vi.useFakeTimers();
+
+    component.handleClickDown(pointerAt(10, 10));
+    // 25px/5px move — under the 30px threshold, so it must not cancel.
+    component.handlePointerMove(pointerAt(35, 15));
+    vi.advanceTimersByTime(75);
+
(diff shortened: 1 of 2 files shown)


mairas:
## Review synthesis

**Post-review hardening:** the press test now uses fake timers so no real 100 ms interval leaks into the test run (`ngOnDestroy`→`clearTimers` already covers teardown, but this keeps the test self-contained).

mairas:
## Review synthesis (standard multi-persona pass)

> **This supersedes the earlier preliminary [_Review synthesis_ comment](https://github.com/halos-org/skip/pull/262#issuecomment-4960574294).** That was a single-persona automated pass; this is the standard review skill output — five independent persona reviewers (correctness, testing, maintainability, project-standards, reliability) run in fresh contexts, then synthesized (confidence-gated, deduplicated, agreement-weighted).

**Scope:** `svg-boolean-button` momentary port — PR #262 (#241a, half of #241).
**Intent:** emit the pressed state immediately on `pointerdown` (remove the 200ms hold delay + `holdTimeoutId` field), start a 100ms repeat interval while held; a swipe past the 30px threshold sets `pressed=false` and clears timers; the null-safe `data()` guard is preserved (file stays strict-clean); new focused vitest spec drives fake timers.
**Diff:** `svg-boolean-button.component.ts` (rewritten `handleClickDown`, simplified `handlePointerMove`/`handleClickUp`/`clearTimers`) + new `svg-boolean-button.component.spec.ts`.

Line references are to the PR's post-change files.

---

### P2 — should land (or get a product call) before merge

|---|---|---|---|---|---|
| 1 | `svg-boolean-button.component.spec.ts` (whole spec) / component.ts:95 | **Release path `handleClickUp` is entirely untested.** The momentary contract is emit-while-held, **stop-on-release**. `handleClickUp` is bound to both `pointerup` and `pointerleave` and was refactored in this diff (timer teardown → `clearTimers()`), yet no test invokes it. A regression that fails to clear the 100ms interval on a normal finger-lift leaves a momentary control (the mock is a "Bilge Pump") emitting `value=true` forever, and CI would stay green. Safety-relevant gap on a marine control. | testing | 0.90 | fix-here (add spec case before merge) |
| 2 | `svg-boolean-button.component.ts:67` | **Immediate emit on `pointerdown` fires one actuator PUT on a swipe/scroll that main's 200ms delay suppressed.** On main the first emit was gated behind the hold timeout, so a swipe/scroll starting on the button crossed the 30px threshold and cancelled it before any emit → zero PUTs. Now the emit is synchronous; the parent (`widget-boolean-switch.toggle()` → `signalkRequestsService.putRequest`) turns every emit into a real SK PUT, so the leading edge of any touch that lands on the control actuates it (horn/windlass/bilge) before the swipe branch can cancel. The swipe branch (`pressed=false` + `clearTimers()`) only stops the repeat — it cannot un-send the first `true`, and no compensating `false` is emitted. **Direct trade-off of the "emit immediately" intent — needs a product call, not necessarily a revert.** | correctness + reliability *(independent agreement — treat as higher-confidence)* | 0.62 | discuss / product-call; mitigations below |

**On #2 — mitigation options if the leading-edge actuation is unacceptable:**
- `touch-action: none` on the interactive `<g>` in the SVG so a scroll cannot originate on the control (pointer stays captured; swipe-cancel governs), **or**
- emit a compensating `value:false` in the `handlePointerMove` swipe-cancel branch so an accidental leading-edge `true` is explicitly released. This pairs naturally with the release-emit work deferred to #241b (today `handleClickUp` never emits `false` on a normal release either).

---

### P3 — nice to fix

|---|---|---|---|---|---|
| 3 | `svg-boolean-button.component.ts:78` | Sub-threshold move (no-cancel) branch untested. Test 2 covers `deltaX=50` (cancel); nothing verifies a small jitter below 30px keeps the emit running. An inverted condition or lowered threshold would silently stop a held control on minor movement, uncaught. | testing | 0.82 | fix-here (add spec case) |
| 4 | `svg-boolean-button.component.ts:60` | Null `controlData` guard error path untested. PR explicitly preserves `if (!data) return`, but no test drives `controlData=null` through `handleClickDown` (call the handler directly, without `detectChanges`, since the template dereferences `data()`). Dropping the guard would `cloneDeep(null)`-throw, uncaught. | testing | 0.60 | fix-here (optional) |
| 5 | `svg-boolean-button.component.spec.ts:52` | Spec comment narrates removed behaviour ("Old behaviour deferred the first emit behind a 200ms hold timeout"). Violates the repo comment standard (explain why, never what changed; write for a reader who never saw the diff). The test title already conveys the intent — delete or restate as the current contract. | maintainability | 0.60 | fix-here (trivial) |
| 6 | `widget-boolean-switch.component.ts:17` *(pre-existing)* | `IDimensions` is a data-only type defined inside `widget-boolean-switch.component.ts` but imported by `svg-boolean-button`, `svg-boolean-light`, `svg-boolean-switch`, `svg-zone-states` — and now this new spec (5th importer). Consuming a type from an unrelated component's module couples those widgets to that file. Belongs in `core/interfaces/widgets-interface.ts` alongside `IDynamicControl`. | maintainability | 0.60 | file-issue (touches 5 files, out of scope for this PR) |

---

### Residual risks

- **Momentary "off" is never transmitted.** Neither `handleClickUp` (normal release) nor the swipe path emits `value=false`, so the actuator only ever receives repeated `true` then silence — off-semantics depend entirely on server-side auto-reset or the deferred #241b work. Pre-existing (main behaved the same after its delay); explicitly the other half of #241.
- **Emit/PUT rate:** while held, ~10 PUT/s per button to the SK server, no client-side coalescing; multiple held buttons multiply it. Unchanged from main, but now reachable with zero hold delay (interval starts on every press).
- **Runaway interval if all release events are missed.** The interval stops only via `pointerup`/`pointerleave`/swipe/`ngOnDestroy`. `pointercancel` is unbound; compliant browsers fire `pointerleave` after `pointercancel`, so covered in practice, but a non-conforming environment would leave the interval emitting forever. Pre-existing, mitigated.
- **Stale snapshot:** the interval closure captures one `cloneDeep(data)`; a server-side value change while held emits the stale state. Harmless for a momentary press; note if #241b reuses the pattern.
- **Betterer/strict-null:** `svg-boolean-button.component.ts` has zero `.betterer.results` entries, and the `if (!data) return` guard is retained, so the file stays strict-clean — no baseline regeneration needed.

### Testing gaps (consolidated)

- `handleClickUp` normal release (stops interval) — **no test** (P2 #1).
- `handleClickUp` swipe early-return branch (`isSwiping` reset) — no test.
- `handlePointerMove` sub-threshold branch (small move must not cancel) — no test (P3 #3).
- `handleClickDown` null-`controlData` guard — no test (P3 #4).
- Test 2 is titled "…and releases" but asserts only the **absence** of further emits — it never asserts any release/`false` emission, so "releases" is unverified.
- Handlers are invoked directly on the instance; the SVG template event bindings (`pointerdown`/`move`/`up`/`leave` → handlers) are never exercised, so a broken/removed binding would pass.
- `ngOnDestroy`/`clearTimers` teardown while pressed — no test.
- **CI is the source of truth for vitest** (Node 24, `npm ci`, `@angular/build:unit-test`); the spec is statically well-formed and touches no `localStorage`, but its passing status must be confirmed by CI before merge.

---

**Verdict: ready-with-fixes.** No P0/P1. Two P2s gate a clean merge: (1) add the `handleClickUp` release-path test — a green CI must not be able to hide a momentary control that never stops emitting; and (2) get a product call on the leading-edge actuation-on-swipe trade-off (accept it, or apply one of the mitigations above). The P3s are quick follow-ons in this PR except #6, which should become its own refactor issue. Post-synthesis counts: **P0 0 · P1 0 · P2 2 · P3 4** (1 pre-existing).

*No approval / change-request submitted — comment only, per review policy.*

mairas:
**Review findings addressed.** Per a product decision, this now uses a **~75ms hold delay** rather than zero-delay immediate emit (deviating from upstream #1107): a deliberate press actuates in ~75ms while a swipe crossing the threshold within the window cancels before any emit — resolving the leading-edge spurious-actuation **P2**. Added tests for hold/swipe-cancel/quick-tap/repeat/sub-threshold/**release-teardown** (the safety P2) / null paths, and reworded the history-narrating comment. Pre-existing `IDimensions` coupling filed as #263. Local gate green; CI re-running.

PR #264, opened 2026-07-13
refactor(charts): shared Chart.js registration; gate annotations on finite values

### LINKED ISSUE #239: Port Kip #1104: centralize Chart.js registration + fix data-chart annotation display
## Upstream

mxtommy/Kip PR: https://github.com/mxtommy/Kip/pull/1104 (merged as `74c11fdb`, "Rollback Chart.js and annotation plugin registration").

## What upstream changed

Two coupled changes:

1. **Registration reorg.** Removes the per-component `Chart.register(...)` calls (widget-history-chart-dialog, minichart, widget-data-chart, widget-windtrends-chart) and registers every Chart.js building block + the `annotation` and `streaming` plugins **once** in `src/main.ts`, before any chart is created. Also drops the now-dead empty `annotation: { annotations: {} }` blocks from minichart/windtrends plugin options.
2. **Data-chart annotation display fix (the real bug).** In `widget-data-chart`, min/max/average annotation lines were configured with a placeholder value and `label.display: true`, so an enabled line could render at value `0` with an empty label before real data arrived. The fix tracks `lastAverageValue`/`lastMinimumValue`/`lastMaximumValue` and gates both the line `display` and the label `display`/`content` on the presence of a finite value.

## Why it needs a manual port

The patch does not apply to skip. `git apply --check` is clean for `src/main.ts` and `widget-history-chart-dialog.component.ts`, but fails for `minichart`, `widget-windtrends-chart`, and especially `widget-data-chart`, because skip diverged those files (its own `HistoryChartStreamService` streaming path, `resolveWindowMs`/`deriveDataSourceInfo`, `value: null` base config, and an `applyAnnotationValues` path that reads `point.data.lastAverage` and gates only on `Number.isFinite`). The annotation-display gating must be re-expressed against skip's current structure rather than cherry-picked.

Note: skip's `signalk.component.ts` uses `chart.js/auto`, so it self-registers and is unaffected by the registration move.

## Affected skip files

- `src/main.ts` — add the one-time global `Chart.register(...)` (union of what all four widgets need: annotation, streaming, CategoryScale, Filler, Legend, LineController, LineElement, LinearScale, PointElement, SubTitle, TimeScale, Title, Tooltip). Verified this union covers every plugin the four registrars currently list.
- `src/app/core/components/widget-history-chart-dialog/widget-history-chart-dialog.component.ts` — drop local `Chart.register` (line 29).
- `src/app/widgets/minichart/minichart.component.ts` — drop local `Chart.register` (line 15) and the `ChartStreaming` import (line 12); skip already lacks the empty annotation block.
- `src/app/widgets/widget-windtrends-chart/widget-windtrends-chart.component.ts` — drop local `Chart.register` (line 19) and the `ChartStreaming` import (line 17); skip already lacks the empty annotation block.
- `src/app/widgets/widget-data-chart/widget-data-chart.component.ts` — drop local `Chart.register` (line 18) plus the `annotationPlugin`/`ChartStreaming` imports, then reimplement the annotation gating (base config at lines 367-410, `updateAnnotationVisibility` at 546-553, value update at 815-830) so min/max/average line **and** label `display` follow a tracked finite value.

## Suggested approach

1. Land the registration move first as one unit: add the global `Chart.register` to `main.ts`, then remove all four per-component registrations and the dead plugin imports. Removing local registrations only after the global one is in place avoids a window where a chart is created with no controller registered.
2. Separately reimplement the widget-data-chart annotation gating against skip's diverged `applyAnnotationValues`/`updateAnnotationVisibility`, mirroring upstream: hide line and label until the corresponding value is finite; set label `content` from the tracked value.
3. Extend skip's existing `widget-data-chart.component.spec.ts` (skip already has its own) with a case asserting annotation lines/labels stay hidden with no data — do not copy upstream's new spec files verbatim, since skip's test setup differs.

mairas:
## Why
Chart.js was registered per-component in four consumers (data-chart, minichart, windtrends, history-dialog), and the data-chart threshold annotation line/label rendered at a placeholder 0 before finite data arrived (#239).

## How
- One idempotent `registerChartComponents()` (registerables ∪ annotation ∪ streaming) in `src/app/core/utils/chart-registration.util.ts`, called once by each of the four consumers; the per-component `Chart.register` calls are deleted.
- data-chart annotation lines **and** labels start hidden and show only once their tracked value is `Number.isFinite`.

## Verification
- Full vitest suite green locally (1158 tests); `lint` ✓, `betterer:ci` 183 ✓.
- New data-chart specs: a default-enabled line stays hidden with no data, then shows with its value once a finite value streams.

Closes #239

### DIFF
--- a/src/app/core/components/widget-history-chart-dialog/widget-history-chart-dialog.component.ts
+++ b/src/app/core/components/widget-history-chart-dialog/widget-history-chart-dialog.component.ts
@@ -2,8 +2,9 @@ import { AfterViewInit, ChangeDetectionStrategy, Component, DestroyRef, ElementR
 import { MAT_DIALOG_DATA, MatDialogModule } from '@angular/material/dialog';
 import { MatButtonModule } from '@angular/material/button';
 import { toSignal } from '@angular/core/rxjs-interop';
-import { Chart, ChartConfiguration, ChartDataset, Color, LegendItem, LineController, LineElement, LinearScale, PointElement, TimeScale, Tooltip, Legend } from 'chart.js';
+import { Chart, ChartConfiguration, ChartDataset, Color, LegendItem } from 'chart.js';
 import 'chartjs-adapter-date-fns';
+import { registerChartComponents } from '../../utils/chart-registration.util';
 import { IWidget } from '../../interfaces/widgets-interface';
 import { IKipSeriesDefinition } from '../../services/kip-series-api-client.service';
 import { isKipTemplateSeriesDefinition, type IKipConcreteSeriesDefinition, type IKipTemplateSeriesDefinition } from '../../contracts/kip-series-contract';
@@ -26,7 +27,7 @@ import { MatIconModule } from '@angular/material/icon';
 import { MatButtonToggleModule } from '@angular/material/button-toggle';
 import { FormsModule} from '@angular/forms';
 
-Chart.register(LineController, LineElement, LinearScale, PointElement, TimeScale, Tooltip, Legend);
+registerChartComponents();
 
 interface ChartPoint {
   x: number;
--- a/src/app/core/utils/chart-registration.util.spec.ts
+++ b/src/app/core/utils/chart-registration.util.spec.ts
@@ -1,0 +1,22 @@
+import { describe, it, expect, vi } from 'vitest';
+
+// Chart.js cannot instantiate under jsdom; the only behaviour under test is the module-level
+// register-once guard. The suite shares module state, so other chart specs may have already
+// imported (and tripped) the util — reset the module registry inside the test and re-import
+// fresh so the guard starts clean regardless of run order.
+vi.mock('chart.js', () => ({ Chart: { register: vi.fn() }, registerables: [] }));
+vi.mock('chartjs-plugin-annotation', () => ({ default: {} }));
+vi.mock('@aziham/chartjs-plugin-streaming', () => ({ default: {} }));
+
+describe('registerChartComponents', () => {
+  it('registers the Chart.js components only once across repeated calls', async () => {
+    vi.resetModules();
+    const { Chart } = await import('chart.js');
+    const { registerChartComponents } = await import('./chart-registration.util');
+
+    registerChartComponents();
+    registerChartComponents();
+
+    expect(vi.mocked(Chart.register)).toHaveBeenCalledTimes(1);
+  });
+});
--- a/src/app/core/utils/chart-registration.util.ts
+++ b/src/app/core/utils/chart-registration.util.ts
@@ -1,0 +1,19 @@
+import { Chart, registerables } from 'chart.js';
+import annotationPlugin from 'chartjs-plugin-annotation';
+import ChartStreaming from '@aziham/chartjs-plugin-streaming';
+
+let registered = false;
+
+/**
+ * Registers every Chart.js building block plus the annotation and streaming
+ * plugins exactly once for the whole app. Chart consumers call this at module
+ * load instead of each running their own `Chart.register`, so a chart is never
+ * created before its controller/scale/plugin set is present.
+ */
+export function registerChartComponents(): void {
+  if (registered) {
+    return;
+  }
+  registered = true;
+  Chart.register(...registerables, annotationPlugin, ChartStreaming);
+}
--- a/src/app/widgets/minichart/minichart.component.spec.ts
+++ b/src/app/widgets/minichart/minichart.component.spec.ts
@@ -18,11 +18,13 @@ vi.mock('chart.js', () => {
   }
   return {
     Chart: MockChart,
+    registerables: [],
     TimeScale: {}, LinearScale: {}, LineController: {}, PointElement: {},
     LineElement: {}, Filler: {}, CategoryScale: {}
   };
 });
 vi.mock('chartjs-adapter-date-fns', () => ({}));
+vi.mock('chartjs-plugin-annotation', () => ({ default: {} }));
 vi.mock('@aziham/chartjs-plugin-streaming', () => ({ default: {} }));
 
 import { ComponentFixture, TestBed } from '@angular/core/testing';
--- a/src/app/widgets/minichart/minichart.component.ts
+++ b/src/app/widgets/minichart/minichart.component.ts
@@ -7,12 +7,11 @@ import { CanvasService } from '../../core/services/canvas.service';
 import { ITheme } from '../../core/services/app-service';
 import { UnitsService } from '../../core/services/units.service';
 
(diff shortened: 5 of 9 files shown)


mairas:
## Review synthesis — PR #264

**Scope:** `refactor(charts): shared Chart.js registration; gate annotations on finite values` (branch `refactor/chartjs-shared-registration`, closes #239).
**Intent:** (1) Consolidate Chart.js registration into one idempotent `registerChartComponents()` (registerables ∪ annotation ∪ streaming plugin) in `chart-registration.util.ts`, called once by the 4 chart consumers; per-component `Chart.register` calls deleted. (2) Hide the widget-data-chart threshold annotation line *and* its label until the tracked value is `Number.isFinite`.
**Reviewers:** correctness, testing, maintainability, project-standards, adversarial (5 independent personas). Synthesis dedups cross-reviewer agreement and suppresses findings below 0.60 confidence.

All surviving findings are **P3 (advisory)** — none block merge. No P0/P1/P2.

### P3 — advisory

|---|---|---|---|---|---|
| 1 | `widget-data-chart.component.ts:154` | Theme / display-config change hides enabled annotation lines until the next stream emission. In the `else if (this.chart)` branch, `updateAnnotationVisibility()` (154) re-applies the cached line values, then `setChartOptions(cfg)` (156) rebuilds `plugins.annotation` wholesale with `display:false`/`value:undefined`/`label.display:false`, and `chart.update('none')` renders the hidden state. Since the chart was built with `options: this.lineChartOptions`, the reassignment lands live. Enabled avg/min/max lines vanish until the next delta — seconds on minute-scale, minutes/indefinitely on hour-day or static SK paths (e.g. a dusk day→night theme switch). Verified on-branch: effect order is exactly `updateAnnotationVisibility()` → `setChartOptions()`. Fix: move `updateAnnotationVisibility()` **after** `setChartOptions(cfg)` (matches the emission path at ~870). | correctness + adversarial | **0.75** (agreed) | fix here |
| 2 | `widget-data-chart.component.ts:570` | `updateAnnotationVisibility` now also writes each line's `value` + `label.content` (via `applyAnnotationLine`), not just toggles `display`; the name understates scope, inviting a future maintainer to reorder/drop the call in the value-application path and silently lose value/label updates. Rename (e.g. `refreshAnnotationLines`) or split visibility vs value application. | maintainability | 0.62 | fix here |
| 3 | `chart-registration.util.ts:13` | New shared util has no unit test despite an established `*.util.spec.ts` convention (8 sibling specs). The `if (registered) return` guard + the registerables∪annotation∪streaming union are the only new logic; a guard inversion or dropped `registerables` would pass CI (4 consumer specs mock chart.js) and only break at runtime. Add a spec: call twice, assert `Chart.register` invoked once with the expected union. | testing | 0.60 | fix here |
| 4 | `widget-data-chart.component.spec.ts:118` | The annotation-hidden test is tautological: with no data emitted, `setChartOptions` is the last writer and hardcodes `display:false`/`label.display:false`, so the asserted `false` holds regardless of the `Number.isFinite` gate. A regression dropping the gate would not be caught here — only the companion "shows once finite" test proves it. Drive the gate directly: emit a non-finite tracked stat with the line enabled and assert it stays hidden. | testing | 0.60 | fix here |

**Dedup note:** findings from correctness and adversarial at `widget-data-chart.component.ts:154` describe the same ordering defect and propose the identical one-line fix; merged as #1 with elevated confidence. **Pre-existing status:** the ordering predates this PR (on `main`, `setChartOptions` kept `display:true`, so the line rendered at a stale position rather than hidden). This PR's new `lastAverage/Min/Max` backing fields exist precisely to preserve the "enabled+finite ⇒ visible" invariant, and this path violates it — so it is PR-relevant and cheap to fix here rather than filing.

### Residual risks (non-blocking)

- **Full `registerables` now pulled into every chart consumer** (all controllers/scales/elements incl. bar/radar/doughnut, plus the Tooltip and Colors plugins) where each widget previously registered only its line-chart subset. Behavior-safe today — datasets set `pointHitRadius:0`/`pointRadius:0` and explicit `borderColor`/`backgroundColor`, so tooltips have nothing to hit and auto-palette short-circuits — but a future dataset adding a hit radius or omitting a color would surface behavior that did not exist before. Enlarges the registered set / bundle for a Pi-class target; **route a bundle/tree-shaking check to a perf reviewer.** (correctness, maintainability, adversarial)
- `widget-history-chart-dialog` now has annotation + streaming plugins registered globally though it renders a static time-scale chart; both should no-op, but it is a new coexistence surface. (correctness)
- `signalk.component.ts` (options page) creates a 5th Chart via `chart.js/auto` (self-registering) and is **not** consolidated into `registerChartComponents()`, so the "single idempotent registration" is not literally singular app-wide. Harmless today; consistency gap. (adversarial)
- The `registered` flag is set `true` **before** `Chart.register(...)` runs, and idempotency relies on that process-global singleton. Harmless (a module-load throw crashes loudly), but the flag-before-call order offers no benefit and would mask any future per-instance/registration-order need. (adversarial, correctness)

### Testing gaps (non-blocking)

- No unit test for `registerChartComponents` idempotency or that it registers the registerables ∪ annotation ∪ streaming union (see finding #3).
- **`widget-history-chart-dialog.component.spec.ts` was not updated and does not mock chart.js or its plugins** — via the shared util it now transitively imports and registers the real annotation + streaming plugins at module load, whereas the other 3 chart specs mock all three and the data-chart spec comments call the plugins jsdom-hostile. It never instantiates a `Chart` (no `detectChanges`), so registration-only likely holds, but this is **unverified on this host (boat box runs live services; tests deferred to CI)** and would red CI if plugin import has jsdom side effects. (correctness, testing, maintainability, project-standards)
- New widget-data-chart tests cover only `averageLine` (hidden → shown-after-finite). No coverage for `minimumLine`/`maximumLine` finite-gating + value/label content, the `enabled → finite(shown) → disabled(hidden)` transition, the vertical-chart `scaleID` branch, the config/theme-change path (finding #1 scenario), or the reset of `last*Value` to `NaN` on `rebuildForDataset`. (all)
- No test for the `applyAnnotationLine` `enabled=false` branch (a disabled line must stay hidden even when the tracked value is finite). (testing)

### project-standards

Zero findings. Betterer ratchet confirmed not triggered (none of the 8 changed files are keyed in `.betterer.results`; the new util has no nullable-access constructs) — verified statically, not by running `betterer:ci` on this live host. `COPILOT.md` / `.github/instructions/*` are disclaimed by CLAUDE.md as stale upstream and were not applied.

---

**Verdict: ready-to-merge.** No P0/P1/P2. The four surviving items are all P3 advisory. Recommend landing finding #1 first — it is a genuine (if edge-case) user-visible defect that contradicts the PR's own stated goal, and the fix is a one-line reorder. Findings #2–#4 (rename + two test additions) are cheap to fold in but do not gate merge. Tests were not executed here (live-services host); CI is the source of truth.

<sub>Synthesized from 5 independent persona reviews. This is a review summary, not an approval.</sub>

PR #265, opened 2026-07-13
test(plugin): standalone vitest for the Freeboard panel plugin + manifest fix

### LINKED ISSUE #221: chore: harden Freeboard-SK panel manifest and add plugin tests
Deferred P3 items from the #218 review (panel v1 shipped without them by design).

- **Declare host capabilities in `requires`.** `plugin/index.js` uses `requires: []`. It renders on Freeboard-SK 3.0.0-beta.0 because the beta doesn't gate on `requires`, but a GA host that enforces it could drop the panel/button. Declare the capabilities the manifest actually uses (per plotter-extensions-api: panel-iframe + button support) once the GA contract is known. Changing it now risks hiding the extension against the beta, so it needs re-testing against the host.
- **Add plugin tests.** `plugin/index.js` encodes three contracts with no coverage: the running gate (`listResources` returns {} / `getResource` throws until `start()`), the read-only rejections (`setResource`/`deleteResource` throw), and the id match. `plugin/` is outside the Angular vitest scope, so this needs a small standalone node/vitest config (the `test:mcp-schema` script is a precedent).
- **Revisit button `slot`.** `slot: 'primary'` rendered into a secondary Freeboard toolbar group in testing — placement is host-version-dependent. Confirm the correct slot value against the host doc.

Also worth considering: an automated check that the emitted manifest conforms to the Freeboard-SK plotter-extension schema, so upstream field/enum drift fails CI instead of silently dropping the button.

mairas:
## Why
`plugin/index.js` (the bundled Freeboard-SK panel plugin) had no tests, and its manifest advertised `requires: []` / `slot: 'primary'` rather than the intended `requires: ['panels.iframe','buttons']` / `slot: 'mapToolbar'` (#221).

## How
- New standalone **Node** vitest (`plugin/vitest.config.ts`, mirroring the `gen-mcp-schema` precedent) + `plugin/index.spec.ts` covering the running-gate, read-only rejection, plugin id, and manifest shape.
- Corrected the manifest `requires`/`slot`; wired `test:plugin` into `npm run ci`; CHANGELOG noted.

## Verification
- `npm run test:plugin` green (7 tests); full app suite + lint + betterer unaffected.
- ⚓ **boat-verify (Stage-1 close):** confirm the button lands in `mapToolbar` on the deployed Freeboard beta and the `requires` change doesn't hide the extension (pre-agreed escalation: if the slot misrenders, revert the slot value, keep the tests).

Closes #221


mairas:
## Review synthesis — PR #265

**Scope / intent:** Add a standalone Node vitest for the Freeboard-SK panel plugin — `plugin/vitest.config.ts` (mirroring `tools/gen-mcp-schema`) and `plugin/index.spec.ts` covering `plugin/index.js`: running-gate, read-only rejection, plugin id, manifest shape. Wire `test:plugin` into `npm run ci`. Correct the manifest to `requires=['panels.iframe','buttons']` and `slot='mapToolbar'` (was `[]` / `'primary'`). The manifest change is a Freeboard-SK external contract, boat-verified separately. Closes #221.

**Reviewers (5 independent personas):** correctness, testing, maintainability, project-standards, api-contract. Two returned clean (correctness, project-standards).

**Post-synthesis:** 3 findings survived (all P3 / advisory); no P0/P1/P2. Nothing blocking.

### P3 — advisory

| # | file:line | issue | reviewer(s) | conf. | route |
|---|-----------|-------|-------------|-------|-------|
| 1 | `plugin/index.spec.ts:146` | Manifest-shape test omits two behavioral fields: `panel.lifecycle` (`'keepAlive'`, controls whether Freeboard keeps the iframe alive across toggles) is unasserted, and nothing ties `panel.id` (`'skip-panel'`) to `button.action.panel` — a rename of one but not the other leaves the toggle pointing at a non-existent panel and the suite stays green. | testing | 0.68 | fix-here (cheap: 2 assertions) |
| 2 | `CHANGELOG.md:3` | Corrected manifest targets the GA `mapToolbar` slot + `requires` capability enforcement, but line 3 still promises 'Freeboard-SK 3.0.0-beta or later'. The PR's own new bullet scopes the fix to a *GA* Freeboard-SK — the tell that beta/GA may differ. If `mapToolbar`/enforcement is GA-only, the documented compatibility window overstates reality. | api-contract | 0.60 | fix-here (narrow the min-version claim once boat verification confirms it) |
| 3 | `plugin/index.spec.ts:1` | New plugin TS (`plugin/index.spec.ts`, `plugin/index.js`) sits outside the lint + strict-null / no-any gates: `ng lint` globs `src/**` only and `tsconfig.strict.json` includes `src/**` only, while vitest runs via esbuild with no type-check. Complies today, but nothing keeps it compliant — a future edit can reintroduce `any`/`!` with CI green. This PR grows the unchecked TS surface (same gap already exists for `tools/gen-mcp-schema`). | maintainability | 0.60 | file-issue (config-scope change; extend `lintFilePatterns` to `plugin/**`) |

All three are new to this PR (none pre-existing). None is a merge blocker.

### Residual risks

- **External Freeboard-SK contract is unverifiable in-repo — cited by all 5 reviewers (strongest cross-reviewer agreement).** `requires=['panels.iframe','buttons']` / `slot='mapToolbar'` are Freeboard-SK's contract; `test:plugin` only pins Skip's own emitted strings, so a self-consistent-but-wrong manifest stays green while the toolbar button silently fails to appear. Correctness rests entirely on the separate boat verification named in the intent — not on CI.
- **Beta vs GA divergence (correctness, api-contract).** One manifest is emitted to every host and can carry only one `slot`. If the beta enforced `requires=[]` and GA enforces the new tokens, the change assumes all supported hosts accept `mapToolbar`; the beta path is untested here. Directly related to finding #2.
- **Disable/re-enable leaks a provider — pre-existing, unchanged by this diff (correctness, testing, maintainability).** `stop()` clears `running` but never unregisters the provider; a start→stop→start re-registers a second `plotterExtensions` provider. The `running` flag masks stale providers for list/get, but duplicate registration is neither tested nor prevented.
- **npm tarball ships the test files (project-standards, api-contract).** `package.json` `files: plugin/**` publishes `plugin/index.spec.ts` + `plugin/vitest.config.ts`. Dead weight only — `main: plugin/index.js` still resolves. Packaging hygiene, not a contract break.

### Testing gaps

- `panel.lifecycle` (`'keepAlive'`) and manifest `version` fields are typed in the spec but never asserted — a regression in either passes CI (lifecycle is finding #1).
- No start→stop→start / double-`start()` lifecycle coverage (ties to the provider-leak risk above).
- No integration test against a real or mocked Freeboard-SK that enforces capability declarations — the exact behavior the change fixes is only boat-verified, never exercised by `test:plugin`.
- Unknown-id `getResource('other')` is asserted only to throw, not to return the documented 'No such plotterExtensions resource' message/shape.

### Verdict

**Ready to merge.** No blocking findings — zero P0/P1/P2; the three surviving items are P3 advisory. Finding #1 is a cheap in-PR hardening (two assertions) worth folding in; #2 is a doc-accuracy tweak to make once boat verification pins the true minimum Freeboard-SK version; #3 is a repo-wide lint/type-gate scope gap worth a tracking issue rather than blocking this PR. The load-bearing correctness of the manifest change is covered by the separate boat verification, not by CI — this suite locks the contract in against regression, which is exactly its intended role.

<sub>Synthesized from 5 independent persona reviews. This is a review comment, not an approval — no code changed, no merge action taken.</sub>

Issue #21, opened 2026-06-30
[SK-02] Object/array value flattening in delta parser fabricates non-spec Signal K child paths


mairas:
**Severity:** medium | **Confidence:** confirmed | **Area:** signalk | **Source:** codebase review 2026-06-30, finding `SK-02`

## Problem

When a Signal K delta update carries an object value (e.g. `navigation.position` with `{latitude, longitude}`), `parseUpdates` in the delta service recursively flattens it into dotted child paths such as `navigation.position.latitude` and `navigation.position.longitude`. These child paths do not exist in the SK specification; the canonical model has `navigation.position` as the path whose value is the compound object. As a result Kip's data browser and path pickers present a namespace that diverges from the server's. Additionally, there is no `Array.isArray` guard anywhere in the flattening logic: because `typeof [] === 'object'`, array-valued paths are iterated via `Object.keys` and exploded into numeric-index child paths (`path.0`, `path.1`, ...). Whether a value is deep-flattened, single-level flattened, or kept whole depends on runtime `FLATTEN_CONFIG` depth and size limits, making the resulting path set non-deterministic from the caller's perspective. Only paths containing the substring `'displays.'` are exempted from flattening.

## Impact

Any widget or consumer that subscribes to a path by its canonical SK name (e.g. `navigation.position`) will not receive data if the service has already split it into synthetic child paths. Conversely, path pickers surface fabricated paths that do not correspond to real SK paths on the server, confusing users and potentially causing silent data loss when a widget is pointed at a synthetic path that disappears across reconnects or server restarts. The array index-explosion behaviour can produce unbounded numbers of synthetic paths for array-valued SK keys. The finding notes the issue is fixable upstream (neutral fork relevance), but it affects Kip deployments including HaLOS marine installations.

## Affected code

- `src/app/core/services/signalk-delta.service.ts:375-412`
- `src/app/core/services/signalk-delta.service.ts:434-477`

## Suggested direction

Add an `Array.isArray` guard before any object-flattening branch so array values are emitted whole rather than index-exploded. Reconsider whether deep object flattening is appropriate for canonical compound-value paths (e.g. `navigation.position`): emitting the object at its canonical path and letting individual widgets read sub-fields is more spec-compliant. If flattening is retained for specific cases, make the whole/flatten decision deterministic — remove the size-based branching in `canFlattenCompletely` or replace it with an explicit allowlist of paths that should be flattened.

## Verification

[...]

PR #266, opened 2026-07-13
feat(config): migrate older configs on import (Kip→Skip onboarding)

### LINKED ISSUE #175: Support importing older-version configs by migrating on import
## Problem

`ProfileService.importProfile` rejects any uploaded config whose `app.configVersion` is not exactly the current `LATEST_APP_CONFIG_VERSION`:

> This configuration is version X, but this version of KIP supports version Y. Re-export it from a current KIP and try again.

So a config file exported from an older Skip (or from upstream KIP, where version-compatible) cannot be imported at all — the user must first load it into a matching-version install and re-export. As the user base grows and configs get shared across devices/versions, this becomes a real friction point.

## Opportunity

The in-place upgrade path already knows how to migrate configs version-by-version: `ConfigurationUpgradeService` (the `runUpgrade(version)` dispatcher + the per-version transforms) plus the trigger in `app.component`. Import could route a **shape-valid but older-version** uploaded config through the same migration chain and then accept it, instead of hard-rejecting.

## Considerations

- Import operates on an **uploaded in-memory blob**, not a server-stored slot, so the migration would need a path that transforms a config object directly rather than the `listConfigs`/`getConfig`/`setConfig` slot flow the boot-time upgrade uses.
- Decide how far back to support (min importable version) and what to do when the gap is unbridgeable.
- Security posture: running migrations over arbitrary uploaded JSON — keep the existing structural `isValidConfigShape` gate and bound the work.
- Storage-namespace note: Skip uses its own namespace (`SERVER_CONFIG_APPID='skip'`, `skip.*` localStorage), so the legacy pre-fork (v9/v10) migration transforms are effectively dead here; an import-migration feature would realistically target the fork's own version line (currently v13) forward.

## Related

- #64 retired the recorder config model and added the v12→v13 app-config migration.

mairas:
## Why
Importing a config from an older version hard-rejected, blocking onboarding — including Kip users migrating to Skip (Kip is Skip's prospective user pool) (#175).

## How
- New `ConfigurationUpgradeService.migrateImportedConfig()` runs a **pure in-memory** upgrade chain (no slot I/O, no reload — the pinned trap) from the v11 floor up to current, reusing the existing per-version transforms.
- accept / migrate / reject matrix with distinct error messages; the import success toast notes when a migration ran.

## Verification
- Full vitest suite green locally (1158 tests); `lint` ✓, `betterer:ci` 183 ✓.
- Spec matrix: current = accept; ≥ v11 = migrate; below-floor / unknown / malformed = reject with distinct messages.

Prerequisite for S2-0 (#21). Closes #175

### DIFF
--- a/src/app/core/components/options/configuration/config.component.spec.ts
+++ b/src/app/core/components/options/configuration/config.component.spec.ts
@@ -143,6 +143,36 @@ describe('SettingsConfigComponent', () => {
     expect(toastMock.show).toHaveBeenCalledWith('boom', 0, false, 'error');
   });
 
+  it('reports a migration in the import success toast when the config was migrated', async () => {
+    profileMock.importProfile.mockResolvedValueOnce(true);
+    const file = new File(
+      [JSON.stringify({ app: { configVersion: 11 }, theme: { themeName: '' }, dashboards: [] })],
+      'c.json', { type: 'application/json' });
+    const input = { files: [file], value: 'x' } as unknown as HTMLInputElement;
+
+    component.uploadJsonConfig({ target: input } as unknown as Event);
+
+    await vi.waitFor(() => expect(toastMock.show).toHaveBeenCalled());
+    const [message, , , severity] = toastMock.show.mock.calls.at(-1) ?? [];
+    expect(message).toMatch(/migrat/i);
+    expect(severity).toBe('success');
+  });
+
+  it('omits migration wording in the success toast when no migration was needed', async () => {
+    profileMock.importProfile.mockResolvedValueOnce(false);
+    const file = new File(
+      [JSON.stringify({ app: { configVersion: 13 }, theme: { themeName: '' }, dashboards: [] })],
+      'c.json', { type: 'application/json' });
+    const input = { files: [file], value: 'x' } as unknown as HTMLInputElement;
+
+    component.uploadJsonConfig({ target: input } as unknown as Event);
+
+    await vi.waitFor(() => expect(toastMock.show).toHaveBeenCalled());
+    const message = toastMock.show.mock.calls.at(-1)?.[0] as string;
+    expect(message).toMatch(/imported/i);
+    expect(message).not.toMatch(/migrat/i);
+  });
+
   it('exports the server-held config, never a stale localStorage copy', () => {
     const settings = TestBed.inject(SettingsService) as unknown as {
       getAppConfig: ReturnType<typeof vi.fn>;
--- a/src/app/core/components/options/configuration/config.component.ts
+++ b/src/app/core/components/options/configuration/config.component.ts
@@ -189,12 +189,19 @@ export class SettingsConfigComponent {
             return;
           }
           try {
-            await this.profileService.importProfile(result.name, parsed);
-            this.toast.show(`Profile "${result.name}" imported`, 1000, true, 'success');
+            const migrated = await this.profileService.importProfile(result.name, parsed);
+            const message = migrated
+              ? `Profile "${result.name}" imported and migrated to the current version`
+              : `Profile "${result.name}" imported`;
+            this.toast.show(message, 1000, true, 'success');
           } catch (err) {
             this.reportError(err);
           }
         });
+    };
+    reader.onerror = () => {
+      this.toast.show('Could not read the selected file.', 0, false, 'error');
+      console.error('File read error:', reader.error);
     };
     reader.readAsText(file);
     input.value = '';
--- a/src/app/core/services/configuration-upgrade.service.spec.ts
+++ b/src/app/core/services/configuration-upgrade.service.spec.ts
@@ -1,8 +1,17 @@
 import { TestBed } from '@angular/core/testing';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
-import { ConfigurationUpgradeService } from './configuration-upgrade.service';
+import { ConfigurationUpgradeService, MIN_IMPORTABLE_APP_CONFIG_VERSION } from './configuration-upgrade.service';
 import { StorageService } from './storage.service';
 import { SettingsService } from './settings.service';
+import { IConfig } from '../interfaces/app-settings.interfaces';
+import { LATEST_APP_CONFIG_VERSION } from '../constants/config-versions.const';
+
+const importConfig = (version: unknown): IConfig =>
+    ({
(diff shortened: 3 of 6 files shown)


mairas:
## Review synthesis — PR #266 `feat(config): migrate older configs on import`

**Scope/intent:** Add `ConfigurationUpgradeService.migrateImportedConfig(config)` — a pure in-memory upgrade chain (no slot I/O, no page reload) from the v11 floor (`MIN_IMPORTABLE_APP_CONFIG_VERSION`) to current (`LATEST_APP_CONFIG_VERSION`), reusing the per-version boot transforms; an accept/migrate/reject matrix with distinct error messages; a migration-aware import success toast. `importProfile` now returns `Promise<boolean>` (sole caller `config.component` updated). Kip→Skip onboarding. Closes #175.

**Reviewers (6 independent personas):** correctness, testing, maintainability, project-standards, api-contract, reliability.

**Synthesis:** 8 raw findings → 1 suppressed (confidence < 0.60), 2 merged as one (cross-reviewer agreement), leaving **3 surviving findings, all P3**. Two persona passes (project-standards, api-contract) returned zero findings.

---

### P3 findings

| # | file:line | issue | reviewer(s) | confidence | route |
|---|---|---|---|---|---|
| 1 | `configuration-upgrade.service.ts:316` | `migrateOneAppVersion`'s version switch (`case 11`/`case 12`) duplicates `runUpgrade`'s per-version dispatch. A future `LATEST` bump that adds a v13→v14 transform to the boot path but forgets the switch case makes **today's current-format exports non-importable** ("could not be migrated from version 13"). Fails loud, but a user-facing regression exactly at a version boundary; coupling is undocumented and unguarded by any test. | correctness + maintainability | 0.60 (×2 reviewers) | fix-in-PR (recommended): single ordered transform table both paths consult, or a chain-coverage test over `[MIN_IMPORTABLE .. LATEST)` |
| 2 | `configuration-upgrade.service.ts:307` | The migration-loop reject branch (`throw "could not be migrated from version N"`) is untested. It's exactly the sad path this feature exists to handle gracefully — a shape-valid config whose dashboards make a reused transform catch-and-return-null would hit it in production — yet ships with no test guarding the wording or the no-write behavior. `migrateOneAppVersion`'s `default: return null` arm is the same untested branch. | testing | 0.78 | add test in PR |
| 3 | `config.component.ts:174` | *(pre-existing)* `FileReader` in `uploadJsonConfig` wires only `reader.onload`, no `reader.onerror`. A raw read failure (locked/unreadable file, I/O error) silently dead-ends the import: no dialog, no toast, `importProfile` never called — the one failure path in this function with zero user feedback. This diff edits this function (toast wording) but leaves the read boundary unguarded. | reliability | 0.78 | fix-here (trivial + related): add `reader.onerror` toast before `readAsText` |

---

### Residual risks (below finding bar, tracked)

- **Version-constant ↔ transform-switch coupling is unenforced** (correctness, maintainability, api-contract, reliability all flag it — the theme behind finding #1). Floor + switch cases must be kept in lockstep with the version constants by hand; only the import path fails loud on a missing case, so boot-vs-import divergence is possible until a chain-coverage test exists.
- **"Purely in memory" import still mutates shared `messages()`/`error()` signals** (maintainability #2, suppressed at 0.55; echoed by correctness, testing, reliability). The reused transforms call `pushMsg`/`pushError`, so a successful import injects boot-flavored `[Upgrade]` lines and a mid-chain failure sets the boot `error()` signal. No unconditional UI binding found today, so impact is nil — but the doc's "purely in memory" contract is misleading, and a stale `error()` could surface if the boot overlay and import ever coexist. Low-cost fix: tighten the doc; durable fix: decouple the signals.
- **Clone-vs-reference inconsistency**: the already-current branch returns the input by reference (`{ config, migrated: false }`) while the migrate path `cloneDeep`s. Benign (sole caller writes then discards), but a future caller mutating the result would corrupt its own parsed object only in the no-op case.
- **Import trusts self-reported `app.configVersion`**: a file claiming v13 but carrying v11-era widget shapes is written unmigrated (parity with main); no post-upgrade schema re-validation.
- **`KIP` in new user-facing error strings** ("newer than this version of KIP supports", "Load it into an older KIP") in the rebranded Skip app. Matches pre-existing copy in this file and `profile.service.ts`; no written standard mandates rebranding these, so a consistency note, not a violation.
- **Single-pass import vs two-reload boot**: identical only because both transforms are pure functions of the config object; a future transform depending on runtime-applied defaults between boot reloads would make single-pass import diverge.

### Testing gaps

- **No end-to-end seam test**: both `config.component` toast tests mock `importProfile`'s boolean, so no test proves a real v11 file drives `migrated=true` through the real `ProfileService` + `ConfigurationUpgradeService` into the migration toast.
- **v11 widget migration never verified through import**: every import/migration test uses `dashboards: []`. The v11 `widget-host2` selector rewrite, w/h/x/y doubling, and `minW/minH` backfill are never asserted on real widget data via the import path — a silent regression would ship. (v12→v13 stripping *is* covered with real widgets.)
- **Migration-loop failure path untested** (finding #2): mid-chain transform returning null.
- **Chain-coverage test missing** (finding #1): nothing walks `MIN_IMPORTABLE..LATEST-1` asserting each step advances the version, so switch/constant drift surfaces only as a runtime import failure.
- **Non-integer / string `configVersion`** (`"11"`, `11.5`) untested; only `undefined` exercises the guard, so the `!Number.isInteger` sub-condition is unverified.
- **Clone isolation** verified only for v11, not v12; `storage.setConfig` rejection/timeout during import untested (must surface an error toast, not a false success); `FileReader` read-error path untested (finding #3); signal-cleanliness untested.
- **Static-only for CI checks**: lint (`no-explicit-any`, unused params), the `strictNullChecks` betterer ratchet, and spec pass/fail were reasoned statically — not run on this live-services host. CI (Node 24) remains the source of truth. New code uses null-safe optional-chaining/`typeof` guards, so ratchet-break risk is low but unverified.

---

**Verdict: ready-to-merge.** No P0/P1/P2 findings; nothing blocks merge. Three P3s, all low-severity: the pre-existing `FileReader` gap (#3) is a trivial, in-scope one-liner worth fixing here; the dispatch-drift coupling (#1) and the untested reject branch (#2) are best closed with a single chain-coverage test that pins the version-constant/switch lockstep and the loop-reject path together. None require a fresh review cycle.

PR #269, opened 2026-07-13
test(setup): drop DI-inert global app-service stubs (#159)

### LINKED ISSUE #159: src/test.ts global app-class stubs are DI-inert under the vitest builder

mairas:
## Why
`src/test.ts` registered global app-class stubs (`SettingsServiceStub`, …) as DI providers, but under `@angular/build:unit-test` those are **DI-inert** — the setup file's classes are different module instances than the app bundle's, so **any** provider there for an app class does nothing (#159). The stubs had silently drifted with no spec failing, and the CLAUDE.md "stub drift" guidance described a mechanism that no longer operates.

## How
- **Empirically confirmed** (against the same builder CI uses): the full suite stays green with every app-class provider removed.
- Deleted **all** global app-class providers — the 9 DI-inert stubs (`SettingsService`, `AuthenticationService`, `SignalKConnectionService`, `ConnectionStateMachine`, `PluginConfigClientService`, `AppNetworkInitService`, the 3 widget host directives) **and** the equally-inert real `UnitsService` provider — plus their dead classes, imports, and the widget-directive override in the TestBed patch. `src/test.ts` **537 → 271 lines**, and it now provides **only** framework tokens, env/DOM shims, and icon registration.
- Rewrote the CLAUDE.md **"Testing reality"** section: real-services-plus-spec-local-fakes model, with the accurate discriminator — `providedIn: 'root'` services resolve real regardless; non-root services (e.g. `UnitsService`) are provided locally by the spec that needs them.

## Verification
- Full suite **1161 pass** locally; `lint` ✓, `betterer:ci` 183 ✓. CI re-confirms on Node 24.

Unblocks the `src/test.ts` chain (A2/D1/E3/B5). Closes #159


mairas:
## Review synthesis — PR #269 (`test(setup): drop DI-inert global app-service stubs`)

Five independent persona reviewers (correctness, testing, maintainability, project-standards, adversarial) ran against the diff. The code change itself — deleting the 9 DI-inert stub providers, dead stub classes/imports, and the widget-directive override branch from `src/test.ts` — drew **no findings**: every reviewer independently traced the deletions as safe (root-provided services resolve to the real root instance regardless of an inert stub; the two non-root cases, `AppNetworkInitService` and the widget host directives, are provided locally by every consumer spec). CI is green and `mergeStateStatus` is CLEAN.

All surviving findings are in the accompanying **CLAUDE.md rewrite**, which is internally inconsistent, carries a now-stale sibling instruction, and mis-states the CI gate. All are doc-only, `pre_existing: false`, and fixable in this PR.

### P2 — should land before merge

| # | File:Line | Finding | Conf | Reviewers |
|---|-----------|---------|------|-----------|
| 1 | CLAUDE.md:83 | Stale "Fork-specific gotchas" line still says the fork's fixes "live in `src/test.ts` stubs — re-apply them", directly contradicting the rewritten line 63 ("do **not** re-add a global app-service stub — it is inert anyway"). Following it re-introduces exactly the stubs this PR deletes. | 0.72 | project-standards |
| 2 | CLAUDE.md:61 | The "different module instances → DI-inert" mechanism is applied uniformly to justify deleting every app-service stub, yet the same sentence lists "the real `UnitsService`" among providers that "actually take effect". `UnitsService` is a bare non-root `@Injectable()` — the same DI category as the deleted `AppNetworkInitService` stub — so the stated mechanism cannot explain why one is kept-effective and DI-identical ones are deleted-inert. The real discriminator is `providedIn:'root'` (deleted stubs targeted root services whose real instance always wins; `UnitsService`/`AppNetworkInitService` are non-root and must be provided explicitly). A dev reasoning from the rule as written will mispredict provider behavior. | 0.62 | maintainability (P2), adversarial (P3), testing (P3), correctness (P3, corroborating) — **4-reviewer agreement** |

### P3 — nice to fix in the same pass

| # | File:Line | Finding | Conf | Reviewers |
|---|-----------|---------|------|-----------|
| 3 | CLAUDE.md:64 | `npm run ci` is documented as "lint + betterer:ci + test", but `package.json`'s `ci` is `lint && betterer:ci && test:headless && test:plugin`. The standalone Freeboard plugin vitest suite (`test:plugin`) is dropped from the description, so a dev running `npm test` to "match CI" silently skips it. | 0.68 | maintainability (0.68), project-standards (0.62) — 2-reviewer agreement |

**Suggested reconciliation (findings 1 & 2 are the same doc telling two contradictory stories):** state the real discriminator once — providers overriding a `providedIn:'root'` service are inert here (component still resolves the real root service), which is why the root-provided stubs were dead; `UnitsService` is kept because it is *not* root-provided and has no fallback, so any spec constructing a non-root consumer (`UnitsService`, `AppNetworkInitService`) must provide it locally. Then either drop `UnitsService` from the "takes effect" list (it is a redundant safety net every consumer spec already overrides locally) or confirm it is load-bearing and explain why the module-instance framing does not apply to it. Fix the stale line 83 to point re-apply guidance at the framework-token/env-shim/icon setup plus spec-local fakes, not "src/test.ts stubs". Fold `test:plugin` into the line-64 CI description.

### Residual risks (unverifiable in this read-only, live-services host — CI on Node 24 remains the authoritative gate)
- The full-suite greenness (1161 tests) that is the *sole* safety proof for the deletions could not be re-run here; static traces support it and the branch's CI is CLEAN, but no in-code assertion locks the DI-inert invariant.
- The DI-inertness itself is an unpinned `@angular/build:unit-test`/vitest bundling property (setup-file module graph separate from specs). A future builder/Angular upgrade that de-duplicates that graph would silently flip *all* `src/test.ts` app-token providers from inert to effective — and with the stubs now deleted there is no fallback. Nothing guards this invariant.
- The doc reverses the prior "local jsdom lacks usable localStorage → Settings/Auth/Storage crash at construction" warning to "Local runs work (~90s)". Since the stubs were always inert, specs already received the real services, so this likely held before too — but local-run behavior and the ~90s figure (which matches the CI job time) could not be confirmed here.
- Whether the retained global `UnitsService` provider is load-bearing or effectively dead (every consumer spec provides it locally; the one exception drives the component prototype without TestBed) is unresolved — this is the same ambiguity underlying finding #2.

### Testing gaps
- No automated guard prevents a future dev from re-adding an inert global app-service stub — it would pass CI while silently doing nothing, re-creating the false-isolation trap this PR removes.
- No test pins the DI-inert invariant the whole model rests on; a tooling upgrade that breaks it surfaces as diffuse suite failures, not one clear signal.
- The mocking model is now convention-only: nothing enforces that a spec constructing a non-root-service consumer supplies the required local provider; omission surfaces only as a runtime NG0201 in whichever spec builds that graph.
- The ~30 spec files whose greenness now depends on real-service construction resilience are neither enumerated nor guarded.

---

**Verdict: ready-with-fixes.** No P0/P1. The code deletion is sound and CI-green; merge is gated only on the CLAUDE.md accuracy fixes — two P2 (a stale instruction that tells devs to re-add exactly what this PR deletes, and a mechanism paragraph that contradicts itself over `UnitsService`) plus one P3 (CI-command description). All are doc edits in a file this PR already rewrote; landing them in the same PR keeps the "read before touching specs" guidance internally consistent.

<sub>Synthesized from 5 persona reviewers; dedup by file+line±3+title with cross-reviewer agreement noted; confidence <0.60 suppressed (1 correctness P3 folded in as corroboration). Do-not-approve/merge orchestrator pass.</sub>

mairas:
**Review findings addressed** (all 3 were in the CLAUDE.md rewrite; the code deletion drew zero findings):
- **P2 mechanism/UnitsService contradiction:** resolved empirically — removing the `UnitsService` provider keeps the suite green (1161), confirming it was DI-inert too. Removed it, so `src/test.ts` now provides **no** app services. The doc now states the accurate discriminator: `providedIn:root` services resolve real regardless; non-root services are provided locally per spec.
- **P2 stale gotcha:** the "re-apply the `src/test.ts` stubs" line is rewritten (no stubs remain).
- **P3 ci command:** corrected to `lint + betterer:ci + test:headless + test:plugin`.

Re-verified: full suite 1161 ✓, lint ✓, betterer 183 ✓.

Issue #257, opened 2026-07-13
Tracking: backlog execution plan — Gate A decisions, lanes & work packets (supersedes #76)


mairas:
Execution tracker for the full open backlog. Supersedes #76 (retired; its settled-decision record remains valid history). Basis: the 2026-07-13 full-backlog triage — every open issue re-validated against main @ c420ebeb with file:line evidence, closure verdicts adversarially verified, lanes proven file-disjoint from per-issue touch-lists.

**Gate A (the single planning interview) was held 2026-07-13.** All blocking decisions below are settled; the ~35 non-blocking design defaults were delivered for async review — objections land as comments here and amend the affected packet. A same-day follow-up interview settled the two flagged external-contract items: the embed URL contract (#216 — now including URL profile addressing) and the units precedence + formula scope (#246 — per-widget unit overrides retire in Phase 2). Stages 1–2 are designed to run without further questions.

**Standing principle (Gate A):** Skip is independent of HaLOS. No packet is gated on HaLOS image contents, deployment state, or any particular boat's server configuration — server-side capabilities are adopted with graceful degradation when absent.

## ✅ Progress — updated 2026-07-16

**Stage 1 is ~85% landed.** Packets merged (issue closed):

- **Lane A:** A1 (#159) ✅, A2 (#106) ✅
- **Lane B:** B4 (#151) ✅, B5 (#85) ✅, B1 (#243 stage 1 — refcount foundation, PR #256) ✅, B2 (#243 release sites, PR #311) ✅, B3 (#254, PR #312) ✅ — **Lane B code complete** (the #243 off-box AIS-churn soak still feeds the #101 rule)
- **Lane C:** C1 (#154) ✅
- **Lane D:** D1 (#79) ✅, D2 (#170) ✅
- **Lane E:** E1 (#190) ✅, E2 (#217) ✅, E3 (#9) ✅, E4 (#202) ✅, E5 (#221) ✅, E6 (#216 — chromeless embed + ephemeral URL profile, PR #305) ✅ — **Lane E complete**
- **Lane F:** F1 (#126+#247) ✅, F3 (#198) ✅; F2 → #179 ✅ #180 ✅ but **#128 (path rename) deferred** (collides with the existing `/settings` route; belongs with the #38 internal naming swap) — F4 (#197) open (boat-verify)
[...]

PR #270, opened 2026-07-14
refactor(settings): retire the *AsO() observable bridges (#79)

### LINKED ISSUE #79: Migrate settings.service consumers to signals; retire the *AsO() observable bridge

mairas:
## What & why
Packet D1 of the #257 backlog. The settings-store rewrite (#17) left two `BehaviorSubject` bridges — `getDefaultUnitsAsO()` and `getNotificationServiceConfigAsO()` — as a transitional API over the `unitDefaults` / `notificationConfig` signals. This retires them: their only two consumers now read the signals directly.

## Changes
- **units.service** — `_defaultUnits` seeds synchronously (matching the former `BehaviorSubject` replay) and tracks changes via `effect()`. Drops the subscription + `OnDestroy`.
- **notifications.service** — config applied synchronously at construction (preserving the same-tick stream-setup contract), then tracked via `effect()`. A reference-equality guard skips the effect's redundant initial run while still applying a config that changed before the first flush.
- **settings.service** — the two subjects, their write-path `.next()` mirroring, and both getters are deleted.
- Specs updated: the notifications spec stubs the config as a signal and flushes effects with `TestBed.tick()`; the settings spec asserts the signals directly.

## Verification
Local `npm run ci` green: lint, betterer:ci (183 issues, unchanged), 1161 vitest, 7 plugin.

Closes #79.

### DIFF
--- a/src/app/core/services/notifications.service.spec.ts
+++ b/src/app/core/services/notifications.service.spec.ts
@@ -1,5 +1,6 @@
 import { TestBed } from '@angular/core/testing';
-import { BehaviorSubject, Subject } from 'rxjs';
+import { signal, WritableSignal } from '@angular/core';
+import { Subject } from 'rxjs';
 import { describe, expect, it, vi } from 'vitest';
 import { INotification, INotificationInfo, NotificationsService } from './notifications.service';
 import { DataService } from './data.service';
@@ -46,7 +47,7 @@ function makeDelta(path: string, value: ISignalKNotification | null): ISignalKDa
 }
 
 describe('NotificationsService', () => {
-  let configSubject: BehaviorSubject<INotificationConfig>;
+  let configSignal: WritableSignal<INotificationConfig>;
   let notificationMsg$: Subject<ISignalKDataValueUpdate>;
   let notificationMeta$: Subject<IMeta>;
   let isReset$: Subject<boolean>;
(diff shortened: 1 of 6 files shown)


mairas:

Standard workflow: 6 reviewers in fresh contexts (correctness, testing, maintainability, project-standards, reliability, adversarial). **4 clean, 2 findings** — both test-coverage gaps, both addressed.

| Sev | Finding | Resolution |
|---|---|---|
| P2 | The new `effect()`-driven unit-defaults tracking had zero coverage (no `units.service.spec.ts`); a broken effect would silently stop propagating a runtime units change. | **Added `units.service.spec.ts`** — asserts the synchronous seed and that a runtime `unitDefaults` change propagates through the effect (`TestBed.tick()`). |
| P3 | The notifications reference-guard seed-repeat branch (`if (config === this._notificationConfig) return`) was never exercised. | **Added a test** — subscribe + count emissions, `TestBed.tick()` with no config change, assert exactly one emission (the seed). Pins the guard against removal. |

No correctness, maintainability, project-standards, reliability, or adversarial findings.

Fixes verified: local `npm run ci` green (lint · betterer 183 · full vitest suite · 7 plugin).

PR #271, opened 2026-07-14
fix(bootstrap): recover in place on connection failure, no /options redirect (#190)

### LINKED ISSUE #190: No-connection bootstrap redirects to the legacy Connectivity/URL options page
## Problem

When bootstrap can't reach the Signal K server, the app redirects to the **Connectivity** options page — the one with a "Signal K URL" text field and a **Connect** button (`/options`, Connectivity tab).

This is legacy Kip behavior from when Kip was a standalone app that connected to an arbitrary, user-entered Signal K server URL (with CORS). **It no longer makes sense** now that connectivity has been simplified: Skip is a Signal K *webapp*, served by the Signal K server and connecting **same-origin only** (no CORS, no user-entered URL — the URL is always `window.location.origin`). There is no scenario where the user should type a Signal K URL, so bouncing them to a URL-entry page on a connection failure is dead-end/confusing.

## Where

`src/app/core/services/app-initNetwork.service.ts`, in the bootstrap `catch`:

- line ~212: `error.status === 0` (network unreachable, after HTTP retries) → `this.router.navigate(['/options'])`
- line ~219: unknown error → `this.router.navigate(['/options'])`

The `/options` route (`src/app/app.routes.ts`) loads the options `TabsComponent`, whose Connectivity tab (`src/app/core/components/options/signalk/signalk.component.*`) hosts the "Signal K URL" + Connect UI.

## Repro

Load the webapp with no reachable Signal K backend (e.g. run the dev server without a server behind it). After the HTTP retries exhaust, the app lands on `/options` showing "Signal K URL: http://localhost:… / Connect" and "Not connected, Not signed in".

## Suggested direction

- On a same-origin connection failure, show a **connection-error / retry** state (on the dashboard, or a minimal reconnect prompt) instead of the URL-entry options page — the user can't fix a same-origin failure by typing a URL.
- Reassess whether the **Connectivity "Signal K URL" config** (and the standalone/CORS connection path it implies) is still needed at all, or should be removed/hidden now that connectivity is same-origin-only.

## Context

Found while visually verifying the navigation UX redesign (#183). Unrelated to that work — filing separately.

mairas:
## What & why
Packet E1 of the #257 backlog. A failed network bootstrap redirected the user to the legacy Connectivity/URL options page — a dead end on a boat kiosk.

## Changes
- **app-initNetwork.service** — drops the two `router.navigate(['/options'])` fallbacks (and the now-unused `Router` injection). The `network-unreachable` / `unknown` bootstrap issues are still emitted and now drive recovery in place; the 401 → re-auth path is untouched.
- **app.component** — surfaces those issues as a persistent **Retry** toast, mirroring the existing `missing-shared-config` / `auth-blocked` recovery effects; the action reloads the app for a clean re-bootstrap.
- Tests: app-initNetwork asserts no navigation on network-unreachable/unknown failure (regression guard); app.component asserts the Retry toast appears for those reasons and not on a clean bootstrap.

`signalk.component` (listed in the packet) needed no change — it has no coupling to the removed redirect.

## Verification
Local `npm run ci` green: lint, betterer:ci (183 issues, regenerated), 1165 vitest, 7 plugin.

Closes #190.

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -5,8 +5,8 @@
 //
 exports[`strictNullChecks`] = {
   value: `{
-    "src/app/app.component.ts:1462426010": [
-      [112, 19, 25, "Object is possibly \'undefined\'.", "255143637"]
+    "src/app/app.component.ts:4055976915": [
+      [113, 19, 25, "Object is possibly \'undefined\'.", "255143637"]
     ],
     "src/app/core/components/dashboard/dashboard.component.ts:3995263170": [
       [121, 12, 22, "Object is possibly \'undefined\'.", "2187394131"],
--- a/src/app/app.component.spec.ts
+++ b/src/app/app.component.spec.ts
@@ -8,6 +8,8 @@ import { DashboardService } from './core/services/dashboard.service';
 import { uiEventService } from './core/services/uiEvent.service';
 import { AppService } from './core/services/app-service';
 import { ChromeVisibilityService } from './core/services/chrome-visibility.service';
+import { ToastService } from './core/services/toast.service';
+import { SettingsService } from './core/services/settings.service';
 
 // Access the component's private surface without widening the class API.
 interface AppComponentHotkeyApi {
@@ -44,6 +46,7 @@ describe('AppComponent', () => {
     hide: ReturnType<typeof vi.fn>;
     pulsePeek: ReturnType<typeof vi.fn>;
   };
+  let toast: { show: ReturnType<typeof vi.fn> };
 
   beforeEach(async () => {
     dashboard = {
@@ -63,6 +66,8 @@ describe('AppComponent', () => {
     };
     appService = { toggleNightMode: vi.fn() };
     chrome = { revealed: signal(false), reveal: vi.fn(), hide: vi.fn(), pulsePeek: vi.fn() };
+    toast = { show: vi.fn().mockReturnValue({ onAction: () => new Subject() }) };
+    appNetworkInitServiceStub.bootstrapIssue$.next({ reason: 'none' });
 
     await TestBed.configureTestingModule({
       imports: [AppComponent],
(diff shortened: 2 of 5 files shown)


mairas:

Standard workflow: 6 reviewers in fresh contexts. **1 clean, 5 findings across two root issues** — all addressed or tracked.

### Root issue A — false "cannot reach" toast on recovery
Flagged independently by **5 of 6 reviewers** (correctness P3, testing P2, maintainability **P1**, reliability P2, adversarial P2). The `network-unreachable` issue was emitted *before* the HTTP retry wait, so a link that recovered during retry still latched the persistent "Cannot reach the server" Retry toast; the recovery branch only logged.

→ **Fixed:** merged the duplicate if-chains so the status-0 issue is emitted *after* `waitForHttpRetryCompletion()` with the outcome-accurate reason — `network-unreachable` only when the link stays down, `unknown` ("startup failed, retry to reload") when HTTP recovered but the bootstrap never completed. Added a recovered-during-retry test.

### Root issue B — Retry mechanism
- **(maintainability P2)** Raw `window.location.reload()` bypassed the app's test-guarded `settings.reloadApp()` seam and left the action untested → **switched to `this.settings.reloadApp()`**; added a spec asserting the Retry action invokes it.
- **(adversarial P1)** A reload can dead-end to a browser error page if the server is still down (no service worker / offline shell). This affects all five `reloadApp()` sites — a pre-existing architectural gap → **filed #272** as the follow-up.

### Also
- **(reliability P3)** `expect(mockRouter.navigate).not.toHaveBeenCalled()` was trivially true once Router was uninjected → **removed the dead Router mock**; the status-0 / unknown tests now assert the positive recovery state (`degraded`).

Fixes verified: local `npm run ci` green (lint · betterer 183 regenerated · full vitest suite · 7 plugin).

PR #274, opened 2026-07-14
fix(sso): suppress framed auto-redirect; break out to top on sign-in (#217)

### LINKED ISSUE #217: fix: Skip's SSO redirect breaks when framed without a session

mairas:
## What & why
Packet E2 of the #257 backlog. Found during the #213 Freeboard-panel spike: when Skip is embedded in an iframe **without** a Signal K session, boot auto-redirected the frame to the SSO login (`/sso?flow=openid_connect…`), which sets `frame-ancestors 'none'` — the browser blocks the navigation and the panel renders broken.

## Changes (all in `SsoRedirectService`)
- **Framed detection** (`isFramed()`, `window.self !== window.top`, fail-safe to framed on cross-origin access error).
- **`attemptAutoRedirect` returns a new `'framed'` outcome** when embedded, spending no budget. The existing `attemptCookieRedirect` treats anything ≠ `'redirected'` as blocked, so the bootstrap surfaces the existing **auth-blocked recovery** ("Sign in" toast) instead of a broken framed redirect — no changes needed in app-init or app.component.
- **`manualSignIn` breaks out to the top window** (`navigate` targets `window.top` when framed), so the explicit sign-in loads the login at the top level. `window.top` navigation is permitted even cross-origin.
- **Unframed behavior and the redirect budget are untouched.**

## Tests
Added: framed `attemptAutoRedirect` returns `'framed'` with no navigation and no budget spend; framed `manualSignIn` navigates the (swapped) `window.top`. Existing budget/fallback/fail-closed tests unchanged (jsdom is unframed by default).

## Verification
Local `npm run ci` green (lint · betterer 183 · full vitest suite · 7 plugin).

Closes #217.

### DIFF
--- a/src/app/core/services/app-initNetwork.service.spec.ts
+++ b/src/app/core/services/app-initNetwork.service.spec.ts
@@ -248,6 +248,16 @@ describe('AppNetworkInitService', () => {
             expect(latestIssue()).toEqual({ reason: 'auth-blocked', cause: 'budget-exhausted' });
         });
 
+        it('framed boot: framed outcome -> auth-blocked/sign-in-required even if a stale shared budget is exhausted (#217)', () => {
+            mockSsoRedirect.attemptAutoRedirect.mockReturnValue('framed');
+            mockSsoRedirect.isBudgetExhausted.mockReturnValue(true);
+
+            expect(handleCookieAuth({ status: 'notLoggedIn', authenticationRequired: true, oidcAutoLogin: true })).toBe('auth-blocked');
+
+            // The framed prompt must never inherit a stale standalone budget's 'budget-exhausted' wording.
+            expect(latestIssue()).toEqual({ reason: 'auth-blocked', cause: 'sign-in-required' });
+        });
+
         it('oidcAutoLogin disabled: does not auto-redirect, surfaces sign-in-required', () => {
             const status: ILoginStatus = { status: 'notLoggedIn', authenticationRequired: true, oidcAutoLogin: false, oidcEnabled: true };
 
--- a/src/app/core/services/app-initNetwork.service.ts
+++ b/src/app/core/services/app-initNetwork.service.ts
@@ -89,13 +89,17 @@ export class AppNetworkInitService implements OnDestroy {
    * auth-blocked recovery state (budget exhausted, or a manual sign-in is required).
    */
   private attemptCookieRedirect(status: ILoginStatus | null): 'redirecting' | 'blocked' {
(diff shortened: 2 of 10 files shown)


mairas:

Standard workflow: 6 reviewers in fresh contexts (correctness, testing, maintainability, project-standards, security, adversarial). 3 clean (correctness, project-standards, security); 5 findings, **all addressed**.

### Fixed
- **Real bug — framed prompt mislabeled "budget-exhausted" (adversarial P2):** the redirect budget is shared per-origin-per-tab in sessionStorage, so a stale standalone budget could make the framed sign-in prompt read *"Automatic sign-in did not complete"* instead of *"Sign in to Signal K."* → `attemptCookieRedirect` now derives the cause from the `framed` outcome and forces `sign-in-required` when framed, regardless of the stale budget. App-init test added.
- **DRY (maintainability P2):** `isFramed()` byte-for-byte duplicated the tested `isEmbeddedInIframe()` in uiEvent.service. → extracted to **`core/utils/iframe.util.ts`** (with its tests); sso-redirect, uiEvent, and widget-login now share the one canonical detector, and sso-redirect no longer pulls uiEvent's `screenfull`/`no-sleep` into the bootstrap path.
- **Frame-bust (adversarial P3):** WidgetLoginComponent auto-called `manualSignIn()` on init; now that it breaks out to `window.top`, a framed `/login` load could auto-navigate the embedder with no gesture. → gated the auto-redirect on `!isEmbeddedInIframe()`.
- **Testing (P2×2, P3):** framed detection was only ever mocked → added a **real-`isFramed`** test (swaps `window.top`), an end-to-end framed→`sign-in-required` app-init test, and a null-`window.top` guard test.

Fixes verified: local `npm run ci` green (lint · betterer 183 · full vitest suite · 7 plugin).

Issue #246, opened 2026-07-13
Adopt Signal K unit-preferences API for display units + symbols (server-sourced units)


mairas:
## Motivation

Skip currently owns its **entire** unit story client-side: a hand-authored
conversion table (`_conversionList`) and a `js-quantities`-backed conversion engine
(`unitConversionFunctions`) in `src/app/core/services/units.service.ts`, plus a
per-user "default unit per group" preference stored in Skip's own config
(`SettingsService.unitDefaults`). That table is the source of the label
inconsistencies tracked in #245, and its measure strings double as both conversion
keys and on-screen labels.

Signal K now offers a server-side **unit-preferences** layer
(`signalk-units-preference`; the guide lives in the `signalk-server` repo) that can
supply, per path, the user's preferred display unit, its abbreviation/symbol, and the
conversion — as metadata. Adopting it would let Skip stop maintaining its own unit
table and per-widget symbol hacks, and would make Skip honour a boat-wide unit
preference set once on the server instead of separately in every client.

This is the larger follow-up flagged in #245. #245 stays the near-term fix (make
Skip's own labels consistent); this issue is the strategic direction.

## What Signal K provides

Source: [signalk-server unit-preferences guide](https://github.com/SignalK/signalk-server/blob/master/docs/guides/unitpreferences.md).

- Per path, the server adds a `displayUnits` object to the path's **metadata** with:
  `category` (e.g. `"speed"`), `targetUnit` (e.g. `"kn"`), `symbol` (the display
  abbreviation), `formula` (a math.js expression converting the SI value → target,
  e.g. `"value * 1.94384"`), `inverseFormula`, and optional `displayFormat`
  (e.g. `"0.0"`).
- Delivery: embedded in path meta — over the WebSocket via a `sendMeta=all`
  subscription, or via REST `.../meta`. **No separate per-path call needed.**
- Config REST surface (for a settings UI): `GET /signalk/v1/unitpreferences/`
  `config` · `categories` · `definitions` · `presets` · `presets/:name` · `active`.
- Built-in presets: Metric, Imperial (US/UK), Nautical (Metric/Imperial US/UK), with
  per-user settings and per-path overrides layered on top — all server-side.

## Why this fits Skip well

- **Transport already exists.** Skip already subscribes with `sendMeta=all`
  (`signalk-delta.service.ts:79`) and already parses per-path meta (`parseSkMeta`,
  ~line 486). Receiving `displayUnits` is mostly a matter of capturing the new field
[...]

PR #279, opened 2026-07-14
feat(units): render consistent unit symbols everywhere via a central resolver (#245)

### LINKED ISSUE #245: Widget unit labels: use consistent, valid abbreviations everywhere (no spelled-out words; kph→km/h)
## Summary

Widget displays are inconsistent about how they label units. Some show a proper
abbreviation or symbol (`m`, `m/s`, `mph`), while others show the unit spelled out
(`knots`, `celsius`, `fahrenheit`, `liter`, `gallon`, `feet`, `Minutes`). They
should all use short, standard abbreviations/symbols.

A second, related defect: some of the abbreviations that *are* used aren't valid —
`kph` is not a standard abbreviation for kilometres per hour (it should be `km/h`).
So the fix is twofold: (1) abbreviate the spelled-out ones, and (2) correct the
wrong/nonstandard ones.

## Where it comes from

The label a widget draws next to a value is the path's `convertUnitTo` string,
rendered verbatim. E.g. `widget-numeric` draws it directly:

- `src/app/widgets/widget-numeric/widget-numeric.component.ts:290` draws `unit`,
  where `unit = cfg.paths['numericPath'].convertUnitTo`.

`convertUnitTo` is one of the `measure` values from the conversion table in
`src/app/core/services/units.service.ts` (`_conversionList`, ~line 113). That
`measure` string is doing double duty: it's both the **lookup key** into
`unitConversionFunctions` **and** the **on-screen label**. Because the table was
authored inconsistently, some measures are symbols and some are English words, and
that inconsistency shows up directly in every widget that renders a unit.

## Concrete offenders (in `_conversionList`)

| Group | Current `measure` | Should read like |
|---|---|---|
| Speed | `knots` | `kn` |
| Speed | `kph` (invalid) | `km/h` |
| Temperature | `celsius` | `°C` |
| Temperature | `fahrenheit` | `°F` |
| Volume | `liter` | `L` |
| Volume | `gallon` | `gal` |
| Volume | `m3` | `m³` |
| Length | `fathom` | `ftm` |
| Length | `feet` | `ft` |
| Length | `inch` | `in` |
| Time | `Minutes` | `min` |
| Time | `Hours` | `h` |
| Time | `Days` | `d` |

(Time also has a capitalisation inconsistency: `Minutes`/`Hours`/`Days` are
capitalised while `s` for seconds is not.) While sweeping, the other abbreviations
should be sanity-checked for correctness too — e.g. `g/min` / `g/h` use `g` for
gallons, which collides with grams.

The exact target strings are a call for whoever picks this up; the list above is the
set of places that are inconsistent, not a prescribed spelling.

## Design consideration (not a prescription)

`measure` is also the value persisted in saved dashboard configs
(`convertUnitTo: "celsius"`), and it's the key `convertToUnit()` looks up. So simply
renaming the `measure` strings would (a) invalidate existing user configs and (b)
require touching every conversion-function key in lockstep. A less disruptive route
is to add a separate short **display symbol** to the `IUnit` shape and render *that*
in widgets, leaving `measure` as the stable internal key. The `skBaseUnits` array in
the same file already carries a `display` field (`m³`, `°`, `lx`, …) for the SK base
units, so there's precedent for a display-vs-key split. Whoever implements decides
between rename-with-migration and add-a-display-field.

Note the symbol is *already* hand-mapped ad hoc in at least one widget —
`src/app/widgets/widget-bms/widget-bms.component.ts:1234`
(`displayUnit === 'celsius' ? '°C' : '°F'`). A single source of symbols would let
that hack, and any like it, go away.

## Related: could Signal K supply the symbols?

Partly, and worth recording for a possible larger follow-up. Core SK `meta.units`
only carries the **SI base unit** (`m/s`, `K`, `Pa`), no preferred unit and no
symbol beyond that code. But the Signal K **unit-preferences** layer
(`signalk-units-preference`; its guide now lives in the `signalk-server` repo) adds a
per-path `displayUnits` object over the metadata API containing exactly this:
`category`, `targetUnit`, **`symbol`** (the display abbreviation), `formula` /
`inverseFormula`, and `displayFormat`. So the server *can* hand a client both the
preferred target unit and its abbreviation.

Adopting that is a **much bigger, separate change** than this issue: Skip currently
owns its whole conversion engine (`unitConversionFunctions` via `js-quantities`) plus
per-user unit defaults, and would have to defer conversion + formatting + symbols to
the server API and degrade gracefully when the plugin isn't installed. Out of scope
here — this issue is just "make Skip's own labels consistent and correct." Sourcing
symbols/conversions from SK should be its own issue if we want to pursue it.

## Acceptance criteria

- Every unit shown in a widget uses a short abbreviation or symbol — no
  spelled-out English words.
- `kph` no longer appears; kilometres per hour reads as `km/h`.
- Existing saved dashboards keep working (no silent loss of a widget's configured
  unit).

mairas:
## What & why
Packet G3 of the #257 backlog. Widget unit labels were inconsistent — some symbols, some spelled-out words (`knots`, `celsius`, `liter`, `feet`, `Minutes`), some invalid (`kph`). The `measure` string did double duty as both the conversion lookup key and the on-screen label.

## Changes
- **`IUnit` gains an optional display `symbol`** (measure stays the internal key → **no config migration**) and a central **`UnitsService.getUnitDisplaySymbol(measure)`** resolver. Symbols added for the offenders (`knots→kn`, `kph→km/h`, `celsius→°C`, `fahrenheit→°F`, `liter→L`, `m3→m³`, `gallon→gal`, `fathom→ftm`, `feet→ft`, `inch→in`, `Minutes→min`/`Hours→h`/`Days→d`, the `g`-for-gallons collisions → `gal/*`, `ohm→Ω`, `percent→%`, and the lat/lon minute/second marks `lat ′` / `lon ″`).
- **Every on-screen unit label routes through the resolver.** The two ad-hoc `getUnitsLabel` switches (data-chart, history dialog) and the 5 celsius ternaries (alternator/bms/inverter/solar-charger/charger) are deleted; the gauge / simple-linear / windsteer / racer-line / numeric render sites resolve symbols.
- **Conversion-*key* usages are left untouched** — `convertToUnit`, `getHighlights`, cache signatures, and the latitude / `feet` comparisons still use the raw measure. Saved dashboards keep their configured units.

This is also the seam #246-P1 (S2-2) swaps server unit-preference sources into.

## Verification
Local `npm run ci` green (lint · betterer 183 regenerated · 1177 vitest · 7 plugin).
**Boat-verify:** rendered unit-label spot-check across widgets — batched into the Stage-1 session.

Closes #245.

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -219,8 +219,8 @@ exports[`strictNullChecks`] = {
       [44, 9, 9, "Type \'null\' is not assignable to type \'string\'.", "2966443234"],
       [148, 40, 27, "Object is possibly \'undefined\'.", "3506675417"]
     ],
-    "src/app/widgets/widget-gauge-ng-compass/widget-gauge-ng-compass.component.ts:1976286112": [
-      [235, 30, 9, "Argument of type \'string | undefined\' is not assignable to parameter of type \'string\'.\\n  Type \'undefined\' is not assignable to type \'string\'.", "4119212820"]
+    "src/app/widgets/widget-gauge-ng-compass/widget-gauge-ng-compass.component.ts:3154966941": [
+      [237, 30, 9, "Argument of type \'string | undefined\' is not assignable to parameter of type \'string\'.\\n  Type \'undefined\' is not assignable to type \'string\'.", "4119212820"]
     ],
     "src/app/widgets/widget-iframe/widget-iframe.component.ts:2602616758": [
       [30, 4, 9, "Type \'null\' is not assignable to type \'string | undefined\'.", "2966443234"]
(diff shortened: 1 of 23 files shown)


mairas:

Standard workflow: 5 reviewers in fresh contexts. **All 4 findings addressed.**

### Fixed
- **Incomplete sweep (maintainability P2 + project-standards P2):** the **Classic Steel gauge** (`gauge-steel`) was missed — it still rendered the raw spelled-out measure. → routed its `unitString` through `getUnitDisplaySymbol`, keeping the `units` input raw for `convertToUnit` so zone conversion is untouched. Spec stub updated.
- **`getUnitDisplaySymbol` untested (testing P2):** the central resolver every render site depends on had no direct tests (widget specs stub it). → added direct tests to `units.service.spec.ts` covering the symbol mappings, the no-symbol fallback (`mph`→`mph`), unknown-key passthrough, and the null/undefined/empty guard.
- **simple-linear `abr` mode (correctness / maintainability P3):** `substring(0,1)` of the resolved symbol collapsed `°C`/`°F` to a bare `°`. → skip truncation for short symbols (≤2 chars) so C/F (and `kn`) stay distinct.
- **Numeric `percentraw` double-`%` (adversarial P3):** the value already appends `%` via decorations, and `percentraw` now resolves a `%` symbol too → `45% %`. → added `percentraw` to the numeric label-exclusion list (mirroring `percent`).

The display-vs-key split held up: no conversion key was routed through the resolver.

Fixes verified: local `npm run ci` green (lint · betterer 183 · 1180 vitest · 7 plugin).

PR #280, opened 2026-07-14
Replace connection-status magic-number operation codes with typed per-layer enums

### LINKED ISSUE #9: [ARCH-01] Three divergent `operation: number` enumerations with a dead UI branch in connection status
**Severity:** medium | **Confidence:** confirmed | **Area:** architecture | **Source:** codebase review 2026-06-30, finding `ARCH-01`

## Problem

Connection status is represented by a raw `operation: number` field on three separate interfaces (`IEndpointStatus`, `IStreamStatus`, and `IConnectionStatus`) that each assign subtly different meanings to the same numeric values. `IEndpointStatus` defines 4 as "Resetting", `IStreamStatus` uses 4 as "closing" (internal, never pushed to observers), and `IConnectionStatus`'s `stateToOperationCode()` explicitly labelled `// Legacy operation codes for compatibility` never emits 4 at all — instead it uses 5 for `PermanentFailure`. As a result, `AppComponent.displayConnectionsStatusNotification()` contains a `case 4:` branch that is dead code (the only producer, `stateToOperationCode()`, cannot reach it), while the actual 4=Resetting state from the connection service has no corresponding UI notification. Consumers in the delta, storage, and authentication services all branch on `operation === 2` sourced from `IEndpointStatus`, a different enum than the one the UI reads.

## Impact

The overlapping-but-inconsistent numeric vocabulary obscures the real connection contract, making it hard to reason about which enum a value comes from at any given call site. The dead `case 4:` in the UI silently swallows a notification that was presumably intended, while the 4=Resetting state from the connection service goes unhandled. Any future change to connection state handling risks hitting the wrong numeric constant. This is an internal consistency issue with neutral relevance to SKip/HaLOS-specific deployment, but it argues for an upstream PR to prevent the confusion from compounding.

## Affected code

- `src/app/core/services/signalk-connection.service.ts:38-47`
- `src/app/core/services/signalk-delta.service.ts:24-28`
- `src/app/core/services/connection-state-machine.service.ts:23-30`
- `src/app/core/services/connection-state-machine.service.ts:315-336`
- `src/app/app.component.ts:302-325`

## Suggested direction

Collapse to the single `ConnectionState` enum that already exists in `connection-state-machine.service.ts` as the sole source of truth. Derive any legacy numeric codes from it in one mapping function, and remove the per-interface numeric `operation` duplicates. Delete the dead `case 4:` in `displayConnectionsStatusNotification()`. Replace all remaining `operation === 2` numeric checks across delta, storage, and authentication services with references to the canonical enum value. This cleanup is well-suited for an upstream PR rather than a fork-local change.

## Verification

Verified against the codebase: all three divergent interfaces, the dead `case 4:` UI branch, and the cross-enum `operation === 2` consumers were confirmed to exist exactly as described (confidence: confirmed).

mairas:
## Motivation

The three connection-status interfaces — `IConnectionStatus`, `IEndpointStatus`, `IStreamStatus` — each carried an untyped `operation: number` field whose meaning lived only in a stale `0–4` doc comment. Consumers compared against bare magic numbers (`operation === 2`), and the app's connection-status toast switch dispatched on those codes. This is packet **E3** of the #257 backlog (issue #9): give each connection layer a typed vocabulary.

## Approach

Per-layer typed enums, because the three fields describe **different** layers:

- **`IConnectionStatus`** (overall lifecycle): the `operation` field and the `stateToOperationCode()` mapping are **deleted**. `AppComponent.displayConnectionsStatusNotification` now switches on `connectionStatus.state` (the existing `ConnectionState` enum) directly.
- **`IEndpointStatus`** (HTTP discovery): `operation: number` → `state: EndpointStatus` (`Disconnected` / `Connecting` / `Connected` / `Error`).
- **`IStreamStatus`** (WebSocket delta stream): `operation: number` → `state: StreamStatus` (adds `Closing` for the internal `closeWS` state).
- Consumers in `signalk-delta.service` and `storage.service` now key off `EndpointStatus.Connected` instead of `=== 2`.

## Behavior preservation

This is a typing/consistency refactor. Every toast (message, duration, silent flag, level), the delta `WsServiceUrl` gating, and the storage-readiness gate map to the **same observable outcomes** as the numeric codes. The two intentional deltas:

1. The dead `case 4` (a "Resetting" info toast) is removed — `stateToOperationCode` never produced code 4 for `IConnectionStatus`, so it was unreachable.
2. `PermanentFailure` now has an explicit named branch (was numeric code 5).

Note: `operation === 2` was **not** dead — `processEndpointResponse` sets `operation: 2` on HTTP-discovery success, so the `EndpointStatus.Connected` mapping preserves the live delta `WsServiceUrl` capture.

## Tests

Adds `AppComponent` notification-switch coverage (the state→toast mapping was previously untested), including the bootstrap-silent path, and repoints the endpoint/stream status fixtures across the affected specs.

## Verification

Local `npm run ci` equivalent: lint ✓ · betterer 183 (unchanged, baseline regenerated for content hashes) ✓ · 1171 unit tests ✓ · 7 plugin tests ✓.

Reviewed via the standard multi-persona pass (correctness, testing, maintainability, project-standards, reliability, adversarial): 5/6 lenses clean; one confirmed P3 (bootstrap-silent coverage) addressed; one P2 refuted as a nitpick (100% switch-branch coverage already).

Closes #9

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -5,7 +5,7 @@
 //
 exports[`strictNullChecks`] = {
   value: `{
-    "src/app/app.component.ts:2963952310": [
+    "src/app/app.component.ts:281943337": [
       [111, 19, 25, "Object is possibly \'undefined\'.", "255143637"]
     ],
     "src/app/core/components/dashboard/dashboard.component.ts:3995263170": [
--- a/src/app/app.component.spec.ts
+++ b/src/app/app.component.spec.ts
@@ -3,6 +3,7 @@ import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { BehaviorSubject, Subject } from 'rxjs';
 import { signal } from '@angular/core';
 import { AppComponent } from './app.component';
+import { ConnectionState, IConnectionStatus } from './core/services/connection-state-machine.service';
 import { AppNetworkInitService } from './core/services/app-initNetwork.service';
 import { DashboardService } from './core/services/dashboard.service';
 import { uiEventService } from './core/services/uiEvent.service';
@@ -16,6 +17,10 @@ interface AppComponentHotkeyApi {
   dashboardVisible: { set: (v: boolean) => void };
   handleKeyDown: (key: string, event: { shiftKey: boolean }) => void;
   onShellPointerDown: (event: { target: { closest: (sel: string) => unknown } }) => void;
+}
+
+interface AppComponentNotifyApi {
+  displayConnectionsStatusNotification: (status: IConnectionStatus) => void;
 }
 
 describe('AppComponent', () => {
@@ -68,6 +73,7 @@ describe('AppComponent', () => {
     chrome = { revealed: signal(false), reveal: vi.fn(), hide: vi.fn(), pulsePeek: vi.fn() };
     toast = { show: vi.fn().mockReturnValue({ onAction: () => new Subject() }) };
     appNetworkInitServiceStub.bootstrapIssue$.next({ reason: 'none' });
+    appNetworkInitServiceStub.bootstrapStatus$.next('ready');
 
     await TestBed.configureTestingModule({
       imports: [AppComponent],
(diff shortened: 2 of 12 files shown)


mairas:

Standard mairas-workflow fan-out over the branch diff (`main...refactor/connection-status-enums`), each persona adversarially verifying its own findings. Core claim under test: **behavior preservation**.

**Lenses run:** correctness · testing · maintainability · project-standards · reliability · adversarial (connection-critical + ≥50 changed lines).

**Result: 5 of 6 lenses clean.** Correctness, maintainability, project-standards, reliability, and adversarial each returned zero findings. Correctness independently traced all 10 `ConnectionState` values through the old `stateToOperationCode` → old switch vs. the new `state` switch and confirmed identical `toast()` calls; adversarial could not construct an observable divergence.

**Findings (both from the testing lens):**

| # | Severity | Verdict | Disposition |
|---|----------|---------|-------------|
| 1 | P3 | **CONFIRMED** | Fixed |
| 2 | P2 | **REFUTED** → nitpick | Not actioned |

1. **`silentDuringBootstrap` only asserted in the ready direction** — the silent-during-bootstrap path was untested, so hardcoding `false` would keep every test green while regressing bootstrap-quiet behavior this refactor claims to preserve. **Addressed:** added a test driving `bootstrapStatus = 'starting'` and asserting `silent = true` for a retry state and `PermanentFailure`, plus a `beforeEach` reset so the shared stub can't leak between tests.

2. **"6/10 `ConnectionState` values have no per-value test."** Facts confirmed, but refuted at severity: every behavioral *branch* of the switch (transient / no-toast / warn / permanent / default) is exercised, so branch coverage is 100%; the failure scenario is a hypothetical future edit (defense-in-depth), and the change moves this handler from 0 → 6 tests. Nitpick tier under the #257 merge gate — not actioned.

Verification after the fix: lint ✓ · betterer 183 ✓ · app.component.spec 22/22 ✓ (full suite 1171 + 7 plugin green pre-fix).

PR #282, opened 2026-07-14
Hydrate all app-config holders before persist-on-missing writes (#170)

### LINKED ISSUE #170: pushSettings persist-on-missing writes serialize mid-hydration, reverting later-hydrated fields

mairas:
## Motivation

`SettingsService.pushSettings()` fired its persist-on-missing bootstrap writes (for an absent `autoNightMode` / `redNightMode` / `nightModeBrightness`) **mid-hydration**, interleaved with field hydration. Each write serializes the full `IAppConfig` from the reactive holders, so any holder not yet hydrated was captured at its default and **reverted on the server** (#170):

- a stored profile missing `autoNightMode` but with a custom `browserTabTitle` had the title silently reverted to `'Skip'` on the next reload; and
- the same defect one field over — a missing `autoNightMode` also reverted a stored `redNightMode` / `nightModeBrightness` to their init defaults, because those blocks ran *after* the write.

Only sparse configs are affected (persist-on-missing fires only when a night-mode field is absent), which is why it went unnoticed.

## Approach

Hydrate **every** app-config holder (`browserTabTitle` + the three night-mode fields; `dashboards` too, though it isn't part of the app blob) from stored-or-default state **first**, then persist **once** from the fully-hydrated blob if any night-mode field was absent — coalescing what were up to three separate mid-hydration writes into one correct write. This closes the revert class for every field rather than just `browserTabTitle`.

## Tests

The parametrized persist-on-missing loop now asserts, for each single-absent permutation, that every *other* app-blob field (the sibling night-mode fields + `browserTabTitle`) rides along at its loaded value — locking the whole revert class symmetrically. The all-three-absent test asserts the single coalesced write carries every default.

## Verification

Local gate: lint ✓ · betterer 183 (unchanged) ✓ · full unit suite ✓ · plugin ✓.

Reviewed via the standard multi-persona pass. The first review caught that an initial reorder-only fix left the night-mode fields reverting each other; this revision (hydrate-all-then-coalesce) closes that, and a re-review confirmed the code correct with the test coverage extended symmetrically.

Closes #170

### DIFF
--- a/src/app/core/services/settings.service.spec.ts
+++ b/src/app/core/services/settings.service.spec.ts
@@ -391,11 +391,12 @@ describe('SettingsService — hydration (pushSettings) characterization', () => 
   });
 
   // The persist-on-missing bootstrap fields: exactly these three (widgetHistoryDisabled was
-  // removed by #157). Each missing field fires its own whole-app patch during startup.
-  const persistOnMissing: { field: string; bootstrapDefault: unknown; read: (s: SettingsService) => unknown }[] = [
-    { field: 'autoNightMode', bootstrapDefault: false, read: (s) => s.getAutoNightMode() },
-    { field: 'redNightMode', bootstrapDefault: false, read: (s) => s.getRedNightMode() },
-    { field: 'nightModeBrightness', bootstrapDefault: 0.2, read: (s) => s.getNightModeBrightness() }
+  // removed by #157). A single missing field fires one whole-app patch built from fully-hydrated state;
+  // `loaded` is that field's value in the DEFAULT_LOADED fixture, used to pin sibling preservation.
+  const persistOnMissing: { field: string; bootstrapDefault: unknown; loaded: unknown; read: (s: SettingsService) => unknown }[] = [
+    { field: 'autoNightMode', bootstrapDefault: false, loaded: true, read: (s) => s.getAutoNightMode() },
+    { field: 'redNightMode', bootstrapDefault: false, loaded: true, read: (s) => s.getRedNightMode() },
+    { field: 'nightModeBrightness', bootstrapDefault: 0.2, loaded: 0.65, read: (s) => s.getNightModeBrightness() }
   ];
 
   for (const { field, bootstrapDefault, read } of persistOnMissing) {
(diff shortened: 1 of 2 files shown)


mairas:

Standard mairas-workflow fan-out, then a focused re-review after revision. Core claim under test: the hydrate-then-persist reorder is behavior-preserving apart from the fix, and closes the whole #170 revert class.

**Round 1** (correctness · testing · maintainability · project-standards): **1 confirmed P2** — the initial reorder only hoisted `browserTabTitle`/`dashboards`; the three night-mode blocks stayed interleaved with each other's persists, so a missing `autoNightMode` still reverted a stored `redNightMode`/`nightModeBrightness` (the identical bug, one field over), and the test only asserted `browserTabTitle` — false coverage. **Addressed:** hydrate all holders, then persist once (coalesced).

**Round 2** (correctness · testing · project-standards) on the revision: **2 confirmed P3, both "shipped code is correct — coverage gap"** — sibling-preservation was asserted only for the `autoNightMode`-absent permutation. **Addressed:** the parametrized loop now pins sibling + title preservation for all three single-absent permutations. No production defect survived.

Net behavior change: up-to-three separate bootstrap writes → one coalesced write from fully-hydrated state; every app-blob field preserved for every sparse-config permutation.

PR #283, opened 2026-07-14
Reveal the toolbar by hovering the top peek band (#202)

### LINKED ISSUE #202: feat(nav): hover the peek strip to expose the toolbar, auto-close after leaving
## Request

On mouse/desktop, moving the cursor over the top **peek area** should fully expose the auto-hiding navigation toolbar — no click required. When the toolbar was opened this way, it should auto-close again once the cursor has been away from it for more than *N* seconds.

## Current behavior

The toolbar's visibility lives in `ChromeVisibilityService` (`src/app/core/services/chrome-visibility.service.ts`) with two independent bits:

- `revealed` — full toolbar shown; auto-hides after `CHROME_IDLE_HIDE_MS` (4000 ms) unless hiding is suppressed.
- `peeking` — a transient 8 px edge cue (`.peek-strip`) flashed on input, retracting after `CHROME_PEEK_MS` (1300 ms).

To reveal today you **click/tap** the peek strip, and the strip is only pointer-targetable *during* the transient peek flash (`.toolbar-host.peeking:not(.revealed) .peek-strip { pointer-events: auto }` in `toolbar.component.scss`). Hovering the *revealed* toolbar suppresses auto-hide (`onPointerEnter` → `suppressHide()`); leaving it re-arms the 4 s idle-hide (`onPointerLeave` → `allowHide()` → `scheduleHide(CHROME_IDLE_HIDE_MS)`).

## What's actually new here

The auto-close-after-leaving is **already implemented** — leaving the revealed toolbar re-arms the 4 s idle-hide. So the genuinely new work is **hover-to-expose**: making the top edge a persistent hover hotzone that calls `reveal()`, rather than requiring a click on the transiently-clickable peek strip.

## The N question

Reuse the existing **`CHROME_IDLE_HIDE_MS` (4 s)**. One timeout is simpler than two, and it's already the "tune on real hardware later" default, so hover-exposed and deliberately-revealed toolbars retract on the same clock.

Introduce a *separate, shorter* N (~2 s) only if testing shows accidental hover-reveals are annoying — a toolbar you brushed open by accident arguably deserves to retract faster than one you deliberately opened. Decision: ship the shared 4 s first; split the constant only if the accidental-hover case proves real. Final values to be tuned on real hardware, consistent with the existing timing comment in the service.

## Design notes

- **Hover intent / flicker.** An 8 px top-edge hotzone firing on any cursor pass near the top will feel twitchy. Require a short dwell (~150–250 ms) in the peek zone before revealing.
- **Touch devices have no hover.** This is a mouse/desktop enhancement; the existing tap-the-peek-strip path stays for touch. Gate the hover trigger on a hover-capable pointer.
- **Don't double-close.** Hover-reveal must go through the same `revealed`/suppress-hide/idle-hide machinery so leaving cleanly re-arms one timer — no competing close paths.
- **Accessibility.** Keep the current keyboard/focus reveal on `.peek-strip` intact; hover is additive.

## Acceptance criteria

- On a hover-capable pointer, dwelling the cursor in the top peek zone fully reveals the toolbar without a click.
- After the toolbar is hover-exposed, moving the cursor away and keeping it away for N seconds auto-closes it; returning within N keeps it open.
- Touch tap-to-reveal is unchanged; keyboard focus reveal is unchanged.
- No regression to auto-hide suppression while a dialog/menu is open or focus is inside the toolbar.

mairas:
## Motivation

The auto-hiding navigation toolbar could only be re-exposed by clicking the transient peek strip (or via keyboard/touch). On a mouse/desktop, moving the cursor to the top edge should expose it without a click (#202). The auto-close-after-leaving already exists (leaving the revealed toolbar re-arms the 4s idle-hide), so the genuinely new work is hover-to-expose.

## Approach

`ToolbarComponent` registers a **passive** `document` `pointermove` listener. On each move it:

- returns early for `pointerType === 'touch'` (touch has no hover — the tap-the-peek-strip path stays);
- computes `inZone = clientY <= PEEK_HOTZONE_PX` (8px, matching `--peek-height`);
- on entering the band (and only if not already revealed), arms a short dwell (`CHROME_HOVER_DWELL_MS = 200ms`) that debounces twitchy passes, then calls the existing `reveal()`;
- clears the dwell when the cursor leaves the band.

`reveal()` arms the existing 4s idle-hide, so leaving auto-closes on the same clock — **no second close timer**, per the issue's N-decision. `ngOnDestroy` removes the listener and clears the dwell timer. Using `clientY` off a document listener means the peek strip's `pointer-events` don't change (no click interception).

## Tests

Dwell reveals; leave cancels; touch is a no-op; already-revealed doesn't arm; destroy stops a pending dwell; the listener is bound on construct and the same handler removed on destroy; the dwell survives small in-zone moves (measures time in the band, not stillness); leaving and re-entering arms a fresh dwell.

## Verification

Local gate: lint ✓ · betterer 183 (unchanged) ✓ · full unit suite ✓ · plugin ✓. Reviewed via the standard multi-persona pass (correctness, reliability, testing, maintainability, project-standards) — no production defect; the listener/timer lifecycle and re-arm latch coverage were added on review feedback.

**Manual on-device check pending** (hover behavior on real hardware) before merge.

Closes #202

### DIFF
--- a/src/app/core/components/toolbar/toolbar.component.spec.ts
+++ b/src/app/core/components/toolbar/toolbar.component.spec.ts
@@ -4,7 +4,7 @@ import { Router } from '@angular/router';
 import { of } from 'rxjs';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { AppService } from '../../services/app-service';
-import { ChromeVisibilityService } from '../../services/chrome-visibility.service';
+import { ChromeVisibilityService, CHROME_HOVER_DWELL_MS } from '../../services/chrome-visibility.service';
 import { DashboardService } from '../../services/dashboard.service';
 import { DialogService } from '../../services/dialog.service';
 import { NotificationsService } from '../../services/notifications.service';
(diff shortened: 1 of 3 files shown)


mairas:

Standard mairas-workflow fan-out (correctness · reliability · testing · maintainability · project-standards), each finding adversarially verified. Core claims: the dwell state machine is correct and the listener/timer lifecycle doesn't leak.

**Result: no production defect.** 3 findings, all test-coverage / maintainability (the sole P2 was corrected to P3 on verify — the code is correct and `ToolbarComponent` is a shell singleton):

1. **Listener add/remove lifecycle untested** (direct-call tests skipped it). **Addressed:** added a test spying on `document.addEventListener`/`removeEventListener` asserting the same handler is bound on construct and removed on destroy.
2. **The `!inPeekZone` re-arm latch untested** (no in-zone→in-zone or leave→re-enter). **Addressed:** added a test that the dwell survives small in-zone moves (measures time in the band, not stillness — distinguishes a guard-removal regression) and a leave-then-re-enter test.
3. **`PEEK_HOTZONE_PX` (8px) duplicated with SCSS `--peek-height`.** Accepted with the reciprocal cross-reference comments (the reviewer's stated minimum); a runtime CSS-var read / equality test isn't reliably computable under jsdom.

Follow-up verification after hardening: lint ✓ · 16 toolbar specs ✓ · full suite ✓.

Issue #281, opened 2026-07-14
History-chart reconnect re-backfill: bound the coalesced-chain hold and cancel the abandoned fetch under a slow provider


mairas:
## Context

Follow-up from the review of #85 (History-engine trend-chart reconnect re-backfill, `history-chart-stream.service.ts`). Behind the non-default History engine flag.

To keep the append-only chart consumers chronological, the live tail is held while a reconnect re-backfill fetch is in flight (the seam-ordering gate in `startLive`). Without it a live point stamped at `now` could append ahead of the older-timestamped batch and draw backwards.

## Landed in #85 (PR #284) — single-fetch freeze is now capped

The re-backfill fetch races a `RECONNECT_HOLD_MS` (3s) wall-clock deadline via `Promise.race`, well under the 30s history timeout. On deadline the live tail is released, an honest gap is drawn, and the superseded fetch's late result is ignored. A hung/slow history provider (InfluxDB/parquet — a separate plugin that commonly restarts alongside the SK server) no longer freezes the trend chart for up to ~30s after a reconnect. Original item 1 of this issue → **done**.

## Residual (deferred)

1. **Coalesced-chain hold budget.** The 3s cap is *per run*. Under sustained WS flapping, `reconnectPending` re-runs the re-backfill back-to-back in the `finally`; each run gets its own fresh 3s hold, so a coalesced chain can still hold the tail for roughly N×3s. Direction: one budget spanning the coalesced chain rather than per-run.

2. **Abandoned fetch is not cancelled** (P3, surfaced in the round-5 review). When the hold cap wins the race the superseded fetch is dropped but its HTTP GET is not aborted — it self-terminates only at the 30s `HISTORY_VALUES_TIMEOUT_MS`. Under WS flapping faster than `RECONNECT_HOLD_MS` against a provider in the 3–30s response band, up to ~10 concurrent History GETs can stack onto an already-struggling provider before the earliest self-terminates. Bounded and self-limiting, but it amplifies load exactly when the provider is weakest. Fix: thread an `AbortController` through `HistoryApiClientService.getValues` and abort the abandoned fetch on cap.

   Items 1 and 2 share the same reconnect-storm + slow-provider trigger and are best fixed together.

[...]

PR #284, opened 2026-07-14
History-engine trend charts: re-backfill the gap after a WebSocket reconnect (#85)

### LINKED ISSUE #85: History-chart engine: live tail does not re-backfill after WebSocket reconnect (data gap)

mairas:
## Motivation

When the Signal K WebSocket drops and reconnects, the History-engine trend charts resumed their live delta tail but **never re-fetched the interval that elapsed during the outage**. The chart drew a straight interpolated line across the gap — silently hiding a data dropout on a marine display, exactly what a trend chart must not do. Fixes #85.

## Approach

`HistoryChartStreamService` now watches the connection lifecycle and, on a transition back into `Connected`, re-backfills the window and splices the missed interval in as a batch:

- **Reconnect re-backfill.** On reconnect it re-fetches the window from the History API and appends only the points newer than the current seam (`> lastEmittedTs`, with a skew margin on `from`), so the gap is filled without duplicating the tail.
- **First-connect handling.** The initial `Connected` is skipped only while the constructor backfill is *contemporaneous* with it (the stream stayed connected since seeding). If a disconnect was seen first (chart mounted mid-outage) or the initial backfill failed (cold boot), the first `Connected` is treated as a reconnect and the elapsed interval is fetched.
- **Cadence-aware gaps.** The staleness threshold scales with the learned source interval, so a legitimately slow source is not false-gapped; the outage duration itself is not folded into the learned cadence.
- **Bounded live-tail hold.** The live tail is briefly held during the re-backfill fetch to keep the append-only chart chronological. That hold is capped at `RECONNECT_HOLD_MS` (3s) via `Promise.race` so a slow/hung history provider cannot freeze the chart — on cap it draws an honest gap and resumes live, letting the superseded fetch's late result be ignored.

Scoped to the non-default History engine flag. The delta-stream tail itself is unchanged; this is additive reconnect recovery on top of it.

## Testing

`history-chart-stream.service.spec.ts` gains 17 reconnect specs (fake-timer driven): no-double-backfill, missed-interval re-fetch, hold-while-disconnected, teardown safety, gap-on-error / gap-on-empty, no-gap-blip, cold-boot recovery, coalesced-seam advance, slow-cadence no-gap, known-cadence gap, cadence-race, hold-cap release, seam-advance ordering, mount-mid-outage re-backfill. Full gate green locally (lint + betterer:ci + 1183 unit + 7 plugin).

## Review

Five rounds of multi-persona + adversarial review, each surfacing a real defect (design flaws, cadence threshold, cadence-pollution race, freeze/data-loss, ordering, and the mount-mid-outage miss). Residual bounded items (coalesced-chain hold budget, abandoned-fetch cancellation, seam offset-jitter) are tracked in #281.

Closes #85.

### DIFF
--- a/src/app/core/services/history-chart-stream.service.spec.ts
+++ b/src/app/core/services/history-chart-stream.service.spec.ts
@@ -1,10 +1,11 @@
 import { TestBed } from '@angular/core/testing';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
-import { Subject, firstValueFrom } from 'rxjs';
+import { BehaviorSubject, Subject, firstValueFrom } from 'rxjs';
 import { HistoryChartStreamService, IHistoryChartStreamParams, isHistoryUnavailable } from './history-chart-stream.service';
 import { HistoryApiClientService, HistoryRequestError } from './history-api-client.service';
 import { HistoryToChartMapperService } from './history-to-chart-mapper.service';
 import { DataService, IPathUpdate } from './data.service';
+import { ConnectionState, ConnectionStateMachine } from './connection-state-machine.service';
 import { IDatasetServiceDatapoint } from '../interfaces/dataset.interfaces';
 
 const PARAMS: IHistoryChartStreamParams = {
@@ -18,6 +19,7 @@ const PARAMS: IHistoryChartStreamParams = {
 
 describe('HistoryChartStreamService', () => {
   let path$: Subject<IPathUpdate>;
+  let state$: BehaviorSubject<ConnectionState>;
   const history = { getValues: vi.fn() };
   const mapper = { mapValuesToChartDatapoints: vi.fn() };
   const data = { subscribePath: vi.fn(), getPathUnitType: vi.fn() };
@@ -28,7 +30,8 @@ describe('HistoryChartStreamService', () => {
         HistoryChartStreamService,
         { provide: HistoryApiClientService, useValue: history },
         { provide: HistoryToChartMapperService, useValue: mapper },
-        { provide: DataService, useValue: data }
+        { provide: DataService, useValue: data },
+        { provide: ConnectionStateMachine, useValue: { state$ } }
       ]
     });
     return TestBed.inject(HistoryChartStreamService);
@@ -36,6 +39,9 @@ describe('HistoryChartStreamService', () => {
 
   beforeEach(() => {
     path$ = new Subject<IPathUpdate>();
+    // Charts mount while already connected; the reconnect re-backfill fires only on LATER Connected
+    // transitions, so a stuck-Connected stream leaves the existing live-tail behavior unchanged.
+    state$ = new BehaviorSubject<ConnectionState>(ConnectionState.Connected);
     history.getValues.mockReset();
     mapper.mapValuesToChartDatapoints.mockReset().mockReturnValue([]);
     data.subscribePath.mockReset().mockReturnValue(path$);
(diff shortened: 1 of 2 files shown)


mairas:
## Review synthesis — round 5 (final)

Multi-persona + adversarial fan-out (correctness / testing / maintainability / project-standards / reliability / adversarial, each fresh-context), followed by an adversarial verify pass (default-refute unless a concrete failing sequence). **3 confirmed, 0 refuted.**

### P2 — mount-during-outage swallowed the first reconnect re-backfill — **FIXED in this PR**
`adversarial`, CONFIRMED. If a trend chart **mounts while the WS is down but HTTP is up** (an intermittent marine link + dashboard navigation), the constructor backfill succeeds and seeds the window, but the first `Connected` arrives minutes later. The old guard (`hasConnectedBefore || !ctx.seeded`) treated that first `Connected` as the free first-connect skip because the window was already seeded — so the elapsed outage interval was never re-fetched and the chart drew a straight interpolated line across it: **the exact #85 defect, silently reproduced** for the first post-mount reconnect.

Fix (commit in this PR): the first `Connected` is skipped only when it is *contemporaneous* with the seed (no disconnect observed since `beginLive` started). The guard is now `!ctx.seeded || sawDisconnected`, so a chart mounted mid-outage re-backfills on its first reconnect. New spec: `re-backfills the first connect when mounted during an outage — the seed is not contemporaneous (#85)`.

### P3 ×2 (duplicate) — hold-cap abandons the in-flight fetch without cancelling it — **deferred to #281**
`reliability`, CONFIRMED. When the 3s hold cap wins the `Promise.race`, the superseded History fetch is dropped but its HTTP GET is not aborted (it self-terminates at the 30s `HISTORY_VALUES_TIMEOUT_MS`). Under WS flapping faster than `RECONNECT_HOLD_MS` against a provider in the 3–30s band, up to ~10 GETs can stack onto an already-struggling provider before the earliest self-terminate. Bounded and self-limiting, but it amplifies load exactly when the provider is weakest. The fix needs an `AbortController` threaded through `HistoryApiClientService.getValues` — a cross-service API change, disproportionate for a bounded P3 in this PR. Tracked in #281.

### Rounds 1–4 (addressed in-branch)
Five P2 design flaws (fill/gate/ordering); flat gap-threshold ignoring source cadence; cadence-pollution race (mid-fetch delta inflating the threshold); live-tail freeze/data-loss under a slow provider (now capped at 3s); P3 seam-ordering (gap now advances `lastEmittedTs`). All fixed and spec'd on this branch.

Issue #240, opened 2026-07-10
Port Kip #1106: resizable boolean switches/lights/buttons


mairas:
## Summary

Upstream Kip PR [mxtommy/Kip#1106](https://github.com/mxtommy/Kip/pull/1106) ("Resizable switches and indicators for improved usability", squash `5a8c0895`) makes the boolean-switch panel's controls (switch, light, indicator button) scale their SVG geometry to the available widget size instead of rendering at a fixed 180×35 viewBox. It introduces a shared layout utility that computes all shape coordinates from `(width, height)` and a binary-search routine that picks the largest per-control height at which every control's label still fits (measured against a real canvas font). This fixes cramped/clipped controls on large or multi-control panels.

Skip does **not** have this. Skip still renders the old fixed-geometry version:
- `svg-boolean-switch.component.ts:19-20,25,37` — fixed `toggleOff = "0 35 180 35"` / `toggleOn = "0 0 180 35"` with a `viewBox` swap on state change.
- `svg-boolean-switch.component.svg:4` — `[attr.viewBox]="viewBox"`.
- `widget-boolean-switch.component.ts:119-125` — `onResized` derives control height with the old `35 * width / 180` proportional formula; there is no text-fit measurement.

## What the upstream change contains

Feature-relevant files (the only ones we want):
- **New:** `src/app/widgets/widget-boolean-switch/boolean-control-layout.util.ts` — `getBooleanControlLayout(type, width, height)` returning all switch/light/button coordinates scaled by `height / 35`, and `measureBooleanControlsHeight(...)` (binary search over per-control height that keeps labels within `labelWidth`).
- **New:** `src/app/widgets/widget-boolean-switch/boolean-control-layout.util.spec.ts` — vitest spec for the util.
- `svg-boolean-switch/`, `svg-boolean-button/`, `svg-boolean-light/` — each `.component.ts` gains a `layout` getter (plus `isEnabled`/`nextState` where relevant) and drops the fixed `viewBox`; each `.component.svg` is rewritten to bind shape geometry to `layout.*`.
- `widget-boolean-switch.component.ts` — injects `CanvasService`, tracks `hostSize`, and replaces `onResized`'s proportional math with `updateCtrlDimensions()` that calls `measureBooleanControlsHeight` using a canvas 2D context and `CanvasService.DEFAULT_FONT`.

[...]

PR #286, opened 2026-07-14
Momentary boolean button: emit on press for low control latency (#241a)

mairas:
## Motivation
The momentary boolean button waited a hold delay before actuating — added latency to a boat control. Fixes the momentary-button half of #241.

## Approach
Emit synchronously on `pointerdown` (no hold timer), then repeat every 100ms while held; a swipe (>30px in x **or** y) sets `pressed=false` and clears the repeat so a scroll gesture starting on the button stops firing. Matches upstream `f9d657cf`'s post-PR semantics. The `data()` null-guard is preserved (file stays strict-clean).

## Review
Multi-persona + adversarial verify. One P3 coverage gap fixed: the vertical-swipe (`deltaY>30`) cancel branch is now tested. Emit-on-press means a swipe that begins with a press necessarily fires one emit (the Gate-A latency-first decision), so the test asserts the swipe stops the *repeat*, not zero emits.

The OFF-state visual half of #241 (packet G4 / #240, boat-verify) is intentionally **not** in this PR — sanctioned carve-out.

Part of #257. Refs #241.

### DIFF
--- a/src/app/widgets/svg-boolean-button/svg-boolean-button.component.spec.ts
+++ b/src/app/widgets/svg-boolean-button/svg-boolean-button.component.spec.ts
@@ -43,41 +43,57 @@ describe('SvgBooleanButtonComponent momentary emit', () => {
     vi.clearAllMocks();
   });
 
-  it('actuates after the short hold, not before', () => {
+  it('emits immediately on press, with no hold delay', () => {
     setup();
     vi.useFakeTimers();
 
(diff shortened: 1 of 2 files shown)


mairas:

**1 P3 confirmed, fixed in-PR:** the swipe-cancel condition is `deltaX>30 || deltaY>30`, but every swipe test moved only in X — the vertical-swipe branch was uncovered. Added a Y-only swipe-cancel test. No other findings survived verification.

Issue #106, opened 2026-07-02
Rename the Signal K plugin id kip → skip for Kip/Skip coexistence


mairas:
## Problem

Skip is a fork of Kip. It ships as `@halos-org/skip` (the webapp + a bundled Signal K server plugin — `main: plugin/index.js`), but the plugin still registers under **plugin id `kip`**.

Signal K plugin ids must be unique on a server, and each plugin owns `/plugins/<id>/` for both its HTTP routes **and** its own config storage. So on any server that also runs Kip, the two `kip` plugins **hard-collide**: the `/plugins/kip/` mount, the plugin config slot, and the History-API-provider registration all clash. Kip and Skip cannot be installed simultaneously.

This is the plugin-tier sibling of #84, which gave Skip its own config-storage namespace on the applicationData + browser tiers. #84 deliberately left the plugin id alone (flagged out of scope there) because renaming it means changing a published server component's identity.

> **Updated after #112** (image feature removal): the plugin's image subsystem and its `/plugins/kip/images` routes are gone, and the client `image-asset.service.ts` / `kip-plugin-url.util.ts` were deleted. The plugin's **only** remaining responsibility is now history/series. Scope below is trimmed accordingly.

## Scope

Rename the plugin id `kip` → `skip` (→ served at `/plugins/skip/`) at the single server declaration and every client reference, single-sourced as a constant (mirroring #84):

- **Server:** `kip-plugin/src/index.ts` `plugin.id`
- **Client URL builder:** `signalk-plugin-url.util.ts`
- **Client consumers:** `KipSeriesApiClientService` (`/plugins/kip/series`), `display.component.ts` (`KIP_DATA_PLUGIN_ID`, `providerMode`), `plugin-config-client.service.ts` (`getKipRuntimeModeConfigCached('kip')` / `resolveKipRuntimeModeConfig`), `dashboard-history-series-sync.service.ts`, `history-api-client.service.ts`
- **Specs** asserting `/plugins/kip/...`

## Migration

Cold-turkey, matching #84 — the plugin does not read the old `kip` plugin config. The plugin's data dir is keyed by plugin id, so renaming orphans any **recorded history** under the old id (image assets are no longer a concern post-#112). Acceptable pre-GA; confirm cold-turkey vs. a one-time move and document in the release notes.

## Out of scope

[...]

PR #287, opened 2026-07-14
History dialog: enumerate template concretes over a retention window (#278)

### LINKED ISSUE #278: History dialog: enumerate template concretes over a generous getPaths window (not the chart display window)
## Context
Surfaced by the A2 (#106) adversarial review. With the kip-series reconcile deleted, Skip is a pure History-API reader: template/electrical widget history (BMS/solar/charger/inverter/alternator/ac) is now expanded by `widget-history-chart-dialog.expandTemplateSeries` via `historyApiClient.getPaths({ duration })` instead of the plugin's registered-series list.

## Problem
`expandTemplateSeries` derives the `getPaths` enumeration window from `resolveDuration(series)`, which returns the **chart display window** (`timeScale × period`, or the short `selectedPeriod()` fallback) — not a generous retention window. `getPaths` only returns paths that have logged a sample inside that window, so:
- a configured device that hasn't reported within the display window is **absent from the enumeration and silently dropped** from the template chart (the old plugin `getSeriesDefinitions` returned registered series regardless of recent data);
- the fallback `selectedPeriod()` can be as short as the default chart period.

For continuously-reporting electrical telemetry (battery voltage/current, solar output) the practical risk is low — they log every few seconds, so any display window contains them. The gap bites only infrequently-reporting devices.

## Fix direction
Enumerate template concretes over a **generous** window (e.g. `series.retentionDurationMs` / `AUTO_RETENTION_MS` = 24h) independently of the chart display window, then query each concrete's data over the display window as today. Add a test that a BMS/solar template resolves concretes from a `getPaths` result whose window is the retention window, not the short default.

## Also
Boat-verify (A2 / #277): confirm on the live server that a configured BMS/solar widget's long-press dialog resolves concrete paths from `getPaths` over a realistic window — i.e. that deleting the Skip registration did not blank template-widget history.

Discovered during the #257 campaign (A2 / #106 review).

mairas:
## Motivation
Since the kip-series reconcile was removed (#106), template/electrical widget history is enumerated via the History API's `getPaths`. It used the chart **display** window, so a configured device that hadn't logged a sample within that (possibly short) window was silently dropped from the template chart. Fixes #278.

## Approach
`expandTemplateSeries` now derives the `getPaths` enumeration window from the series' `retentionDurationMs` (~24h), falling back to a named 24h constant when it's absent/invalid. The chart **data** fetch window (`resolveDuration` / `getValues`) is unchanged — only the enumeration widens, so a briefly-quiet device is still listed while the chart still renders over the display window.

## Review
Multi-persona + adversarial verify. One P3 test gap fixed: the fallback-window branch is now covered (absent / 0 / negative / NaN retention all collapse to the 24h default).

Part of #257.

### DIFF
--- a/src/app/core/components/widget-history-chart-dialog/widget-history-chart-dialog.component.spec.ts
+++ b/src/app/core/components/widget-history-chart-dialog/widget-history-chart-dialog.component.spec.ts
@@ -765,4 +765,154 @@ describe('WidgetHistoryChartDialogComponent', () => {
     expect((scales['yCurrent'] as { position?: string }).position).toBe('right');
     expect((scales['yFrequency'] as { position?: string }).position).toBe('left');
   });
+
+  it('enumerates a template concrete visible only in the retention window, not the short display window', async () => {
+    (component as unknown as {
+      data: {
+        title: string;
+        widget: IWidget;
+        seriesDefinitions: IKipSeriesDefinition[];
+      };
+    }).data = {
+      title: 'Solar History',
+      widget: {
+        uuid: 'widget-solar-1',
+        type: 'widget-solar-charger',
+        config: {
+          displayName: 'Solar Charger'
+        }
+      } as IWidget,
+      seriesDefinitions: [
+        {
+          seriesId: 'widget-solar-1:solar-template',
+          datasetUuid: 'widget-solar-1:solar-template',
+          ownerWidgetUuid: 'widget-solar-1',
+          ownerWidgetSelector: 'widget-solar-charger',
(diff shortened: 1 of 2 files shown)


mairas:

**1 P3 confirmed, fixed in-PR:** both new specs set an explicit `retentionDurationMs`, so the `DEFAULT_ENUMERATION_WINDOW_MS` fallback branch (absent/0/negative/NaN retention) was untested. Added a fallback-window spec covering all falsy inputs. No other findings survived verification.

Issue #151, opened 2026-07-03
notifications.service: stale alarms survive reset; unknown states inflate alarmCount


mairas:
Surfaced by the #27 characterization pass (PR #147, `notifications.service.spec.ts`).

## Defect 1: stale alarms survive a DataService reset while notifications are enabled

`notifications.service.ts:128-134` — `reset()` only clears the notifications list when `disableNotifications` is **true**. On a DataService reset (server reconnect / URL switch) with notifications enabled — the common case — stale notifications from the previous connection stay in the list and in the alarm aggregate. On a marine display, stale alarms are a trust problem in both directions (phantom alarms, or masking that a real one went silent).

Pinned by the "keeps existing notifications on reset while notifications are enabled" test.

## Defect 2: unknown-state notifications inflate the badge count

`notifications.service.ts:198` — `activeNotifications` is incremented *before* `getNotificationSeverity()` validates the state, so a notification with an unknown state inflates `alarmCount` (badge) while contributing zero audio/visual severity.

Pinned by the "counts notifications with an unknown state at zero severity" test.

mairas:
Notification functionality will be reviewed at a later stage.

PR #288, opened 2026-07-14
Notifications: reconcile on reconnect instead of blind-clearing (#275)

### LINKED ISSUE #275: Harden notification reconnect-clear: reconcile instead of blind clear (avoid transient alarm blank/silence)
## Context
Surfaced by the B4 (#151) multi-persona review. On a connection reconnect, `NotificationsService` clears its entire notification list and relies on the Signal K server re-delivering still-active notifications in its post-resubscribe snapshot (see `notifications.service.ts`, the `ConnectionStateMachine.state$` subscription).

## Problem (flagged P2 by reliability, adversarial, and maintainability reviewers)
The blind clear has transient/edge-case safety gaps on a marine display:
- **Transient blank + audio restart:** an active alarm (bilge, engine, MOB) is dropped from the badge/count and its looping tone stops during the reconnect-snapshot window, then restarts from zero when the delta re-arrives. Boat WiFi flaps repeatedly, so an unacknowledged alarm stutters.
- **No fallback:** repopulation depends solely on the server volunteering a full snapshot on resubscribe. If a non-default server/plugin only emits notifications on change (not on subscribe), the active alarm is cleared and never restored until its state next changes — a silently dropped safety alarm.

B4 is a net improvement over the prior stale-alarm bug (phantom alarms that survived reset); this issue tracks closing the transient/fallback gaps.

## Options
- **Reconcile (mark-stale + sweep):** on reconnect, mark existing entries pending; as the fresh snapshot arrives, confirm re-sent paths; after a short grace/debounce window, sweep only the unconfirmed ones. Active alarms never blank or restart audio; only truly-stale ones are removed.
- **Re-seed from DataService's retained cache** (which survives the reconnect) rather than clearing to empty.
- At minimum, **do not restart an active alarm's audio** when the same notification is re-sent.

## Scope
Larger than B4's clear-on-reconnect (needs reconciliation state + a grace timer on safety-critical code). Filed as a follow-up per the review.

Discovered during the #257 campaign (B4 / #151 review).

mairas:
## Motivation
On a WebSocket reconnect, `NotificationsService` blind-cleared its entire list and relied on the server re-delivering a full snapshot. On a marine display that transiently blanked an active alarm (bilge/engine/MOB) and **restarted its tone from zero on every WiFi flap**, and if the server never volunteered a snapshot the alarm never returned. Fixes #275.

## Approach
On reconnect, keep the list intact and mark all value-bearing paths *pending*; arm a 3s grace window. A delta that re-delivers a **pending (carried-over)** path removes it from pending and records that a snapshot was seen. At timeout, still-pending paths are swept **only if** a carried path was re-delivered (evidence the server actually re-streamed). On a change-only server (no re-delivery) nothing is swept — active alarms are retained and their tone is never interrupted.

## Review
Multi-persona + adversarial verify found a **P1/P2 hole** in the first cut: the snapshot-seen flag was armed by *any* delta, so an unrelated notification could authorize the sweep to drop still-active alarms on a change-only server — the exact failure this packet exists to prevent. Fixed with **per-path gating** (flag set only when a *carried* path is re-delivered) plus a survival test (unrelated-delta → carried alarm retained). The reviewer's "never sweep active alarms" alternative was rejected: on a snapshot server a resolved alarm is simply *absent*, so that would leave a resolved alarm sounding forever.

Design note: the cleanest signal would be an explicit "resubscribe snapshot complete" event from the connection layer; the packet deliberately avoids that coupling — per-path gating is correct for realistic servers.

Part of #257. Refs #151.

### DIFF
--- a/src/app/core/services/notifications.service.spec.ts
+++ b/src/app/core/services/notifications.service.spec.ts
@@ -1,7 +1,7 @@
 import { TestBed } from '@angular/core/testing';
 import { signal, WritableSignal } from '@angular/core';
 import { Subject } from 'rxjs';
-import { describe, expect, it, vi } from 'vitest';
+import { afterEach, describe, expect, it, vi } from 'vitest';
 import { INotification, INotificationInfo, NotificationsService } from './notifications.service';
 import { DataService } from './data.service';
 import { SettingsService } from './settings.service';
@@ -99,6 +99,10 @@ describe('NotificationsService', () => {
 
   function activeTrack(): number | null {
     return (service as unknown as { _activeAlarmSoundtrack: number | null })._activeAlarmSoundtrack;
+  }
+
+  function alarmPlayer(track: number): HTMLAudioElement {
+    return (service as unknown as { _players: Map<number, HTMLAudioElement> })._players.get(track)!;
   }
 
   describe('alarm state aggregation from deltas', () => {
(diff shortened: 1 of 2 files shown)


mairas:

**P1/P2 confirmed, fixed in-PR** (five findings converged): `_reconcileSawSnapshot` was armed by *any* delta in the grace window, so on a change-only server an unrelated notification could authorize the sweep to drop still-active carried alarms (bilge/engine/MOB) — the exact failure this packet prevents. Fixed with **per-path gating** (the flag is set only when a *carried* path is re-delivered) plus a survival test (unrelated delta → carried alarm retained). Rejected the reviewer's "never sweep active alarms" alternative: on a snapshot server a resolved alarm is simply absent, so that would leave a resolved alarm sounding forever. Design note: an explicit resubscribe-snapshot-complete signal from the connection layer would be cleaner, but the packet deliberately avoids that coupling — per-path gating is correct for realistic servers.

Issue #290, opened 2026-07-14
Wind Trends: TWS hardcoded to environment.wind.speedTrue — blank on ground-referenced-only vessels


mairas:
## Problem

`widget-windtrends-chart` hardcodes its two series paths as module constants (`widget-windtrends-chart.component.ts:23-24`), with no per-widget path configuration:

- TWD → `self.environment.wind.directionTrue`
- TWS → `self.environment.wind.speedTrue`

On a vessel that publishes true wind referenced to **ground** (`environment.wind.speedOverGround`) but not to **water** (`environment.wind.speedTrue`), the TWS trend is **silently empty** — the chart and its axes draw, TWD renders, but the TWS line never appears and nothing tells the user why.

Two problems compound:

1. **Not configurable.** The paths are constants; a user whose sensor/derivation emits a different wind path can't repoint the widget.
2. **Inconsistent reference frame.** TWD uses the ground-referenced `directionTrue` while TWS uses the water-referenced `speedTrue`. Signal K's common derived-wind set is fully ground-referenced (`directionTrue` + `angleTrueGround` + `speedOverGround`), so `speedTrue` is frequently absent while `speedOverGround` is present.

## Evidence

Observed on a dev vessel's Signal K (`derived-data` source):

```
environment.wind.directionTrue    present   → TWD path, renders
environment.wind.speedOverGround  present   → ground-referenced true wind speed
environment.wind.speedTrue        ABSENT    → TWS path, blank
environment.wind.angleTrueGround  present
environment.wind.speedApparent / angleApparent  present (wind sensor, CAN)
```

TWD populates and TWS is blank on exactly this common ground-referenced setup.

## Suggested direction

- Make the widget's wind paths **configurable** (path pickers, like other widgets), defaulting to today's constants — so a user can point TWS at whatever their vessel publishes.
- And/or **degrade gracefully**: when `speedTrue` is absent, fall back to `speedOverGround` (which also aligns the reference frame with the ground-referenced `directionTrue`). When neither is present, keep the current empty state — never crash.
- Consider pairing the defaults consistently (a single reference frame) so out-of-the-box TWD + TWS both populate on a standard SK derived-wind server.

Adopt whatever wind paths the server publishes and degrade gracefully when one is absent (Skip stays independent of any particular server config).

mairas:
## Settled implementation plan (design pass + maintainer decision 2026-07-15)

[...]

PR #291, opened 2026-07-14
Register the minimal Chart.js union, drop chart.js/auto (#267)

### LINKED ISSUE #267: chart-registration: register the minimal component union, not all registerables
Surfaced by the review of PR #264 (#239). `src/app/core/utils/chart-registration.util.ts` registers `...registerables` — **every** Chart.js controller/element/scale (bar, pie, doughnut, radar, polar, scatter, bubble, …) — but Skip only renders line/time charts. #239 asked for the *union* of what the four consumers actually registered; the broad set was chosen as "simpler/future-proof."

On a Pi-class HaLOS target, bundle size matters (see `perf-harness/`). Follow-up: replace `...registerables` with the explicit minimal union the four consumers need (line controller/element, point element, linear + time scales, filler/legend/tooltip as used) plus the annotation + streaming plugins, and measure the bundle delta with the perf-harness. Keep one idempotent registration.

mairas:
## Motivation
`chart-registration.util.ts` registered `...registerables` — **every** Chart.js controller/scale/element (bar, pie, radar, financial, …) — but Skip only renders line/time charts. And `signalk.component` imported `chart.js/auto`, which self-registers everything at module load. Both pull the full Chart.js surface into the bundle on a Pi-class target. Fixes #267.

## Approach
Register the explicit **minimal union** the five chart consumers actually use — `LineController, LineElement, PointElement, LinearScale, TimeScale, Filler, Legend, Tooltip, Title, SubTitle` plus the annotation and streaming plugins (the streaming plugin also contributes the `realtime` scale) — once, idempotently. Migrate `signalk.component` off `chart.js/auto` onto the shared `registerChartComponents()` util so nothing self-registers the full set; its needs are a strict subset of the union.

## Verification
jsdom mocks `<canvas>`, so unit tests cannot catch a missing registration. Verified two ways: (a) exhaustive static enumeration of all five consumers (minichart, data-chart, windtrends, history-dialog, signalk) with file:line evidence, independently re-audited under adversarial review; and (b) an on-device deploy + render check of every chart surface. Betterer regenerated for `signalk.component`'s hash (stayed 183 — no new strict-null issues).

Closes #267.

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -71,11 +71,11 @@ exports[`strictNullChecks`] = {
       [54, 6, 17, "Object is possibly \'undefined\'.", "2944426029"],
       [55, 6, 17, "Object is possibly \'undefined\'.", "27783321"]
     ],
-    "src/app/core/components/options/signalk/signalk.component.ts:3082783318": [
-      [101, 10, 6, "Type \'null\' is not assignable to type \'Chart<keyof ChartTypeRegistry, (number | [number, number] | Point | BubbleDataPoint | null)[], unknown>\'.", "3491622934"],
-      [110, 45, 20, "Object is possibly \'undefined\'.", "235433357"],
-      [171, 28, 20, "Object is possibly \'undefined\'.", "235433357"],
-      [171, 28, 51, "Argument of type \'CanvasRenderingContext2D | null\' is not assignable to parameter of type \'ChartItem\'.\\n  Type \'null\' is not assignable to type \'ChartItem\'.", "38422540"]
+    "src/app/core/components/options/signalk/signalk.component.ts:2801402843": [
+      [102, 10, 6, "Type \'null\' is not assignable to type \'Chart<keyof ChartTypeRegistry, (number | [number, number] | Point | BubbleDataPoint | null)[], unknown>\'.", "3491622934"],
+      [111, 45, 20, "Object is possibly \'undefined\'.", "235433357"],
+      [172, 28, 20, "Object is possibly \'undefined\'.", "235433357"],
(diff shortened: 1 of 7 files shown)


mairas:
## Review synthesis

Multi-persona review — union-completeness (adversarial), correctness, maintainability — plus an adversarial verify pass. **0 confirmed findings.** The 12-member union was independently re-enumerated against all five Chart.js consumers (minichart, data-chart, windtrends, history-dialog, signalk) and held; the `signalk.component` migration (module-load registration before construction, `chartjs-adapter-date-fns` retained for the time axis) verified correct.

**Render verification (on-device, https://halpi.hurma:4430):** deployed and checked across chart surfaces. An initial "Settings blank" report was a stale cached bundle (forced reload resolved it) — not a #267 defect. A blank TWS line on Wind Trends is a data mismatch (the boat publishes `speedOverGround`, not the widget's hardcoded `speedTrue`) → tracked as #290, unrelated to chart registration.

No follow-ups required.

mairas:
Force-pushed `e012c17e`: fixed CI. The windtrends spec uses a **self-contained** `vi.mock('chart.js')` factory (deliberately, to avoid a hoisting TDZ), and vitest factory mocks are strict — it was missing the `Legend` and `Tooltip` exports that this PR's util newly imports, so `registerChartComponents()` threw at module load. Added both sentinels to the mock. (The other three consumer specs use `importOriginal`, so they were unaffected; local runs didn't enforce the strict-export check — CI/Node-24 caught it.)

mairas:
Force-pushed `ce29fc53`: the previous fix exposed the same gap in the **minichart** and **data-chart** self-contained `chart.js` mocks (the util's module-level `registered` guard persists across files in a vitest worker, so only whichever consumer spec loads first in its worker hits the strict-export access — non-deterministic sharding masked it locally and rotated it between CI runs). All four self-contained `chart.js` mocks (minichart, data-chart, windtrends, chart-registration.util) now export the full member set the union registers.

Issue #293, opened 2026-07-14
gen:mcp-schema broken on main (WidgetAcComponent import) — schema artifact can't be regenerated


mairas:
## Problem

`npm run gen:mcp-schema` (`UPDATE_SCHEMA=1 npm run test:mcp-schema`) fails on `main`:

```
Error: Could not find an import for "WidgetAcComponent" in src/app/core/services/widget.service.ts
  ❯ findImportModuleSpecifier tools/gen-mcp-schema/ast.ts:186
  ❯ extractWidgetSchemas   tools/gen-mcp-schema/generate.ts:89
```

Three suites (`schema-artifact`, `schema`, `widget-schema`) fail to load and `generate.spec.ts > excludes the commented-out widgets (alternator, inverter, ac)` fails. The generator's AST walker can't resolve `WidgetAcComponent`'s import — the schema-gen tooling has drifted from `widget.service.ts`: the ac/inverter/alternator electrical widgets are registered and live (per CLAUDE.md), but the generator still expects them commented out/excluded.

## Impact

`src/assets/kip-dashboard-schema.json` cannot be regenerated. Any change to a widget's name/description/selector in `widget.service.ts` leaves the committed schema **stale** — and because `test:mcp-schema` is **not** part of `npm run ci`, nothing fails the PR; the artifact silently drifts. Surfaced by F1 (#126/#247): its `Realtime Data Chart` → `Realtime Data Plot` rename cannot be reflected in the schema until this is fixed.

## Fix

Update the schema-gen AST walker and its widget-list expectation to match current `widget.service.ts` (resolve `WidgetAcComponent`'s import; reconcile the "commented-out ac/inverter/alternator" assumption with their live registration), then regenerate `kip-dashboard-schema.json` (which will also pick up F1's plot rename). Consider adding `test:mcp-schema` to the CI gate once green so the artifact can no longer drift unnoticed.

Issue #247, opened 2026-07-13
Realtime Data Chart config still cites the removed ~120-sample limit; make the description user-facing


mairas:
## Summary

The Realtime Data Chart's config help text still describes an old "keep ~120
points" sampling limit that we've since removed. Two problems: the number is now
factually wrong, and the sentence exposes internal sampling mechanics a casual user
doesn't need — they just want to set how much time-series history the chart shows.
The same stale 120-datapoint claim also survives in the History API help doc.

## Current behaviour (why the 120 claim is stale)

The chart window no longer caps at 120 points. Point count is derived from the
window and the sample time, uncapped:

- `src/app/core/utils/chart-window.util.ts:54` —
  `maxDataPoints = Math.max(1, Math.ceil(windowMs / sampleTime))`.
- History-backfill requests by **resolution** (= sample time), not a fixed count:
  `src/app/core/services/history-chart-stream.service.ts:200` —
  `resolutionSeconds = Math.max(1, Math.round(params.sampleTime / 1000))`, passed as
  the History API `resolution`. No 120 anywhere in the path.

## Stale spots

**1. Realtime Data Chart config (the one this issue is about)** —
`src/app/widget-config/dataset-chart-options/dataset-chart-options.component.html:99`:

> Time scale and duration set how much history the chart shows (the chart window).
> Data is sampled to keep roughly 120 points across the window; very short windows
> may show fewer points for performance.

The first sentence is fine and useful. The second is both stale (no 120 cap) and too
implementation-focused for a casual user.

**2. History API help doc** — `src/assets/help-docs/history-api.md:59-63`:

> For performance reasons, KIP charts display a maximum of 120 datapoints. KIP uses
> the chart's configured time window and asks the provider to return data for that
> window as up to 120 datapoints.
> - Slower sampling rates produce fewer datapoints (lower visual resolution).
> - Faster sampling rates do not produce more than 120 datapoints in KIP charts.

Not just a number to swap — the whole rationale is inverted now: with
`maxDataPoints = window / sampleTime`, faster sampling **does** produce more points.
The "Provider Plugin Configuration" guidance built on the 120 premise needs
rethinking, not find-and-replace.

## What to change

- **Config text (#1):** rewrite so it explains the user-facing knobs (time scale +
  duration = how far back the chart looks) without the sampling internals or any
  fixed point count. Directional example, not required copy: *"Time scale and
  duration set how far back the chart shows — the chart window. Finer sample times
[...]

PR #294, opened 2026-07-14
Rename time-series "charts" to "plots" + fix stale sampling copy (#126, #247)

### LINKED ISSUE #126: Rename time-series "charts" to "plots" — "chart" means nautical chart in a marine app
## Problem

"Chart" has a specific meaning in a marine application: a nautical chart. Skip currently uses the word for both meanings at once, sometimes side by side:

- The widget list offers **Realtime Data Chart** (a time-series visualization) two entries below **Freeboard-SK**, described as "the Freeboard-SK **chart plotter**" (`widget.service.ts`).
- **Chartplotter Mode** refers to the nautical chart; the **history chart dialog** and the Numeric widget's background **minichart** refer to time series.
- Help docs mix both: "paths you want to chart", "chart time span" (`history-api.md`) vs. "Freeboard-SK chart plotter" (`dashboards.md`).

For a sailor, "chart" defaults to the nautical map. Time-series visualizations should use a different word.

## Proposal

Rename time-series "charts" to **plots**: "Realtime Data Plot", "history plot", plot-related settings and help-doc language. "Graph" was considered as an alternative; both are standard, "plot" reads better in instrument contexts and is the proposal here. Note that "plot" also appears in "chartplotter", but there it means position plotting — no practical confusion with a data plot widget.

## Scope tiers

**Tier 1 — user-facing strings (do this):** widget names and descriptions in `widget.service.ts`, dialog titles, settings labels, help docs (`dashboards.md`, `history-api.md`, `time-series.md`), README. Cheap, no compatibility impact, resolves the user-visible ambiguity.

**Tier 2 — code identifiers and persisted keys (decide separately):** component/file/selector names (`widget-data-chart`, `widget-history-chart-dialog`, `widget-windtrends-chart`), persisted widget type `'widget-data-chart'` in dashboard configs, and config properties (`showMiniChart`, `datachartPath`, `datachartSource`, ...). Costs:
- Persisted identifiers need a migration; `configuration-upgrade.service.ts` already has the pattern (`'WidgetDataChart': 'widget-data-chart'`).
- Every renamed file and identifier is a standing conflict surface for upstream KIP PR replication (#38). The chart widgets are actively developed upstream, so this friction is recurring, not one-time.

Skip has already accepted divergence of this kind (kip→skip config namespace rename), so Tier 2 is consistent with project direction — but it should be weighed against the replication burden and done as its own change if at all.

**Out of scope — must keep "chart":**
- chart.js library API references (`Chart`, `ChartData`, `chartjs-adapter-date-fns`, `chartjs-plugin-annotation`).
- "Chartplotter" / "chart plotter" for Freeboard-SK and Chartplotter Mode — there the nautical meaning is the correct one.

Current footprint for reference: ~1000 case-insensitive "chart" matches across 73 files (excluding lockfile), a large share of which are chart.js API usage that stays.

mairas:
## Motivation
In a marine app "chart" means a **nautical chart**, so the time-series widgets should read as **plots** in user-facing wording (#126). Separately, the Realtime Data Chart's config still cited a removed "~120 sample" limit — factually wrong and exposing internal sampling mechanics (#247). Fixes #126, #247.

## Approach
Per-context "chart" → "plot" rename of **user-facing wording** across `widget.service.ts` (widget names/descriptions), the config + history-dialog templates, and the help docs + README. **Deliberately kept** (verified per occurrence): nautical "chartplotter" / "nautical chart", code identifiers & component selectors, persisted config keys (`datachartPath`, `showMiniChart`, `verticalChart`, … — renaming would break saved dashboards), CSS class names, and chart.js types.

**#247:** replaced the stale "~120 sample" copy with the real model — point count is **window-derived and uncapped** (the Realtime Data Plot targets ~500 points, 100ms floor). Nuance preserved: the pop-up history **dialog** still targets ~120 points (`widget-history-chart-dialog.component.ts:815`), so both densities are described and the still-valid 7.5s/15-min guidance kept; only the *universal* 120 claim was removed.

## Note — schema not regenerated (blocked by #293)
`widget.service.ts` drives `src/assets/kip-dashboard-schema.json`, but `gen:mcp-schema` is **broken on `main`** (can't resolve `WidgetAcComponent`'s import — pre-existing, filed as #293). So the schema artifact stays stale for the plot rename and will update once #293 lands. `test:mcp-schema` is not in the `npm run ci` gate, so this doesn't block CI.

## Known residual
The widget-config panel still labels the mini-plot toggles "Graph" (a third term) — outside the chart→plot mandate here; flag if the whole time-series UI should converge on "plot".

Fixes #126, #247.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Skip is designed for sailors and boaters who want:
 - A **modern, polished interface** optimized for marine displays.
 - **Touch-optimized design**: touch-first, intuitive design for tablets, phones, and other touch-enabled devices.
 - **Cross-platform support**: runs on phones, tablets, laptops, Raspberry Pi, Web Enabled TV or other fixed displays - anywhere you can run a web browser.
-- **Instant access to all Signal K data**: displays gauges, charts, switches, and other widgets right out of the box.
+- **Instant access to all Signal K data**: displays gauges, plots, switches, and other widgets right out of the box.
 - **Flexible dashboards**: customize layouts, drag-and-drop widgets, night/day mode, kiosk/fullscreen and remote control support.
 
 With Skip, you get the **familiar feel of professional Multi-Function Displays** combined with the flexibility of Signal K. It’s simple, reliable, and highly usable — a modern, touch-first Multi-Function Display for [Signal K](https://signalk.org) vessels.
@@ -115,7 +115,7 @@ All Skip widgets are visual presentation controls that are very versatile, with 
 - **AC/DC Charger**- Monitor charging performance at a glance with a compact AC/DC Charger Widget. View single or multiple chargers with charge mode, voltage, current, power and temperature. Chargers are discovered automatically.
 - **Freeboard-SK** – Adds the Freeboard-SK chart plotter as a widget with automatic sign-in.
 - **Autopilot Head** – Typical autopilot controls for compatible Signal K Autopilot devices.
-- **Realtime Data Chart** – Visualizes data on a real-time chart with actuals, averages, and min/max.
+- **Realtime Data Plot** – Visualizes data on a real-time plot with actuals, averages, and min/max.
 - **AIS Radar**: Display AIS targets with range rings, interactive target details, and quick zoom and filtering controls.
 - **Embed Webpage Viewer** – Embeds external web apps (Grafana, Node-RED, etc.) into your dashboard.
 - **Racesteer** – Race steering display fusing polar performance data with live conditions for optimal tactics.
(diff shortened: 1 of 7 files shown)


mairas:

Claims-accuracy + rename-safety lenses, adversarial-verified. **0 findings.** The keep-list was honored — no TypeScript identifier, component selector, **persisted config key**, CSS class, chart.js type, or nautical "chartplotter"/"nautical chart" was renamed; no user-facing time-series "chart" was missed. The #247 rewrite checks out against `chart-window.util.ts` (window-derived ~500-point model) and the history-dialog's still-live ~120 density (`widget-history-chart-dialog.component.ts:815`). Schema regeneration is blocked by the pre-existing #293, which is not part of the CI gate.

Issue #198, opened 2026-07-04
Rebrand user-facing docs KIP → Skip (help-docs + README body)


mairas:
## Problem

The product was rebranded to **Skip**, but the user-facing documentation still calls it **KIP** almost everywhere. This is the prose/content tier — distinct from the code-identifier rename in #192, which is deliberately scoped to runtime identifiers (CSS vars, selectors, symbols, log strings) and does not touch docs.

Two surfaces are affected:

### 1. In-app help docs (`src/assets/help-docs/*.md`) — shipped as Skip's own help

These render inside the app as Skip's help pages, yet refer to the product as KIP throughout. `welcome.md` — the first page a user sees — opens with "You just installed KIP". Files with `kip`/`KIP` references (occurrence counts):

| File | refs | File | refs |
|---|---|---|---|
| `welcome.md` | 14 | `putcontrols.md` | 7 |
| `embedwidget.md` | 16 | `kiosk.md` | 6 |
| `nodered-control-flows.md` | 12 | `contact-us.md` | 6 |
| `zones.md` | 11 | `configuration.md` | 4 |
| `community.md` | 9 | `contributing-widgets.md` | 4 |
| `history-api.md` | 8 | `datainspector.md` | 4 |
| `time-series.md` | 4 | `dashboards.md` | 2 |
| `video-widget.md` | 3 | | |

`community.md` and `contact-us.md` also point users to the **KIP Discord** and **`mxtommy/Kip` GitHub issues** for support.

### 2. `README.md` body

Below the Skip fork banner, line 9 reads *"Original Kip readme follows."* and lines 17–298 are the upstream Kip README verbatim — "KIP is the most popular Signal K MFD…", `KIPDemo.png`, the whole feature tour, contribution guide, and `mxtommy/Kip` links.

## This is a review, not a blind find-replace

`sed s/KIP/Skip/g` would introduce false claims:

- **Removed features.** Some docs describe upstream features Skip dropped — e.g. the Widget Historical Charts / history provider (removed in #176/#177) and the image widget (removed in #107/#108). `history-api.md`, `time-series.md`, and the README's history-charts paragraph should be **corrected or removed**, not rebranded.
- **Support links.** `mxtommy/Kip` issue links should point at `halos-org/skip` (users filing Skip bugs shouldn't land upstream). Skip has no Discord of its own — decide whether to drop the Kip Discord link or keep it as an upstream-community pointer, rather than relabel it "Skip Discord".

## Scope

**In:**
[...]

Issue #128, opened 2026-07-02
Settings entry button is labeled "Options" — rename to "Settings"


mairas:
## Problem

The button that opens the settings page is labeled **Options**, but the page it opens calls itself **Settings**. The same destination has two names depending on where you look:

- Actions page button: `label="Options"` with `svgIcon="settings"` (`src/app/core/components/settings/settings.component.html:5`)
- Route: `/options`, browser title "KIP - Options" (`app.routes.ts:25`)
- Destination page header: `pageTitle = "Settings"` with the same settings icon (`options/tabs/tabs.component.ts:27`)
- All tab components on that page are named `settings-*` (settings-signalk, settings-display, settings-notifications, settings-units, settings-config)

## Proposal

Standardize on **Settings**, matching the destination page's own title, the icon, and platform convention:

- Button label "Options" → "Settings"
- Browser title "KIP - Options" → "Skip - Settings" (the KIP prefix looks stale for Skip anyway)
- Route path `/options` → `/settings`. No redirect needed — Skip has no users yet. Update the references: `app.routes.ts`, two navigations in `app-initNetwork.service.ts`, and the tutorial widget's `routerLink="/options"`.
- Tutorial widget link text "Connectivity options" can stay — lowercase "options" as a generic word is fine

**Optional, decide separately:** the internal naming swap — `settings.component` is actually the actions overlay while `options/` is the settings page. That's a file/directory rename that adds upstream KIP replication friction (#38), so it shouldn't block the fix above.

mairas:
Scope update: dropped the redirect (Skip has no users yet) and moved the `/options` → `/settings` route rename into the main proposal — without the redirect it is a four-reference change. Internal file naming swap remains a separate decision due to upstream replication friction.

mairas:
Partially addressed in #296 (merged) — the user-facing half is done:

- ✅ Actions-overlay button label **"Options" → "Settings"** (icon was already `settings`; the page already titles itself "Settings").
- ✅ Browser title **"KIP - Options" → "Skip - Settings"** for the `/options` route (and the other stale `KIP - …` titles de-KIP'd).

[...]

PR #295, opened 2026-07-15
Fix false history-provider claim and rebrand user-facing KIP prose in templates

### LINKED ISSUE #285: User-facing KIP prose in app templates (incl. false "built-in KIP History API provider" claim)
## Context

Surfaced by the F3 (#198) docs-rebrand review. #198 rebranded the help-docs + README (prose) but scoped **app-template** strings to #192 (internal `kip` identifier rename). However several Angular templates carry **user-facing "KIP" prose** — distinct from `--kip-*` CSS identifiers — and one is a **stale false claim**, so they warrant a prose pass (the internal-identifier rename in #192 won't necessarily fix user-visible wording, and shouldn't be conflated with a correctness fix).

## The false claim (priority)

- `src/app/core/components/widget-history-chart-dialog/widget-history-chart-dialog.component.html:22` — reads *"the built-in **KIP** History API provider."* **Skip has no built-in history provider** — it is a pure History-API reader served by an external provider (`signalk-to-influxdb2` / `signalk-parquet`) and renders an empty state when none is present (see CLAUDE.md "History & charts"). This misleads users about where trend data comes from. Fix the wording to reflect the external-provider reality.

## Other user-facing "KIP" prose in templates

- `remote-control`, `display`, `config`, `signalk.component` (`<pre>KIP: {{version}}`), `upgrade-config` (incl. a `Signal K #kip` Discord link), `root-modal-widget-config`, and the app route titles (`KIP - Settings`).

## Disposition

Rebrand the user-facing prose KIP → Skip and correct the false provider claim — either folded into #192 (same files) with a distinct commit, or as a small dedicated prose pass. The false "built-in KIP History API provider" line should be prioritized regardless of scheduling.

Refs #198, #192.

mairas:
## Why

Two related issues in user-facing Angular template prose:

1. **A false claim (priority).** The history-chart dialog's empty state told users about *"the built-in KIP History API provider"* — but Skip ships **no** server-side history provider. It is a pure History-API *reader*; trend data comes from an **external** Signal K provider (`signalk-to-influxdb2` or `signalk-parquet`). The old wording misled users about where their data comes from and what to fix when a chart is empty.
2. **Stale "KIP" product prose.** F3 (#198) rebranded the help docs + README but deliberately scoped app-template strings out. Several templates still say "KIP" in user-visible sentences.

## What

- **Corrected the empty-state message** (`widget-history-chart-dialog.component.html`) to accurately point users at an external Signal K history provider they must install/configure to support the widget's path.
- **Rebranded user-facing "KIP" → "Skip"** across `signalk`, `configuration`, `display`, `remote-control`, and `upgrade-config` templates (version label, prose sentences).
- In `remote-control`, also updated the instruction path "Options → Display" → "**Settings → Display**" to stay consistent with the sibling Options→Settings rename (#128).

## Deliberately preserved

- The external **Signal K `#kip` Discord channel** link (a real upstream community channel — there is no `#skip` channel to point to).
- The persisted on-disk config dir names `.../users/(username)/kip/` and `.../global/kip/` (storage identifiers, not prose).
- All `--kip-*` CSS variables, TypeScript identifiers (`IKipDisplayInfo`, `KipUUID`, `self.kip.remote.*`), selectors, and config keys.

Pure prose in `.html` templates — no behavior change, no `.ts` touched (betterer baseline unaffected). Lint passes.

Fixes #285.

### DIFF
--- a/src/app/core/components/options/configuration/config.component.html
+++ b/src/app/core/components/options/configuration/config.component.html
@@ -71,7 +71,7 @@
         </div>
 
         <div class="upload-txt">
-          Import a KIP configuration file as a <strong>new profile</strong>. This never overwrites an
+          Import a Skip configuration file as a <strong>new profile</strong>. This never overwrites an
           existing profile.
         </div>
         <div class="upload-btn btn-div">
--- a/src/app/core/components/options/display/display.component.html
+++ b/src/app/core/components/options/display/display.component.html
@@ -82,14 +82,14 @@
           </mat-panel-title>
           @if (!isPhonePortrait().matches) {
             <mat-panel-description>
-              Configure remote dashboard operations between KIP instances.
+              Configure remote dashboard operations between Skip instances.
             </mat-panel-description>
           }
         </mat-expansion-panel-header>
         <mat-slide-toggle #isRemoteControlCb class="full-width"
           [(ngModel)]="isRemoteControl"
           name="isRemoteControl">
-            Allow remote KIP instances to view and control this dashboard.
+            Allow remote Skip instances to view and control this dashboard.
         </mat-slide-toggle>
         <mat-form-field class="optional-field">
           <mat-label>Name of this instance</mat-label>
--- a/src/app/core/components/options/signalk/signalk.component.html
+++ b/src/app/core/components/options/signalk/signalk.component.html
@@ -53,7 +53,7 @@
       <canvas class="activity-graph" #activityGraph></canvas>
     </div>
     <div class="info-container">
-      <pre>KIP: {{ app.appVersion() }}
+      <pre>Skip: {{ app.appVersion() }}
 {{ app.browserVersion() }}, {{ app.osVersion() }}<br />
 {{ endpointServiceStatus().serverDescription }}
 {{ streamStatus().message }}, {{ isUserSession() ? 'Signed in' : 'Not signed in' }}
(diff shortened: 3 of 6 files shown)


mairas:
## Multi-persona review synthesis (claims + scope-standards, adversarial verify)

### Claims-truth: the false-provider fix is accurate ✓
The rewritten empty-state message was verified against `CLAUDE.md` "History & charts" and the history-api client/mapper: Skip ships **no** built-in/server-side history provider, data comes from an external SK provider (signalk-to-influxdb2 / signalk-parquet), and the widget renders an empty state when none is present. The new wording is factually correct and keeps the actionable "supports the widget's path" guidance. No overclaim introduced by the other rebrands.

### Scope-standards: 1 confirmed P3 — now addressed
A reviewer flagged the remote-control help breadcrumb. After the sibling rename (#296) renames the config tile "Options"→"Settings", the original "Options → Display" wording would be stale, and the reviewer's own suggested "Settings → **Options** → Display" fix would *also* be stale (that tile no longer says "Options"). The toolbar entry to that overlay is an **icon-only** gear (aria-label "Settings", no visible text), so a chained "Settings → Display → Remote Control" breadcrumb is ambiguous.

**Fix applied:** reworded to *"on that Skip device open Settings, then enable Remote Control on the Display tab."* — unambiguous and accurate against the merged UI (the config page is titled "Settings", has a "Display" tab, which contains a "Remote Control" panel with the enable toggle). Amended into the prose commit.

Integrity checks passed: only prose changed (no identifiers/CSS/config-keys/selectors); the external Signal K `#kip` Discord channel and the on-disk `kip/` config paths were correctly preserved; no `.ts` touched. Remaining raw findings refuted.

Issue #180, opened 2026-07-03
Include the widget type in the Widget Settings dialog title


mairas:
The widget configuration dialog uses a single static title for every widget. When several widgets are on a dashboard, the dialog gives no indication of *which* widget is being configured. The title should include the widget type, e.g. **"Numeric — Widget Settings"** or **"Widget Settings: Numeric"**.

## Where

- Title is set/rendered in `src/app/widget-config/root-modal-widget-config/root-modal-widget-config.component.ts` (`titleDialog`, bound in `root-modal-widget-config.component.html:2`).
- The dialog receives the widget config via `MAT_DIALOG_DATA` (`widgetConfig: IWidgetSvcConfig`). The opening host has the widget instance and its `type`:
  - `src/app/core/components/widget-host2/widget-host2.component.ts` — `this.widgetProperties.type`
  - `src/app/core/components/group-widget/group-widget.component.ts`

## Friendly name source

The human-readable widget name lives in the widget registry, keyed by `type`:

- `src/app/core/services/widget.service.ts` — each entry has `{ name, category, ... }` (e.g. type → `name: 'Numeric'`, `'Autopilot'`, `'Date & Time'`). Resolve `widgetProperties.type` to that `name` and compose the title, rather than showing the raw type string.

The widget type may need to be passed through the dialog-open data (alongside `config`) so the modal can render it.

## Related

Builds on #179 (the "Widget Options" → "Widget Settings" rename). Coordinate the final wording so both land consistently.

mairas:
Implemented in #296 (merged). The widget config dialog title now composes the widget's friendly registry name in front of the base title — e.g. **"Numeric — Widget Settings"** — resolved via `WidgetService.getWidgetName(selector)`, falling back to plain "Widget Settings" for types without a registry entry (e.g. the group widget). (Did not auto-close from the PR's "Fixes #179, #180" — GitHub only parsed the first number.)

Issue #38, opened 2026-06-30
[triage] 79 lower-confidence review findings for inspection


mairas:
## Lower-confidence review findings — parked for triage

This issue consolidates the **79 findings from the 2026-06-30 codebase review that were _not_ independently confirmed** — i.e. those tagged *not verified*, *subjective*, or *info* in the review. They are collected here (rather than filed individually) so they can be inspected and either promoted to their own issue or dismissed.

The **35 confirmed / partially-confirmed findings** are filed as individual issues (label `kip-review`, with `severity:*` and `area:*`). Each item below carries its review severity and confidence tag; treat severity as provisional until the finding is verified against the current fork.

Checkboxes are for triage tracking: tick once a finding has been inspected (promoted to an issue or closed out).

<sub>Source: `Kip/_local/2026-06-30-kip-codebase-review.md` (full per-dimension appendix). Findings reference paths on the `profiles` branch the fork was created from.</sub>

### Architecture & data flow

- [ ] **ARCH-04** - medium, not-verified - AppNetworkInitService (~490 lines) concentrates all bootstrap phases behind interacting boolean flags in a single try/catch/finally, making correct sequencing hard to verify. _Fix:_ Extract discrete phases (loadConfig, establishConnection, resolveAuth, bootstrapStorage, seedDatasets, finalizeConnection) returning explicit outcomes, replacing the boolean-flag finally coordination with a sequenced promise chain. `src/app/core/services/app-initNetwork.service.ts`
- [ ] **ARCH-05** - medium, not-verified - AppComponent injects 19 services and performs imperative orchestration of notification overlays, bootstrap-recovery toasts, connection toasts, and sidenav state in the constructor, mixing cross-cutting concerns in the root component. _Fix:_ Move notification-overlay sequencing and bootstrap-issue/connection toast policies into focused services, leaving AppComponent as a thin shell that wires gestures and renders. `src/app/app.component.ts`
- [ ] **ARCH-06** - medium, not-verified - DataService (~797 lines) acts as a god-service carrying the path map, notification routing, dual full-tree caches with manual index maps and debounce timers, meta merging, and delta-rate statistics. _Fix:_ Extract the full-tree projection and delta-statistics into separate services that observe DataService, keeping DataService focused on the path map and per-path subjects. `src/app/core/services/data.service.ts`
[...]

PR #296, opened 2026-07-15
Standardize on 'Settings' terminology and add widget type to config dialog title

### LINKED ISSUE #179: Rename widget config dialog title "Widget Options" to "Widget Settings"
The widget configuration dialog is titled **"Widget Options"**. It should read **"Widget Settings"** for consistency with the rest of the UI's "Settings" terminology (see #128, which renames the app-level Settings entry button away from "Options").

## User-facing string

The rendered title comes from `titleDialog` in the root modal config component:

- `src/app/widget-config/root-modal-widget-config/root-modal-widget-config.component.ts:52` — `public titleDialog = "Widget Options";` (bound in the template at `root-modal-widget-config.component.html:2` via `<h6 mat-dialog-title>{{ titleDialog }}</h6>`)

The `title: 'Widget Options'` string is also passed when the dialog is opened; update these for consistency even if the rendered value is driven by `titleDialog`:

- `src/app/core/components/widget-host2/widget-host2.component.ts:303`
- `src/app/core/components/group-widget/group-widget.component.ts:85`

## Scope

User-facing label only. Internal identifiers, method names (`openWidgetOptions`), and code comments referencing "widget options" can stay as-is; this issue is about the visible dialog title.

mairas:
Bundles #128, #179, #180 — all UX-consistency work around the "Settings" wording.

## What

**#128 — "Options" → "Settings":**
- The actions-overlay button that opens the config page: `label="Options"` → `label="Settings"` (the page already titles itself "Settings"; the icon is already `settings`).
- The `/options` route's browser title `'KIP - Options'` → `'Skip - Settings'`, and the other stale KIP browser titles de-KIP'd (`'Skip - Remote Control'`, `'Skip - Help'`, `'Skip - Data Inspector'`).

**#179 — "Widget Options" → "Widget Settings":** the config dialog's base title and the `title:` strings passed at both open sites. Also rebrands the Embed-widget warning prose "normal KIP" → "normal Skip".

**#180 — widget type in the dialog title:** the dialog now composes the widget's friendly registry name in front of the base title, e.g. **"Numeric — Widget Settings"**, falling back to plain "Widget Settings" for types without a registry entry (e.g. the group widget). Added `WidgetService.getWidgetName(selector)` next to the existing selector-keyed lookups.

## Deliberate decisions (please sanity-check)

1. **Route path `/options` → `/settings` is NOT done — deferred.** #128 proposed it, but `/settings` **already exists** (the actions/home overlay, `settings.component`). Renaming the path would collide, and resolving that requires the internal actions/settings naming swap that #128 itself defers (#38 friction). So the actions-overlay route (`/settings`, title left as `'KIP - Settings'`) is untouched — leaving that one stale "KIP" is intentional; de-KIPing it to "Skip - Settings" would duplicate the `/options` title. The path is invisible to users; the user-facing button label + tab title are fixed. Route-path rename tracked with the #38 swap.
2. **How `widgetName` reaches the dialog.** `DialogService.openWidgetOptions` only forwards `data.config`, and `DialogWidgetOptionsData.config` is typed `object`. Rather than widen the dialog-service data type (out of this packet's file scope), `widgetName` is carried **on** the config payload (spread into a fresh object — the live widget config is never mutated) and **stripped before form generation** (`delete formConfig.widgetName`), so it is never turned into a form control and never persisted: `submitConfig()` returns `formMaster.getRawValue()`, which is built from the stripped config. It is purely a title hint.

## Verification

Lint ✓, betterer:ci ✓ (183 → **182**, one issue incidentally fixed), and the new `root-modal-widget-config.component.spec.ts` title-composition tests pass locally (6/6). Pure user-facing string + title-composition change; no data-model or persistence change.

Fixes #179, #180. Addresses #128 (button label + browser title; route-path rename deferred to the #38 naming swap — see decision 1).

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -62,9 +62,6 @@ exports[`strictNullChecks`] = {
     "src/app/core/components/data-inspector/data-inspector.component.ts:2942185648": [
       [71, 4, 20, "Type \'Signal<BreakpointState | undefined>\' is not assignable to type \'Signal<BreakpointState>\'.\\n  Type \'Signal<BreakpointState | undefined>\' is not assignable to type \'() => BreakpointState\'.\\n    Type \'BreakpointState | undefined\' is not assignable to type \'BreakpointState\'.\\n      Type \'undefined\' is not assignable to type \'BreakpointState\'.", "1507091964"]
     ],
-    "src/app/core/components/group-widget/group-widget.component.ts:2421937404": [
-      [85, 8, 6, "Type \'IWidgetSvcConfig | undefined\' is not assignable to type \'object\'.\\n  Type \'undefined\' is not assignable to type \'object\'.", "1544266319"]
-    ],
     "src/app/core/components/options/notifications/notifications.component.ts:3253454142": [
       [41, 4, 20, "Type \'Signal<BreakpointState | undefined>\' is not assignable to type \'Signal<BreakpointState>\'.\\n  Type \'Signal<BreakpointState | undefined>\' is not assignable to type \'() => BreakpointState\'.\\n    Type \'BreakpointState | undefined\' is not assignable to type \'BreakpointState\'.\\n      Type \'undefined\' is not assignable to type \'BreakpointState\'.", "1507091964"],
       [47, 4, 24, "Object is possibly \'undefined\'.", "2222700196"],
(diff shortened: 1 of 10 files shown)


mairas:
## Multi-persona review synthesis (5 lenses + adversarial verify)

Lenses: correctness, **adversarial**, rename-safety, maintainability, testing. Each raw finding was independently adversarially verified (default-refute).

### Headline: the `widgetName`-on-config design is sound
An adversarial reviewer was tasked specifically to **prove `widgetName` can leak into persisted widget config** — tracing every path from `config: {…liveConfig, widgetName}` through the dialog to `submitConfig()`. **Verdict: REFUTED.** The spread creates a fresh object (live config never mutated), the dialog strips `widgetName` before form generation, and `submitConfig()` persists `formMaster.getRawValue()` (built from the stripped config). No leak path exists.

### Confirmed findings against this diff: 0
The only confirmed finding (a P3) is that `remote-control.component.html`'s help text ("open Options → Display → …") goes stale once this PR renames that tile to "Settings". That file isn't in this diff — it's owned and fixed in the sibling PR #295 (reworded to "open Settings, then enable Remote Control on the Display tab"). No change needed here.

6 of 8 raw findings were refuted (theoretical, pre-existing, or already-handled).

### CI fix applied
The initial push red-failed 3 existing `widget-host2.component.spec.ts` tests: this PR adds a `widgetService.getWidgetName(...)` call to the dialog-open path, and the spec's **mock** `WidgetService` lacked that method, so the call threw before `openWidgetOptions`. Fixed by adding `getWidgetName` to the mock (commit in `d970652e`). `group-widget`/`widget-embedded`/`dashboard-history-series-sync` specs were unaffected (real root `WidgetService` or no exercise of the path); both touched components' specs now pass locally (19/19).

### Deliberate decisions reviewers were asked to sanity-check (both held up)
- **Route path `/options`→`/settings` deferred** (collides with the pre-existing `/settings` actions overlay; entangled with the #38 naming swap). No user-facing "Options" label was left mismatched — the only such string was the remote-control help, fixed in #295.
- **`/settings` overlay title left `KIP - Settings`** to avoid duplicating the new `Skip - Settings` on `/options`. Reviewers found no place both titles surface confusingly.

Issue #25, opened 2026-06-30
[TS-04] Widget-config layer fully opts out of typed reactive forms, silently breaking on config key renames


mairas:
**Severity:** medium | **Confidence:** confirmed | **Area:** type-safety | **Source:** codebase review 2026-06-30, finding `TS-04`

## Problem

All 13 widget-config components use `UntypedFormGroup`, `UntypedFormControl`, `UntypedFormArray`, or `UntypedFormBuilder` — the Angular escape hatches that carry no compile-time type information. The config forms map directly onto `IWidgetSvcConfig`, but the form layer has no compile-time link to that interface: control names such as `convertUnitTo` and `pathSkUnitsFilter` are bare string literals inside `fb.group({...})` calls, so renaming or removing a key in `IWidgetSvcConfig` produces no type error and silently desyncs the form. `paths-options.component.ts` compounds this with `protected pathsFormGroup!: any` (lines 36–37, both a non-null assertion and an `any` type), then calls the `FormArray`-only `.push()` method on that field regardless of its runtime shape; only a runtime `isArray()` flag prevents a crash when the value is actually a `FormGroup`.

## Impact

Any refactor of the large `IWidgetSvcConfig` interface — adding, renaming, or removing a config key — will not produce a compiler error in the form layer. Bugs introduced this way will surface only at runtime, potentially presenting as silent data loss or misconfigured widgets to end users. The `any`-typed field in `paths-options.component.ts` introduces a latent crash path: if the `isArray()` guard is ever bypassed or incorrectly evaluated, calling `.push()` on a `FormGroup` will throw at runtime. In a marine/SKip deployment where the widget configuration layer drives critical instrument display, silent desync between the form and the backing interface is a reliability risk.

## Affected code

- `src/app/widget-config/paths-options/paths-options.component.ts:25-46`
- `src/app/widget-config/paths-options/paths-options.component.ts:36-37` (`protected pathsFormGroup!: any`)
- `src/app/widget-config/paths-options/paths-options.component.ts:49-50` (`.push()` on `any`-typed field)
- `src/app/widget-config/root-modal-widget-config/root-modal-widget-config.component.ts`
- `src/app/widget-config/path-control-config/path-control-config.component.ts`

## Suggested direction

[...]

PR #297, opened 2026-07-15
Type the paths-options form group, removing the widget-config `any` (#25 Phase 0)

mairas:
First slice of the #25 typed-forms migration (**Phase 0 — quick win**). See the [plan](https://github.com/halos-org/skip/issues/25#issuecomment-4977756811).

## Why

`paths-options.component.ts` held `protected pathsFormGroup!: any` — the **sole** `@typescript-eslint/no-explicit-any` suppression in the widget-config layer, and a latent crash path (`.push()` is FormArray-only but the field can be a FormGroup). The `any` existed because the same field is a `FormArray` for multi-control widgets and a `FormGroup` otherwise, and the template binds it four ways that a naive union breaks under `strictTemplates`.

## What

- Field retyped to a discriminated `UntypedFormArray | UntypedFormGroup`.
- Typed accessors so `strictTemplates` is satisfied with **zero runtime change** (casts are compile-time only; the bound values are identical): `pathsGroup` (FormGroup view for the `[formGroup]` container + the Record/single-path branch), `pathControlGroups` (the per-path element groups for the array branch), `pathGroupByKey(key)` (the Record branch's element lookup).
- `addPath()`'s FormArray-only `.push()` is now guarded by an `instanceof UntypedFormArray` check (defends the FormGroup case that previously would have thrown).

## Premise correction (from the #25 analysis)

The issue lists `path-control-config.component.ts` as also carrying an `any` — it does **not** (already `input.required<UntypedFormGroup>()`). There is exactly one `any` in the named pair, and it's the one removed here. A genuine typed conversion of `path-control-config` is hard (runtime-added `sampleTime`, numeric-only `convertUnitTo`, `parent.parent` sibling access) and is out of scope for #25.

## Verification

Lint clears (no-explicit-any gone), **`build:dev` passes** (the real strictTemplates gate), the paths-options spec passes, and `.betterer.results` regenerated (182 held; content-hash re-anchored). No behavior change — the discriminant (`isArray()`) provably tracks the runtime shape the parent builder produces.

Part of #25 (Phase 0). Does not close the issue.

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -131,13 +131,13 @@ exports[`strictNullChecks`] = {
       [119, 4, 20, "Object is possibly \'undefined\'.", "2522653710"],
       [120, 4, 20, "Object is possibly \'undefined\'.", "2522653710"]
     ],
-    "src/app/widget-config/paths-options/paths-options.component.ts:3600643100": [
+    "src/app/widget-config/paths-options/paths-options.component.ts:3797853945": [
       [31, 51, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'IAddNewPathObject\'.", "2620553983"],
       [32, 40, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'string\'.", "2620553983"],
       [33, 54, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'IDynamicControl[]\'.", "2620553983"],
-      [42, 28, 49, "Object is possibly \'null\'.", "3725617003"],
-      [66, 26, 49, "Object is possibly \'null\'.", "3725617003"],
-      [70, 26, 49, "Object is possibly \'null\'.", "3725617003"]
+      [58, 28, 49, "Object is possibly \'null\'.", "3725617003"],
+      [85, 26, 49, "Object is possibly \'null\'.", "3725617003"],
+      [89, 26, 49, "Object is possibly \'null\'.", "3725617003"]
     ],
     "src/app/widget-config/root-modal-widget-config/root-modal-widget-config.component.ts:104802355": [
       [80, 4, 11, "Type \'{ label: string; value: string; }[]\' is not assignable to type \'never[]\'.\\n  Type \'{ label: string; value: string; }\' is not assignable to type \'never\'.", "3027418051"],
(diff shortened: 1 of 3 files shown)


mairas:
## Review synthesis (correctness + adversarial, both read-only)

**Zero confirmed findings.** Both reviewers independently traced the full path and confirmed the zero-runtime-behavior-change claim holds:

- **Accessor equivalence:** `pathsGroup` / `pathControlGroups` / `pathGroupByKey(key)` return byte-identical runtime values (same object references) to the old `any`-typed bindings — the `as` casts are compile-time only. `[formGroup]="pathsGroup"` binds the identical FormArray instance in the multi-control case, exactly as before; children receive their group via explicit `[pathFormGroup]` `@Input` (not `formControlName`), so the FormGroup-vs-FormArray directive path that could diverge is never exercised.
- **`isArray()` ↔ runtime shape cannot disagree:** in the parent builder, both `isPathArray` and the `paths` control's runtime type are decided in the same pass by the same condition (`multiChildCtrls !== undefined` → `fb.array([])` + `isPathArray=true`; else `fb.group({})`). So each branch's cast always matches the runtime type.
- **The `instanceof` guard is pure defense:** `addPath()` is reachable only via `boolean-multicontrol-options`'s `(addPath)` output, rendered only under `@if(multiChildCtrls !== undefined)` — the same condition that makes `paths` a FormArray. So the guard's early-return is unreachable; it never drops a path add. Verified `instanceof UntypedFormArray` is exact (`const UntypedFormArray = FormArray`, single bundle).
- **betterer diff** is only a content-hash rekey + line-number shifts of the three pre-existing "possibly null" entries — no logic change.

### Documented residual (not a defect in this diff)
The `instanceof` guard couples correctness to the parent-template invariant "`multiChildCtrls` present ⟺ `paths` is a FormArray." Today that holds by construction. If a *future* change ever wires `addPath` for a single-path widget, the failure mode shifts from a loud `push is not a function` crash to a **silent no-op**. Worth a note for whoever touches that wiring; no action needed now.

CI green (Node-24 suite), lint, `build:dev` (strictTemplates), spec, and betterer all pass.

PR #298, opened 2026-07-15
Type the leaf chart/datetime config inputs to FormControl<T> (#25 Phase 1)

mairas:
Second slice of the #25 typed-forms migration (**Phase 1 — leaf input typing**). See the [plan](https://github.com/halos-org/skip/issues/25#issuecomment-4977756811).

## What

Retyped the form-control `@Input`s of the three leaf config children — `dataset-chart-options`, `display-chart-options`, `display-datetime` — from `UntypedFormControl` to Angular's typed `FormControl<T>`, and retyped the **29 feeder getters** in `root-modal-widget-config` that supply them to the matching `FormControl<T>`. Each `T` is derived from the corresponding `IWidgetSvcConfig` field (e.g. `color: string`, `numDecimal: number`, `filterSelfPaths: boolean`, `convertUnitTo: string | null`, `datachartAngleRange: 'signed' | 'direction' | null`). This gives compile-time value typing inside the leaf components; `.value` is now `T` instead of `any`.

`formMaster` itself stays untyped (that's Phase 3), so the getters still cast from it — but a child input and its feeder now share the same `FormControl<T>`, which `strictTemplates` enforces at the binding.

## Scope notes
- **`boolean-control-config` is NOT here** — it receives a `FormGroup` (IDynamicControl shape), not named FormControls, so it moves to Phase 2 (a plan refinement vs. the original grouping).
- `paths-options` untouched (Phase 0); its inputs stay `UntypedFormControl` and still accept the now-typed getters (`FormControl<T>` widens to `FormControl<any>`).
- No template changes — only TS types.

## One real fix surfaced by the stronger typing
Typing `datachartPath` as `FormControl<string | null>` made `.value` correctly `string | null`, which exposed that `dataset-chart-options` passed it to `getPathObject`/`getConversionsForPath` (which want `string`) without narrowing. Fixed by capturing the guarded value in a `const` so the truthy check narrows it to `string` (also de-duplicates three `this.datachartPath()` reads; behavior identical). This kept betterer at **182** (no strictNullChecks regression).

## Verification
Lint ✓ · `build:dev` (strictTemplates, the real gate) ✓ · betterer `182` held ✓ · all 4 modified components' specs pass (16/16). No behavior change.

Part of #25 (Phase 1). Does not close the issue.

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -113,15 +113,15 @@ exports[`strictNullChecks`] = {
       [37, 10, 26, "Type \'null\' is not assignable to type \'Subscription\'.", "273497119"],
       [79, 8, 13, "Type \'null\' is not assignable to type \'string | undefined\'.", "2466344431"]
     ],
-    "src/app/widget-config/dataset-chart-options/dataset-chart-options.component.ts:2604195000": [
-      [82, 26, 10, "Argument of type \'ISkPathData | null\' is not assignable to parameter of type \'ISkPathData\'.\\n  Type \'null\' is not assignable to type \'ISkPathData\'.", "384538813"]
+    "src/app/widget-config/dataset-chart-options/dataset-chart-options.component.ts:155376543": [
+      [83, 26, 10, "Argument of type \'ISkPathData | null\' is not assignable to parameter of type \'ISkPathData\'.\\n  Type \'null\' is not assignable to type \'ISkPathData\'.", "384538813"]
     ],
-    "src/app/widget-config/display-chart-options/display-chart-options.component.ts:333710162": [
+    "src/app/widget-config/display-chart-options/display-chart-options.component.ts:2641875525": [
       [48, 4, 11, "Type \'{ label: string; value: string; }[]\' is not assignable to type \'never[]\'.\\n  Type \'{ label: string; value: string; }\' is not assignable to type \'never\'.", "3027418051"]
     ],
-    "src/app/widget-config/display-datetime/display-datetime.component.ts:1505324315": [
-      [45, 50, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'UntypedFormControl\'.", "2620553983"],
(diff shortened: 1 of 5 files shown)


mairas:
## Review synthesis (correctness, read-only)

**Zero findings.** The reviewer cross-checked the full T-mapping and traced the one refactor:

- **T-mapping — all 29 getters + child inputs match `IWidgetSvcConfig`.** Every control's `T` was verified against its declared field type by the `.get('key')` string: the `| null` fields (`convertUnitTo`/`datachartPath`/`datachartSource`) carry `| null`; `datachartAngleRange` is exactly `'signed' | 'direction' | null`; `datasetAverageArray` is correctly `string` (a string enum, not an array, despite the name); every `yScale*`/`numDecimal` is `number`, every `show*`/`enable*` is `boolean`. No wrong-type rows — the "both-sides-consistently-wrong-T" hazard (which strictTemplates can't catch) does not occur.
- **The narrowing fix is behavior-identical.** Between the three original `this.datachartPath().value` reads the old code only ran the truthy guard, `getPathObject` (pure), and `setPathSources` (mutates `pathSources`/`datachartSource`, never `datachartPath`); `valueChanges` is `debounceTime(300)` async and can't fire synchronously in `ngOnInit`. So the single captured `const` returns the same value; guard semantics identical.
- **No runtime dependence on the old `any`** — annotation-only; every `.value` returns the same runtime value.

### Documented residuals (not defects; relevant to later phases)
1. strictTemplates enforces that a getter's `T` and its bound input's `T` *agree*, but not that the template feeds the *right* getter to each input — a same-typed mis-wire would still compile. No templates changed here, so any such case is pre-existing.
2. The scalar controls are now typed non-nullable while `formMaster` is still built untyped and `.reset()` sets `null`. Invisible with strictNullChecks off; a future per-file strict migration would surface a `.value`-typed-non-null-but-runtime-null gap. Worth remembering when strict is eventually turned on for these files.

CI + lint + `build:dev` (strictTemplates) + betterer (182) + 16/16 specs all green.

PR #299, opened 2026-07-15
Bring plugin/ under lint + type-check gates (#268)

### LINKED ISSUE #268: Extend lint / strictNullChecks / no-any coverage to plugin/
Surfaced by the review of PR #265 (#221). The bundled Freeboard-SK panel plugin under `plugin/` — now including `plugin/index.spec.ts` and `plugin/vitest.config.ts`, plus `plugin/index.js` — sits **outside** the app gates: `ng lint`, the `strictNullChecks` betterer ratchet, and `@typescript-eslint/no-explicit-any` only cover `src/`. So type/lint regressions in the shipped plugin (which is referenced by `package.json` `main`) are unguarded.

Follow-up: bring `plugin/` under lint + type checking (extend the eslint flat config / a tsconfig include, or a dedicated lint step wired into `npm run ci`), so the plugin code and its tests are held to the same bar as the app.

mairas:
The bundled Freeboard-SK panel plugin under `plugin/` — referenced by `package.json` `main` and shipped in the npm package — sat **outside** the app quality gates: `ng lint`, `@typescript-eslint/no-explicit-any`, and the type-checker only covered `src/`, and `test:plugin` ran plain Vitest (esbuild transpile, **no type-check**). So `plugin/index.spec.ts` got zero type-checking and neither `.js` nor `.ts` was linted. This closes that gap.

## Changes (config-only — no plugin code touched)
- **`angular.json`** — extend `lintFilePatterns` with `plugin/**/*.ts` and `plugin/**/*.js`.
- **`eslint.config.js`** — add a `plugin/**/*.js` block (`eslint.configs.recommended` + `sourceType: commonjs` + `globals.node`) so `index.js` is linted (CommonJS `require`/`module` allowed; genuine undefined globals flagged). The existing unanchored `**/*.ts` block already covers the plugin specs.
- **`plugin/tsconfig.json`** (new) — a `noEmit` type-check surface for the plugin `.ts` files at the app's default (non-strict) bar, wired into `test:plugin` via `tsc --noEmit && vitest run`.
- **`package.json`** — declare `globals` as an explicit devDependency (it was transitive-only; `eslint.config.js` now `require`s it directly). Lockfile updated with just that one root-dep line.

## Why not the betterer/strictNullChecks route
The design considered extending the strictNullChecks ratchet to `plugin/`, but `plugin/`'s only `.ts` file is a **spec** — and the ratchet excludes all `*.spec.ts` everywhere — so that would add zero files (a no-op that looks like coverage). The real gap was "not type-checked **at all**," which a plain `tsc --noEmit` fixes.

## Verification (local, Node 22; CI re-runs on Node 24)
- `ng lint` — plugin now included, all files pass; **coverage proven** by a temporary `no-undef` probe in `index.js` (caught) then reverted.
- `tsc --noEmit -p plugin/tsconfig.json` — clean.
- `npm run test:plugin` — new `tsc` step + 7 vitest tests pass.
- `npm run betterer:ci` — unchanged (scope untouched).

Closes #268.


mairas:
## Review synthesis (project-standards, read-only)

**Verdict: CI-safe and convention-correct, zero blocking findings.** The reviewer audited against `CLAUDE.md` and the CI wiring:

- **No CI regression.** The plugin specs are clean under the existing `**/*.ts` block (no `any` — the spec uses `unknown`; `processInlineTemplates` is a no-op with no `@Component`s). The new `plugin/**/*.js` block correctly resolves `require`/`module` and binds the only free identifiers; `.js`/`.ts` globs are disjoint.
- **The `globals` promotion is `npm ci`-safe** — it was already resolved as a transitive dev dep of `@eslint/eslintrc` satisfying `^14.0.0`, so the 1-line root-devDependency addition is byte-identical to what `npm install` would emit; the lockfile sync check passes.
- **`plugin/tsconfig.json`** extending the root config is fine — `angularCompilerOptions` are ignored by plain `tsc`, `es2020`+`bundler` is valid, `types:[node]` suffices (vitest symbols are explicitly imported). Non-strict bar is correct (the strict ratchet is `src/`-scoped; no betterer impact).
- **Conventions correct** — `test:plugin` is already in `ci` (Node 24), `tsc`-direct matches the `snc` pattern, CHANGELOG present.

**Addressed:** the reviewer noted the CHANGELOG line read slightly broad — `index.js` (the shipped runtime) gets **lint** coverage but not `tsc` type-checking (no `allowJs`). Reworded to state precisely: lint across the whole plugin incl. `index.js`, type-check on the TypeScript sources.

Non-blocking residual (left as-is): `.npmignore`'s bare `tsconfig.json` pattern prunes the new `plugin/tsconfig.json` from the npm tarball (desired); pre-existing `index.spec.ts`/`vitest.config.ts` still publish (pre-existing, immaterial to plugin load).

PR #300, opened 2026-07-15
Typed path/control factories, removing the last widget-config any (#25 Phase 2a)

mairas:
Third slice of the #25 typed-forms migration (**Phase 2a — typed leaf factories**). See the [plan](https://github.com/halos-org/skip/issues/25#issuecomment-4977756811) and the [Phase 2 split note](https://github.com/halos-org/skip/issues/25#issuecomment-4978351083).

## What

Typed the two genuinely-closed, single-behavior leaf shapes the reflective builder constructs:
- **`IWidgetPath`** (array-mode paths): `generatePathArray` + `generatePathFields` now build a typed `FormGroup<IWidgetPathControls>` with a `FormControl<T>` per field, `source`/`sampleTime` required, others plain. **This removes the last `any` in the widget-config layer** (`generatePathFields`'s `value: any` + its eslint-disable — #25's second `any`).
- **`IDynamicControl`** (multi-control entries): `generateCtrlArray` returns a typed `FormGroup<IDynamicControlGroup>` with `ctrlLabel` required. The shared `IDynamicControlGroup` control-map is exported from `widgets-interface.ts`.
- **`boolean-control-config`** is now fully typed — its `ctrlFormGroup` input is `FormGroup<IDynamicControlGroup>`. The parent binding needed no change (`UntypedFormGroup` = `FormGroup<any>` assigns cleanly).

## Byte-identical, guarded by a characterization test
The factories are `as`-typed views over the **exact same dynamically-built control trees** — same keys, validators, `disable()` behavior, and initial values (`new FormControl(value)` is runtime-identical to `new UntypedFormControl(value)`, only the static type differs). A new **characterization test** (`test(widget-config): lock leaf control/path shapes`) locks both shapes against drift: it asserts the multi-control `ctrlLabel` is required, and that a paths-array element makes `source`/`sampleTime` required but `path` not. It passes on the pre-refactor code and after.

## Deliberately deferred to Phase 2b
The recursion-built shapes (gauge/displayScale/autopilot/ais/electrical family) and record-mode path unification — record-mode paths intentionally lack the `source`/`sampleTime` required validators that array-mode applies, so unifying them is behavior-sensitive and gets its own PR.

## Verification
No `any` remains in `widget-config` · lint ✓ · `build:dev` (strictTemplates) ✓ · **betterer 182 → 179** (the stronger typing let three latent strictNullChecks issues be removed) · the characterization test + `boolean-control-config` + `boolean-multicontrol-options` specs pass.

Part of #25 (Phase 2a). Does not close the issue.

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -102,11 +102,10 @@ exports[`strictNullChecks`] = {
       [72, 6, 5, "Type \'number | null\' is not assignable to type \'number\'.\\n  Type \'null\' is not assignable to type \'number\'.", "173263814"],
       [78, 6, 5, "Type \'number | null\' is not assignable to type \'number\'.\\n  Type \'null\' is not assignable to type \'number\'.", "183473191"]
     ],
-    "src/app/widget-config/boolean-control-config/boolean-control-config.component.ts:3033536446": [
-      [25, 40, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
-      [26, 39, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
-      [33, 4, 11, "Type \'{ label: string; value: string; }[]\' is not assignable to type \'never[]\'.\\n  Type \'{ label: string; value: string; }\' is not assignable to type \'never\'.", "3027418051"],
-      [37, 79, 34, "Object is possibly \'null\'.", "3704244789"]
+    "src/app/widget-config/boolean-control-config/boolean-control-config.component.ts:37806614": [
+      [26, 40, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
+      [27, 39, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
+      [34, 4, 11, "Type \'{ label: string; value: string; }[]\' is not assignable to type \'never[]\'.\\n  Type \'{ label: string; value: string; }\' is not assignable to type \'never\'.", "3027418051"]
     ],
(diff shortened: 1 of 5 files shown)


mairas:
## Review synthesis (adversarial, read-only)

**Zero findings — could not break the byte-identical claim.** The reviewer traced every changed line against the original and walked the full delete flow:

- **Control-tree preserved.** All three factories keep the exact building logic — same keys (`Object.keys(formData)`), same validators (`source`/`sampleTime`/`ctrlLabel` required, nothing else), same `disable()` on `isPathConfigurable === false` (in the untouched caller), same initial values. `new FormControl(value)` and `new UntypedFormControl(value)` are the same runtime constructor (`UntypedFormControl` re-exports `FormControl`) — identical nullability/value/validators.
- **Untouched branches confirmed** — record-mode paths, gauge, autopilot, ais all still flow through the unchanged `generateFormGroups`.
- **No typed-handle hazard** — both typed FormGroups are immediately pushed into `UntypedFormArray`s (type erased) and later read via untyped `.get(...)`, so the `IWidgetPathControls` optional-key over-approximation is inert.

Two residuals, both rated **unreachable** (not defects):
1. `deleteControl`'s `.controls.pathID.value ?? ''` differs from the original only if `pathID` were null — but every `pathID` is a UUID (`UUID.create()`, persisted+reloaded), and `IDynamicControl.pathID` is a non-null `string`, so the branch never fires.
2. `IWidgetPathControls` types optional path keys as present controls, but no code reads the typed handle (discarded into the FormArray).

CI green; betterer improved 182→179; the characterization drift-guard + boolean-control-config + boolean-multicontrol-options specs pass.

PR #301, opened 2026-07-15
Make bootstrap Retry / recovery reloads offline-resilient (reachability-gated) (#272)

### LINKED ISSUE #272: Bootstrap Retry / reloadApp is not offline-resilient (reload can dead-end when the server is down)
## Context
Surfaced by the E1 (#190) multi-persona review. AppComponent's new bootstrap Retry toast — and the app's other four `SettingsService.reloadApp()` call sites — recover by reloading the page (`location.replace('./')`).

## Problem
On the exact scenario the bootstrap Retry targets — the Signal K server unreachable at startup — a reload re-fetches `index.html` from the (same-origin) server. Skip registers **no service worker / offline shell** (`manifest.json` alone gives no offline caching, and `index.html` is served no-cache). So if the server is still down when the user taps Retry, the reload fails and the browser replaces the loaded, still-useful degraded UI with a "site can't be reached" error page — strictly worse than staying put.

This affects every `reloadApp()` site, not just the new Retry toast; it is a pre-existing architectural gap that E1 makes more visible on the boat-critical no-server-at-startup path.

## Options
- **Offline shell:** register a service worker (ngsw) so the app shell is available offline; a reload then re-serves the cached shell, which re-attempts the connection.
- **In-place retry:** re-attempt the connection/bootstrap (e.g. re-run `AppNetworkInitService.initNetworkServices()` or trigger the state machine's HTTP retry) instead of a full reload, keeping the loaded app and clearing the toast only when the server actually returns. Needs a re-entrancy audit of `initNetworkServices()`.
- **Reachability-gated reload:** only reload once a lightweight reachability probe succeeds.

## Scope
Larger than E1 (touches the app's reload strategy and/or offline story). Filed as a follow-up per the review; E1 ships the `reloadApp()` seam and the corrected recovery messaging.

Discovered during the #257 campaign (E1 / #190 review).

mairas:
Fixes #272. See the [design](https://github.com/halos-org/skip/issues/272#issuecomment-4978130357).

## Problem
The bootstrap "Retry" toast and the app's reconfig/reset recoveries reload via `location.replace('./')`. On the exact scenario Retry targets — SK unreachable at startup — that reload re-fetches `index.html` from the down server; with no service worker, the browser replaces the working degraded shell with its "site can't be reached" page. Strictly worse than staying put.

## Fix — reachability-gated recovery
New `ReloadService.reload()` probes the server (same-origin `GET './'`, 4 s timeout, `no-store`) **before** reloading: reachable → the normal `location.replace('./')`; unreachable → keep the degraded shell and show a persistent 'warn' Retry toast that re-runs the gated reload (unbounded, user-driven, never leaves the app). `__KIP_TEST__` no-op preserved so specs never navigate.

Wired the bootstrap Retry toast and the three interactive `SettingsService` reloads (`resetConnection`, `setActiveProfile`, `resetSettings`) through the seam. **`reloadApp()` itself is unchanged** (raw `location.replace`) — the four config-upgrade completion reloads keep calling it, because they rely on the reload destroying the document to tear down their blocking `upgrading()` overlay; gating them would strand the overlay. Manual retry only; no service worker, no in-place re-hydration (both rejected in the design, with reasons).

## ⚠️ Design correction caught by on-boat testing
The original probe treated **any** HTTP response (incl. 4xx/5xx) as reachable. **On the boat, with SK stopped, Traefik (the reverse proxy that fronts SK at `:4430`) returns `502` — not a connection failure.** So "any response = reachable" would classify a down SK as reachable and reload straight into the 502 page — the very dead-end this guards against. Verified live:

| Signal K | `GET /@halos-org/skip/` via Traefik |
|---|---|
| up | **200** |
| stopped | **502** |

Fix: the probe treats **5xx as unreachable** (proxy/backend-down) while a 2xx/3xx/4xx means SK itself answered and the reload will render. This is the load-bearing behavior unit tests cannot cover (they mock `fetch`; they can't see Traefik).

## Verification
- **On-boat:** SK up → probe 200 (reachable); SK stopped → probe 502 (now correctly unreachable). SK restarted cleanly (healthy in ~3 s).
- Lint ✓ · `build:dev` ✓ · betterer 182 held ✓ · specs: `reload.service` (incl. explicit 401→reachable / 502→unreachable / network-error→toast + retry-loop), `app.component`, `settings.service` — all pass.

## Follow-up (not here)
`signalk.component.ts`'s own `location.reload()` (cross-origin connect flow) is the same failure class but a separate, lower-priority case — noted in the design, deferred.

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -5,8 +5,8 @@
 //
 exports[`strictNullChecks`] = {
   value: `{
-    "src/app/app.component.ts:281943337": [
-      [111, 19, 25, "Object is possibly \'undefined\'.", "255143637"]
+    "src/app/app.component.ts:1479801840": [
+      [113, 19, 25, "Object is possibly \'undefined\'.", "255143637"]
     ],
     "src/app/core/components/dashboard/dashboard.component.ts:3995263170": [
       [121, 12, 22, "Object is possibly \'undefined\'.", "2187394131"],
--- a/src/app/app.component.spec.ts
+++ b/src/app/app.component.spec.ts
@@ -10,7 +10,7 @@ import { uiEventService } from './core/services/uiEvent.service';
 import { AppService } from './core/services/app-service';
 import { ChromeVisibilityService } from './core/services/chrome-visibility.service';
 import { ToastService } from './core/services/toast.service';
-import { SettingsService } from './core/services/settings.service';
+import { ReloadService } from './core/services/reload.service';
 
 // Access the component's private surface without widening the class API.
 interface AppComponentHotkeyApi {
@@ -52,6 +52,7 @@ describe('AppComponent', () => {
     pulsePeek: ReturnType<typeof vi.fn>;
   };
   let toast: { show: ReturnType<typeof vi.fn> };
+  let reloadService: { reload: ReturnType<typeof vi.fn> };
 
   beforeEach(async () => {
     dashboard = {
@@ -72,6 +73,7 @@ describe('AppComponent', () => {
     appService = { toggleNightMode: vi.fn() };
     chrome = { revealed: signal(false), reveal: vi.fn(), hide: vi.fn(), pulsePeek: vi.fn() };
     toast = { show: vi.fn().mockReturnValue({ onAction: () => new Subject() }) };
+    reloadService = { reload: vi.fn() };
     appNetworkInitServiceStub.bootstrapIssue$.next({ reason: 'none' });
     appNetworkInitServiceStub.bootstrapStatus$.next('ready');
 
@@ -84,6 +86,7 @@ describe('AppComponent', () => {
         { provide: AppService, useValue: appService },
         { provide: ChromeVisibilityService, useValue: chrome },
         { provide: ToastService, useValue: toast },
+        { provide: ReloadService, useValue: reloadService },
       ],
     }).compileComponents();
   });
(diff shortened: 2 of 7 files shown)


mairas:
## Review synthesis (reliability, read-only)

**Verdict: reliable — no concrete defect.** Every protection the design relies on is present:

- **Timeout timer always cleared** via `finally { clearTimeout }` — no dangling timer / spurious post-resolve abort.
- **No toast/subscription leak** — `MatSnackBarRef` completes `_onAction` on every dismissal and MatSnackBar shows one toast at a time, so the retry chain holds at most one live subscription; rapid taps can't stack probes (the toast is dismissed on tap, so none is visible during the probe window).
- **DI cycle genuinely broken** — `ReloadService`'s only eager dep is `Injector`; `ToastService` (→`SettingsService`→`ReloadService`) is resolved lazily via `injector.get` only on the unreachable branch, after construction. Even the construction-time `resetConnection` path is safe (probe is async; the toast resolves on a later microtask after the constructor returns).
- **Reachable path byte-identical** — same `location.replace('./')` behind the same `__KIP_TEST__` guard; only a bounded (≤4 s) extra probe round-trip added to normal reloads.
- **Config-upgrade reloads untouched** — all four still call raw `reloadApp()`, preserving overlay-teardown-by-navigation.

Low residuals noted for awareness, none blocking: a cross-origin SSO IdP without CORS would make the probe reject (spurious Retry) — can't fire on the boat (same-origin Authelia); a proxy answering 200-with-error-body would read reachable (unusual config; Traefik returns 5xx for backend-down, which we handle); and `signalk.component.ts`'s own `location.reload()` is the same failure class on a different path (pre-existing, the deferred follow-up).

Testing gaps documented (branch logic is covered; the 3-line `setTimeout→abort` wiring is confirmed correct by inspection and the finally-clear is present): a fake-timers test exercising the real 4 s abort path would strengthen it further.

Rebased onto post-#300 main; betterer:ci consistent at **179**.

PR #302, opened 2026-07-15
Compile-time config-key-rename protection for the widget-config form (#25 Phase 3)

mairas:
The **headline** of the #25 typed-forms migration (**Phase 3**). Delivers the issue's actual stated goal: the widget-config form layer no longer silently desyncs when a config key is renamed — it fails to compile.

## What
Routed every `this.formMaster.get('literal')` in `root-modal-widget-config` (39 sites) through a key-constrained accessor:
```ts
private configControl<K extends keyof IWidgetSvcConfig>(key: K): AbstractControl | null {
  return this.formMaster.get(key);
}
```
The `K extends keyof IWidgetSvcConfig` bound is the whole point: a key that isn't a field on `IWidgetSvcConfig` is a compile error. All 39 routed keys were verified to be real interface fields (none dropped or forced).

## Design choice — lower-risk than the planned control-map
The [plan](https://github.com/halos-org/skip/issues/25#issuecomment-4977756811) proposed `FormGroup<WidgetConfigControls>` + **one load-bearing cast** over the reflective builder — which the design itself flagged as the phase's main risk (the cast can silently hide runtime/type drift). This achieves the **same rename-protection goal** with a key-constrained accessor instead: **no `WidgetConfigControls` type, no cast over the builder, `formMaster` stays `UntypedFormGroup`, every existing value cast preserved.** Zero runtime change, zero drift risk.

## Proof the guard actually fires
Temporarily calling `this.configControl('totallyFakeKeyXYZ')` produces exactly:
```
error TS2345: Argument of type '"totallyFakeKeyXYZ"' is not assignable to parameter of type 'keyof IWidgetSvcConfig'.
```
A guard that can't catch a rename would be worthless — this one does. (Probe reverted; not in the diff.)

## Verification
Lint ✓ · `build:dev` (strictTemplates) ✓ · betterer **179 held** ✓ · root-modal spec (incl. the Phase-2a characterization drift-guard) 8/8 ✓.

Part of #25 (Phase 3). Remaining: Phase 2b (recursion-built gauge/autopilot/ais/electrical shapes) + Phase 5 (dead-code cleanup) — both optional polish now that the `any`s are gone and rename protection is in.

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -138,7 +138,7 @@ exports[`strictNullChecks`] = {
       [85, 26, 49, "Object is possibly \'null\'.", "3725617003"],
       [89, 26, 49, "Object is possibly \'null\'.", "3725617003"]
     ],
-    "src/app/widget-config/root-modal-widget-config/root-modal-widget-config.component.ts:393658683": [
+    "src/app/widget-config/root-modal-widget-config/root-modal-widget-config.component.ts:2322637437": [
       [85, 4, 11, "Type \'{ label: string; value: string; }[]\' is not assignable to type \'never[]\'.\\n  Type \'{ label: string; value: string; }\' is not assignable to type \'never\'.", "3027418051"]
     ],
     "src/app/widget-config/solar-charger-setup/solar-charger-setup.component.ts:3993417046": [
--- a/src/app/widget-config/root-modal-widget-config/root-modal-widget-config.component.ts
+++ b/src/app/widget-config/root-modal-widget-config/root-modal-widget-config.component.ts
@@ -1,5 +1,5 @@
 import { Component, OnInit, inject, signal, DestroyRef } from '@angular/core';
-import { UntypedFormGroup, UntypedFormControl, FormControl, FormGroup, Validators, UntypedFormBuilder, UntypedFormArray, FormsModule, ReactiveFormsModule } from '@angular/forms';
+import { AbstractControl, UntypedFormGroup, UntypedFormControl, FormControl, FormGroup, Validators, UntypedFormBuilder, UntypedFormArray, FormsModule, ReactiveFormsModule } from '@angular/forms';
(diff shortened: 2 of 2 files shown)


mairas:
## Review synthesis (correctness, read-only)

**Behavior-preserving and complete — zero findings.**

- **Every routed site is runtime-identical.** `configControl(key)` does `return this.formMaster.get(key)` typed `AbstractControl | null` — exactly what `UntypedFormGroup.get()` returned — and each of the 39 sites keeps its original cast and `| null` handling, so `configControl('x') as <cast>` ≡ the old `formMaster.get('x') as <cast>`.
- **Completeness confirmed.** All 39 `formMaster.get('literal')` sites routed (4 windsteer, 1 updatePath, 2 deletePath, 32 getters). The only remaining bare `formMaster.get(` is the accessor body itself; `getRawValue()` is a different method, correctly untouched. Non-formMaster `.get()` (`groups.get(key)` builder-local, `fg.get('pathID')` on child path groups — `pathID` verified NOT a config key) correctly left bare.
- **All 37 unique keys are genuine `keyof IWidgetSvcConfig`** (spot-checked the less-obvious `compassModeEnabled`/`courseOverGroundEnable`/`waypointEnable`/`driftEnable`/`datasetAverageArray`).

**Honest coverage boundary (documented, not a defect):** the accessor protects the key **name** — renaming/removing a config field breaks every call site at compile time (#25's actual "silently desyncs on rename" problem). It does **not** couple the value **type**: the per-site casts (`as FormControl<boolean>` etc.) aren't checked against `IWidgetSvcConfig[K]`, so a field's *type* change would compile silently. The plan's `FormGroup<WidgetConfigControls>` control-map would have carried that too — at the cost of the load-bearing cast + drift risk this approach deliberately avoids. Reflected in the accessor's doc comment.

lint ✓ · build ✓ · betterer 179 held ✓ · root-modal spec 8/8 ✓ · rename-guard proven to fire.

PR #305, opened 2026-07-15
Chromeless embed mode + ephemeral profile URL addressing (E6 / #216)

### LINKED ISSUE #216: feat: chromeless/embedded Skip route for the Freeboard panel

mairas:
E6 / #216. Adds a chromeless "embed" render mode for the Freeboard-SK panel iframe, plus ephemeral URL-selected profile addressing. See the [design](https://github.com/halos-org/skip/issues/216#issuecomment-4979643814).

## Two boot-latched pre-hash query flags
`/@halos-org/skip/?embed=1&profile=day#/page/0`. **Key insight:** the app uses `withHashLocation`, so `router.navigate()` only touches the fragment — `window.location.search` is structurally untouched by in-app navigation. So "boot-latched" = a new `EmbedModeService` reads `window.location.search` **once** at construction (forced by the first blocking app-initializer). No `queryParams` plumbing anywhere.

## Behaviors — each grafts onto an existing gate (minimal new surface)
- **Chrome off** — the toolbar mount gate becomes `@if (dashboardVisible() && !embed())` — DOM-*unmount*, not CSS-hide.
- **Read-only / edit off** — one choke point: `DashboardService.setStaticDashboard(false)`/`toggleStaticDashboard()` no-op under embed, pinning `isDashboardStatic` true. That single signal already gates grid drag/resize, widget tap→config, group edit, and every unlock entry point. Plus a `!embed()` gate on the empty-state "Unlock and Customize" button.
- **Kept, zero change** — alarm badge (CDK overlay, already static-gated) + swipe/wheel page-nav (shell div, already static-gated).
- **Ephemeral profile** — resolved in `initNetworkServices()` to a **local** `effectiveSharedConfigName`, threaded into `getConfig` + `bootstrapRemoteContext` **only** — **never assigned to `this.config`, so it never persists to localStorage** (asserted in a spec). Unknown/invalid → persisted default + `console.warn`. Ordering guaranteed (blocking app-initializer completes before the router/dashboards exist).
- **Reload** — `performReload()` preserves `window.location.search` so the flags survive a Retry-reload inside the keepAlive iframe.
- **Route guard** — `embedBlockedGuard` redirects `/settings`,`/options`,`/remote` to `/page/0` under embed (defense-in-depth; closes the `setActiveProfile`-persistence hazard).
- **Plugin** — the Freeboard panel URL is `…?embed=1` (chromeless, default profile). The `?profile=` plumbing is fully built for manual/future use but no specific profile is baked in (per the maintainer decision).

## Verification
Lint ✓ · `build:dev` (strictTemplates) ✓ · betterer **179 held** ✓ · **198 unit tests + 7 plugin tests pass** (incl. the ephemerality assertion, embed-flag parsing, the `isDashboardStatic` pin, toolbar unmount, and the guard). No behavior change when not embedded (`embed()` false is a pure no-op everywhere).

**Boat-verify (batched to the boat session):** chromeless render inside the Freeboard drawer on the plotter, correct default profile, edit gestures inert, alarm badge + swipe nav present, and the flags surviving a keepAlive reload — the end-to-end checks unit tests can't cover.

Closes #216.


mairas:
## Multi-persona review synthesis (read-only-escape / ephemeral-leak / correctness / boot-reliability + adversarial verify)

**8 findings raised, 4 confirmed after adversarial verify** (2×P2, 2×P3). Both load-bearing invariants had real, reachable holes — being fixed before merge:

### P2 — read-only defeated by the auto config-upgrade (`app.component.ts:93`)
Under `?embed`, a v11/v12 stored config still triggers `upgrade.runUpgrade()`, which writes migrated config to every user slot **and** calls `settings.reloadApp()` → `location.replace('./')`, dropping `?embed` → the iframe reloads into the **fully editable** app. A transient (config lagging the deployed version) but a complete read-only defeat.
**Fix:** gate the auto-upgrade effect behind `!embed()` (defer migration to a full-app session); make `reloadApp()` preserve `window.location.search` so the flags survive any reload path.

### P2 — ephemeral profile leaks on rename (`settings.service.ts`)
`?profile=day` (honored with or without embed); renaming the ephemerally-active profile persists the URL-selected name into the saved per-device default.
**Fix:** the rename's "is this the device default → repersist" decision keys off the **persisted** name, not the ephemeral active name.

### P3 — missing-config toast writes under embed (`app.component.ts:159`)
The `missing-shared-config` bootstrap toast (CDK overlay, shows despite the unmounted toolbar) offers a "Create" button → `resetSettings()` → a `setConfig` write.
**Fix:** gate the toast behind `!embed()`.

### P3 — pre-v13 migration leaks the ephemeral profile's remote-control identity (`app-initNetwork.service.ts:202`)
On a pre-v13 device, first boot with `?profile=day` makes `migrateRemoteControlToDevice` read `isRemoteControl`/`instanceName` from the **ephemeral** slot and persist them.
**Fix:** source the migration from the **persisted** profile's config, never the ephemeral effective slot.

The other 4 findings were refuted on adversarial verify. Fixes in progress; I'll re-verify (build/betterer/specs, with new specs pinning each invariant) and update this PR before merge.

mairas:
## Review resolution — both load-bearing invariants confirmed

Two adversarial re-attack rounds on the fixed code, plus a final complete enumeration. **6 real bugs found and fixed** across the cycle; both invariants now verified:

### Read-only invariant — HOLDS (complete enumeration)
Every `storage.patchConfig`/`setConfig` caller in the app (only 3 services: `settings.service`, `profile.service`, `configuration-upgrade.service`) was traced for embed reachability. **All 6 escapes are gated behind `!embed()`:** the config-upgrade effect, the missing-config "Create" toast, the dashboards boot write-back, the app-config persist-on-missing self-heal — plus the user-triggered write surfaces (profile CRUD, settings tabs) are unreachable (routes guarded, toolbar unmounted, `isDashboardStatic` pinned; the Ctrl+Shift+E hotkey routes through the gated unlock). A **zero-`patchConfig`/`setConfig`-during-embed-boot** guard test (with a self-heal-triggering config) pins the whole invariant.

### Ephemeral-profile invariant — HOLDS
The URL-selected `?profile` is threaded only to a local `effectiveSharedConfigName`; the rename keys off the *persisted* name and the pre-v13 migration skips the ephemeral slot — a re-attack could construct no path where the ephemeral name reaches localStorage.

### Noted residuals (out of scope; documented for a possible follow-up)
- `RemoteDashboardsService` publishes remote-control **deltas** (`self.kip.remote.*`) — not an applicationData config/dashboard write, and only when the device is configured as a remote-control target. Doesn't breach E6's read-only-config promise; worth deciding separately whether embed should also suppress remote-control publication.
- Pre-v13 `connectionConfig` **localStorage** device-stamp (not a server write; ephemeral-override correctly excluded).

Verification: lint · `build:dev` · betterer **179** · **245 unit tests** + 7 plugin tests. Boat-verify (chromeless render in the Freeboard drawer) batches to the boat session.

PR #307, opened 2026-07-15
fix(gen-mcp-schema): resolve widgets via lazy-loader map, regen artifact, gate drift in CI

### LINKED ISSUE #293: gen:mcp-schema broken on main (WidgetAcComponent import) — schema artifact can't be regenerated

mairas:
## What

`npm run gen:mcp-schema` was broken on `main` — it threw `Could not find an import for "WidgetAcComponent"` and no schema could be regenerated, so the committed `kip-dashboard-schema.json` had silently gone stale. Fixes #293.

## Root cause

`widget.service.ts` loads every widget through a **dynamic lazy-loader map** — `ComponentClass: () => import('…').then(m => m.ComponentClass)` — with **zero static imports**. The schema generator resolved each widget's source file via `findImportModuleSpecifier`, which only understood static `import {} from '…'`. It therefore failed on the alphabetically-first widget (`widget-ac` → `WidgetAcComponent`; not special, just first) and produced nothing. Because `test:mcp-schema` wasn't part of `npm run ci`, the failure never blocked a PR.

## Changes

1. **Resolve widgets from the lazy-loader map** (`findLazyLoadedModuleSpecifier`): finds the `import('SPECIFIER').then(m => m.ComponentClass)` entry, tying the module specifier to the actual exported class via the `.then` callback (not just the map key). Replaces the now-dead `findImportModuleSpecifier`.
2. **Handle the non-literal `DEFAULT_CONFIG` values current widgets use** (surfaced only once resolution worked, because the generator resolves all 36 widgets for the first time since the lazy-loader conversion):
   - `undefined` (an identifier in the AST, not a keyword) → the unset value; keys valued `undefined` drop out of the JSON artifact. *(widget-data-chart yScale keys)*
   - **same-file module-scope constant references** → resolved one hop to their literal. *(widget-windsteer `windSectorWindowSeconds: DEFAULT_WIND_SECTOR_WINDOW_SECONDS`)*
   - Every **other** identifier still throws loudly — and now **names the offending component**, so the next drift is diagnosable instead of an anonymous `Unsupported syntax: Identifier`.
3. **Flip the stale spec** — the ac/inverter/alternator electrical family is live in `_widgetDefinition`, so `excludes the commented-out widgets` becomes `includes the live electrical widgets`.
4. **Regenerated the artifact**: **31 → 36 widgets** (ac, alternator, inverter, hoekens-anchor-alarm, video were all absent because the generator was fully broken), and it picks up F1's **"Realtime Data Chart" → "Realtime Data Plot"** rename (#126/#247) that couldn't reach the schema until now.
5. **Gate against recurrence** — added `test:mcp-schema` to `npm run ci` (alongside `test:plugin`), so a widget name/`DEFAULT_CONFIG` change that leaves the artifact stale now fails CI instead of drifting silently.

## Testing

- `npm run test:mcp-schema` — 6 files / 24 tests green (drift gate passes against the committed artifact).
- `npm run lint` clean; `npm run betterer:ci` unchanged (179; no `src/**/*.ts` touched); generator type-checks under `--strict`.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -8,3 +8,10 @@ tag = False
 [bumpversion:file:package.json]
 search = "version": "{current_version}"
 replace = "version": "{new_version}"
+
+# The generated MCP schema artifact stamps the app version as meta.kipVersion,
+# and test:mcp-schema gates it against drift in CI. Keep it in step on a bump so
+# a version change alone never fails CI (no schema regen needed).
+[bumpversion:file:src/assets/kip-dashboard-schema.json]
+search = "kipVersion": "{current_version}"
+replace = "kipVersion": "{new_version}"
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "snc": "tsc -p tsconfig.strict.json",
     "betterer": "betterer --tsconfig tsconfig.betterer.json",
     "betterer:ci": "betterer ci --tsconfig tsconfig.betterer.json",
-    "ci": "npm run lint && npm run betterer:ci && npm run test:headless && npm run test:plugin",
+    "ci": "npm run lint && npm run betterer:ci && npm run test:headless && npm run test:plugin && npm run test:mcp-schema",
     "prepublishOnly": "npm run build:prod",
     "dev": "ng serve --configuration=dev --serve-path=/@halos-org/skip/",
     "build:dev": "ng build --configuration=dev",
--- a/src/assets/kip-dashboard-schema.json
+++ b/src/assets/kip-dashboard-schema.json
@@ -548,6 +548,29 @@
   "widgets": [
     {
       "bindingKind": "none",
+      "category": "Gauge",
+      "componentClassName": "WidgetAcComponent",
+      "defaultConfig": {
+        "ac": {
+          "optionsById": {},
+          "trackedDevices": []
+        },
+        "color": "contrast",
+        "ignoreZones": false
+      },
+      "defaultHeight": 6,
+      "defaultWidth": 6,
+      "description": "Monitor AC bus and line-level loads with real-time voltage, current, frequency, and power metrics.",
+      "icon": "acMonitor",
+      "minHeight": 2,
+      "minWidth": 2,
+      "name": "AC Monitor",
+      "pathSlots": [],
+      "requiredPlugins": [],
+      "selector": "widget-ac"
+    },
+    {
+      "bindingKind": "none",
       "category": "Component",
       "componentClassName": "WidgetAisRadarComponent",
       "defaultConfig": {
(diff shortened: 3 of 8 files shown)


mairas:

**12 findings → 6 confirmed / 6 refuted.** All 6 confirmed are addressed in the force-pushed update; the confirmed set collapses to **3 distinct issues** (four lenses independently flagged the same one). The review caught a real regression this PR had introduced and would have **CI-locked** — exactly the failure mode #293 exists to prevent.

### ✅ Fixed

**1. [P1] `meta.configVersion` regressed 13 → 12** — flagged by correctness, adversarial, maintainability, and project-standards.
The generator hardcoded `CONFIG_VERSION = 12`, stale against `LATEST_APP_CONFIG_VERSION = 13` (`config-versions.const.ts`). On `main` the artifact correctly read 13 — hand-patched by commit `36e9506e` *because the generator was broken and couldn't regenerate*. Making the generator runnable and regenerating silently reverted it to 12, and the new CI drift-gate would have cemented the wrong value (the external kip-mcp-server would then stamp generated configs as v12, one version behind, triggering a spurious v12→v13 migration).
**Fix:** the generator now **reads** `REMOTE_CONFIG_FILE_VERSION` + `LATEST_APP_CONFIG_VERSION` from their single definition site via AST (no app-module import — the generator only parses source), so it can't drift again. `schema.spec.ts` imports the same constants and asserts `meta` matches them. Artifact regenerated → `configVersion: 13` (unchanged vs `main`; the regression is gone).

**2. [P3] every version bump would break CI** — adversarial.
The gated artifact embeds `kipVersion` from `package.json`, so `./run bumpversion` would leave the committed artifact stale and fail CI with a misleading widget-drift message.
**Fix:** added the artifact to `.bumpversion.cfg` (`"kipVersion": "{current_version}"` search/replace, mirroring the existing `package.json` entry), so a version bump syncs the field in the same commit — no schema regen needed at release.

**3. [P3] orphaned/stale JSDoc** — maintainability.
The new `IdentifierResolver` type was inserted between `literalToValue`'s doc block and the function, and the doc's "an identifier … throws" claim contradicted the new handling.
**Fix:** moved the type above the doc so it rejoins the function; the doc now describes the `undefined` and same-file-const handling and the remaining fail-loud contract.

### Also hardened (from refuted-but-valid observations)
- **Unit tests added** (`ast.spec.ts`) pinning the new behaviors — lazy-loader resolution, `undefined`-drop, const-resolution, and the fail-loud-with-component-name path — which were previously covered only indirectly by the full-artifact snapshot (testing lens, 3 findings).
- **`collectModuleConstants` tightened to `const`-only** (`ts.NodeFlags.Const`) so it matches its name and can't resolve a reassignable `let`/`var` (maintainability lens).

### Deferred (real, but out of #293's scope) → filed
- **[P2] `tools/` is neither linted nor type-checked in CI** (maintainability). Pre-existing gap (the generator escapes the repo's strict-typing gates); the #293 changes were verified type-clean by hand. Closing it is analogous to #268's `plugin/` work and is its own change → **filed as #308.**

### Refuted (verified non-issues)
The remaining refutations were theoretical/unreachable in this diff: a `datachartPath: undefined` edge (no such widget), and three test-hardening suggestions (the underlying gap is addressed by the new `ast.spec.ts`).

*Note: the maintainability lens ran during a brief safety-classifier outage; its findings (#1 corroborated by three other lenses, plus the doc + const-only nits) were each re-verified directly against the code.*

**Post-fix verification:** 7 mcp-schema suites / 30 tests green, drift gate passes, generator type-checks under `--strict`, lint clean, betterer unchanged (179).

PR #309, opened 2026-07-16
feat(windtrends): make wind series paths user-configurable (fix blank TWS on ground-referenced vessels)

### LINKED ISSUE #290: Wind Trends: TWS hardcoded to environment.wind.speedTrue — blank on ground-referenced-only vessels

mairas:
## What

The Wind Trends widget hardcoded its two series paths (TWD `self.environment.wind.directionTrue`, TWS `self.environment.wind.speedTrue`) as module constants with no `config.paths`. On a ground-referenced-only vessel (`speedOverGround` present, `speedTrue` absent) the TWS line was **permanently blank** with no way to repoint it. Fixes #290.

## Approach (maintainer decision)

Make both series **user-configurable** via the standard `config.paths` record — reuse the existing paths-config machinery, build no new UI. **The shipped defaults are unchanged** (`speedTrue` kept as the TWS default; `directionTrue` for TWD; **no auto-fallback**): existing installs behave identically, and a ground-referenced vessel repoints TWS to `speedOverGround` through the now-available Paths tab. The fix is *configurability*, not a default change.

## Changes (four coordinated sites, reuse-only)

1. **`DEFAULT_CONFIG`** — a two-slot `paths` record (`trueWindDirection`, `trueWindSpeed`, shaped like windsteer's). Declaring `paths` alone surfaces the generic **Paths** config tab (`root-modal` renders `<paths-options>` on `paths !== undefined` — zero widget-specific UI).
2. **Consumption** — streaming reads each slot's `path`/`source` (→ `'default'` when unset) and subscribes **per series independently**, gated on its own non-empty path, so clearing one slot (`pathRequired: false`) still renders the other. Stays on `HistoryChartStreamService.getBackfillThenLive`.
3. **Rebuild trigger** — broadened from `timeScale`-only to a signature over `[timeScale, both slots' path/source]` (mirrors `widget-data-chart`'s `pathSignature`), so edited paths take effect. The y-axis age recompute is gated on *either* series so a cleared direction slot still positions the speed series.
4. **History-dialog sync** — `mapWindTrendsWidget` reads the same slots and omits a series when its path is cleared (mirrors `mapDataChartWidget`).

MCP schema regenerated: windtrends `bindingKind` `none → paths-record` with the two slots.

## Notes (deliberate, for review)

- **`showConvertUnitTo: false` on both slots** — the widget's axes are fixed deg/kts and the conversion is hardcoded regardless of the slot, so exposing a Format dropdown that does nothing would mislead. (windsteer's speed slot similarly omits it.)
- **Known minor limitation (out of scope):** the center-tick "Data acquisition…" overlay clears only when *both* series have ≥2 points, so clearing one slot leaves the overlay box over the chart center while the other series still renders. Fixing needs the plugin to know which slots are active — left per the minimal-scope decision.

## Testing

- Windtrends component spec (7): default **unchanged** (`directionTrue` + `speedTrue`) + configured-paths + both NIL-gating directions + rebuild-on-path-change. Sync-service spec (15): default **unchanged** + configured + NIL.
- Local gate on halpi: `lint` clean, `betterer:ci` unchanged (179), schema drift gate green (7 suites/31 tests), both modified specs green.

### DIFF
--- a/src/app/core/services/dashboard-history-series-sync.service.spec.ts
+++ b/src/app/core/services/dashboard-history-series-sync.service.spec.ts
@@ -3,12 +3,43 @@ import { TestBed } from '@angular/core/testing';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { DashboardHistorySeriesSyncService } from './dashboard-history-series-sync.service';
 import { IKipSeriesDefinition } from '../contracts/kip-series-contract';
-import { IWidget } from '../interfaces/widgets-interface';
+import { IWidget, IWidgetPath, IWidgetSvcConfig } from '../interfaces/widgets-interface';
 import { WidgetService } from './widget.service';
 
 function seriesIds(series: IKipSeriesDefinition[]): string[] {
     return series.map(item => item.seriesId).sort();
 }
+
+// Mirrors the two configurable slots the widget's DEFAULT_CONFIG ships. resolveWidgetConfig merges
+// these into a windtrends widget config that lacks its own paths, so the default mapping still
+// resolves the shipped TWD/TWS paths.
+const WIND_DIRECTION_SLOT: IWidgetPath = {
+    description: 'True Wind Direction',
+    path: 'self.environment.wind.directionTrue',
+    source: 'default',
+    pathType: 'number',
+    isPathConfigurable: true,
+    pathRequired: false,
+    sampleTime: 1000,
+};
+const WIND_SPEED_SLOT: IWidgetPath = {
+    description: 'True Wind Speed',
+    path: 'self.environment.wind.speedTrue',
+    source: 'default',
+    pathType: 'number',
+    isPathConfigurable: true,
+    pathRequired: false,
+    sampleTime: 1000,
+};
+const WINDTRENDS_DEFAULT_CONFIG: IWidgetSvcConfig = {
+    filterSelfPaths: true,
+    color: 'contrast',
+    timeScale: 'Last 30 Minutes',
+    paths: {
+        trueWindDirection: WIND_DIRECTION_SLOT,
+        trueWindSpeed: WIND_SPEED_SLOT,
+    },
+};
 
 describe('DashboardHistorySeriesSyncService', () => {
     let widgetServiceMock: {
@@ -17,7 +48,9 @@ describe('DashboardHistorySeriesSyncService', () => {
 
     beforeEach(() => {
         widgetServiceMock = {
-            getDefaultConfig: vi.fn().mockReturnValue(undefined)
+            getDefaultConfig: vi.fn().mockImplementation((selector: string) =>
+                selector === 'widget-windtrends-chart' ? structuredClone(WINDTRENDS_DEFAULT_CONFIG) : undefined
+            )
         };
 
         TestBed.configureTestingModule({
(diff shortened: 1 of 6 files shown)


mairas:

**8 findings → 2 confirmed / 6 refuted.** The two confirmed are the same defect (correctness + adversarial), now fixed.

### ✅ Fixed — [P2] loading overlay stuck under a single-configured-series
The `centerTickPlugin` readiness gate was `dirVals>=2 && spdVals>=2` (AND). Clearing one slot — the exact configuration this PR enables for a ground-referenced-only vessel clearing TWS — leaves that series' datasets empty forever, so the "Data acquisition…" overlay was drawn **permanently** over the series that *does* stream. Smoking gun the review caught: the sibling y-axis age-recompute gate was broadened to OR (`dirData.length>0 || speedData.length>0`) in the same diff, but this gate was left AND — an internal inconsistency, not a scope call.

**Fix:** extracted `isChartReady()` gated per-configured-series — `(dirActive || spdActive) && (!dirActive || dirVals>=2) && (!spdActive || spdVals>=2)`, with the active flags derived from the same slot-path truthiness the per-series stream gating uses. Both configured → clears once both have data (unchanged); one cleared → clears once the active series has data; **both cleared → stays acquiring** (leading term). Added 4 tests including a both-configured lock so nobody "simplifies" it to a pure OR. Also documented the load-bearing `showConvertUnitTo: false` (fixed deg/kts axes).

### Refuted (verified non-issues in this diff)
- **[P3] backfill ignores configured source** — pre-existing `fetchBackfill` behavior in `history-chart-stream.service` (shared, outside this diff); the live tail honors source, backfill builds `path:last`. Real but not introduced here and shared across all history-chart consumers — a separate concern.
- **[P2] rebuild test drives via theme** — the test is weaker than ideal, but in production `WidgetRuntimeDirective.options` is a computed signal, so a real path change *does* re-fire the rebuild; not a reachable defect.
- **[P3] history-dialog symmetric/both-cleared untested; [P2] slot-key literal duplicated across component+sync; [P3] `showConvertUnitTo` deviation; [P3] paths whole-replace vs deep-merge** — test-symmetry / theoretical-future-drift; all currently consistent, none reachable.

**Verification (post-fix):** windtrends spec 11 green (incl. the 4 readiness tests), sync spec 15 green, schema drift gate green (artifact unchanged — the doc comment carries no literal value), lint clean, betterer 179.

Issue #192, opened 2026-07-04
chore: rename internal `kip` identifier namespace to `skip`


mairas:
## Problem

The fork rebranded the product to **Skip** (`@halos-org/skip`) on the surface — package name, serving path, PWA manifest, branding — but the code-level identifier namespace was never rebranded. Internally most things are still `kip`. There are **zero** `skip-`-prefixed runtime identifiers, so every new identifier that follows the existing convention (like `.kip-app-shell`) deepens the inconsistency. We're Skip, not Kip.

## Scope decision (updated 2026-07-16)

**Breaking change accepted; NO migration, dual-write, compat window, or version gate anywhere.** There are no users yet, which voids the sole rationale for the original "wire-format / applicationData config keys would break existing user configs" carve-out. This issue now covers the **full** internal `kip → skip` rename, including the formerly-excluded `self.kip.remote.*` wire paths and the persisted `kipUUID` field. Everything the product owns becomes `skip`; only genuine upstream references and the already-correct serving contract stay.

Basis: a full-codebase enumeration (2026-07-16). Coupling traps flagged ⚠️ below **break silently — no build/type/lint error** if a rename is split.

### Pure-internal (no external contract)
- **CSS custom properties** — 39 `--kip-*` tokens across ~47 consumer files: defs in `src/styles.scss` (~102 decl lines) + the theme maps `$kip-{light,dark,night}-color` in `_m3{light,dark,night}.scss`; 35 `.scss`, 10 `.ts` (D3 attr strings; `app-service.ts` get/setProperty for `--kip-nightMode*`), 2 `.html` inline styles. ⚠️ A name split across two casings breaks CSS silently. ⚠️ **`tools/gen-mcp-schema/generate.ts:195` hardcodes a regex `/\$kip-dark-color:.../` and throws if unmatched — rename in lockstep with the `_m3dark.scss` map.**
- **Directive selectors** — `[kipGestures]`, `[kipScrollNav]`, `[kipResizeObserver]` (+ `KipResizeObserverDirective`, `(kipResize)` output, `IKipResizeEvent`, file `kip-resize-observer.directive.ts`→`skip-*`). Selector strings + every template usage + the specs' `querySelector('[kipScrollNav]')` move together (an un-updated querySelector silently matches nothing).
- **Runtime classes + generated DOM ids** — `.kip-app-shell`, `.kip-widget-card-colors`, `kip-gesture-inject-${id}` (4 widgets). Self-contained.
[...]

PR #310, opened 2026-07-16
feat(embed): suppress remote-control delta publication under embed mode

### LINKED ISSUE #306: Consider suppressing remote-control delta publication under embed mode
Surfaced by the E6 (#216) read-only re-review as an out-of-scope residual (E6 is merged).

`RemoteDashboardsService` (`src/app/core/services/remote-dashboards.service.ts`) publishes this device's screen list + active-dashboard index to the Signal K server as deltas on `self.kip.remote.*` paths. It is **not** gated by the E6 `embed()` flag.

This does **not** breach E6's read-only invariant as scoped — those are transient remote-control pub/sub **deltas**, not `applicationData` config/dashboard writes — and it only activates when the per-device `isRemoteControl` flag is set (unlikely for a fresh Freeboard embed viewer). But a strictly read-only chromeless panel arguably should not advertise itself as a remote-control target / publish its screen state.

**Decision to make:** should embed mode also suppress remote-control delta publication (and/or the remote-control feature entirely)? If yes, gate `RemoteDashboardsService`'s publication behind `!embedMode.embed()`. Low priority; conditional on a device being configured as a remote-control target.

mairas:
## What

A chromeless **embed** panel is strictly read-only and should not advertise itself as a remote-control target. `RemoteDashboardsService` published this device's screen list + active-dashboard index to the SK server on `self.kip.remote.*`, and acted on incoming `activeScreen` requests — **none of it gated by embed**. Closes #306.

## Why it matters (beyond hygiene)

The embed iframe shares the device `KipUUID` + `connectionConfig` (localStorage) with the full Skip app on the same origin. So:
- The three **ungated** constructor "clear" PUTs fire on *every* embed boot regardless of `isRemoteControl`.
- If remote control is enabled in the full app, the embed panel inherits `isRemoteControl=true` and both publishes and acts as a nav target — and its boot clears can **stomp the remote paths the full app just published for the same UUID**.

## Change

Inject the leaf `EmbedModeService` (no DI-cycle risk — it injects nothing) and **early-return at the top of the constructor**, before the boot clears and before the effects are wired. Under embed there is therefore zero remote-control delta publication *and* no incoming-command navigation (effect 4 is never created). The passive `activeScreen` subscription (a read) runs in a field initializer and is left in place — a read is consistent with read-only embed.

9 lines of source, one guard.

## Notes for review

- This is *not* a breach of the E6 persisted-write invariant (these are transient pub/sub deltas, not `applicationData` writes) — it closes the shared-UUID stomp + "advertises as remote-control target" gap.
- The `self.kip.remote.*` command channel is currently publish-only-orphaned (its server-side handler was removed in `7ccc3990`), so this suppresses publication of an already-consumer-less channel plus the still-live incoming-nav side.
- Sequencing: the kip→skip wire-path rename (#192, Stage 2) will later rename `self.kip.remote.*` in this same file; this lands first, so no collision.

## Testing

- `remote-dashboards.service.spec.ts` — 28 green (25 existing + 3 new): under embed, zero `putRequest` even with remote control enabled at construction; no `navigateTo` on an incoming `activeScreen`; no publish when remote control is toggled on mid-session. Existing tests unaffected (the real `EmbedModeService` reads empty jsdom `location.search` → `embed=false`).
- Local gate on halpi: lint clean, betterer:ci unchanged (179).

### DIFF
--- a/src/app/core/services/remote-dashboards.service.spec.ts
+++ b/src/app/core/services/remote-dashboards.service.spec.ts
@@ -8,6 +8,7 @@ import { Dashboard, DashboardService } from './dashboard.service';
 import { DataService, IPathUpdate } from './data.service';
 import { SettingsService } from './settings.service';
 import { SignalkRequestsService } from './signalk-requests.service';
+import { EmbedModeService } from './embed-mode.service';
 import { States } from '../interfaces/signalk-interfaces';
 
 const KIP_UUID = 'test-kip-uuid';
@@ -410,6 +411,40 @@ describe('RemoteDashboardsService', () => {
 
       expect(errorSpy).toHaveBeenCalledTimes(3);
       errorSpy.mockRestore();
+    });
+  });
+
+  describe('under embed mode', () => {
+    beforeEach(() => {
+      TestBed.overrideProvider(EmbedModeService, { useValue: { embed: () => true, profile: () => null } });
+    });
+
+    it('publishes no remote-control deltas at all, even with remote control enabled at construction', () => {
+      settings.isRemoteControl.set(true);
+      dashboard.activeDashboard.set(1);
+
+      createService();
+
+      expect(requests.putRequest).not.toHaveBeenCalled();
+    });
+
+    it('does not act as a remote-control target (ignores incoming activeScreen requests)', () => {
+      settings.isRemoteControl.set(true);
+      createService();
+
+      data.push(OWN_ACTIVE_SCREEN_PATH, 2);
+      TestBed.tick();
+
+      expect(dashboard.navigateTo).not.toHaveBeenCalled();
+    });
+
+    it('does not publish when remote control is toggled on during the session', () => {
+      createService();
+      requests.putRequest.mockClear();
+
+      enableRemoteControl();
+
+      expect(requests.putRequest).not.toHaveBeenCalled();
     });
   });
 });
--- a/src/app/core/services/remote-dashboards.service.ts
+++ b/src/app/core/services/remote-dashboards.service.ts
@@ -4,6 +4,7 @@ import { DashboardService, Dashboard } from './dashboard.service';
 import { DataService } from './data.service';
 import { toSignal } from '@angular/core/rxjs-interop';
 import { SignalkRequestsService } from './signalk-requests.service';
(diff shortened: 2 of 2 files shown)


mairas:
## Review synthesis (3 lenses → adversarial verify)

**1 finding → 0 confirmed / 1 refuted.** The change is clean; no code changes required.

### Completeness check — the load-bearing question, answered: boundary is closed
The review specifically enumerated whether remote-control deltas could still publish under embed by a path *other* than the gated constructor. `RemoteDashboardsService`'s public publish methods (`setActiveDashboardOnRemote` / `setScreensOnRemote` / `clearActiveScreenOnRemote`) have exactly one external caller — the `/remote` controller component — which is **unreachable under embed** (the E6 embed route guard redirects `/remote` → `/page/0`). So the constructor early-return is a *complete* suppression: nothing publishes remote-control deltas under embed, and the incoming-command navigation (effect 4) is never wired. Correctness and testing lenses found nothing.

### Refuted (considered, correctly not actioned)
- **[P2→P3] Stale advertisement not retracted under embed** — because the embed early-return skips the boot clears, a remote-control advertisement left by a *prior* same-UUID non-embed session is never retracted. **Refuted / out of scope:** (a) #306's contract is "the embed panel must not advertise *itself*" — it makes zero writes and self-advertises nothing; a ghost left by a different prior session is a *pre-existing* missing destroy-time clear, independent of this diff; (b) the fix the finding wants (an embed-time clear) is itself a **PUT/write** that would violate #306's read-only invariant and break the new zero-`putRequest` test; (c) the pre-PR path it cites as "lost" also *re-advertised* the embed panel when remote control was on — the exact bug #306 fixes. Harm is bounded/cosmetic under niche preconditions.

### Feature-health note (surfaced, not blocking)
The refutation touches a broader, pre-existing reality worth flagging: the `self.kip.remote.*` command channel is currently **publish-only-orphaned** (its server-side handler was removed in `7ccc3990`), and `RemoteDashboardsService` has no destroy-time advertisement retract. Both are independent of #306 and belong to a future remote-control-health revisit — raised to the maintainer rather than auto-filed.

**Verification:** 28 spec tests green (25 existing + 3 embed), lint clean, betterer 179, CI green.

PR #311, opened 2026-07-16
fix(data): wire refcounted path-registration release + prune evicted-context caches (#243 stage 2)

mairas:
## What

Stage 2 of #243 — wire the release sites and harden the release contract so `DataService`'s per-path caches stop growing without bound (the "app goes unresponsive after a while" freeze, port of Kip #1129). Stage 1 (the `refCount` + `unsubscribePath(path, source)` + `removePathsForContext` foundation) landed in #256; this makes the churning consumers actually release, and prunes evicted-context path data.

## Design decisions (settled in a design pass; flagged here for review)

- **Additive `acquirePath(path, source)` → `{ data$, release }`** — an **idempotent, forgery-proof** disposable handle wrapping the stage-1 `unsubscribePath` primitive (`release` captures a private `released` flag → tears down exactly once; the caller passes no path/source, so it can't aim a release at another registration or over-decrement its own). **This takes the maintainer's explicitly-allowed "(or a companion acquire API)" branch rather than changing `subscribePath`'s signature.** Verified rationale: `subscribePath` has exactly **3 churning consumers** (widget-streams, remote-control, history-chart-stream — migrated here) and **2 bounded app-lifetime singletons** (`app-service`, `remote-dashboards` — both `toSignal(subscribePath(CONST,'default'))`, left untouched). Additive keeps the blast radius on the 3 sites that matter and is strictly lower-risk than breaking every caller. A doc note on `subscribePath` steers future churning consumers to the handle.
- **Meta decoupled from `registration[0]` (mandatory, lands first)** — `getPathMetaObservable` now serves from a new `_pathMetaByPath` subject map, not `registration[0].pathMeta$`, so releasing one source can never `complete` a live path's meta stream and freeze the metadata directive's zones. Bonus: fixes a pre-existing `of(null)` dead-stream (meta subscribed before any registration) and drops meta fan-out from O(sources) to O(1) per delta. `_pathMetaByPath` is pruned in `removePathsForContext` (delete-only) or it would reproduce the same unbounded growth.
- **D2 → keep skip's AIS constants** (TTL 10 min, sweep 15 s, cap 500). Upstream's 60 min/5 min would hold an evicted target's ~25-30 `_skData` entries **6× longer** — skip has no authoritative AIS-removal signal, so TTL is the *primary* eviction path, and a longer TTL works *against* the leak fix.

## The five release sites (6-commit chain)

1. **meta-decouple** (DataService) — `_pathMetaByPath`, rewrite `getPathMetaObservable`/`setMeta`, delete `pathMeta$` from `IPathRegistration`, prune on context removal.
2. **`acquirePath`** (DataService) — the handle.
3. **widget-streams.directive** — handle held 1:1 with the `baseSignatures` keyset; release-before-overwrite on base-key change, per removed path, on empty/invalid path, on bulk reset, and a **new `ngOnDestroy`** (the primary current leak vector). **Never** releases on the RxJS pipeline swap (sample-time/unit/timeout changes keep the same base key).
4. **remote-control.component** — release the two display registrations before rebind + a new `ngOnDestroy`.
5. **history-chart-stream.service** — release on `stop()` teardown after the live unsub; start/stop race handled (handle var stays null if disposed before backfill → release is a no-op; `ctx.disposed` already blocks late acquisition; idempotency is a second layer).
6. **ais-processing.service** — `removeTrack → removePathsForContext(track.context)` (single funnel for all 3 eviction reasons); the three `Partial<DataService>` spec mocks gain the method in the same commit (mandatory — else 8 eviction tests `TypeError`).

## Testing

- **Full local suite green: 148 files / 1282 tests.** New unit coverage locks the invariants: handle idempotency / over-release with a live co-acquirer / cannot-decrement-below-own-acquire; meta-survives-sibling-release + meta-before-registration + foreign-context meta prune; per-site acquire/release balance incl. the **no-release-on-pipeline-swap** trap and `ngOnDestroy` release-all; `removePathsForContext` fired on each of the 3 AIS eviction paths.
- Lint clean; betterer unchanged (179 — the touched files are strict-clean).

## ⚠️ CI-green ≠ leak-fixed

Unit tests prove per-site acquire/release balance and the meta invariant, but **cannot** prove the end-to-end leak is closed. The real proof is an **off-box soak**: sustained AIS churn (continuous create/evict of foreign contexts) while watching `_pathRegisterByPath.size` and `_skData.size` **stabilize** rather than grow. That needs a dev-machine/boat run with a small size-reading hook (perf-harness heap MB is noise-dominated per the repo docs and won't substitute). **Do not read green CI as "leak fixed"** — soak-verify separately before relying on it.

### DIFF
--- a/src/app/core/components/remote-control/remote-control.component.spec.ts
+++ b/src/app/core/components/remote-control/remote-control.component.spec.ts
@@ -1,6 +1,10 @@
 import { ComponentFixture, TestBed } from '@angular/core/testing';
-import { beforeEach, describe, expect, it } from 'vitest';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { Observable, Subject } from 'rxjs';
 import { RemoteControlComponent } from './remote-control.component';
+import { DataService, IPathUpdate, IPathUpdateWithPath } from '../../services/data.service';
+import { SettingsService } from '../../services/settings.service';
+import { SignalkRequestsService } from '../../services/signalk-requests.service';
 
 describe('RemoteControlComponent', () => {
   let component: RemoteControlComponent;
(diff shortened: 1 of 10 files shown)


mairas:
## Multi-persona review synthesis (6 lenses → adversarial verify)

**7 findings → 0 confirmed / 7 refuted.** No reachable defect: the correctness + adversarial-leak lenses enumerated every acquire/release path (the leak boundary), the history-chart start/stop race, and the AIS eviction funnel, and found the accounting balanced. Three refuted testing findings converged on incomplete **release-balance coverage**, so I hardened the tests (additive commit `e0ef52d4`, code untouched) — the right investment for a memory-safety packet where release balance is the invariant.

### ✅ Test coverage hardened (from the convergent testing findings)
- **widget-streams timeout-branch:** a `dataTimeout` change takes the distinct `rootChanged` force-rebuild path — added a test asserting it still does **not** release or re-acquire (base key unchanged). This was the one that mattered: a real distinct code path, previously unexercised for release behavior.
- **widget-streams multi-base:** new tests holding two concurrent bases prove `releaseAllBases` releases **all** on both bulk-reset and `ngOnDestroy` (the `forEach`, not just a single entry).
- **remote-control balance:** the site had only a `should create` smoke test — added a local-fake-`DataService` harness asserting a display switch releases both prior handles before acquiring two new, and `ngOnDestroy` releases the current pair.
- Cleanups: removed a dead `subscribePath` mock member (history spec); tightened the history-teardown comment to state the real balance guarantee (null `releaseLive` + idempotent release, not the disposed guard).

### Refuted (verified non-issues / by-design)
- **[P3] history re-entrant unsubscribe during backfill emission** — requires a consumer that synchronously unsubscribes inside `next()`; all three actual consumers store the subscription and tear down from a separate stack. Unreachable; the idempotent handle + null-var cover it regardless.
- **[P3] meta subject deleted while a widget subscribes (delete-only prune, no re-feed)** — real asymmetry, but `removePathsForContext` only prunes *other-vessel AIS-target* contexts, reaching a gauge requires non-default `filterSelfPaths` + pinning to a transient AIS MMSI path, **and** AIS meta is static schema-derived so the pre/post value can't diverge (no observable freeze). Delete-only is deliberately documented; "fixing" it (get-or-create on write) would reintroduce the unbounded growth this packet removes.
- **[P3] `baseReleases` 1:1 coupling enforced by convention** — a design observation; the coverage above pins the balance behaviorally.

### ⚠️ CI-green ≠ leak-fixed (restated)
Unit tests prove per-site acquire/release balance and the meta invariant; they cannot prove the end-to-end leak is closed. The real proof is an **off-box soak** — sustained AIS churn while `_pathRegisterByPath.size` / `_skData.size` **stabilize** — which needs a dev-machine/boat run with a size-reading hook. Do not read green CI as "leak fixed."

**Verification:** full local suite 148 files green (incl. the new release-balance tests), lint clean, betterer:ci unchanged (179).

PR #312, opened 2026-07-16
fix(data): clear stale path only when all sources silent + deep-reset resurfacing (#254)

### LINKED ISSUE #254: enableTimeout=false widgets no longer cleared by a sibling source's timeout (fallout of #206 source-scoping)
## Context

PR #252 (#206) scoped `DataService.timeoutPathObservable` to the source that actually timed out, so one dead source no longer nulls a sibling widget bound to a different, still-live source of the same path. That fix is correct for widgets that run the timeout operator.

It exposes a behavior reversal for widgets with **`enableTimeout: false`** (a shipped default for `widget-boolean-switch` and `widget-autopilot`, among others).

## The reversal

A widget with `enableTimeout: false` has no `timeout()` operator in its pipeline (`widget-streams.directive.ts` ~150-165), so it **never self-clears** — it only ever went blank because the *old* null-all reset pushed `null` through the shared `(path, source)` `pathDataUpdate$` subject when **any** sibling on the same path timed out.

With source-scoped clearing:

- Widget A on `(path, gps-primary)`, `enableTimeout: true`.
- Widget B on `(path, gps-secondary)`, `enableTimeout: false`.
- A shared upstream failure kills both feeds.
- At `dataTimeout`, A times out and clears only `gps-primary`. **B keeps rendering its last value indefinitely** — previously A's null-all cleared it too.

## Assessment (needs a product-intent decision)

This is arguably **more correct**: `enableTimeout: false` means "keep showing the last value even if data stops," and the old cross-source clearing was an incidental side effect that silently overrode that opt-out. But it's a silent reversal, and for safety-adjacent widgets (autopilot state; a position/depth widget a user explicitly set to no-timeout) a frozen-but-plausible reading is a hazard.

Options:
1. **Accept + document** — state that safety-critical widgets should enable timeout; treat `enableTimeout: false` as a deliberate "never blank" choice. (Lowest effort; most consistent with the flag's meaning.)
2. When a **specific** source times out and **no live source remains** for the path, also clear the `default` bucket / siblings (re-adds some cross-clearing, but only in the genuinely-all-silent case — avoids re-introducing #206's live-sibling flicker).
3. Give `enableTimeout: false` widgets a passive staleness signal driven by the path's last-update timestamp rather than an active timeout.

## Related (separate, pre-existing)

Reliability + correctness review also noted that the reset is **shallow**: `timeoutPathObservable` nulls the derived `pathDataUpdate$` but leaves the upstream `_pathData$` holding the stale value, so a later `_pathState$` emission (e.g. a notification/zone change for the path) re-pairs the stale value via `combineLatest` and can resurface it. That predates this PR and affects the "did it really clear?" question independently — worth its own look.

Surfaced by the adversarial + reliability review of PR #252.

mairas:
## What

Two linked fixes in the timeout/clearing path of `DataService`. Fixes #254. Builds on B2 (#243, #311).

**1. The `enableTimeout:false` frozen-widget reversal.** After #252 scoped timeout clearing to the source that actually timed out, a widget with `enableTimeout:false` (shipped default for `widget-boolean-switch`, `widget-autopilot`) bound to a dead source renders its **last value forever** — it has no `timeout()` operator, so it only ever cleared as a side effect of the old null-all reset. A frozen-but-plausible reading is a hazard for safety-adjacent widgets.

**2. A pre-existing shallow reset.** `timeoutPathObservable` nulled the *derived* `pathDataUpdate$` but left the upstream `_pathData$` holding the stale value, so a later `_pathState$` emission (a notification/zone change) re-paired the stale value via `combineLatest` and **resurfaced** it.

## Approach (Gate A Option 2 — clear only when *no source is live*)

Re-add cross-clearing of the sibling/default registrations, but **only in the genuinely all-silent case**, so #252's live-sibling fix is never reintroduced (cross-clearing provably cannot fire while any source is live).

- **Liveness is a per-path wall-clock receipt, not `refCount`.** `refCount` counts live *subscribers*, not live *data* — the frozen widget keeps refCount>0 while on-screen with a dead sensor, so refCount could never fire the cross-clear. New `_lastObservedByPath: Map<path, number>` is stamped with `Date.now()` on every delta. A path is "stale" iff nothing fed it within the timing-out widget's TTL. A live sibling produces deltas → refreshes the clock → `stale=false` → cross-clear is never entered (the exact #252 guarantee). `_pathRegisterByPath` is used only to *enumerate* the sibling registrations to clear.
- **`_lastObservedByPath` is pruned in `removePathsForContext`** alongside `_skData`/`_pathMetaByPath` — a new per-path map would otherwise reintroduce the unbounded-growth leak B2 just fixed.
- **Deep reset (shallow-fix core):** a shared `resetRegistrationValue` helper pushes onto the **upstream** subjects — `_pathData$.next({value:null,timestamp:null})` **then** `_pathState$.next(States.Normal)` — and lets the standing `combineLatest` produce the derived emission. Nulling the upstream cache means a later `_pathState$` emission re-pairs *null* with the new state; the value stays cleared. Order is load-bearing (null-data first → the transient is `{null, staleState}`, a blank widget with a briefly-old badge; the reverse would flash the stale value with a fresh `Normal` state — the exact resurrection being fixed). Same helper is used for the direct reset and every cross-cleared sibling, so cross-cleared siblings can't resurface either.

## Notes for review

- **`timeoutPathObservable` gains a required 4th param `dataTimeoutMs`** (threaded from `widget-streams.directive` — the sole production caller); the liveness window is the timing-out widget's own TTL.
- **Double emission:** the deep reset now produces two synchronous `combineLatest` emissions per reset (was one) — final value identical, documented in the helper's JSDoc.
- With multiple `enableTimeout:true` widgets of different TTLs on one path, the shortest window wins the cross-clear check — bounded and self-healing (a later delta recovers). Recovery is automatic everywhere: any new delta re-populates `_pathData$` and refreshes the receipt clock.

## Testing

- **Full local suite green (148 files.)** New coverage: the **#252 live-sibling guard** (a source fed within the TTL → no cross-clear, siblings keep their value); all-silent cross-clear fires; the `enableTimeout:false` victim registration (which never calls `timeoutPathObservable` itself) is reached; the shallow-reset resurface bug (null, then push a notification state change → value stays null); a cross-cleared sibling stays null after a notification; and `_lastObservedByPath` foreign-context prune. Recency-dependent cases use vitest fake timers (`setSystemTime`).
- Lint clean; betterer:ci 179 (a `paths-options.component.ts` hash update for the new `MatHint` import is regenerated in the docs commit).
- A plain-English `<mat-hint>` documents the semantic in the path config; the `enableTimeout` JSDoc got a one-line shared-path addendum.

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -130,7 +130,7 @@ exports[`strictNullChecks`] = {
       [119, 4, 20, "Object is possibly \'undefined\'.", "2522653710"],
       [120, 4, 20, "Object is possibly \'undefined\'.", "2522653710"]
     ],
-    "src/app/widget-config/paths-options/paths-options.component.ts:3797853945": [
+    "src/app/widget-config/paths-options/paths-options.component.ts:2311600985": [
       [31, 51, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'IAddNewPathObject\'.", "2620553983"],
       [32, 40, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'string\'.", "2620553983"],
       [33, 54, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'IDynamicControl[]\'.", "2620553983"],
--- a/src/app/core/directives/widget-streams.directive.spec.ts
+++ b/src/app/core/directives/widget-streams.directive.spec.ts
@@ -20,6 +20,7 @@ class FakeDataService {
         path: string;
         source: string;
         pathType: string;
+        dataTimeoutMs: number;
     }[] = [];
 
     subscribePath(path: string, source?: string): Observable<IPathUpdate> {
@@ -46,8 +47,8 @@ class FakeDataService {
         return { data$, release };
     }
 
-    timeoutPathObservable(path: string, source: string, pathType: string): void {
-        this.timeoutCalls.push({ path, source, pathType });
+    timeoutPathObservable(path: string, source: string, pathType: string, dataTimeoutMs: number): void {
+        this.timeoutCalls.push({ path, source, pathType, dataTimeoutMs });
     }
 }
 
@@ -411,7 +412,9 @@ describe('WidgetStreamsDirective', () => {
         // Do not emit anything; advance virtual time beyond 20ms to trigger timeout
         await vi.advanceTimersByTimeAsync(30);
         expect(dataSvc.timeoutCalls.length).toBe(1);
-        expect(dataSvc.timeoutCalls[0]).toEqual({ path: 'env.to', source: 'default', pathType: 'string' });
+        // dataTimeout (0.02 s) is threaded through as the ms window; compute it the same way the
+        // directive does so the assertion is exact regardless of float representation.
+        expect(dataSvc.timeoutCalls[0]).toEqual({ path: 'env.to', source: 'default', pathType: 'string', dataTimeoutMs: 0.02 * 1000 });
     });
 
     it('forwards a configured non-default source into timeoutPathObservable', async () => {
@@ -426,7 +429,7 @@ describe('WidgetStreamsDirective', () => {
         // The stream's effective source must reach the reset so a source-bound widget
         // clears its own registration, not the default bucket (#206).
         await vi.advanceTimersByTimeAsync(30);
-        expect(dataSvc.timeoutCalls[0]).toEqual({ path: 'env.to', source: 'n2k-1', pathType: 'string' });
+        expect(dataSvc.timeoutCalls[0]).toEqual({ path: 'env.to', source: 'n2k-1', pathType: 'string', dataTimeoutMs: 0.02 * 1000 });
     });
 
     it('applies convertUnitTo to numeric values (initial + sampled)', async () => {
@@ -657,7 +660,7 @@ describe('WidgetStreamsDirective', () => {
  */
 class TtlFakeDataService {
     subjects = new Map<string, BehaviorSubject<IPathUpdate>>();
-    timeoutCalls: { path: string; source: string; pathType: string }[] = [];
+    timeoutCalls: { path: string; source: string; pathType: string; dataTimeoutMs: number }[] = [];
 
     private keyFor(path: string, source?: string): string {
         return `${path}|${source?.trim() || 'default'}`;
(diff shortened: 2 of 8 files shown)


mairas:

**7 findings → 4 confirmed / 3 refuted.** The review caught a genuine regression this PR had introduced — the reason to run it before merge, not after. Fixed in `df5861d6`.

### ✅ Fixed — [P1/P2] deep reset clobbered a persistent alarm state (correctness + adversarial + api-contract, same root)
The deep reset pushed `_pathState$ = States.Normal`. Concrete failure: a battery-voltage path in active **Alarm** (widget red) times out → reset nulls the value *and* forces state Normal; `_skData.state` stays Alarm. The feed resumes still-low (alarm still active on the server → **no new notification delta**), so `updatePathData` pushes only the value → `combineLatest` emits `{value, Normal}` → **the alarm coloring is gone for the rest of the alarm's life** (the notification handler dedupes on state *transition*, and `_skData.state` never changed). On `main` the shallow reset left `_pathState$` at Alarm, so recovery re-paired `{value, Alarm}` — this PR regressed it.

**Fix (a simplification):** the resurface bug depends *only* on nulling the upstream `_pathData$`; the state is a separate server-notification channel the reset has no business touching. `resetRegistrationValue` now pushes **only** `_pathData$.next({value:null,timestamp:null})`. Reset yields `{null, realState}`, recovery yields `{value, realState}` — a still-active alarm survives. Bonus: this **removes the double-emission** the PR originally flagged (one `combineLatest` emission per reset now) and moots the null-before-Normal ordering concern. New regression test asserts an Alarm survives a timeout→value-only-resume cycle.

### #3 [P3] documented, behavior kept — shortest-window cross-clear
On an all-silent shared path, the strictest-tolerance (shortest-TTL) widget's timeout cross-clears a longer-tolerance sibling's value early. **Kept per-path** (safety-conscious + robust) and documented at the site + JSDoc. The reviewer's per-(path,source) alternative is *worse*: the timeout operator is one-shot (never re-arms), so a sibling skipped because its source was momentarily fresh would never later get cross-cleared — leaving the `enableTimeout:false` widget frozen forever (the original #254 bug). Rejected with that rationale in a comment.

### Refuted (verified)
- **[P3] null-before-Normal ordering has no test** — moot: the fix removes the state push entirely, so there is no intra-call ordering left.
- **[P3] `lastObserved === undefined` disjunct untested** — valid gap, **closed**: added a test that deletes the receipt and times out, exercising the `=== undefined` branch.
- **[P3] snapshot getters (`getPathObject`/`_skData`) still report the stale value after a reset** — pre-existing from #252 (the reset deliberately never writes `_skData`); B3 extends the same by-design property to siblings. Left as-is.

**Verification:** full local suite green (148 files), incl. the alarm-survives + undefined-recency + #252-live-sibling-guard + cross-clear + resurface tests; lint clean; betterer:ci 179.

PR #314, opened 2026-07-16
chore(ci): type-check and lint the tools/ generator directory

### LINKED ISSUE #308: Type-check + lint the tools/ generator directory in CI (gen-mcp-schema escapes strict-typing gates)
## Problem

The `tools/` directory (notably `tools/gen-mcp-schema/`) is **neither linted nor type-checked in CI**:

- `angular.json` `lintFilePatterns` covers only `src/**` and `plugin/**/*.{ts,js}` — not `tools/**`.
- `tsconfig.strict.json` (the betterer scope) and the app build both include only `src/**/*.ts`.
- `test:mcp-schema` runs via vitest, which transpiles with esbuild and **strips types** — it never type-checks.

So a type error, an `any`, or an unused var in the generator tooling passes `npm run ci` and only surfaces at runtime (if at all). This is the same gap #268 closed for `plugin/` (which now gets `tsc --noEmit -p plugin/tsconfig.json` inside `test:plugin`, plus a lint block).

Surfaced during the #293 review (PR #307): the reviewer correctly flagged the generator changes had no type/lint gate. The #293 changes themselves were verified type-clean by hand, but the gap should be closed so the next change to the generator can't drift.

## Proposed fix (mirror the `plugin/` precedent from #268)

1. Add `tools/gen-mcp-schema/tsconfig.json` (extends the root tsconfig, `noEmit`, strict) covering `tools/gen-mcp-schema/**/*.ts`.
2. Prepend `tsc --noEmit -p tools/gen-mcp-schema/tsconfig.json &&` to the `test:mcp-schema` script (exactly as `test:plugin` does).
3. Extend `angular.json` `lintFilePatterns` with `tools/**/*.ts` (the generator obeys the repo's `no-explicit-any`/`no-unused-vars` rules) and confirm `ng lint` passes.

Small, self-contained, and it hardens the exact tooling #293 repaired.

mairas:
## Why

The `tools/` directory (notably `tools/gen-mcp-schema/`) was neither linted nor type-checked in CI: `angular.json` linted only `src/**` and `plugin/**`, no tsconfig included `tools/`, and `test:mcp-schema` runs through vitest/esbuild which strips types without checking them. A `configVersion` regression in the generator was only caught by hand during #293 (flagged in the #307 review). This closes that gap so the next change to the generator can't drift.

## How

Mirrors the `plugin/` gating precedent from #268:
- New `tools/gen-mcp-schema/tsconfig.json` (extends root, `noEmit`), run via `tsc --noEmit` prepended to `test:mcp-schema` — the same pattern `test:plugin` uses.
- `tools/**/*.ts` added to `angular.json` `lintFilePatterns`.
- Schematic EJS templates under `tools/schematics/**/files/**` are eslint-ignored (not parseable TypeScript).
- Fixed the 3 `@typescript-eslint/array-type` violations the new lint gate surfaced in `generate.ts`.

Verified locally: `tsc --noEmit` on the new tools tsconfig is clean; `npm run lint` passes; a deliberately-injected `Array<T>` made the lint gate fail (then reverted), confirming `tools/**/*.ts` is genuinely enforced.

Closes #308.

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -111,7 +111,8 @@
               "src/**/*.ts",
               "src/**/*.html",
               "plugin/**/*.ts",
-              "plugin/**/*.js"
+              "plugin/**/*.js",
+              "tools/gen-mcp-schema/**/*.ts"
             ]
           }
         }
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "test": "ng test --watch=false",
     "test:interactive": "ng test",
     "test:headless": "CI=1 ng test --watch=false",
-    "test:mcp-schema": "vitest run --config tools/gen-mcp-schema/vitest.config.ts",
+    "test:mcp-schema": "tsc --noEmit -p tools/gen-mcp-schema/tsconfig.json && vitest run --config tools/gen-mcp-schema/vitest.config.ts",
     "test:plugin": "tsc --noEmit -p plugin/tsconfig.json && vitest run --config plugin/vitest.config.ts",
     "gen:mcp-schema": "UPDATE_SCHEMA=1 npm run test:mcp-schema",
     "lint": "ng lint",
--- a/tools/gen-mcp-schema/generate.ts
+++ b/tools/gen-mcp-schema/generate.ts
@@ -208,7 +208,7 @@ function extractColorHex(root: string): Map<string, string> {
 function extractColors(root: string): DesignSystem['colors'] {
   const file = path.join(root, APP_SERVICE);
   const raw = literalToValue(findArrayLiteral(parseSourceFile(file), 'configurableThemeColors'));
-  const list = raw as Array<{ value: unknown; label: unknown }>;
+  const list = raw as { value: unknown; label: unknown }[];
   const hexByToken = extractColorHex(root);
   return list.map((c) => {
     const value = String(c.value);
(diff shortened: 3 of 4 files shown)


mairas:
## Review synthesis (correctness + project-standards → findings addressed)

**Correctness — clean.** Traced the full CI chain: `pr.yml` → shared `pr-checks` → local `run-tests` (`npm run ci`) → `test:mcp-schema` → `tsc --noEmit -p tools/gen-mcp-schema/tsconfig.json && vitest`. `--listFiles` confirms the tsconfig compiles the real generator sources (not zero files / wrong root); the `&&` short-circuit makes a type error fail the script; lint's `no-explicit-any: error` is active. The `Array<T>`→`T[]` edits are pure type syntax, erased at compile time. Gate is genuinely enforcing, not a no-op. *(Non-blocking notes: the base tsc is non-strict — consistent with how the app itself is base-checked, `tools/` is out of the betterer strict ratchet by design; and CLAUDE.md's `ci`-script description is pre-existing stale, unrelated to this PR.)*

**Project-standards — 2 medium findings, both addressed.** The original diff linted `tools/**/*.ts` (whole tree) while type-checking only `tools/gen-mcp-schema/` — breaking the scope-parity the `plugin/` precedent keeps, and leaving a trap: `create-host2-widget.spec.ts` and any *future* tool outside gen-mcp-schema would be linted but escape the type-check this packet exists to add.

**Fix (pushed):** narrowed the lint glob to `tools/gen-mcp-schema/**/*.ts` so lint and type-check cover the same set — which also matches the issue's named target ("the generator directory") and the plugin precedent exactly. Dropped the now-moot `tools/schematics/**/files/**` eslint ignore. The separate schematics scaffolder (its runtime `.js` + orphaned spec) is genuinely distinct work, explicitly left out of scope rather than half-gated. Re-verified locally: `tsc` gate green, `npm run lint` → "All files pass linting."

PR #315, opened 2026-07-16
fix(gauge): stop value box flicker on fast values (RPM) (#222)

### LINKED ISSUE #222: Radial gauge value box flickers on fast values (RPM): animatedValue tween flashes through intermediate numbers
## Symptom

A radial gauge bound to a fast, large-magnitude value (e.g. engine RPM) with a 500 ms sample time appears to update in **bursts**: every ~500 ms the numeric readout rapidly flashes through a run of intermediate values, then sits still until the next update. It reads as a flicker rather than a clean update.

## Root cause

Not the data pipeline. Two independent stages:

1. **Sampling is correct.** `WidgetStreamsDirective` throttles the path with RxJS `sampleTime(500)` (`src/app/core/directives/widget-streams.directive.ts:138`), so the widget receives exactly one value per 500 ms window (plus one fast initial emission). No bursting here.

2. **The gauge tweens the value box.** The radial gauge enables `@godind/ng-canvas-gauges` value animation. In `src/app/widgets/widget-gauge-ng-radial/widget-gauge-ng-radial.component.ts:306`:

   ```
   g.animation = ...; g.animatedValue = ...; g.animationRule = 'linear';
   ```

   With `animatedValue: true`, the digital value box does not snap to the new number — it linearly interpolates from the previous value to the new one, rendering every intermediate integer through the gauge's own rAF loop. Duration is `gaugeAnimationDurationMs(500)` = `min(120, 500)` = **120 ms** (`src/app/core/utils/gauge-animation.util.ts:16`).

So each cycle: new sample lands → readout counts old→new over 120 ms (the "flashing burst") → idle ~380 ms → repeat.

## Why RPM makes it obvious

RPM moves fast and covers a large range. Between two 500 ms samples the value can jump hundreds of RPM, and the linear tween walks through every integer in that gap in 120 ms. On a slow, small-range value (depth, temperature) the same animation is imperceptible; on RPM it's a visible flicker.

## Current constraints

- Animation is **hardcoded on** after the first frame: `animatedValue`/`animation` are bound to `animationEnabled()` = `gaugeBootstrapped()` (`widget-gauge-ng-radial.component.ts:97,209`). There is no per-widget config toggle.
- The needle animation itself is desirable — it's specifically the animated *value box text* that produces the flicker.

## Options

1. **Set `animatedValue: false`** for the radial gauge — canvas-gauges then shows the actual value in the box immediately while the needle still animates smoothly. Simplest fix if the library behaves as documented; needs verification.
2. **Add a config toggle** ("smooth value" / "snap value") so users pick per widget.
3. **Fork `@godind/ng-canvas-gauges`** (currently `^6.2.1`) if the library can't cleanly decouple needle animation from value-text animation, or if `animatedValue: false` has side effects we can't live with. This is the fallback if 1/2 don't give clean behavior.

Filing to track; likely start with option 1 + verify on a real RPM feed before deciding whether a gauge-lib fork is warranted.

mairas:
## Why

A radial gauge bound to a fast, large-range value (e.g. engine RPM at ~500 ms sample time) appears to **flicker**: each new sample makes the numeric value box count from the old value to the new one across the 120 ms animation window, then sit idle until the next sample. Root cause is the gauge library's `animatedValue`, which tweens the value-box text through every intermediate integer instead of snapping to the new number. The **needle** animation is fine — it's specifically the animated value-box text that reads as flicker (#222).

## How

Introduce a single `gaugeAnimationOptions` policy factory (`src/app/core/utils/gauge-animation.util.ts`) and route **all three** ng-canvas gauges (radial, linear, compass) through it, at both the initial-options build **and** the post-bootstrap `gauge.update()` site that previously turned `animatedValue` back on (the known "line-209 trap"). The factory keeps `animatedValue` permanently off while leaving `animation` (needle) gated on the post-first-frame flag and the 120 ms duration cap (`gaugeAnimationDurationMs`) untouched. The only behavioural change at each site is `animatedValue` → `false`; `animation`, `animateOnInit`, and `animationRule` keep their prior values.

The value now snaps to each sample while the needle still animates smoothly. A pure unit spec pins the decoupling (value box never animates in either flag state; needle animation follows the flag; never animates on init) so the tween can't be silently re-enabled by a future edit to any of the six option sites.

`.betterer.results` carries only the compass file's content-hash key update — its one pre-existing strictNullChecks issue and the 179 total are unchanged.

**Boat-verify:** the no-flicker behaviour on live RPM is batched to the Stage-1-close boat session.

Closes #222.

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -213,7 +213,7 @@ exports[`strictNullChecks`] = {
       [44, 9, 9, "Type \'null\' is not assignable to type \'string\'.", "2966443234"],
       [148, 40, 27, "Object is possibly \'undefined\'.", "3506675417"]
     ],
-    "src/app/widgets/widget-gauge-ng-compass/widget-gauge-ng-compass.component.ts:3154966941": [
+    "src/app/widgets/widget-gauge-ng-compass/widget-gauge-ng-compass.component.ts:716655136": [
       [237, 30, 9, "Argument of type \'string | undefined\' is not assignable to parameter of type \'string\'.\\n  Type \'undefined\' is not assignable to type \'string\'.", "4119212820"]
     ],
     "src/app/widgets/widget-iframe/widget-iframe.component.ts:2602616758": [
--- a/src/app/core/utils/gauge-animation.util.spec.ts
+++ b/src/app/core/utils/gauge-animation.util.spec.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest';
-import { GAUGE_ANIMATION_MS, gaugeAnimationDurationMs } from './gauge-animation.util';
+import { GAUGE_ANIMATION_MS, gaugeAnimationDurationMs, gaugeAnimationOptions } from './gauge-animation.util';
 
 describe('gaugeAnimationDurationMs', () => {
   it('uses a short fixed window instead of ~the whole sample interval', () => {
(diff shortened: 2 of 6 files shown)


mairas:
## Review synthesis (correctness + maintainability → clean, one nit fixed)

**Correctness — no findings.** Traced every animation-option write across all three components + the library `update()` semantics. The invariant holds: `animatedValue` is provably `false` on every path in both flag states. The only two sites per component that write it (`buildGaugeOptions` and the post-bootstrap trap) both route through `gaugeAnimationOptions()`; every other `.update()` passes either `gaugeOptions` (carries `false`) or a partial payload that omits the field — and `base-gauge.update()` was confirmed to *merge* keys, never reset, so a partial can't re-enable the tween. Type safety confirmed: the four fields are declared on the shared `GenericOptions` base both option types extend.

**Maintainability — net win.** The factory is proper DRY, not premature abstraction: it consolidates a 4-field policy that was duplicated identically across 6 sites, and #222 was precisely one of those sites (the post-bootstrap `update`) diverging to re-enable `animatedValue`. It also corrected an actively-wrong compass comment. One low nit: the `Pick<RadialGaugeOptions,…>` label read as radial-specific though the policy is gauge-agnostic — **fixed** with a comment noting the fields are the shared `GenericOptions` base (the package doesn't re-export it, so the Pick is a pragmatic proxy).

**Deliberate scope call (testing gap):** both reviewers flagged that the spec pins the invariant at the factory (its single source — all 6 sites verifiably consume it) but not via a component-level lifecycle spy asserting no `update()` ever carries `animatedValue:true`. Closing that means three brand-new component specs (none exist today) that can't be run locally on the boat Pi (component specs need the full Angular builder). Given the single-source factory guard already holds and the gap is rated low, the factory-level spec is kept as the guard rather than expanding the packet with unverifiable specs. The invariant lives at one chokepoint by design.

**Verification:** lint clean; `betterer:ci` 179 (unchanged; only the compass content-hash key moved); factory spec green (6/6). Full suite + component behaviour = CI. **Boat-verify** (no-flicker on live RPM) batched to the Stage-1-close session.

PR #316, opened 2026-07-16
feat(boolean): resizable controls + OFF-state visibility (port Kip #1106/#1107)

### LINKED ISSUE #240: Port Kip #1106: resizable boolean switches/lights/buttons

mairas:
## Why

The boolean switch-panel controls (switch, indicator button, light) rendered at a fixed 180×35 viewBox, so labels clipped on large or multi-control panels, and an OFF switch/light was only *dimmed* rather than clearly off — hard to read at a glance on the water.

## How

Ports two upstream Kip commits as a stacked pair (kept separate, never squashed):

- **Resize (`5a8c0895`, #240):** new `boolean-control-layout.util` (copied verbatim from upstream) derives every shape coordinate from `(type, width, height)` and binary-searches the largest per-control height whose label still fits, measured against a real canvas font via `CanvasService`. The three `svg-boolean-*` components move to a single layout-driven `foreignObject` template. The fork's fixed `35·w/180` proportional resize is replaced; the now-dead `nbCtrl` field is dropped (no-dead-code).
- **OFF-state (`f9d657cf`, #241b):** the OFF indicator gets a dedicated `offColor` (theme `*Dimmer`) instead of reusing the label color, the OFF label is dimmed, and the light gains layered specular/depth gradients for a 3D lit look; stroke widths scale with height.

**Fork divergence preserved:** all three components keep `ChangeDetectionStrategy.OnPush`; the fork's swipe-guard pointer logic is untouched. Upstream's rewritten templates use `data()?.ctrlLabel` (null-safe) vs the fork's old `data().ctrlLabel`, so adopting them is strictly safer. `IDimensions` was *not* re-added to the interfaces (the fork dedups it).

**Scope guard:** the momentary-button half of Kip #1107 (#241a) already shipped separately (#286) and is untouched here.

## Verification

`npm run lint` clean; `betterer:ci` 179 (unchanged — `widget-boolean-switch` entry re-hashed only, same 2 errors at shifted lines); the runnable layout-util spec passes 4/4. The two new `svg-boolean-{switch,light}` offColor specs are CI-verified (component specs require the Angular builder, which the boat Pi doesn't run), mirroring the proven `svg-boolean-button` spec pattern — they inject no services.

**Boat-verify:** label fit across panel sizes/counts, OFF-state colors, and the 3D light appearance are batched to the Stage-1-close boat session. One thing to eyeball there: the switch template's `foreignObject` div omits the XHTML `xmlns` the button/light divs carry (ported faithfully from upstream `5a8c0895`).

Closes #240. Addresses #241 (the 241b visual half; 241a already shipped in #286).

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -193,9 +193,9 @@ exports[`strictNullChecks`] = {
       [1358, 21, 11, "\'b.longitude\' is possibly \'undefined\'.", "3370058570"],
       [1358, 35, 11, "\'a.longitude\' is possibly \'undefined\'.", "2985066057"]
     ],
-    "src/app/widgets/widget-boolean-switch/widget-boolean-switch.component.ts:800795185": [
-      [71, 38, 9, "Argument of type \'string | undefined\' is not assignable to parameter of type \'string\'.\\n  Type \'undefined\' is not assignable to type \'string\'.", "4119212820"],
-      [94, 10, 12, "Object is possibly \'null\'.", "4261782498"]
+    "src/app/widgets/widget-boolean-switch/widget-boolean-switch.component.ts:1261729337": [
+      [77, 38, 9, "Argument of type \'string | undefined\' is not assignable to parameter of type \'string\'.\\n  Type \'undefined\' is not assignable to type \'string\'.", "4119212820"],
+      [100, 10, 12, "Object is possibly \'null\'.", "4261782498"]
     ],
     "src/app/widgets/widget-datetime/widget-datetime.component.ts:3033260432": [
       [60, 40, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'string\'.", "2620553983"],
--- a/src/app/widgets/svg-boolean-button/svg-boolean-button.component.svg
+++ b/src/app/widgets/svg-boolean-button/svg-boolean-button.component.svg
@@ -1,71 +1,45 @@
 <svg
    [attr.height]="dimensions().height"
    [attr.width]="dimensions().width"
-   [attr.viewBox]="viewBox"
+   [attr.viewBox]="'0 0 ' + dimensions().width + ' ' + dimensions().height"
+  (pointerup)="handleClickUp()"
+  (pointerleave)="handleClickUp()"
+  (pointerdown)="handleClickDown($event)"
+  (pointermove)="handlePointerMove($event)"
+  (pointercancel)="handlePointerCancel()"
    version="1.1"
    id="svg9"
    xmlns="http://www.w3.org/2000/svg"
    xmlns:svg="http://www.w3.org/2000/svg">
-  <defs id="defs9" />
-  <g (pointerup)="handleClickUp()" (pointerleave)="handleClickUp()" (pointerdown)="handleClickDown($event)" (pointermove)="handlePointerMove($event)">
-    <g id="disabled" transform="translate(0,35)">
-      <rect
-        id="clickRecDisabled"
-        width="180"
-        height="35"
-        x="0"
-        y="0"
-        pointer-events="all"
-        style="display:inline;opacity:1;fill:transparent;fill-opacity:1;stroke:none;stroke-width:0;stroke-dasharray:none;stroke-opacity:0" />
-      <rect
-        id="buttonDisabled"
-        width="165.02518"
-        height="25.025183"
-        rx="3.6672263"
-        x="7.4874101"
-        y="4.9874086"
-        [attr.fill]="labelColorDisabled"
-        style="display:inline;mix-blend-mode:normal;fill-opacity:1;stroke:#ffffff;stroke-width:1.5;stroke-dasharray:none;stroke-opacity:0.2" />
-      <svg x="12" y="0" width="156" height="35" viewBox="0 0 156 35">
-        <text
-          id="textDisabled"
-          [attr.fill]="labelColorDisabled"
-          style="text-anchor: middle;text-align: center;font-style:normal;font-variant:normal;font-weight:bold;font-stretch:normal;font-size:14px;font-family:arial;font-variant-ligatures:normal;font-variant-caps:normal;font-variant-numeric:normal;font-variant-east-asian:normal"
-          x="77.968102"
-          y="22">
-            {{ data().ctrlLabel }}
-        </text>
-      </svg>
-    </g>
-    <g id="enabled">
-      <rect
-        width="180"
-        height="35"
-        id="rect1"
-        x="0"
-        y="0"
-        pointer-events="all"
-        style="display:inline;opacity:1;fill:transparent;fill-opacity:1;stroke:none;stroke-width:0;stroke-dasharray:none;stroke-opacity:0"/>
-      <rect
-        id="buttonEnabled"
-        width="165.02518"
-        height="25.025183"
-        rx="3.6672263"
-        x="5.9874086"
-        y="4.9874086"
-        [attr.fill]="valueColor"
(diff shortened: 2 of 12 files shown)


mairas:
## Review synthesis (correctness + adversarial + maintainability)

**Port fidelity — verified.** Both correctness and adversarial independently confirmed `boolean-control-layout.util.ts` and all three `.svg` templates are **byte-identical to upstream `f9d657cf`**, and the three component `.ts` files match upstream semantics with exactly the three declared fork adaptations (IDimensions import path, the `if (!this.streams) return` guard, dead `nbCtrl` removal). `switchStrokeWidth` binding matches upstream (knob scales, track stays hardcoded 1.5 — verified against `f9d657cf`, faithful). #241a momentary logic untouched (already shipped in #286).

**Correctness — no findings.** Layout math is robust (`scaleFactor` floors at 1/35; `Math.max(0,…)` guards; the binary search is monotonic and terminates in [1, panel/N]). OFF-state `offColor` sourced per-color from `theme.*Dimmer`, applied only to OFF elements, no ON-path leak. Templates are null-safe under OnPush. The switch `foreignObject` div's missing `xmlns` was investigated and is **inert** — Angular's compiler sets `preventNamespaceInheritance` on `foreignObject`, so the child `<div>` is HTML-namespaced regardless; it also matches upstream. Left as-is (adding it would only diverge the faithful port).

**Adversarial — two edge-case behaviors the port adopts from upstream, tracked as #318:**
1. *Cold-load font race* (~0.62): the private measure-canvas has no font-ready hook, so a first measurement before Roboto loads can mis-size controls until the next resize (persists on a static dashboard).
2. *Long-label panel shrink* (~0.85, reproduced): `measureBooleanControlsHeight` picks one shared height via `.every()`, so one verbose label shrinks every control; long-label + narrow-widget reaches a 1px, untappable sliver.

These are inherent to upstream's measure-based sizing (vs the fork's old `35·w/180` geometry) — not defects in the port. **Filed as #318** with suggested mitigations (min touch-target floor; font-ready re-measure) and flagged on the Stage-1 boat checklist to confirm real-world severity before fixing. This is **boat-verify** and covered by the pre-agreed escalation (misrender → follow-up, not revert).

**Maintainability — addressed / tracked.** `getColors` is now byte-identical between switch and light (this PR deepened it) → **filed as #319** (extract a shared `resolveDimmerColors`), deliberately kept out of this PR to preserve the focused faithful port. Fixed here (commit `39bc3cb7`): anchoring comment on the layout geometry + `scaleValue`→`scaleFactor` rename + 4 degenerate-input robustness tests (0×0, negatives, empty/oversized labels → finite, non-negative, terminates).

**Verification:** lint clean; `betterer:ci` 179 (unchanged); util spec 8/8. Full suite + rendered behaviour = CI + the boat session.

PR #317, opened 2026-07-16
fix(history-chart): bound reconnect re-backfill hold and cancel abandoned fetch (#281)

### LINKED ISSUE #281: History-chart reconnect re-backfill: bound the coalesced-chain hold and cancel the abandoned fetch under a slow provider

mairas:
## Why

Follow-up to #85 (surfaced in its round-5 review). Behind the non-default History engine flag, the reconnect re-backfill holds the live tail while a fetch is in flight to keep append-only chart consumers chronological. Two residual problems under a slow/hung history provider (InfluxDB / parquet, which commonly restart alongside the SK server) with sustained WebSocket flapping:

1. The `RECONNECT_HOLD_MS` (3s) cap was **per run**. `reconnectPending` re-runs the re-backfill back-to-back in the `finally`, each earning a fresh 3s hold, so a coalesced chain could freeze the tail for ~N×3s.
2. When the cap won the race, the superseded fetch was dropped but its HTTP GET was **not aborted** — it self-terminated only at the 30s timeout, so flapping faster than the cap could stack up to ~10 concurrent History GETs onto an already-struggling provider.

## How

- **Shared hold budget:** a single wall-clock deadline (`ctx.holdDeadline`) is armed by the first run of a coalesced chain and shared by every re-run, which races only the *remaining* budget. Once the budget is spent the chain stops and draws the honest gap. Total hold is one `RECONNECT_HOLD_MS`, not N×.
- **Fetch cancellation:** an `AbortController` is threaded through `fetchBackfill` into `HistoryApiClientService.getValues`; on cap (or teardown) the abandoned fetch is aborted, and `getValues` bridges the signal via `takeUntil(fromEvent(signal, 'abort'))` to unsubscribe the request — cancelling the underlying `HttpClient` GET instead of leaving it to time out and stack.

`getValues` gained an optional `signal?: AbortSignal` — additive, no existing caller affected. Single-run behaviour is unchanged (a length-1 chain arms `deadline = now + RECONNECT_HOLD_MS`, so the remaining budget equals the full cap). Adds 3 tests (chain-hold bound, abort-on-cap wiring, real HTTP cancellation via `HttpTestingController`). Lint clean; betterer 179 (neither service was in the baseline).

closes #281

### DIFF
--- a/src/app/core/services/history-api-client.service.spec.ts
+++ b/src/app/core/services/history-api-client.service.spec.ts
@@ -305,6 +305,24 @@ describe('HistoryApiClientService', () => {
     await expect(promise).rejects.toMatchObject({ name: 'HistoryRequestError', status: 0 });
   });
 
+  it('cancels the in-flight HTTP request when the abort signal fires', async () => {
+    connectionStub.serverServiceEndpoint$.next(CONNECTED_ENDPOINT);
+    const controller = new AbortController();
+
+    const promise = service.getValues({ paths: 'navigation.speedThroughWater:avg', resolution: 1 }, controller.signal);
+    await new Promise(resolve => setTimeout(resolve));
+
+    const req = httpMock.expectOne(request => request.url === VALUES_URL);
+    expect(req.cancelled).toBe(false);
+
+    controller.abort();
+
+    // Aborting unsubscribes the request (cancelling the underlying GET) and settles the promise as a
+    // benign request error, rather than leaving the request to hang until the 30s timeout.
+    await expect(promise).rejects.toBeInstanceOf(HistoryRequestError);
+    expect(req.cancelled).toBe(true);
+  });
+
   it('should return null when history paths endpoint returns 500', async () => {
     connectionStub.serverServiceEndpoint$.next({
       state: EndpointStatus.Connected,
--- a/src/app/core/services/history-api-client.service.ts
+++ b/src/app/core/services/history-api-client.service.ts
@@ -1,6 +1,6 @@
 import { Injectable, inject, DestroyRef } from '@angular/core';
 import { HttpClient, HttpErrorResponse, HttpParams } from '@angular/common/http';
-import { firstValueFrom, timeout } from 'rxjs';
+import { firstValueFrom, fromEvent, takeUntil, timeout } from 'rxjs';
 
 /** A backfill request that hangs past this is treated as a transient failure, not left to spin forever. */
 const HISTORY_VALUES_TIMEOUT_MS = 30_000;
@@ -207,6 +207,9 @@ export class HistoryApiClientService {
    *   - from, to, duration: define the time range
    *   - resolution: optional downsampling window
    *   - context: optional Signal K context (defaults to 'vessels.self')
+   * @param {AbortSignal} [signal] - Aborts the in-flight request, cancelling the underlying HTTP GET so
+   *   a caller that has abandoned the fetch (e.g. a reconnect re-backfill past its hold cap) does not
+   *   leave it running until the 30s timeout.
    *
    * @returns {Promise<IHistoryValuesResponse | null>} The history response, or `null` when the server
    *   has no history provider (no service URL, or a 404/501 response).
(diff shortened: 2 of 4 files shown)


mairas:
## Review synthesis (correctness + reliability → sound, findings addressed)

**Reliability — no findings; verdict sound.** Both target failure modes are genuinely bounded: the shared `holdDeadline` caps the coalesced-chain tail freeze at one `RECONNECT_HOLD_MS` (not N×), and the `AbortController` cancels the abandoned XHR so History GETs can't stack up to the 30s timeout. The coalesced-starvation case draws an *honest* gap (the seam advances; `emitReconnectGap` only fires past the source's staleness threshold) — not a silent #85-style hole. Deliberate, bounded trade-off, not a defect.

**Correctness — one low finding, fixed.** All six checks held (shared-budget armed-once-per-chain and reset at every non-disposed end → no leak into a later chain; `!budgetSpent` terminates spent chains; length-1 chains byte-identical to before; the `AbortController` is fresh and aborted only after the `takeUntil(fromEvent(signal,'abort'))` listener registers and after the race settles → no subscribe-after-abort race, no unhandled rejection, no double-settle vs the 30s timeout). The one finding: a comment claimed a torn-down fetch aborts "now," but on teardown the abort actually fires at the next race resolution (≤ remaining budget). **Fixed** the comment to state that accurately (the behavior is bounded + harmless — the disposed guard blocks any post-dispose emission — so no logic change).

**Testing gaps closed** (commit `0eb75f77`): added the three reliability-critical paths the reviewers flagged — **recovery re-arm** (a spent-budget chain clears `holdDeadline`; the next healthy reconnect arms a fresh full budget and backfills — proves no latched state degrades recovery), **teardown abort** (dispose mid-fetch → no post-dispose emission, signal aborts at the hold timer), and **coalesced fast-resolve recursion** (run 1 delivers within budget; run 2 runs under the *remaining* shared budget, aborting at the shared deadline a fresh cap would defer). Also hardened the existing bound test against a fake-timer id-ordering tie.

**Verification:** `history-chart-stream.service.spec` 41/41 (was 38 + 3 new); api-client spec 16/16; lint clean; `betterer:ci` 179 (unchanged). Not betterer-flagged — merges independently.

PR #320, opened 2026-07-16
fix(boolean): floor control height at a tappable minimum, ellipsize long labels (#318)

### LINKED ISSUE #318: Boolean control resize robustness: font-load race + long-label panel shrink (#316 review)
Surfaced by the multi-persona review of #316 (G4, port of Kip #1106/#1107). The port faithfully adopts upstream's **canvas-measured** boolean-control sizing (`boolean-control-layout.util` + `widget-boolean-switch`), replacing the fork's prior font-independent `35·width/180` geometry. Two robustness gaps in that measure-based approach — **confirm real-world severity at the Stage-1 boat session before fixing** (with short real labels on the real display these may be non-issues):

### 1. Cold-load font race (adversarial, conf ~0.62)
`measureBooleanControlsHeight` measures label widths against a private `canvas` 2D context at `700 {px} Roboto`. That private context subscribes to nothing on font load (unlike `CanvasService`, which holds a `fontsReadyPromise` and clears its cache on `document.fonts.ready`). Roboto loads async; on a cold boot the first (often only) measurement can run before it's available, so the binary search picks a height from fallback-font metrics. Nothing re-measures afterward — only `onResized`/the config effect call `updateCtrlDimensions`, not per-value ticks — so on a static dashboard the wrong size persists for the session until a manual resize.
- Compounding: weight **700 is not registered** anywhere (only 300/400/500), so both measurement and render fall back to synthetic-bold 500. They're mutually consistent, but a real 700 face is never used.
- Possible fix: re-measure on `document.fonts.ready` (or reuse `CanvasService`'s font-ready-aware measurement instead of a private context); decide whether to register a 700 face or measure at an available weight.

### 2. One long label shrinks the whole panel (adversarial, conf ~0.85, reproduced)
`measureBooleanControlsHeight` computes **one shared height** for all controls via `controls.every(...)` — the single widest label caps the height, forcing every other control down. Reproduced with the spec's linear text mock: three short labels → per-control height 66; adding one label "Emergency Bilge Pump Override Switch" drops **all four** to height 23 (~65% shrink). Long label + narrow widget reaches the 1px floor → an invisible, **untappable** control (safety-relevant on a marine control panel). The fork's old geometry kept controls tappable and clipped overflowing labels instead.
- Possible fix: a **minimum touch-target height floor** in the layout util (clamp up to a tappable minimum, clip the label if it then overflows) — restores the fork's "shape stays usable" property while keeping upstream's fit behavior in the normal case. This is util-level and unit-testable.

Also route the resize-storm re-measure cost (a fresh binary search per ResizeObserver tick, no debounce/rAF coalescing) through a perf pass on the 10× perf-harness.

Relates to #240 / #241 / #316.

mairas:
## Why

#318 (confirmed by headless render): a boolean-switch panel picks **one shared height** for all its controls — the tallest at which *every* label fits (`measureBooleanControlsHeight`). So a single long label dragged every control in the panel down to ~23px sub-tap-target slivers, wasting most of the tile and rendering an **Emergency Bilge Pump** switch as an unreadable, barely-tappable line. The mechanism shipped in #316 (the Kip #1106/#1107 port).

## How

1. **Tap-target floor** — `MIN_BOOLEAN_CONTROL_HEIGHT = 44`: the shared height floors at a tappable minimum instead of collapsing to fit a long label; the overlong label **ellipsizes** instead (each label wrapped in a `min-width:0` flex-item span with `text-overflow:ellipsis`).
2. **Label-lane backoff** — `MIN_BOOLEAN_LABEL_WIDTH = 24`: the switch/light shape lane is itself height-scaled (`48·height/35`), so a taller control leaves *less* horizontal room for its label. Starting at the floor, the height backs off one px at a time toward the natural fit-height while any **labeled** control's lane would fall below a legible minimum. A narrow tile therefore trades a little height for a visible label (pre-#318 behavior there); a roomy tile keeps the full floor and ellipsizes. Empty/whitespace labels are skipped — there is nothing to keep legible, so they never force a control below the floor.

The backoff is **monotone in both tile dimensions** (the returned height reduces to `max(best, min(height_start, Hs−1))`, a max/min of monotone terms), so a resize can never crash a control to a sliver.

### Review journey (this is why it took the shape it did)

Four adversarial passes, three of which found a real hole and drove a redesign:
1. Naive floor → *labels vanished* on narrow switch/light tiles (the floor inflates the height-scaled shape lane, zeroing the label lane) → added a backoff.
2. Backoff was label-blind → an *empty-labeled* control could drag the panel below the floor (reintroducing #318) → made the trigger label-aware.
3. An interim all-or-nothing gated variant → *non-monotonic collapse* on resize + all-or-nothing sibling drag-down → replaced with the monotone one-px backoff (this version).
4. Final pass: **no findings** — soundness confirmed by analytical proof + a ~1.5M-point sweep.

## Verification

- Util spec **17/17** — including the exact repros the review surfaced: the floor, the narrow-tile backoff, the empty/whitespace-label cases, a **width-91 monotonicity sweep** (the resize-crash case), the sibling drag-down, and a cross-type button+empty-switch guard. Lint clean; `betterer:ci` 179 (the util carries no strict-null baseline entry).
- **Headless render (perf-harness, real prod bundle, element-clipped, toolbar-free)** — control heights + label visibility confirmed by pixels + DOM metrics:

| tile | before | after |
|---|---|---|
| 203px, 3 mixed (one long) | 23px slivers | **44px** tappable; long label "Emergency Bilg…" ellipsized, short labels full |
| 91px, single "Navigation Lights" | (gated variant crashed to ~15px) | **43px** (backoff shed 1px), label "N…" |
| 91px, empty switch + long switch | — | **both 43px** (empty not dragged to a sliver) |
| 65px, single "Nav" | 44px, **label vanished** | **29px**, "Nav" fully visible |

Rendering behavior can't be unit-tested (jsdom canvas is a stub), so the ellipsis, no-vanish, tap-floor, and monotonicity were verified visually.

Two documented, pre-existing degenerate branches (an empty-label shape can still overflow a sub-shape-width tile; a genuinely unfittable label on a sub-30px tile degrades toward 1px) are **not regressions** — the fix only ever *lowers* height from the floor, so they are ≤ #316's behavior.

Closes #318.

### DIFF
--- a/src/app/widgets/svg-boolean-button/svg-boolean-button.component.svg
+++ b/src/app/widgets/svg-boolean-button/svg-boolean-button.component.svg
@@ -38,8 +38,8 @@
       xmlns="http://www.w3.org/1999/xhtml"
       [style.color]="isPressed ? labelColorEnabled : labelColorDisabled"
       [style.fontSize.px]="layout.labelFontSize"
-      style="width:100%;height:100%;display:flex;align-items:center;justify-content:center;font-family:Roboto,sans-serif;font-weight:700;line-height:1;white-space:nowrap;overflow:hidden;text-overflow:clip;">
-      {{ data()?.ctrlLabel }}
+      style="width:100%;height:100%;display:flex;align-items:center;justify-content:center;font-family:Roboto,sans-serif;font-weight:700;line-height:1;overflow:hidden;">
+      <span style="min-width:0;overflow:hidden;white-space:nowrap;text-overflow:ellipsis;">{{ data()?.ctrlLabel }}</span>
     </div>
   </foreignObject>
 </svg>
--- a/src/app/widgets/svg-boolean-light/svg-boolean-light.component.svg
+++ b/src/app/widgets/svg-boolean-light/svg-boolean-light.component.svg
@@ -35,8 +35,8 @@
       [style.opacity]="isEnabled ? 1 : 0.6"
       [style.fontSize.px]="layout.labelFontSize"
       [style.textShadow]="isEnabled ? '0 0 4px currentColor, 0 0 4px currentColor' : 'none'"
-      style="width:100%;height:100%;display:flex;align-items:center;justify-content:flex-start;font-family:Roboto,sans-serif;font-weight:700;line-height:1;white-space:nowrap;overflow:hidden;text-overflow:clip;">
-      {{ data()?.ctrlLabel }}
+      style="width:100%;height:100%;display:flex;align-items:center;justify-content:flex-start;font-family:Roboto,sans-serif;font-weight:700;line-height:1;overflow:hidden;">
+      <span style="min-width:0;overflow:hidden;white-space:nowrap;text-overflow:ellipsis;">{{ data()?.ctrlLabel }}</span>
     </div>
   </foreignObject>
 
--- a/src/app/widgets/svg-boolean-switch/svg-boolean-switch.component.svg
+++ b/src/app/widgets/svg-boolean-switch/svg-boolean-switch.component.svg
@@ -43,8 +43,8 @@
       [style.opacity]="isEnabled ? 1 : 0.6"
       [style.fontSize.px]="layout.labelFontSize"
       [style.textShadow]="isEnabled ? '0 0 4px currentColor, 0 0 4px currentColor' : 'none'"
-      style="width:100%;height:100%;display:flex;align-items:center;justify-content: flex-start;font-family:Roboto,sans-serif;font-weight:700;line-height:1;white-space:nowrap;overflow:hidden;text-overflow:clip;">
-      {{ data()?.ctrlLabel }}
+      style="width:100%;height:100%;display:flex;align-items:center;justify-content: flex-start;font-family:Roboto,sans-serif;font-weight:700;line-height:1;overflow:hidden;">
+      <span style="min-width:0;overflow:hidden;white-space:nowrap;text-overflow:ellipsis;">{{ data()?.ctrlLabel }}</span>
     </div>
   </foreignObject>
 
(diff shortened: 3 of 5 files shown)


mairas:
## Review synthesis (correctness + 4 adversarial passes → sound)

**Correctness — clean** (reviewed at the first floor version): floor math never overflows the tile or regresses the fitting case; spec numbers exact; ellipsis CSS correct in all three templates.

**Adversarial — 4 passes; the reason the design changed three times:**
1. Naive floor → labels *vanished* on narrow switch/light tiles (floor inflates the height-scaled shape lane, zeroing the label lane). **Fixed** → backoff.
2. Backoff was label-blind → an *empty-labeled* control dragged the panel below the floor, reintroducing #318. **Fixed** → label-aware trigger.
3. An interim all-or-nothing gated variant → *non-monotonic collapse* on resize (a 91px-wide switch crashed 43px→~15px when dragged taller) + all-or-nothing sibling drag-down. **Fixed** → monotone one-px backoff (this version).
4. Final pass → **no findings.** Soundness proven analytically (returned height = `max(best, min(height_start, Hs−1))`, monotone by construction) and empirically (~1.5M-point sweep, zero failures).

Every finding was fixed in-PR and pinned by a regression test (util spec now 17/17: floor, narrow backoff, empty/whitespace, width-91 monotonicity sweep, sibling drag-down, cross-type guard). The two residual notes from pass 4 are pre-existing degenerate branches (≤ #316 behavior), not regressions.

**Verification:** util 17/17; lint clean; betterer 179; headless render confirms 44px+ellipsis (203px), 43px monotone (91px), no vanish (65px), no shape clip — pixels + DOM metrics. This fix is machine-verified end-to-end.

Issue #321, opened 2026-07-16
Boolean control label sizing uses a font-race-prone private canvas measure (cold-boot mis-size)


mairas:
Split out of #318: the **long-label panel shrink** half was fixed in #320, which auto-closed #318, but the **cold-load font-race** half (from the same #316 review) was not addressed. Tracking it here so it isn't dropped.

`measureBooleanControlsHeight` (`boolean-control-layout.util.ts`) measures label widths against a private `canvas` 2D context at `700 {px} Roboto` (`widget-boolean-switch.component.ts`). That private context subscribes to nothing on font load — unlike `CanvasService`, which holds a `fontsReadyPromise` and clears its cache on `document.fonts.ready`. Roboto loads asynchronously, so on a cold boot the first (often only) measurement can run against fallback-font metrics, picking a wrong shared height. Nothing re-measures afterward: only `onResized` and the config effect call `updateCtrlDimensions`, not per-value ticks — so on a static dashboard the wrong size persists for the session until a manual resize.

Compounding: weight **700 is not registered** anywhere in the app (only 300/400/500), so both the measurement font and the rendered `font-weight:700` fall back to synthetic-bold 500 — mutually consistent, but a real 700 face is never used.

**Possible fix:** re-measure on `document.fonts.ready` (or reuse `CanvasService`'s font-ready-aware measurement instead of a private context); separately decide whether to register a 700 Roboto face or measure at an available weight.

Severity: low (~0.62 in review; self-corrects on the next resize). Verifiable with the perf-harness boolean render config added on the `test/boolean-render-shots` work.

PR #322, opened 2026-07-16
test(perf-harness): headless boolean-widget render matrix

mairas:
## Why

Verifying boolean-widget rendering (label fit, OFF-state, the #318 long-label panel-shrink) was framed as needing the on-boat visual session. It doesn't — the perf-harness already renders the real Skip prod bundle headless in Chromium against a mock Signal K server. This adds the missing config surface + a probe so the boolean widget can be judged from real pixels here, off the boat.

This is the harness that machine-reproduced #318 during the #320 fix (renders at 203/91/65px width confirmed the collapse and then the fix). Landing it makes future boolean-widget checks repeatable — including verifying #321 (the cold-boot font-race) without a boat trip.

## What

- **`perf-harness/lib/kip-config.mjs`** — new `booleanControlWidget({ controls, w, h, ... })` factory that emits the persisted `widget-boolean-switch` config shape (`multiChildCtrls` list + matching `paths` keyed by `pathID`; per-control `label`/`type`/`color`/`value`). Pure additive; reuses the file's existing `seq`/`uid`/`node` helpers.
- **`perf-harness/shot-boolean.mjs`** — probe that renders a control-count × label-length × tile-size matrix (single/3/4 controls; short/medium/very-long labels; narrow/x-narrow/large tiles) across the day (dark) and config-driven light themes, screenshots each `widget-boolean-switch` tile to `results/shots/boolean/`, and prints per-control shared height with a 44px tap-target flag (`ok`/`sub-tap`/`SLIVER`/`NO-WIDGET`). Includes the exact #318 repro: a 3–4 control panel with one very-long label alongside short ones.

App-config is pinned to the current schema version to skip the migration toast; night mode is documented as a runtime CSS filter (geometrically inert), so the theme axis is the config-driven light theme.

No app-code changes — `perf-harness/` only, self-contained (its own deps), never touches app builds or CI.

## Verify

```
cd perf-harness && npm install
CHROME_BIN=/usr/bin/chromium node shot-boolean.mjs --public ../public
```
(after a `../public` prod build). Writes screenshots + a geometry table to stdout.

Relates to #318 / #320 / #321.

### DIFF
--- a/perf-harness/README.md
+++ b/perf-harness/README.md
@@ -84,6 +84,14 @@ reusing a label for a different build.
 visual before/after; `verify-click.mjs` checks radar hit-testing (overlapping
 targets under view rotation) end-to-end.
 
+`shot-boolean.mjs` renders the boolean-switch (Switch Panel) widget across a
+control-count × label-length × tile-size matrix (day + light themes), writes
+`results/shots/boolean/<name>.png` per tile, and prints per-control shared height
+with a 44px tap-target flag — the #318 long-label panel-shrink probe. It boot-asserts
+that each tile rendered its seeded control count (exits non-zero otherwise). Invoke
+with `CHROME_BIN=/usr/bin/chromium node shot-boolean.mjs --public ../public` after a
+`../public` build.
+
 ## Interpreting the numbers
 
 | Question | Metric | Caveat |
--- a/perf-harness/lib/kip-config.mjs
+++ b/perf-harness/lib/kip-config.mjs
@@ -97,6 +97,39 @@ export function aisRadarWidget() {
   });
 }
 
+/**
+ * Switch-panel (widget-boolean-switch) factory. Mirrors the persisted config
+ * shape: multiChildCtrls holds the IDynamicControl list, paths holds one
+ * IWidgetPath per control keyed by a matching pathID. Each control:
+ *   { label, type ('1' switch | '2' button | '3' light), color, value, isNumeric, path }
+ * Controls render from multiChildCtrls regardless of live data; paths only feed
+ * value streaming (left un-streamed here for deterministic on/off states).
+ */
+export function booleanControlWidget({ controls, w = 4, h = 6, displayName = 'Switch Panel', showLabel = true, color = 'contrast' } = {}) {
+  const uuid = uid('bool');
+  const multiChildCtrls = [];
+  const paths = [];
+  controls.forEach((c, i) => {
+    const pathID = `bctrl-${String(++seq).padStart(4, '0')}`;
+    multiChildCtrls.push({
+      ctrlLabel: c.label, type: c.type ?? '1', pathID,
+      value: c.value ?? false, color: c.color ?? color, isNumeric: c.isNumeric ?? false,
+    });
+    paths.push({
+      description: c.label, path: c.path ?? `self.electrical.switches.bank.0.${i}.state`,
+      source: 'default', pathType: 'boolean', isPathConfigurable: true, sampleTime: 500, pathID,
+    });
+  });
+  return (x, y) => node(w, h, x, y, {
+    type: 'widget-boolean-switch', uuid,
+    config: {
+      displayName, showLabel, filterSelfPaths: true, paths,
+      enableTimeout: false, dataTimeout: 5, color, zonesOnlyPaths: false,
+      putEnable: true, putMomentary: false, multiChildCtrls,
+    },
+  });
+}
+
 /** Lay out widget factories in a grid and wrap them in one dashboard. */
 export function buildDashboards(factories, cols = 12) {
   let x = 0, y = 0, rowH = 0;
--- a/perf-harness/lib/kip-config.mjs
+++ b/perf-harness/lib/kip-config.mjs
@@ -102,8 +102,10 @@ export function aisRadarWidget() {
  * shape: multiChildCtrls holds the IDynamicControl list, paths holds one
  * IWidgetPath per control keyed by a matching pathID. Each control:
  *   { label, type ('1' switch | '2' button | '3' light), color, value, isNumeric, path }
- * Controls render from multiChildCtrls regardless of live data; paths only feed
- * value streaming (left un-streamed here for deterministic on/off states).
+ * Controls render from multiChildCtrls regardless of live data. Each path carries
+ * suppressBootstrapNull so the widget keeps the control's configured on/off value
+ * instead of being clobbered by DataService's synchronous bootstrap null (nothing
+ * streams these paths here, so without it every control would render OFF).
  */
 export function booleanControlWidget({ controls, w = 4, h = 6, displayName = 'Switch Panel', showLabel = true, color = 'contrast' } = {}) {
   const uuid = uid('bool');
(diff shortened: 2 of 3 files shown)


mairas:
## Code review — #322 (headless boolean-widget render matrix)

**Scope:** `perf-harness/lib/kip-config.mjs` (+33), `perf-harness/shot-boolean.mjs` (+169) vs `main`. Diagnostic tooling only — outside app build/lint/CI.
**Intent:** render the boolean-switch widget headless across a control-count × label-length × tile-size matrix so #318-class rendering can be judged from real pixels off the boat.
**Reviewers (5):** correctness, testing, maintainability, project-standards (always-on) + **adversarial** (202-line diff; the tool's value hinges on the emitted config faithfully matching the real widget). Skipped: security / api-contract / reliability (no matching surface).

The harness is structurally sound — the diamond findings are about **trustworthiness of its output**, since a verification tool that renders something plausible-but-wrong is worse than none. Two app-code claims were verified before synthesis: `suppressBootstrapNull` is a real `IWidgetPath` field with a working leading-null-drop gate, and the widget renders exactly one `.svg-widget svg` per control (all sharing one `ctrlDimensions`).

### P2 — Moderate

| 1 | `shot-boolean.mjs` | **No boot/count assertion — a degraded render passes silently.** The script only `waitForSelector`s one host and prints a `flag`; it never exits non-zero. A partial render (3 controls seeded, 1 svg) or a collapsed control (SLIVER) prints a flag a human can miss and exits 0. Sibling `run.mjs` asserts `count === expected` and throws — the harness's own documented "boot is asserted, not assumed" convention, unmet here. | testing, project-standards, adversarial(residual) | **0.95** | fix-here |
| 2 | `kip-config.mjs` | **`value:true` controls render OFF.** Widget subscribes every path; DataService seeds a `{value:null}` BehaviorSubject delivered synchronously (`take(1)`), overwriting the configured value → `Boolean(null)` = OFF. So every `value:true` control (and OFF-state distinctness) screenshots wrong, and the factory comment's "deterministic on/off states" is false. #318 *geometry* is unaffected (height keys off label, not value). **Verified:** `suppressBootstrapNull:true` on each path drops the leading null and preserves the configured state. | correctness | 0.72 | fix-here |
| 3 | `shot-boolean.mjs` | **Cold-font race — reported `ctrlH` can reflect fallback-font metrics, not settled Roboto.** The widget measures label widths against a private canvas and never re-measures on font load; the fixed `waitForTimeout(700)` doesn't guarantee Roboto has settled. A tile can report `ctrlH=44 ok` (fallback font) while the settled render shrinks below the tap floor — the exact #318 regression, hidden. This is the harness face of #321. | adversarial, correctness(residual), testing | 0.72 | partial fix-here (await `fonts.ready`) + #321 |

### P3 — Low

| 4 | `kip-config.mjs:11,34` | **`appConfig()` default `configVersion:12` is stale vs `LATEST_APP_CONFIG_VERSION=13`** (header comment also says 12). `shot-boolean` correctly works around it via `appV13()`, but `run.mjs`/`screenshot.mjs` still seed 12 → boot into the v12→v13 migration toast/reload inside their measurement window. **Pre-existing** (this PR didn't touch those lines). | maintainability, project-standards | 0.92 | **file-issue** |
| 5 | `perf-harness/README.md` | New `shot-boolean.mjs` probe undocumented; README documents every sibling probe (`smoke`, `screenshot`, `verify-click`) — an undocumented probe is effectively invisible. | project-standards | 0.72 | fix-here |
| 6 | `shot-boolean.mjs:98` | `measure()` couples to the widget template (`.svg-widget svg`, `app-svg-boolean-` prefix) with no WHY comment; a template rename degrades it to silent zero-geometry rather than an error. | maintainability | 0.62 | fix-here (comment) |
| 7 | `shot-boolean.mjs:125` | Magic threshold `8` in the SLIVER flag is unnamed while its neighbour `MIN_TAP=44` is a named/commented const. | project-standards | 0.60 | fix-here (named const) |
| 8 | `shot-boolean.mjs:76` | **Light-theme axis adds no geometry coverage for #318.** The layout is a pure function of tile/label/font — theme has no geometric input — so the 3 light-theme rows confirm color only, not an independent geometric result. Don't over-read them as theme-independent confirmation. | adversarial | 0.74 | fix-here (tighten comment) |
| 9 | `shot-boolean.mjs:123` | `flag` reads `svgs[0].hAttr` and `Number(getAttribute('height'))` → a non-numeric height becomes `NaN`, and `NaN<8`/`NaN<44` are both false → falsely 'ok'. (The "svgs[0] hides a later collapsed sibling" half is **refuted** — all controls share one `ctrlDimensions`, so svgs[0] is representative.) The `NO-WIDGET` branch is also dead (the selector wait throws first). | testing | 0.70 | fix-here (`Number.isFinite` guard) |

- **#4** routed **file-issue** — a shared-default fix affecting `run.mjs`/`screenshot.mjs` behavior, needs a harness re-run to verify; out of this PR's focused scope. Also folds in the init-script-injector triplication (maintainability, 0.66) as a small consolidation. Filing before this PR is considered mergeable.

### Coverage
- **Suppressed:** none (all findings ≥ 0.60).
- **Matrix coverage is sound** (testing): all three control types, narrow + x-narrow tiles, one-long-among-short repros at 3 and 4 controls, and a clean A/B pair (`c3-short-narrow` vs `c3-mixed-repro-narrow`, identical but for the SHORT→LONG switch label) that isolates the shared-height drop.
- **Residual risk:** the harness measures against `hAttr` but never asserts it matches the laid-out `getBoundingClientRect` — the cold-font stale-measure (finding 3) is invisible in the table until the `fonts.ready` gate lands.

Findings 1, 2, 3(fonts.ready), 5, 6, 7, 8, 9 fixed on-branch in a follow-up commit; **#4 filed as a follow-up issue** (blocks "ready to merge" until linked). No P0/P1. The core #318-geometry function is trustworthy today; the fixes harden the ON/OFF-state and cold-boot uses and make the tool fail loudly instead of lying.

*🤖 Review synthesized from 5 independent Opus persona reviewers.*

mairas:
## Review fixes applied + machine-verified — `cffa0bc1`

Addressed the findings on-branch and verified against real pixels/behavior (fresh prod build + headless Chromium on the target Pi):

| # | Fix | Verification |
|---|-----|--------------|
| 1 | **Boot-integrity assert** — exit non-zero on a control-count mismatch or unmeasurable height | Injected a bad-type control into a 3-control tile → run printed `c3-short-narrow[COUNT 2/3]` + `boot check failed` and **exited 1**. Clean matrix still **exits 0**. |
| 2 | **`suppressBootstrapNull:true`** on seeded paths so `value:true` renders ON | Screenshots now show the ON state: the `value:true` green switch is lit/knob-right, the `value:true` "Aux" light is a bright sphere beside its OFF siblings (states clearly distinct). Previously all rendered OFF. |
| 3 | **`await document.fonts.ready`** before measuring | Runs without hang; geometry stable. Full app-side font-race remains tracked in #321. |
| 9 | `NaN`/`≤0` height guard (`BAD-H`) | Non-numeric height no longer flags 'ok'. |
| 6,7,8,5 | measure() DOM-coupling comment; named `SLIVER_PX`; light-theme framed as color-only; probe documented in `perf-harness/README.md` | — |

The #318 mechanism reproduces cleanly: `c3-short-narrow` = **72px** vs `c3-mixed-repro-narrow` (same tile, one long label) = **44px** floor, long label ellipsized — the #320 fix holding, no SLIVER/sub-tap anywhere.

Finding **#4** (stale `configVersion:12` default affecting other probes) filed as **#323** — pre-existing, out of this PR's focused scope.

**Verdict: ready to merge** (pending merge authorization). No P0/P1; #323 linked.

PR #325, opened 2026-07-16
fix(boolean-switch): re-measure control height when fonts settle (#321)

### LINKED ISSUE #321: Boolean control label sizing uses a font-race-prone private canvas measure (cold-boot mis-size)

mairas:
## Why

The boolean-switch widget derives one shared control height from **canvas-measured label widths**. It measured against a *private* canvas that never re-ran on web-font load, and nothing re-measures on a static dashboard (only `onResized` and the config effect call `updateCtrlDimensions`). So on a cold boot — before Roboto settles — the panel could be sized against fallback-font metrics and keep the **wrong height for the whole session** until a manual resize. (Split out of #318; the long-label-shrink half was fixed in #320.)

Closes #321.

## What

- **`CanvasService.whenFontsReady()`** — exposes the service's existing `fontsReadyPromise` so callers that lay out against measured text can re-measure once fonts settle.
- **`CanvasService.measureTextWidth(text, font)`** — a shared offscreen-context text measure. The widget now routes through it instead of owning a private canvas (the issue's "reuse CanvasService's measurement" suggestion; also removes the duplicate context).
- **Widget**: re-measures on `whenFontsReady()`; `ctrlDimensions` becomes a **signal** so the post-font re-measure actually propagates to the controls under zoneless change detection (a plain-field mutation wouldn't).

**Scope note:** the issue also mentioned Roboto weight 700 not being registered. Deliberately left out — the *measurement* (`700px`) and the *render* (`font-weight:700`) both use weight 700, so they fall back consistently; it's a cosmetic choice, **not** the mis-size cause, and orthogonal to the race fix.

## Verify

- New `widget-boolean-switch.component.spec.ts` + `canvas.service.spec.ts` additions: **37 tests pass** — the re-measure fires on fonts-ready *and the new height reaches the signal*, the fonts-before-size ordering is safe, and `measureTextWidth`/`whenFontsReady` are covered directly. Lint clean; betterer flat at 179.
- **Machine render** (fresh prod build, headless Chromium on the target Pi): geometry is **identical** to pre-change (`c1`=90, `c3-short`=72, `c3-mixed`=44 floor), boot-assert clean — no rendering regression.

## Review

Reviewed by 4 independent personas (correctness, maintainability, testing, reliability). Correctness/maintainability found no defects (traced sound in every mount/font ordering; propagation correct). Addressed testing + reliability findings on-branch: added a direct `CanvasService.measureTextWidth`/`whenFontsReady` test (P2 — the fix's core primitive was only exercised through a mock), a `.catch` on the re-measure to match every other `fontsReadyPromise` consumer (P3), a stronger assertion that the re-measured height reaches `ctrlDimensions` (P3), and the fonts-before-size ordering test (P3).

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -193,9 +193,9 @@ exports[`strictNullChecks`] = {
       [1358, 21, 11, "\'b.longitude\' is possibly \'undefined\'.", "3370058570"],
       [1358, 35, 11, "\'a.longitude\' is possibly \'undefined\'.", "2985066057"]
     ],
-    "src/app/widgets/widget-boolean-switch/widget-boolean-switch.component.ts:1261729337": [
-      [77, 38, 9, "Argument of type \'string | undefined\' is not assignable to parameter of type \'string\'.\\n  Type \'undefined\' is not assignable to type \'string\'.", "4119212820"],
-      [100, 10, 12, "Object is possibly \'null\'.", "4261782498"]
+    "src/app/widgets/widget-boolean-switch/widget-boolean-switch.component.ts:802906206": [
+      [74, 38, 9, "Argument of type \'string | undefined\' is not assignable to parameter of type \'string\'.\\n  Type \'undefined\' is not assignable to type \'string\'.", "4119212820"],
+      [97, 10, 12, "Object is possibly \'null\'.", "4261782498"]
     ],
     "src/app/widgets/widget-datetime/widget-datetime.component.ts:3033260432": [
       [60, 40, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'string\'.", "2620553983"],
--- a/src/app/core/services/canvas.service.spec.ts
+++ b/src/app/core/services/canvas.service.spec.ts
@@ -95,3 +95,30 @@ describe('CanvasService resize handling (freeze-audit)', () => {
     vi.restoreAllMocks();
   });
 });
+
+describe('CanvasService text measurement (#321 font-race)', () => {
+  let service: CanvasService;
+  beforeEach(() => {
+    TestBed.configureTestingModule({});
+    service = TestBed.inject(CanvasService);
+  });
+
+  it('measureTextWidth measures against the shorthand font and scales with text length', () => {
+    // jsdom canvas shim (src/test.ts) reports measureText width = text.length * 8
+    expect(service.measureTextWidth('abc', '700 14px Roboto')).toBe(24);
+    expect(service.measureTextWidth('abcdef', '700 14px Roboto')).toBe(48);
+  });
+
+  it('measureTextWidth lazily creates and reuses a single offscreen context', () => {
+    const createSpy = vi.spyOn(document, 'createElement');
+    service.measureTextWidth('a', '14px Roboto');
+    service.measureTextWidth('bb', '14px Roboto');
+    const canvasCreations = createSpy.mock.calls.filter(([tag]) => tag === 'canvas').length;
+    expect(canvasCreations).toBe(1); // created once on first use, reused thereafter
+    createSpy.mockRestore();
+  });
+
+  it('whenFontsReady resolves', async () => {
+    await expect(service.whenFontsReady()).resolves.toBeUndefined();
+  });
+});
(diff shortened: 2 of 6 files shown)


mairas:
## Code review — this PR (#321 boolean font-race)

**Scope:** re-measure the boolean-switch control height when web fonts settle. Touches `CanvasService` (shared, +2 public methods) and the widget (signal + re-measure).
**Reviewers (4):** correctness, maintainability, testing, reliability. (Adversarial skipped — <50-line diff; but the shared-service blast radius drove the reliability + a direct-render check instead.)

Correctness and maintainability found **no defects** — the fix was traced sound across every mount/font ordering (no path leaves the height stuck at 0; the removed private-canvas fallback is byte-identical to the old even-split when width is 0; signal `.set` with a fresh object propagates to the OnPush children), and `measureTextWidth` doesn't overlap the existing `calculateOptimalFontSize`.

### Findings — all addressed on-branch

| 1 | `canvas.service.ts` | **`measureTextWidth` — the fix's core primitive — had no direct test** (the widget spec mocks CanvasService out). | testing | 0.85 | **fixed** (added `canvas.service.spec` cases: width scales with length, single-context reuse, `whenFontsReady` resolves) |
| 2 | `widget-boolean-switch.component.ts:119` | The re-measure `.then` omitted the `.catch` that **every other** `fontsReadyPromise` consumer in the file has. | reliability | 0.68 | **fixed** (no-op `.catch`) |
| 3 | `…component.spec.ts` | Test 2 proved the measurer re-fired but **not** that the new height reached `ctrlDimensions` — a re-measure that dropped the result would pass. | testing | 0.66 | **fixed** (vary mock width across the fonts boundary, assert height changes) |
| 4 | `…component.spec.ts` | No coverage for **fonts-resolve-before-host-size** (the early-return guard). | testing | 0.66 | **fixed** (added the ordering test) |

### Residual (noted, not changed — with reasoning)
- **Post-destroy re-measure**: a widget destroyed before fonts settle still runs `updateCtrlDimensions` when the (once-only, early-boot) promise resolves. All reviewers agreed it's harmless (signal `.set` post-destroy is a no-op) and bounded (the closure releases when the singleton promise resolves, seconds after boot). A `DestroyRef` guard would be tidier but is **over-engineering for a once-resolved promise** — declined per the reviewers' own severity calls; the `.catch` covers the unhandled-rejection angle.
- **`measureTextWidth` returns 0 for both no-context and a genuinely empty label** — documented in the JSDoc; the sole caller treats 0 as "fits", benign for both.
- **`ctrlDimensions` is an imperatively-`set` signal** here vs the sibling `widget-zones-state-panel`'s pure `computed` — defensible (canvas measurement is a side effect + a one-shot async trigger, awkward in a `computed`), and this PR moved toward signals, not away.

### Coverage
- Suppressed: none. Betterer flat at 179; lint clean; **37 tests** pass; machine render confirms **no geometry regression**.

No P0/P1. All four findings fixed on-branch; residuals are documented and low.

*🤖 Review synthesized from 4 independent Opus persona reviewers.*

PR #326, opened 2026-07-16
fix(perf-harness): seed configVersion 13 to skip the migration reload (#323)

### LINKED ISSUE #323: perf-harness: appConfig() default configVersion 12 is stale vs LATEST_APP_CONFIG_VERSION 13 (triggers migration reload in other probes)
Surfaced by the #322 review (maintainability 0.82, project-standards 0.70).

`perf-harness/lib/kip-config.mjs` seeds `configVersion: 12` in `appConfig()`'s default (line ~34) and its header comment (line ~11) calls 12 "LATEST_APP_CONFIG_VERSION". The actual source of truth `src/app/core/constants/config-versions.const.ts` sets `LATEST_APP_CONFIG_VERSION = 13`.

**Impact:** every probe that seeds the default `appConfig()` (`run.mjs`, `screenshot.mjs`, `verify-click.mjs`) boots into `ConfigurationUpgradeService`'s v12→v13 migration (toast + reload) *inside its measurement window* — polluting freeze/render numbers. `shot-boolean.mjs` already works around this locally via an `appV13()` override that pins 13 and deletes `dataSets`, which is why #322 is unaffected — but the workaround shouldn't have to exist per-probe.

**Proposed fix:**
- Set `appConfig()`'s default `configVersion` to `13` and correct the header comment.
- Drop `shot-boolean.mjs`'s `appV13()` override once the default is correct (keep the `dataSets` strip only if still needed for the v13 schema).
- Re-run `run.mjs` and `screenshot.mjs` to confirm they still boot clean (boot-assert passes) and no longer take the migration reload — this is why it's a gated change, not a blind default bump.
- While here: the `window.__KIP_TEST__` + `localStorage.setItem` init-script injector is triplicated verbatim across `run.mjs`, `screenshot.mjs`, and `shot-boolean.mjs` (maintainability 0.66) — consider a shared helper in `kip-config.mjs` alongside `localStorageBundle`.

Deliberately kept out of #322 to keep that PR a focused boolean-render-matrix addition (it correctly works around the stale default). Pre-existing; low severity but affects the trustworthiness of the freeze numbers other probes produce.

mairas:
## Why

The perf-harness `appConfig()` default seeded `configVersion: 12`, but the app's `LATEST_APP_CONFIG_VERSION` is **13**. Every probe using the default (`run.mjs`, `screenshot.mjs`) therefore booted into `ConfigurationUpgradeService`'s v12→v13 migration (toast + reload) *inside its measurement window*, polluting the freeze/render numbers. `shot-boolean.mjs` worked around it locally with an `appV13()` override.

Closes #323.

## What

- **`appConfig()`** now seeds `configVersion: 13` and drops `dataSets: []` — a genuine post-v12→v13 config has no `dataSets` (the migration strips it), so the app boots without running the migration. Header comment corrected 12→13 (the other two version numbers already matched the const file).
- **`shot-boolean.mjs`**: removed the now-redundant `appV13()` workaround; uses `appConfig()` directly.
- **DRY** (2nd commit): the triplicated `window.__KIP_TEST__` + `localStorage.setItem` init-script is extracted to a shared `initScriptContent(bundle)` helper, used by `run.mjs`, `screenshot.mjs`, and `shot-boolean.mjs`.

perf-harness-only — no app-code changes, outside app CI/lint.

## Verify

All three probes boot-assert clean at v13 (a mis-seeded config would fail the boot check), run twice — after the primary fix and again after the DRY refactor:
- `shot-boolean.mjs` → exit 0, 20 tiles `ok`.
- `screenshot.mjs` → exit 0, AIS radar rendered.
- `run.mjs --scenarios resize-storm` → exit 0, **boot-asserts 28/28** (this is the key regression check — it seeds the default `appConfig()` via `serverConfigDocument`).

## Review

Correctness-reviewed against app source: confirmed the app skips migration at `configVersion === 13`, the v12→v13 transform strips exactly `dataSets`/`datasetUUID`/`chartEngine`, `IAppConfig`/`DefaultAppConfig` omit `dataSets`, and **no app code reads `app.dataSets`** — so removing it can't misbehave at boot. `initScriptContent` is byte-equivalent to the three inline injectors it replaces (the only difference, run.mjs's `= true` spacing, is inert — readers test truthiness only). No findings.

### DIFF
--- a/perf-harness/lib/kip-config.mjs
+++ b/perf-harness/lib/kip-config.mjs
@@ -8,7 +8,7 @@
  *
  * Three distinct version spaces (all from src/app/core/constants/config-versions.const.ts):
  *  - applicationData URL path segment 11 (REMOTE_CONFIG_FILE_VERSION)
- *  - app.configVersion 12 (LATEST_APP_CONFIG_VERSION)
+ *  - app.configVersion 13 (LATEST_APP_CONFIG_VERSION)
  *  - connectionConfig.configVersion 13 (CONNECTION_CONFIG_VERSION)
  */
 export const SELF_URN = 'vessels.urn:mrn:signalk:uuid:11111111-1111-4111-8111-111111111111';
@@ -30,9 +30,12 @@ const DEFAULT_NOTIF = {
 };
 
 export function appConfig(extra = {}) {
+  // configVersion 13 with no dataSets field: a genuine post-v12->v13 config, so the
+  // boot skips ConfigurationUpgradeService's migration toast/reload (which is what
+  // strips dataSets/datasetUUID/chartEngine) inside the measurement window.
   return {
-    configVersion: 12, autoNightMode: false, redNightMode: false, nightModeBrightness: 0.27,
-    widgetHistoryDisabled: false, dataSets: [], unitDefaults: DEFAULT_UNITS,
+    configVersion: 13, autoNightMode: false, redNightMode: false, nightModeBrightness: 0.27,
+    widgetHistoryDisabled: false, unitDefaults: DEFAULT_UNITS,
     notificationConfig: DEFAULT_NOTIF, browserTabTitle: 'Skip', ...extra,
   };
 }
--- a/perf-harness/lib/kip-config.mjs
+++ b/perf-harness/lib/kip-config.mjs
@@ -160,6 +160,16 @@ export function localStorageBundle({ origin, subscribeAll }) {
   return { 'skip.connectionConfig': JSON.stringify(cc) };
 }
 
+/**
+ * Playwright addInitScript body: sets the boot-time test flag, then seeds the
+ * localStorageBundle() keys before the app script runs. Pass the object from
+ * localStorageBundle().
+ */
+export function initScriptContent(bundle) {
+  return `window.__KIP_TEST__=true;` +
+    Object.entries(bundle).map(([k, v]) => `localStorage.setItem(${JSON.stringify(k)}, ${JSON.stringify(v)});`).join('');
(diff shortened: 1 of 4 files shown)


mairas:
## Code review — this PR (#323 perf-harness configVersion)

**Scope:** perf-harness-only — bump `appConfig()` default to v13, drop `dataSets`, remove `shot-boolean`'s `appV13()` workaround, share the init-script injector. Outside app CI/lint.
**Reviewer:** correctness (focused) — proportionate to a small, mechanical, already-machine-verified change (the author ran all three probes; they boot-assert clean at v13). The two load-bearing claims were checked **against app source**, not the diff's own comments.

### Findings: none

**Config-shape fidelity (verified):** `SettingsService.checkConfigUpgradeRequired` flags upgrade only when `storageVersion !== 13`, and the boot effect runs the migration only for live v11/v12 — so at `configVersion: 13` migration is never entered. The v12→v13 transform strips exactly `dataSets` (+ `datasetUUID`/`chartEngine`); `IAppConfig` and the shipped `DefaultAppConfig` both omit `dataSets`; a whole-tree grep finds **no app-code reader of `app.dataSets`** — removing it can't misbehave at boot.

**`initScriptContent` equivalence (verified):** byte-equivalent to the three inline injectors; the only difference (run.mjs's `__KIP_TEST__ = true` spacing vs no-space) is inert — the four readers test truthiness only — and the double-`JSON.stringify` serialization is preserved exactly.

No findings. Empirically verified (3 probes boot-assert clean at v13) and source-verified.

Issue #197, opened 2026-07-04
Add a fresh toolbar-layout screenshot to the welcome help doc


mairas:
PR-3 (#187) rewrote `welcome.md`'s "General Layout" section to "The Toolbar" and **removed** the annotated `img/general-layout.png` screenshot, which depicted the retired sidenav chrome (Actions/Notification sidenavs, numbered permanent nav buttons) — misleading under the new auto-hiding-toolbar model.

The section now describes the toolbar in text. A fresh annotated screenshot would restore the visual reference:

- Capture the revealed auto-hiding toolbar showing Settings, fullscreen, night-mode, the page navigator, notifications, and the edit button.
- Ideally also show the persistent alarm badge (lower-left) with an active alarm.
- Re-add it to `welcome.md` under "The Toolbar" with matching numbered callouts.

Requires a running instance with the merged navigation chrome (feat/navigation-chrome shipped in PR-1 #193). Tracked under the navigation UX redesign epic #183.

mairas:
See this screenshot for a Skip screen with an active toolbar. Alarms are not present.

<img width="889" height="1036" alt="Image" src="https://github.com/user-attachments/assets/121442c1-bf9c-480f-8f94-08b03f86ba88" />

mairas:
An updated screenshot with the correct actions icon.

<img width="890" height="1037" alt="Image" src="https://github.com/user-attachments/assets/96496bf2-de25-4139-8772-23bf3777467d" />

PR #327, opened 2026-07-17
test(perf-harness): render probes to machine-verify #216 + #245 off the boat

mairas:
## Why

The #257 campaign's Stage 1 is code-complete; the only remaining Stage-1 work is a single on-boat verification session, which **gates all of Stage 2** (Gate B). Two of that session's items are pure *rendering* checks that don't need the boat — this session already proved the boat-verify rendering is machine-runnable here (the #318/#320 boolean work). These probes move #216 and #245 off the boat so session 1 shrinks to what physically needs it (K1 approval, live-RPM/hover feel, the #197 capture).

## What

Two reusable perf-harness probes (perf-harness-only; outside app CI/lint):

- **`shot-embed.mjs` (#216)** — boots Skip with/without the `?embed` pre-hash flag and asserts the chromeless contract: `<app-toolbar>` present normally, **unmounted under embed**, all seeded widgets still render in both, and an unknown `?profile` falls back to the default dashboard (the console output shows the app's fallback warning firing). Exact-count boot-assert; exits non-zero on any failure.
- **`shot-units.mjs` (#245)** — renders a numeric widget per `convertUnitTo` across a unit spread and screenshots the grid; each tile's label resolves to the display symbol (kn, km/h, °C, gal, gal/min, Ω, …), never a spelled-out word.
- `numericWidget` gains a `displayName` param (default `'N'`, existing callers unchanged) so each units tile is labelled by its measure key.

## Verify (run on the target Pi against a current build)

- `shot-embed.mjs` → **all 5 assertions pass**; embed screenshot is cleanly chromeless (no toolbar), content rendered.
- `shot-units.mjs` → **12/12 tiles**, screenshot visually confirms every symbol: `kph→km/h`, `celsius→°C`, `gallon→gal`, `g/min→gal/min`, `ohm→Ω`, plus the measure-key fallbacks (mph, K, V, psi) — no descriptions rendered.

## Coverage limits (honest)

- **#245 symbols are a human eyeball** of the screenshot — the label is canvas-drawn (not DOM), so the machine step asserts only tile count. The symbol-*resolution* logic is already unit-tested in `units.service.spec.ts`; the screenshot backstops the render integration.
- **#216**: verifies the *Skip-side* embed behavior, not rendering inside the real Freeboard-SK drawer (external host). Covers the profile **fallback** path only, not positive profile-switching (needs a second seeded slot). Read-only is proven via toolbar-absence + content-presence, not per-gesture.

Relates to #216, #245, #257 (shrinks the Stage-1-close boat session).

### DIFF
--- a/perf-harness/README.md
+++ b/perf-harness/README.md
@@ -92,6 +92,24 @@ that each tile rendered its seeded control count (exits non-zero otherwise). Inv
 with `CHROME_BIN=/usr/bin/chromium node shot-boolean.mjs --public ../public` after a
 `../public` build.
 
+`shot-embed.mjs` boots Skip with and without the `?embed` pre-hash query flag and
+asserts the #216 chromeless contract: the toolbar (`<app-toolbar>`) is present
+normally but unmounted under embed, while all seeded dashboard widgets still render;
+an unknown `?profile` falls back to the default dashboard. Exits non-zero on any
+failure; writes `results/shots/embed/<label>.png`. It verifies the Skip-side embed
+behavior, not rendering inside the actual Freeboard-SK drawer, and covers only the
+profile-fallback (not positive profile-switching, which needs a second seeded slot).
+
+`shot-units.mjs` renders a numeric widget per `convertUnitTo` across a unit spread
+and screenshots the grid to `results/shots/units/units-grid.png` — the #245 display-
+symbol probe (kn, km/h, °C, gal, gal/min, Ω, …). The unit label is canvas-drawn, so
+the machine step boot-asserts only the tile count; **symbol correctness is a human
+eyeball of the screenshot** (each tile is labelled by its measure key). The symbol
+resolution itself is unit-tested in `units.service.spec.ts`.
+
+Both invoke as `CHROME_BIN=/usr/bin/chromium node <probe> --public ../public` after a
+`../public` build.
+
 ## Interpreting the numbers
 
 | Question | Metric | Caveat |
--- a/perf-harness/lib/kip-config.mjs
+++ b/perf-harness/lib/kip-config.mjs
@@ -59,12 +59,12 @@ function node(w, h, x, y, widgetProperties) {
   return { x, y, w, h, id, selector: 'widget-host2', input: { widgetProperties } };
 }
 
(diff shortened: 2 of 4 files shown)


mairas:
## Code review — this PR (perf-harness render probes for #216 + #245)

**Scope:** two perf-harness probes + a `numericWidget` param. perf-harness-only (outside app CI/lint).
**Reviewers (2):** correctness, testing — proportionate for verified diagnostic tooling; the key risk (the #322 lesson: a harness that greenlights a plausible-but-wrong render) was the focus.

**Correctness — no findings.** Traced every assertion against the app contracts: the `?embed=1#/page/0` URL lands the flag in `window.location.search` (pre-hash) as `EmbedModeService` reads it; `app-toolbar` absence is a clean proxy for the `@if (dashboardVisible() && !embed())` gate (auto-hide is internal CSS, not an unmount); the unknown-profile assertion exercises the real not-present fallback branch; and because no value streams, `getUnitDisplaySymbol` resolves independent of `convertToUnit`, so an incompatible unit-on-speed-path **cannot** error and mask a symbol regression. No false-passes.

### Findings — all fixed on-branch

| 1 | `shot-embed.mjs` | Asserted `widgets >= 1` where the seeded count is known (2) — a dropped widget would pass, weaker than the sibling `shot-boolean`/`shot-units` exact-count convention. | testing | 0.82 | **fixed** (assert `=== WIDGETS.length` on all rows) |
| 2 | `shot-embed.mjs` | Collected console warnings/errors but never used them (dead monitoring — appearance of a health check without one). | testing | 0.76 | **fixed** (print per-probe; now surfaces the profile-fallback warning as positive evidence) |
| 3 | `README.md` | New probes not documented, breaking the sibling-probe convention. | testing | 0.85 | **fixed** (added entries stating the exit contract + coverage limits) |

### Coverage limits — stated in the PR body (per both reviewers)
- #245 symbol correctness is a **human eyeball** (canvas-drawn label; resolution logic is unit-tested separately).
- #216 covers the Skip-side embed behavior (not the real Freeboard drawer) and the profile **fallback** only (not positive profile-switching); read-only via toolbar-absence.

No P0/P1/P2. All three P3s fixed on-branch; the tool now fails on a partial render and surfaces boot console output. Both probes verified green on the target Pi.

*🤖 Review synthesized from 2 independent Opus persona reviewers.*

PR #328, opened 2026-07-17
fix(types): resolve strictNullChecks in core/utils (#6, S2-1 cluster 1)

mairas:
## Why

Opens **S2-1 (#6 strictNullChecks endgame)** — the first directory-cluster sub-PR of the migration that drives the betterer baseline to zero and flips `strictNullChecks` in the app tsconfig. Cluster: `src/app/core/utils` (2 pure helper files, 5 issues). Establishes the per-cluster flow (fix → `snc` clean → betterer regen).

## What

- **`dataScales.util.ts`** `calcNiceNumber`: drop the dead `= null` initializer on `niceFraction`. Both `round` and `!round` paths are complete if/elif/elif/else chains that assign a number before the `return`, so TS definite-assignment covers it — no cast, no suppression, no behavior change.
- **`zones-highlight.utils.ts`** `getHighlights`: type `lower`/`upper` as `number | null` (what `unitsService.convertToUnit` actually returns) and add explicit `== null` checks to the existing skip-guard so TS narrows to `number` for the rest of the loop. **Behavior-preserving:** `Number.isFinite(null)` is already `false`, so a null conversion already hit `continue` — the `== null` clause only adds the type narrowing, not a runtime change.

## Verify

- `npm run snc`: both files clean; total **179 → 174** (−5), no new errors introduced elsewhere.
- `.betterer.results` regenerated (both file blocks removed). `npm run lint` clean. No specs exist for these pure helpers or their gauge consumers; CI runs the full vitest suite per the S2-1 convention.

First of the S2-1 cluster series (#6).

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -93,15 +93,6 @@ exports[`strictNullChecks`] = {
       [54, 4, 14, "Type \'null\' is not assignable to type \'CanvasRenderingContext2D\'.", "912271434"],
       [55, 4, 18, "Type \'null\' is not assignable to type \'HTMLCanvasElement\'.", "1193423419"]
     ],
-    "src/app/core/utils/dataScales.util.ts:2208923770": [
-      [54, 6, 12, "Type \'null\' is not assignable to type \'number\'.", "2083421902"]
-    ],
-    "src/app/core/utils/zones-highlight.utils.ts:1634660371": [
-      [42, 8, 5, "Type \'null\' is not assignable to type \'number\'.", "173263814"],
-      [43, 8, 5, "Type \'null\' is not assignable to type \'number\'.", "183473191"],
-      [72, 6, 5, "Type \'number | null\' is not assignable to type \'number\'.\\n  Type \'null\' is not assignable to type \'number\'.", "173263814"],
-      [78, 6, 5, "Type \'number | null\' is not assignable to type \'number\'.\\n  Type \'null\' is not assignable to type \'number\'.", "183473191"]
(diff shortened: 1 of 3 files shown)


mairas:
## Code review — this PR (#6 S2-1 core/utils cluster)

**Reviewer:** correctness (focused — proportionate to a 3-line pure-type fix; the risk is a masked null / behavior change, which is exactly what a strictNullChecks fix can hide).

**No findings.** Verified both fixes are behavior-preserving across every input class:
- **`dataScales.util.ts`** — the `= null` initializer was truly dead: both `round`/`!round` branches are complete if/elif/elif/**else** chains that assign before `return`; even the degenerate `range <= 0` (`fraction = NaN`) falls through to the final `else` and assigns `10`. Definite assignment holds on all paths; the old `null` was never observable.
- **`zones-highlight.utils.ts`** — the added `lower == null || upper == null` disjuncts are pure redundancy: `Number.isFinite(null)` is already `false`, so null/NaN/±Infinity already hit `continue` under the original guard. Identical continue/no-continue decision for null, NaN, ±Infinity, and finite inputs. After the guard, `== null` narrows `lower`/`upper` back to `number` via control-flow analysis, so the arithmetic typechecks with **no `!`, no cast, no suppression**. `convertToUnit` genuinely returns `number | null`, justifying the widened local type; no downstream consumer affected.

Zero behavior change, sound narrowing, betterer regenerated (179→174). First S2-1 cluster of the series.

Issue #329, opened 2026-07-17
Dataset-chart config: ngOnInit crashes on a configured path with no current data (unguarded null from getPathObject)


mairas:
Surfaced by the S2-1 strictNullChecks pass (#6), which flagged it rather than papering over the null with a cast.

`DatasetChartOptionsComponent.ngOnInit` (`dataset-chart-options.component.ts` ~L81-84):
```ts
const currentPath = this.datachartPath()?.value;
if (currentPath) {
  const pathObject = this.data.getPathObject(currentPath);
  this.setPathSources(pathObject);            // <- pathObject may be null
  this.unitList.set(this.units.getConversionsForPath(currentPath));
}
```

`DataService.getPathObject(path): ISkPathData | null` returns `null` when the path is not (yet) in `_skData` (`cloneDeep(this._skData.get(path)) || null`). `setPathSources` takes a non-null `ISkPathData` and dereferences it — so passing `null` throws.

**The sibling `changePath()` already guards this** (`if (pathObject === null) { … return; }`); `ngOnInit` does not.

**Reachable:** a saved dataset-chart widget config references a configured path whose data hasn't arrived in `_skData` yet at boot (or whose source went offline). Opening that widget's config then crashes `ngOnInit`.

**Fix:** guard `pathObject === null` in `ngOnInit` the same way `changePath()` does — skip `setPathSources`/`unitList` when the path can't be resolved. This is a real behavior change (crash → graceful skip), which is why the S2-1 pass deferred it out of the pure-type cluster (PR for `widget-config`). Relates to #6.

PR #330, opened 2026-07-17
fix(types): resolve strictNullChecks in widget-config (#6, S2-1 cluster 2)

mairas:
## Why

S2-1 (#6 strictNullChecks endgame) — the `src/app/widget-config` cluster. A **quality-over-coverage** pass: fully fix the files whose errors resolve cleanly and behavior-preservingly; **leave the delicate ones untouched and flag them** rather than force a cast that could hide a bug.

## Fully fixed (4 files, baseline 179 → 167)

- **`display-chart-options` + `root-modal-widget-config`**: `colors = []` (inferred `never[]`) → `colors: { label: string; value: string }[] = []`, matching `AppService.configurableThemeColors`. Pure type annotation.
- **`electrical-family-setup` + `solar-charger-setup`**: `hasGroups` → `(this.groupsFormArray?.length ?? 0) > 0` (behaviorally identical — `undefined > 0` and `0 > 0` are both `false`); `addGroup`/`removeGroup` narrow via a local-const guard. The guard is unreachable in practice — both are only invoked from controls inside `@if (supportsGroups())` (= `!!groupsFormArray`) — so behavior is preserved; the guard exists only for the compiler to narrow (a `!`/cast is forbidden by the migration convention).

## Deferred / flagged (left untouched — the important part)

- **`dataset-chart-options`** — flagging surfaced a **real latent bug**: `ngOnInit` passes a possibly-null `getPathObject()` into `setPathSources()` (which derefs it), while the sibling `changePath()` guards `=== null`. A saved chart path with no current data crashes on open. Filed as **#329** (the fix is a behavior change, so it belongs outside this pure-type cluster).
- **`boolean-control-config`, `display-datetime`** — clean fix needs converting `input<T>(undefined)` to `input.required<T>()`, a public input-contract change; `display-datetime` is also #25 typed-forms territory.
- **`boolean-multicontrol-options`** — `convertUnitTo: null` fails because `IWidgetPath.convertUnitTo?: string` omits the `null` its own JSDoc documents as the sentinel; the honest fix widens the shared core interface (#25-adjacent, could ripple).
- **`paths-options`** — excluded (overlaps #25, already partly retyped by H7).

These remain in #6's backlog for careful individual treatment.

## Verify

`npm run snc`: the 4 fixed files clean, total **179 → 167** (−12), no new errors. `.betterer.results` regenerated. `npm run lint` clean. The 4 modified components' specs: **4 files / 28 tests pass**. CI runs the full suite.

**Merge note:** touches `.betterer.results`, so it serializes with #328 — merge #328 first, then this rebases + regenerates onto the new baseline. #6, S2-1 cluster 2.

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -106,20 +106,10 @@ exports[`strictNullChecks`] = {
     "src/app/widget-config/dataset-chart-options/dataset-chart-options.component.ts:155376543": [
       [83, 26, 10, "Argument of type \'ISkPathData | null\' is not assignable to parameter of type \'ISkPathData\'.\\n  Type \'null\' is not assignable to type \'ISkPathData\'.", "384538813"]
     ],
-    "src/app/widget-config/display-chart-options/display-chart-options.component.ts:2641875525": [
-      [48, 4, 11, "Type \'{ label: string; value: string; }[]\' is not assignable to type \'never[]\'.\\n  Type \'{ label: string; value: string; }\' is not assignable to type \'never\'.", "3027418051"]
-    ],
     "src/app/widget-config/display-datetime/display-datetime.component.ts:2719976956": [
       [45, 51, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'FormControl<string>\'.", "2620553983"],
       [46, 53, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'FormControl<string>\'.", "2620553983"],
       [49, 10, 22, "Type \'null\' is not assignable to type \'Subscription\'.", "29842089"]
-    ],
-    "src/app/widget-config/electrical-family-setup/electrical-family-setup.component.ts:4115844403": [
-      [60, 48, 28, "Object is possibly \'undefined\'.", "758680035"],
-      [114, 4, 20, "Object is possibly \'undefined\'.", "2522653710"],
-      [115, 4, 20, "Object is possibly \'undefined\'.", "2522653710"],
-      [119, 4, 20, "Object is possibly \'undefined\'.", "2522653710"],
-      [120, 4, 20, "Object is possibly \'undefined\'.", "2522653710"]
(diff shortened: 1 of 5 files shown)


mairas:
## Code review — this PR (#6 S2-1 widget-config cluster)

**Reviewer:** correctness (focused). **No findings.** Verified all four fixes are behavior-preserving against the app code:

- **`colors` annotation** (display-chart-options, root-modal-widget-config): `AppService.configurableThemeColors` is declared `{ label: string; value: string }[]`, exactly matching the annotation. Runtime value unchanged (`[]` at init, `configurableThemeColors` in ngOnInit). Pure type annotation.
- **`hasGroups`** (electrical-family-setup, solar-charger-setup): `groupsFormArray` is a plain optional field. Case table confirms `(?.length ?? 0) > 0` is identical to `?.length > 0` — undefined→false, 0→false, >0→true.
- **`addGroup`/`removeGroup` guards**: both buttons render only inside `@if (supportsGroups())` (= `!!groupsFormArray`); the methods are `protected` with no other caller; the field is never reset to undefined. So the `if (!groupsFormArray) return;` guard is unreachable in practice (no TOCTOU — it's a field, not a getter), and behavior is preserved. Required for the compiler to narrow (a `!`/cast is forbidden).
- No `!`, `as`, or suppression introduced.

**Residual (noted):** those guards are effectively dead code given the templates — if a *future* caller ever invoked these methods with `groupsFormArray` undefined, the new code silently no-ops where the old threw. Benign, and the correct trade for the narrowing.

**Deferred files** (flagged, not fixed) are documented in the PR body; the `dataset-chart-options` null-deref is filed as **#329**.

Zero behavior change across the 4 fixed files; baseline 179→167. S2-1 cluster 2 (#6). Merges after #328 (shared `.betterer.results`).

PR #331, opened 2026-07-17
fix(types): resolve strictNullChecks in core/components (#6, S2-1 cluster 3)

mairas:
## Why

S2-1 (#6 strictNullChecks endgame) — the `src/app/core/components` cluster (10 files; **`dashboard.component.ts` deliberately excluded** — its ~35 gridstack-null errors get separate careful handling). All 10 fully fixed, **179 → 150** (−29).

## What (all behavior-preserving; no `!`, no `as`, no suppressions)

- **`toSignal(BreakpointObserver.observe(...))` → `{ requireSync: true }`** (data-inspector, notifications, settings ×2): types the signal as non-`undefined`. **Honest caveat:** this is behavior-*adjacent*, not a pure type change — `requireSync` removes the (never-observable) initial `undefined` tick and would throw NG0601 if `observe()` ever failed to emit synchronously. Verified sound: CDK `observe()` emits synchronously (`combineLatest` of `startWith`-seeded queries via `concat(take(1),…)`) in browser, SSR, and the jsdom shim, so it never throws; and the old code already relied on that same sync guarantee (it would have `.matches`-derefed `undefined` otherwise). Net: identical behavior for every reachable input; a *shift in failure timing*, not new fragility.
- **`viewChild<T>()` → `viewChild.required<T>()`** (notifications ×3): the three refs are unconditional in the template and read only in post-view-init user handlers, so `.required` always resolves.
- **`colors: never[]` → typed annotation** (units accumulator); **`_chart`/`canvasCtx`/`canvasElement` → `| null`** with guards matching the pre-existing no-op paths (`CanvasService.drawTitle`/`unregisterCanvas` already early-return on null); **`compType | undefined`** (type-only, every read already guarded); **`drop()` `currentActive !== null`** guards (protective — non-null at drop time via the first `NavigationEnd`).

## Review

Reviewed by **correctness + adversarial** personas (this cluster earned the extra scrutiny for its construction-time-throw surface). **Both returned zero findings** after tracing each invariant — CDK's synchronous-emit contract, the unconditional viewChild refs, and the `activeDashboard` non-null-at-drop ordering. Residuals noted transparently (the `requireSync` CDK-contract dependency; the `drop()` guards being protective dead-code today).

## Verify

`npm run snc`: 10 files clean, **179 → 150**, no new errors. `.betterer.results` regenerated. `npm run lint` clean. 9 of 10 components have specs — **all 9 pass, 37 tests** (the specs `detectChanges()`, exercising the `requireSync` construction path). CI runs the full suite.

**Merge note:** touches `.betterer.results` — serializes with #328 and #330 (merge in sequence, rebase + regen each). #6, S2-1 cluster 3.

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -44,55 +44,6 @@ exports[`strictNullChecks`] = {
       [743, 8, 15, "\'_gridstack.grid\' is possibly \'undefined\'.", "3391422426"],
       [752, 4, 22, "Object is possibly \'undefined\'.", "2187394131"]
     ],
-    "src/app/core/components/dashboards-editor/dashboards-editor.component.ts:976168273": [
-      [80, 39, 40, "Argument of type \'string | undefined\' is not assignable to parameter of type \'string\'.\\n  Type \'undefined\' is not assignable to type \'string\'.", "1983996063"],
-      [94, 6, 4, "Type \'string | undefined\' is not assignable to type \'string\'.\\n  Type \'undefined\' is not assignable to type \'string\'.", "2087876002"],
-      [140, 17, 13, "\'currentActive\' is possibly \'null\'.", "194687392"],
-      [140, 56, 13, "\'currentActive\' is possibly \'null\'.", "194687392"],
-      [142, 44, 13, "\'currentActive\' is possibly \'null\'.", "194687392"],
-      [143, 17, 13, "\'currentActive\' is possibly \'null\'.", "194687392"],
-      [143, 56, 13, "\'currentActive\' is possibly \'null\'.", "194687392"],
-      [145, 44, 13, "\'currentActive\' is possibly \'null\'.", "194687392"]
-    ],
-    "src/app/core/components/data-inspector-row/data-inspector-row.component.ts:3525188709": [
-      [25, 34, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'string\'.", "2620553983"],
-      [34, 4, 10, "Type \'IConversionPathList\' is not assignable to type \'null\'.", "4013178488"],
-      [35, 24, 10, "Object is possibly \'null\'.", "4013178488"]
-    ],
-    "src/app/core/components/data-inspector/data-inspector.component.ts:2942185648": [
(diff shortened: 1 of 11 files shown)


mairas:
## Code review — this PR (#6 S2-1 core/components cluster)

**Reviewers (2):** correctness + **adversarial** — this cluster earned the extra scrutiny because it introduces construction-time-throw surfaces (`requireSync`, `viewChild.required`), not just type annotations. **Both returned zero findings** after independently tracing every behavior-adjacent change.

### Verified invariants (why "zero behavior change" holds)
- **`requireSync` ×4** — CDK `BreakpointObserver.observe()` emits **synchronously**: `combineLatest` of per-query observables each seeded with `startWith(mql)`, wrapped in `concat(…take(1), …)`; the `take(1)` branch emits on subscribe. Holds in browser, SSR (`noopMatchMedia` also `startWith`s), and the jsdom `matchMedia` shim — so NG0601 never fires. The removed `undefined` tick was never observable (the sync value lands during the constructor's `toSignal()` subscribe, before any effect/template read). **Adversarial's sharp nuance:** the *old* code also depended on that sync guarantee (it would have `.matches`-derefed `undefined` otherwise), so this is a shift in *failure timing/site*, not new fragility.
- **`viewChild.required` ×3** (notifications) — `#notificationsForm`/`#statePanel`/`#soundPanel` are unconditional in the template; read only in `saveAllSettings`/`togglePanel` (post-view-init user handlers). `.required` always resolves.
- **`drop()` `currentActive !== null`** (dashboards-editor) — `activeDashboard` is set to ≥0 by the first `NavigationEnd` before the editor is draggable, so it's non-null at drop time. The guard is actually *protective*: the old code did `set(null + 1) = set(1)` in the (unreachable) null case — a latent bug; the new code correctly leaves it unset.
- **signalk/widget-title/host2/embedded** — `getContext('2d')` guards match the pre-existing no-op paths; `dashboards?.[activeIdx]` ternary equals `dashboards?.[null] === undefined`; `compType | undefined` is type-only with every read already guarded.

### Residuals (transparent, no action)
- `requireSync` binds correctness to CDK's synchronous-emit contract (verified across all environments Skip runs in). Documented in the PR body.
- The `drop()` guards are protective dead-code today; noted so the `activeDashboard` ordering invariant is explicit.

Zero behavior change for all reachable inputs, verified by two independent lenses including adversarial. Baseline 179→150. S2-1 cluster 3 (#6). Serializes after #328 and #330.

Issue #333, opened 2026-07-17
S2-1 enabler: widen shared config interfaces (IWidgetPath/IWidgetSvcConfig/IConnectionConfig) where null is a meaningful sentinel


mairas:
Surfaced by the S2-1 strictNullChecks pass (#6). Several widget files can't be cleanly resolved because a **shared interface types a field as non-null while the DEFAULT_CONFIGs and the config UI use `null` as a meaningful, distinct sentinel value.** Widening these specific fields (consistent with how `IWidgetPath.path` is already `string | null`) unblocks all of the following at once, without per-file casts:

- **`IWidgetPath`** — `pathType`, `convertUnitTo`, `pathSkUnitsFilter` are set to `null` in DEFAULT_CONFIGs (e.g. widget-racer-line, widget-multi-state-switch) and the JSDoc documents `null` = "no filter", but the interface types them non-null.
- **`IWidgetSvcConfig.widgetUrl`** — `DEFAULT_CONFIG.widgetUrl: null` is **load-bearing**: the config UI (`root-modal-widget-config.component.html:65`) branches on `widgetUrl !== undefined` to decide whether to show the URL input, so `null` vs `undefined` is semantically distinct (widget-iframe).
- **`IConnectionConfig.signalKUrl`** — `config.blank.const.ts` sets `signalKUrl: null` as the documented sentinel (later overwritten with the host); asserted `toBeNull()` by `app-initNetwork.service.spec.ts`.
- **`ais-processing.service` types** — `Position.latitude/longitude` and `AisTarget.ais.status` are optional there, blocking `widget-ais-radar`'s distance/status code (a related but separate root: optional-vs-null in a shared service type).

**Proposed:** widen exactly these fields to include `null` (matching their real runtime domain and their own docs), then the blocked widget files (widget-iframe, widget-racer-line, widget-multi-state-switch, config.blank.const, and — via the ais-processing types — widget-ais-radar) resolve cleanly. This is a small, targeted interface change best done as its own S2-1 sub-PR before those widget clusters. Relates to #6.

mairas:
**Measured the blast radius** (widened the 5 fields locally against main @ baseline 82, ran `snc`). It's more entangled than a clean unblock — net 82→80, but:

**Fully resolves (2 files off the baseline):**
- `widget-iframe` (via `IWidgetSvcConfig.widgetUrl?: string | null`)
- `config.blank.const` (via `IConnectionConfig.signalKUrl: string | null`)

[...]

Issue #332, opened 2026-07-17
svg-zone-states: constructor effect dereferences nullable theme/data without a guard (crash on init before theme resolves)


mairas:
Surfaced by the S2-1 strictNullChecks pass (#6), which flagged it rather than forcing a type change that would mask it.

`SvgZoneStatesComponent` (`svg-zone-states.component.ts`) declares nullable-defaulted inputs:
```ts
readonly data = input<IDynamicControl>(null, { alias: "controlData" });   // default null
readonly theme = input<ITheme>(null);                                     // default null
```
Its constructor `effect()` then dereferences both with **no guard**:
```ts
effect(() => {
  const data = this.data();
  const theme = this.theme();
  untracked(() => {
    switch (data.notificationState) {          // data may be null
      case States.Emergency:
        this.ctrlStateColor.set(theme.zoneEmergency);   // theme may be null
    ...
      default:
        this.messageTxtColor.set(getColors(data.color, theme).dim);
    }
    this.ctrlLabelColor.set(getColors(data.color, theme).color);
```
The parent `widget-zones-state-panel.component.html:12` binds `[theme]="theme()"` with **no `@if` guard**, and `theme()` is nullable during app init.

**Inconsistent with the sibling SVG controls** (`svg-boolean-switch`/`-light`/`-button`), which all declare `theme = input<ITheme | null>(null)` and start their color logic with `if (!theme) return;`. `svg-zone-states` is missing that guard.

**Impact:** if the effect runs (or re-runs) while `data`/`theme` is null — reachable during init before the theme resolves — it throws a `TypeError` deref, breaking the widget.

**Fix:** guard `if (!data || !theme) return;` at the top of the `untracked` block (mirroring the sibling svg-boolean components), and type the inputs `| null`. This is a behavior change (throw → graceful skip until inputs settle), which is why the S2-1 pass deferred it out of the pure-type cluster. Relates to #6.

PR #334, opened 2026-07-17
fix(types): resolve strictNullChecks in widgets (#6, S2-1 cluster 4)

mairas:
## Why

S2-1 (#6 strictNullChecks endgame) — the `src/app/widgets` cluster. 14 files fully fixed, **133 → 82** (−51), flag-don't-force on the rest.

## Fully fixed (14 files) — all behavior-preserving, no `!`/casts/suppressions

- **Optional SVG inputs** (svg-racesteer, svg-windsteer): `input<number>()` default-`undefined` → `input<number | undefined>(undefined)` — the emitted JS is byte-identical (same `input(undefined)` call), only the type param changed; consumers already guard.
- **Canvas/color widgets** (widget-text, widget-datetime, widget-boolean-switch, widget-gauge-ng-compass, widget-position, widget-race-timer, widget-racer-timer, widget-zones-state-panel, widget-label): `canvasCtx: | null` (existing `!canvasCtx` guards narrow), color-signal types, and `getColors(cfg.color ?? 'contrast', …)` — the `??` never fires (DEFAULT_CONFIG-merged config always carries `color`), and `getColors(undefined)` already maps to `contrast` so it's equivalent even if it did.
- **Narrowing via local capture** (widget-boolean-switch `streams`, svg-windsteer signal reads), **`toHHMMSS(number | null)`**, **`getGMTOffset(string | undefined)`**, **dialog-ais-target** (`convertToUnit` null → `'--'`, matching the existing invalid-value output), **widget-freeboardsk** (`widgetUrl: string`, ctor-assigned; the `signalkUrl?.url` operand is dead).

## Flagged / deferred (untouched — the important part)

- **`svg-zone-states`** — flagging surfaced a **latent crash**: the constructor effect derefs nullable `theme`/`data` with no guard (unlike the sibling svg-boolean components that all `if (!theme) return`). Filed as **#332**.
- **5 files share one root cause → filed as #333**: shared interfaces (`IWidgetPath`, `IWidgetSvcConfig.widgetUrl`, `IConnectionConfig.signalKUrl`) type fields non-null while DEFAULT_CONFIGs + the config UI use `null` as a meaningful sentinel. Widening those exact fields unblocks widget-iframe, widget-racer-line, widget-multi-state-switch, config.blank, and (via ais-processing types) widget-ais-radar at once.
- **`widget-racesteer`** — `targetAngle` is deliberately `null` to hide the target via svg-racesteer's `input.required<number>` contract; a type-lie relied on downstream. Left as-is.

## Review

Reviewed by **correctness + adversarial**. Both converged on one real P3 — `widget-label` used `?? 'green'`/`?? 'grey'` (→ `theme.green`/`grey`) where the old `mapColor(undefined)` fell back to `theme.contrast`; unlike the 8 `getColors` defaults (which correctly map undefined→contrast), this diverged for a null `color`. **Fixed on-branch** → `?? 'contrast'` (both fg/bg), matching the old fallback. Residual noted transparently: several "crash → graceful no-op" guards are behavior-preserving under the current host-ordering invariants (AppService seeds theme sync before widgets; Host2 `initialize` before effects).

## Verify

`npm run snc`: 14 files clean, **133 → 82**, no new errors. `.betterer.results` regenerated. `npm run lint` clean. The 6 modified components that have specs all pass. CI runs the full suite.

**Merge note:** touches `.betterer.results` — serializes with any other open strict PR. #6, S2-1 cluster 4.

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -70,36 +70,12 @@ exports[`strictNullChecks`] = {
       [85, 26, 49, "Object is possibly \'null\'.", "3725617003"],
       [89, 26, 49, "Object is possibly \'null\'.", "3725617003"]
     ],
-    "src/app/widgets/svg-racesteer/svg-racesteer.component.ts:3776013780": [
-      [36, 46, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
-      [37, 47, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
-      [39, 51, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
-      [40, 51, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
-      [41, 57, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
-      [42, 57, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
-      [43, 57, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"]
-    ],
-    "src/app/widgets/svg-windsteer/svg-windsteer.component.ts:2551054731": [
-      [28, 59, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
-      [40, 50, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
-      [45, 46, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
-      [46, 47, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
-      [47, 51, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
-      [49, 57, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
-      [50, 57, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"],
-      [51, 57, 9, "Argument of type \'undefined\' is not assignable to parameter of type \'number\'.", "2620553983"]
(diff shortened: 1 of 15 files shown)


mairas:
## Code review — this PR (#6 S2-1 widgets cluster)

**Reviewers (2):** correctness + **adversarial** — a 14-file/51-fix cluster with behavior-adjacent defaults and crash→guard changes earns both lenses.

### One finding — fixed on-branch
Both reviewers converged on it: **`widget-label`** used `?? 'green'`/`?? 'grey'` where its `mapColor(undefined)` old-fallback was `theme.contrast`. Unlike the 8 `getColors` defaults — where `getColors(undefined)` already maps to `contrast`, so `?? 'contrast'` is exactly equivalent — `widget-label` uses `mapColor`, so `'green'`/`'grey'` diverged (→ `theme.green`/`grey`) for a null `color`. Reachable only via a corrupt stored config (`"color": null`), but a genuine divergence. **Fixed** → `?? 'contrast'` (fg + bg), matching the old fallback. (P3, conf ~0.62.)

### Verified sound (the behavior-adjacent changes)
- **`?? 'contrast'` defaults ×8** — `getColors(undefined) === getColors('contrast')` (unknown keys map to contrast), and the field is always present via the DEFAULT_CONFIG merge, so the `??` never fires anyway.
- **crash→guard changes** (widget-text `setColors` `!theme`, `drawWidget` `!cfg`, windsteer `paths?.`, freeboardsk) — inert today: `AppService` seeds a non-null theme synchronously before any widget renders, and Host2 runs `runtime.initialize` before child effects, so the guarded null is never reached.
- **input retypes / local captures / `'--'` on null / dead operands** — type-only or dead-branch; no runtime effect.

### Residual (noted, no action)
Those crash→no-op guards are behavior-preserving *under the current host-ordering invariants*, not intrinsically — if a future change let a widget construct without `runtime.initialize`, or reset the theme to null post-boot, they'd silently no-op where the old code threw. Inherent to narrowing compiler-unprovable invariants.

### Flagged files (deferred, not forced)
`svg-zone-states` latent crash → **#332**; the shared-interface `null`-sentinel root cause (5 files) → **#333**; `widget-racesteer` intentional-null type-lie left as-is.

Baseline 133→82. The one real divergence is fixed; residuals documented. S2-1 cluster 4 (#6).

PR #335, opened 2026-07-17
fix(types): resolve strictNullChecks in dashboard.component (#6, S2-1 cluster 5)

mairas:
## Why

S2-1 (#6 strictNullChecks endgame) — `dashboard.component.ts`, the single biggest remaining file (34 gridstack null-safety errors). Self-contained (no cascade). **82 → 48.**

## What (all behavior-preserving; no `!`, no new `as`, no suppressions)

The gridstack API surfaces are optional — `GridstackComponent.grid` is `undefined` until the grid mounts, and `GridItemHTMLElement.gridstackNode` / `NgGridStackWidget.input` are structurally optional — but every erroring site is a handler that runs *after* the grid is ready (user actions, post-`ngAfterViewInit`). So the fixes narrow at the use site, matching the `try/catch { grid not ready }` intent the code already had:

- **`_gridstack().grid`** → a local `const grid = …; if (!grid) return;` guard (in `ngAfterViewInit`, `saveDashboard`, `onEmptyAreaTap`, `addWidgetToGrid`, `duplicateWidget`, `pasteCopiedWidget`), or `?.` at the three sites already inside a `try/catch` (the static-toggle effect, `cancelPointerSequence`, `onPointerRelease`, `deleteWidget`).
- **`item.gridstackNode`** → `?.id` / a `const node = …; if (!node) return;` guard before deref; **`source.input`** → `const widgetProps = source.input?.widgetProperties; if (!widgetProps) return;`.
- **`grid.load(dashboard.configuration ?? [])`** (config is optional); **`saveDashboard`** guards `activeDashboard === null`.

Every guard is unreachable once the grid is mounted (the invariant the old `try/catch` relied on), so runtime behavior is unchanged.

## Verify

`npm run snc`: dashboard clean, **82 → 48**, no new errors. `.betterer.results` regenerated (dashboard removed). `npm run lint` clean. **`dashboard.component.spec.ts`: 25 tests pass** — it exercises add/delete/duplicate/paste/save + page transitions, so the guards are exercised and behavior-preserving. CI runs the full suite.

**Merge note:** touches `.betterer.results` — serializes with any other open strict PR. #6, S2-1 cluster 5.


mairas:
## Code review — this PR (#6 S2-1 dashboard.component)

**Reviewers (2):** correctness + **adversarial** — 34 null-guards in a core gridstack-driven component earns both. **Both returned zero code findings.** They independently traced every guard against the gridstack 12.x source and `dashboard.service` state machine and confirmed the null-guards fire only in states where no widget op could succeed anyway:

- **`if (!grid) return;`** (ngAfterViewInit subscribe, saveDashboard, onEmptyAreaTap, addWidgetToGrid, duplicateWidget, pasteCopiedWidget) — unreachable in normal operation: these are user-action/subscription handlers that run after the child GridstackComponent's lifecycle mounts the grid, and `takeUntilDestroyed`/`grid.destroy()` tear subscriptions down before it can go null. `widgetAction$` is a `null`-seeded BehaviorSubject and the `if (action)` rejects the initial null.
- **`saveDashboard` `activeDashboard === null` guard** — behavior-equivalent even if reachable: the old `updateConfiguration(null, …)` was already an internal no-op (no index matches null; the `{ equal: isEqual }` signal suppresses notification), and `grid.save()` is a pure read.
- **`?.` at the try/catch sites** (static-toggle effect, cancelPointerSequence, onPointerRelease, deleteWidget) — exactly the old throw-then-catch-and-ignore.
- **`item.gridstackNode?.id` / `source.input?.` guards** — the undefined case isn't reachable for a settled grid item on a user action (sub-grid children aren't returned by the top grid's `getGridItems()`; `addWidget` assigns `gridstackNode` synchronously).
- **No `!`, no *new* `as`, no suppressions** (pre-existing `as NgGridStackWidget` casts untouched).

### Honest flag (per the migration convention)
The one line that is **not** strictly zero-change: **`grid.load(dashboard.configuration ?? [])`**. Verified from gridstack source that `load(undefined)` throws (and leaves the grid stuck in the already-opened `batchUpdate` with the previous page's widgets), so `?? []` is a **benign behavior change — a latent-bug fix (crash → clean clear), not a regression**. Reachable only via externally-corrupted applicationData with a dashboard object missing its `configuration` key; every in-app creation/migration path sets it to at least `[]`. Surfacing it explicitly rather than filing it under "zero change." (`duplicateWidget`'s `input?.` guard is the same class — crash→graceful-skip on an unexpected node shape.)

### Residual / testing gap (noted, no action)
No spec covers `loadDashboard` with an undefined `configuration` (asserting empty-load, not a stuck batch) — a regression test would lock in that one intended benign change. Deferred (YAGNI: externally-triggered corrupted state, now handled gracefully).

Zero code defects across both lenses; the single behavior change is a benign latent-bug fix, flagged. Baseline 82→48; `dashboard.component.spec` 25 tests pass. S2-1 cluster 5 (#6).

PR #336, opened 2026-07-17
fix(types): widen IWidgetPath + widgetUrl null fields (#6, #333 safe subset)

mairas:
## Why

S2-1 (#6) — the safe subset of the **#333** interface-widen enabler. Several DEFAULT_CONFIGs and the config UI use `null` as a meaningful value for these fields, but the shared interfaces typed them non-null — a **type-lie** (the `null` was already occurring at runtime) that blocked the consumer files.

## What (pure type change — zero runtime effect)

Widen four fields to match their documented runtime domain:
- **`IWidgetPath.pathType` / `pathSkUnitsFilter` / `convertUnitTo`** → allow `null`. Their own JSDoc already states `null` is valid ("Use null for no conversion", "null to list all types of paths").
- **`IWidgetSvcConfig.widgetUrl`** → allow `null` (the config UI branches on `widgetUrl !== undefined`, so `null` ≠ `undefined` is meaningful).

Widening a type has no runtime effect — it only makes the compiler accept the `null` that was already there and *forces* readers to handle it. **No consumer cascaded** (`snc` confirms zero new errors), which proves every reader already handles `null` (unsurprising — the runtime already produced it). So it cannot mask a bug; it can only surface latent ones, and none surfaced.

## Verify

`npm run snc`: **48 → 44**. `widget-iframe` fully resolved; `widget-racer-line` 7→6, `widget-multi-state-switch` 2→1 (they keep other, unrelated errors). No new errors anywhere. `.betterer.results` regenerated, `betterer:ci` clean, `lint` clean. `widget-iframe` spec: 9 tests pass. CI runs the full suite.

## Deferred (documented on #333)

The **`IConnectionConfig.signalKUrl`** widen is intentionally **not** here — it cascades into the load-bearing `settings.service` (`signalkUrl.url` + `validateSignalKUrl`), which needs its own careful connection-flow PR. The measured blast radius is on #333.

#6, S2-1. Serializes on `.betterer.results`.

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -15,8 +15,7 @@ exports[`strictNullChecks`] = {
     ],
     "src/app/widget-config/boolean-multicontrol-options/boolean-multicontrol-options.component.ts:1795746450": [
       [35, 9, 14, "Type \'null\' is not assignable to type \'UntypedFormGroup\'.", "3522031013"],
-      [37, 10, 26, "Type \'null\' is not assignable to type \'Subscription\'.", "273497119"],
-      [79, 8, 13, "Type \'null\' is not assignable to type \'string | undefined\'.", "2466344431"]
+      [37, 10, 26, "Type \'null\' is not assignable to type \'Subscription\'.", "273497119"]
     ],
     "src/app/widget-config/dataset-chart-options/dataset-chart-options.component.ts:155376543": [
       [83, 26, 10, "Argument of type \'ISkPathData | null\' is not assignable to parameter of type \'ISkPathData\'.\\n  Type \'null\' is not assignable to type \'ISkPathData\'.", "384538813"]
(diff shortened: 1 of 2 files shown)


mairas:
## Code review — this PR (#6 / #333 safe subset — interface widen)

**Reviewer:** correctness (focused — a pure shared-interface type-widen; the only real risk is masking a latent bug at a reader, or the field not actually being nullable at runtime). **No findings.**

Verified against the compiler and the real readers:
- **Cannot mask a bug:** widening a type *adds* `null`, making the compiler *stricter* on readers (they must handle it). `snc` reports **zero** new errors (48→44, a pure-removal betterer diff), which proves every reader of these fields already tolerates null.
- **Readers genuinely tolerate null** (spot-checked): `widget-iframe`'s `resolveUrl(rawUrl: string | null | undefined)` explicitly accepts null; `widget-streams.directive`'s `convert ? convertToUnit(...) : val` truthy-guards `convertUnitTo`; the path-form components truthy-guard `pathSkUnitsFilter` / pass it as a null-tolerant FormControl seed.
- **`null` is a real runtime value** for all three (iframe default config sets `widgetUrl: null`; config literals + path forms set `convertUnitTo`/`pathSkUnitsFilter` to null) and each field's **JSDoc already documents null as valid** — the widen makes the type *honest*, not looser than intended.
- **Type-only:** the diff touches only the field annotations + `.betterer.results`; no runtime edit rode along.

The reviewer also caught that `pathType: TWidgetPathType | null` was **redundant** (`TWidgetPathType` already includes `null`) — I dropped that no-op line, keeping the three substantive widens.

Honest type-widen, behavior-neutral, masks nothing, betterer:ci-verified. 48→44; `widget-iframe` off the baseline. `signalKUrl` deferred to its own connection-flow PR (blast radius on #333). #6, S2-1.

PR #344, opened 2026-07-18
Handle whole-object Signal K compound leaf values (S2-0 prep)

mairas:
## Motivation

Preparatory, additive step for **S2-0 / #21** (canonical whole-value emission). Today `SignalKDeltaService` recursively flattens object-valued Signal K deltas into fabricated dotted child paths (`navigation.position.latitude`, `design.length.overall`, …). The flip removes that flattening so compound leaves arrive **whole** at their canonical SK path. Verified against the live boat's SK model, the AIS-relevant compound leaves are `navigation.position`, `design.length`, `design.draft`, `design.aisShipType`, and `navigation.closestApproach`.

Without a whole-object handler, after the flip AIS dimensions, ship type, and CPA would silently drop (their child-path handlers would never fire again).

## Approach

Add `applyAisUpdate` cases that read the sub-fields off the whole compound object at its canonical path (`navigation.position` was already handled whole via `readPositionValue`). Two small helpers — `asRecord` and `readNumericFields` — keep partial updates from clobbering previously-set sibling fields.

These new cases sit **alongside** the existing dotted child-path cases. Because the delta service still flattens today, the child cases stay live and the new whole-object cases are dead until the flip — so this PR is safe to land on its own and shrinks the S2-0 cutover PR. It lands **before** the flip PR (same merge batch).

## Verification

- New characterization tests drive whole objects through the real stream entry point and assert resulting track state (length/draft/aisShipType/closestApproach, CPA collision-risk detection, sibling-merge, ATON type guard).
- `npm run snc` (strictNullChecks, prod + specs) and `npm run lint` clean; ais-processing spec 30/30 green.

Refs #21

### DIFF
--- a/src/app/core/services/ais-processing.service.spec.ts
+++ b/src/app/core/services/ais-processing.service.spec.ts
@@ -5,6 +5,7 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import {
   AisProcessingService,
   AisVessel,
+  AisSar,
   AisAton,
   AisTrack
 } from './ais-processing.service';
@@ -27,6 +28,7 @@ describe('AisProcessingService applyAisUpdate dispatch', () => {
 
   const VESSEL_CONTEXT = 'vessels.urn:mrn:imo:mmsi:123456789';
   const ATON_CONTEXT = 'atons.urn:mrn:imo:mmsi:987654321';
+  const SAR_CONTEXT = 'sar.urn:mrn:imo:mmsi:111222333';
 
   function makeEvent(fullPath: string, value: unknown): IPathUpdateWithPath {
     return {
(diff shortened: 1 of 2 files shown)


mairas:
## Independent review synthesis (multi-persona, adversarially verified)

Ran a 3-lens review (correctness / adversarial / testing) with an adversarial verify pass (default-to-refute) on the diff. 7 findings survived verification; all are addressed in the amended commit. 5 further candidate findings were refuted.

**Correctness (fixed):**
1. **[medium] `navigation.closestApproach` merged instead of replacing** — a later whole update omitting `collisionRiskRating` could never clear a prior value; since `hasCollisionRiskData`/`isNoCollisionRisk` key off `hasOwnProperty('collisionRiskRating')`, a passed threat would keep painting a phantom red collision vector until TTL eviction. Fixed: all whole-object compound leaves now **replace** the sub-object wholesale (matching the `navigation.position` handler), since the emitted object is the complete authoritative value. Sibling `design` leaves (beam/draft/…) stay preserved by merging at the `design` level.
2. **[low] `null` sub-field coerced to 0** — `readNumericFields` ran `toNumberOrUndefined` on present-but-null values, and `Number(null)===0`; a `collisionRiskRating: null` ("value unavailable") became `0`, the most-severe risk. Fixed: `null`/`undefined` sub-fields are now omitted (unavailable ≠ a real value).

**Testing (added):** whole `closestApproach` **without** CRR keeps `hasCollisionRiskData()` false; a later omit **clears** a stale CRR; `null` CRR treated as absent; wholesale-replace of a design sub-object; SAR-target coverage; `aisShipType` id-only / `draft.canoe` / CPA range+bearing partials; whole-position altitude; non-object-value guard. Suite now 39/39.

`npm run snc` + `npm run lint` clean.

Issue #345, opened 2026-07-18
Charting a compound sub-field (e.g. navigation.attitude.roll trend) after the S2-0 canonical-emission flip


mairas:
## Context

S2-0 (#21) removed the delta parser's object-flattening: compound Signal K leaves (`navigation.position`, `navigation.attitude`, `navigation.gnss.satellitesInView`, `design.length/draft/aisShipType`, `navigation.closestApproach`) now arrive **whole** at their canonical path instead of being split into fabricated dotted child paths.

Per the agreed scope, the numeric/gauge widgets that read a sub-field (`widget-position`, `widget-heel-gauge`, `widget-horizon`, autopilot Next-WPT) were adapted in code (they pass a `subField` to `WidgetStreamsDirective.observe()`, which extracts it before unit conversion). **Charting a compound sub-field was deferred** — nothing on the boat and no default charts a sub-field.

## What's deferred (post-flip gaps)

- **Trend/data charts** (`widget-data-chart`, `minichart`, `HistoryChartStreamService`) read their path from flat config fields (`datachartPath`), not `IWidgetPath.paths`, and pull `pkt.data.value` as a number. Pointed at a compound leaf they now get the whole object → `Number(obj)` = NaN → empty/gapped chart. The History-API backfill also needs a dotted-leaf query path reconstructed.
- **`angle-domain.util`** still lists `self.navigation.attitude.{roll,pitch,yaw}` in its signed-angle allowlist — dead entries until charts can target those sub-fields again; needs re-keying on (basePath, subField).
- **`widget-heel-gauge` / `widget-horizon`** had `supportAutomaticHistoricalSeries` turned off in this packet to avoid the auto-history sync requesting a numeric series for an object-typed path; restoring their history-dialog series is part of this work.
- **`units.service` position special-case** (`getConversionsForPath`, the `position.latitude`/`position.longitude` substring test) is now dead — the position widgets' paths are non-configurable so the picker never runs for them; revisit if/when the picker must offer a compound sub-field.

## Direction (when picked up)

Give the chart engines a sub-field selector paralleling the widget `subField`: extract `value[subField]` from the whole compound before `Number()`/finite checks on the live tail, and reconstruct the dotted leaf path (`<compound>.<subField>:<agg>`) for the History-API query so the scalar-column mapper is unchanged. Then restore the attitude widgets' `supportAutomaticHistoricalSeries` and rework `angle-domain` to resolve angularity per sub-field.

Refs #21

PR #346, opened 2026-07-18
Canonical whole-value emission: stop flattening compound Signal K leaves (S2-0)

### LINKED ISSUE #21: [SK-02] Object/array value flattening in delta parser fabricates non-spec Signal K child paths

mairas:
Closes #21. Stacked on #344 (base branch `fix/ais-whole-object-values`) — merge #344 first, then retarget this to `main`.

## Motivation

`SignalKDeltaService` recursively flattened object/array values into fabricated dotted child paths (`navigation.position.latitude`, `navigation.attitude.roll`, …) that don't exist in the SK spec. The path set was non-deterministic (size/depth heuristics), arrays exploded into `path.0`/`path.1` (no `Array.isArray` guard), and an empty `{}` value silently emitted nothing. Path pickers surfaced fabricated paths and canonical subscriptions (`navigation.position`) received no data.

## Approach

Emit every value **whole** at its canonical SK path; meta likewise (`parseSkMeta` stops fanning `properties` into child meta paths). Verified against the live boat, the only consumers of compound sub-fields are four predefined widgets — so rather than a generic accessor, each reads its own sub-field:

- **`WidgetStreamsDirective.observe(pathName, next, subField?)`** extracts one field from the whole compound value at the single raw-value entry point, **before** unit conversion (so `convertUnitTo`, incl. position's `longitudeMin` formatting, keeps working). A scalar value passes straight through; the sub-field stays out of the base key so widgets sharing a path share one registration.
- **widget-position / heel-gauge / horizon / autopilot Next-WPT** point at the whole compound leaf and pass their sub-field. Those sub-field paths are `isPathConfigurable:false` (a numeric widget can't pick an object-typed path).
- **Config migration v13→v14** rewrites stored sub-field widget paths to the whole canonical path (idempotent; both `Record` and array `paths` forms; nested `nextPoint.position` too). Validated against the boat's actual stored config.
- Default dashboard + regenerated MCP widget schema updated to match.

The **generic sub-field accessor** approach (a config-schema field + path-picker UI + units/charts/history cascade) was explored and rejected as over-engineering for four widgets (rationale on #21). **Charting** a compound sub-field is deferred to #345 (nothing on the boat or in defaults charts one; heel-gauge/horizon `supportAutomaticHistoricalSeries` turned off to avoid an object-typed history query).

## Verification

- Full CI gate green (`lint` → `snc` → `test:headless` → `test:plugin` → `test:mcp-schema`): 1360+ tests.
- Re-pinned delta characterization suite (whole-value + whole-meta emission, array-whole, empty-`{}`-whole data-loss fix); new streams sub-field extraction tests (extract-before-convert, scalar passthrough, missing-field→null); new v13→v14 migration tests (Record + array forms, nested compound, genuine-leaf-untouched, idempotency).
- Migration transform validated against the boat's real stored config (all 5 sub-field paths rewrite correctly).
- **Pending pre-merge:** on-boat visual verification that position/heel/horizon render live and a v13 config upgrades cleanly (batched boat session).

### DIFF
--- a/perf-harness/lib/kip-config.mjs
+++ b/perf-harness/lib/kip-config.mjs
@@ -8,7 +8,7 @@
  *
  * Three distinct version spaces (all from src/app/core/constants/config-versions.const.ts):
  *  - applicationData URL path segment 11 (REMOTE_CONFIG_FILE_VERSION)
- *  - app.configVersion 13 (LATEST_APP_CONFIG_VERSION)
+ *  - app.configVersion 14 (LATEST_APP_CONFIG_VERSION)
  *  - connectionConfig.configVersion 13 (CONNECTION_CONFIG_VERSION)
  */
 export const SELF_URN = 'vessels.urn:mrn:signalk:uuid:11111111-1111-4111-8111-111111111111';
@@ -30,11 +30,12 @@ const DEFAULT_NOTIF = {
 };
 
 export function appConfig(extra = {}) {
-  // configVersion 13 with no dataSets field: a genuine post-v12->v13 config, so the
-  // boot skips ConfigurationUpgradeService's migration toast/reload (which is what
-  // strips dataSets/datasetUUID/chartEngine) inside the measurement window.
+  // configVersion at LATEST (14): a genuine current config, so the boot skips
+  // ConfigurationUpgradeService's migration toast/reload entirely inside the
+  // measurement window. Bump this in lockstep with LATEST_APP_CONFIG_VERSION, or
+  // the boot triggers an upgrade+reload and the boot-assert fails.
   return {
-    configVersion: 13, autoNightMode: false, redNightMode: false, nightModeBrightness: 0.27,
+    configVersion: 14, autoNightMode: false, redNightMode: false, nightModeBrightness: 0.27,
     widgetHistoryDisabled: false, unitDefaults: DEFAULT_UNITS,
     notificationConfig: DEFAULT_NOTIF, browserTabTitle: 'Skip', ...extra,
   };
--- a/src/app/app.component.spec.ts
+++ b/src/app/app.component.spec.ts
@@ -453,6 +453,11 @@ describe('AppComponent — embed read-only invariants (#216 E6)', () => {
     expect(runUpgradeSpy).toHaveBeenCalledWith(12);
   });
 
+  it('runs the config migration in the full app for an upgradeable v13 config (the S2-0 gate)', async () => {
+    const { runUpgradeSpy } = await render({ embed: false, configUpgrade: true, configVersion: 13 });
+    expect(runUpgradeSpy).toHaveBeenCalledWith(13);
+  });
+
   it('does NOT show the missing-shared-config create prompt under embed', async () => {
     const { toast, bootstrapIssue$ } = await render({ embed: true });
     toast.show.mockClear();
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -100,7 +100,7 @@ export class AppComponent implements AfterViewInit, OnDestroy {
       if (this.settings.configUpgrade()) {
         const liveVersion = this.settings.getConfigVersion();
 
-        if (liveVersion === 11 || liveVersion === 12) {
+        if (liveVersion === 11 || liveVersion === 12 || liveVersion === 13) {
           this.upgrade.runUpgrade(liveVersion);
         }
 
--- a/src/app/core/constants/config-versions.const.ts
+++ b/src/app/core/constants/config-versions.const.ts
@@ -21,7 +21,7 @@ export const REMOTE_CONFIG_FILE_VERSION = 11;
  * its legacy transforms pin their own MIGRATION_OUTPUT_VERSION so that bumping this constant
  * cannot silently re-label old migration output as current — add a chained migration instead.
  */
-export const LATEST_APP_CONFIG_VERSION = 13;
+export const LATEST_APP_CONFIG_VERSION = 14;
 
 /**
  * Per-device connectionConfig schema version (its own version space, decoupled from the app config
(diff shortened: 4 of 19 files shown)


mairas:
## Independent review synthesis (5-lens, adversarially verified)

Ran correctness / adversarial / migration-reliability / testing / maintainability lenses with an adversarial verify pass (default-to-refute). 8 findings survived verification; 1 refuted. All addressed in `5bd1253a` + `d00b2bf3`.

**Migration correctness — the important one (HIGH + 2 corroborating lenses, fixed):**
- The v13→v14 migration rewrote a compound sub-field path for **every** widget type. A generic widget (numeric/gauge) a user had pointed at `navigation.attitude.pitch`/`.yaw` or `navigation.position.latitude` — a valid numeric leaf before the flip — got its path collapsed to the whole compound leaf and then rendered the raw object as **NaN**, actively downgrading what would otherwise be a clean `--` no-data placeholder. Fix: **scope the rewrite to the sub-field-aware widget types** (`widget-position`/`heel-gauge`/`horizon`); a generic widget is left on its now-inert child path. Fixed the misleading docstring too.

**Migration completeness (MEDIUM, fixed):** the rewrite only touched `path`, leaving a stale stored `isPathConfigurable:true` (and heel/horizon `supportAutomaticHistoricalSeries:true`) which `merge(default,stored)` lets win over the new `false` defaults — so a migrated config would fail path-picker validation / request an object-typed history series. The migration now reconciles both.

**Perf-harness (MEDIUM, fixed):** its mock seed was pinned at `configVersion:13`; after the LATEST→14 bump a run would boot into the upgrade+reload flow it exists to avoid (CI doesn't cover perf-harness). Bumped to 14.

**Test coverage (MEDIUM×2, fixed):** the v13 boot-gate that triggers the migration was untested (only v11/v12 were) → added. The heel-gauge/horizon specs stubbed `observe()` and never fed a path, so the per-widget `subField` — the load-bearing part of the flip — was unverified → they now assert `subField` `'roll'`/`'pitch'` reaches `observe()`.

**Refuted (1):** an adversarial claim that the generic-widget NaN is a defect *of the streams pipeline* — it's a migration-scoping issue (now fixed) plus the intended #345 deferral, not a pipeline bug.

**Nit (not blocking):** the `units.service` `position.latitude`/`.longitude` substring special-case is now dead post-flip — already tracked in the #345 follow-up (which reworks unit resolution for sub-fields).

Full CI gate green after fixes (`lint`→`snc`→`test:headless`→`test:plugin`→`test:mcp-schema`).

mairas:
## On-boat verification ✅ (live, halpi)

Deployed this branch to the boat's Signal K and verified end-to-end against **live data**:

- **App's own boot-time migration ran** (config reverted to v13 first, so the real `runUpgrade(13)` path executed): on-disk `configVersion` → 14, all five sub-field paths rewritten to the compound leaf (`navigation.attitude` / `navigation.position`), `isPathConfigurable` → false, heel/horizon `supportAutomaticHistoricalSeries` → false. Full plumbing (boot-gate → scoped migration → persist) confirmed.
- **Widgets render live**: Position (formatted lat/long), Heel (roll), and Artificial Horizon (pitch + roll) all display correctly from the whole compound values — proving the per-widget `subField` extraction + unit conversion end-to-end.

No regressions observed. Ready for merge (serialized: #344 first, then retarget this to main).

Issue #347, opened 2026-07-18
Units API Phase 2: retire client-side unit ownership in favour of the Signal K unit-preferences API


mairas:
Follow-up to #246. Phase 1 (adopt the server's `displayUnits.targetUnit` as a path's **default** unit, via #246) landed the capture + resolver-source swap while leaving Skip's own machinery in place. Phase 2 retires that machinery so the server preference applies **everywhere**, not just to freshly-configured paths.

## ✅ Settled product decision — do NOT re-open

**All unit management belongs to the Signal K server. Skip does not own, configure, or provide a fallback for units.** The server's unit-preferences API (`meta.displayUnits`: `targetUnit` / `formula` / `inverseFormula` / `symbol` / `displayFormat`) is the single source of every path's display unit and symbol.

This retires — **entirely** — Skip's per-widget `convertUnitTo` overrides, `unitDefaults`, `DefaultUnitsConfig`, and the `options/units` settings screen. There is **no client-side unit fallback and no thin-view over the server prefs.** On a server that runs no units-preference plugin, values display in whatever units the server emits (SI base); Skip does not substitute a preference of its own. Skip retains a conversion *mechanism* only to **apply** the server's specified conversion (targetUnit/formula) — never to *choose* or *store* a unit.

This resolves the "Decide:" fork in scope item 2 and supersedes the earlier "graceful degradation falls back to Skip's engine" non-goal. This is a long-standing product decision; it is not to be re-litigated.

## Scope (decided during Phase 1)

1. **Retire per-widget `convertUnitTo` overrides.** Today every widget path stores a concrete `convertUnitTo` (auto-seeded from the group default), indistinguishable from a deliberate user choice — which is why Phase 1 could only change the *default*, not live rendering. Phase 2 removes the per-widget override (or reframes it as an explicit opt-out) so the server preference drives conversion + label for existing widgets too. Needs a config migration for stored `convertUnitTo` fields. **(Done — Step 2b, merged in #366; display paths now read the server measure via the streams directive.)**
2. **Retire `unitDefaults` + the Units settings screen.** `SettingsService.unitDefaults`, the `Array<IUnitDefaults>` stored config section, `DefaultUnitsConfig`, and `options/units/units.component.*` overlap the server's per-user preset model. **Decided: remove entirely — the server is the only source. Not a thin view over `GET/PUT /signalk/v1/unitpreferences/*`.** Config-version migration to drop the field.
[...]

PR #348, opened 2026-07-18
Adopt Signal K unit-preferences as a path's default unit (S2-2, units Phase 1)

mairas:
Phase 1 of #246. Follow-up scope tracked in #347.

## Motivation

Skip owns its entire unit story client-side (a hand-authored conversion table + per-user group defaults). Signal K's `signalk-units-preference` plugin (installed and active on the boat, preset `nautical-metric`) publishes a per-path `displayUnits` preference in path meta. This lets a path default to the **boat-wide** server preference instead of each client configuring units separately.

## What this does (Phase 1)

- Types `ISkMetadata.displayUnits` (`{category, targetUnit, symbol, formula, inverseFormula, displayFormat}`) — the data already arrives via the existing `sendMeta=all` stream and is deep-merged into `_skData[path].meta`; it was just untyped/unread. Adds `DataService.getPathDisplayUnits(path)`.
- `UnitsService.getConversionsForPath(path).base` now prefers the server's `displayUnits.targetUnit` over the per-group `unitDefaults`, so a **freshly-configured** path defaults to the server preference.

## Design choices

- **Route the server `targetUnit` through Skip's own resolver, not its raw `symbol`.** The boat's preset spells some symbols out (`hour`, `liter`) — exactly what #245 fixed — so adopting the server symbol verbatim would regress label consistency. Mapping `targetUnit`→Skip measure yields Skip's clean symbol.
- **Single-source invariant preserved.** Label (`getUnitDisplaySymbol`) and conversion (`convertToUnit`) still both key off one measure, so "the label provably matches the applied conversion" holds — the server preference is honoured only when its `targetUnit` maps to a measure valid for the path's group.
- **Graceful degradation.** Absent plugin, unit-less path (e.g. `navigation.position`), or an unmapped/unsupported target → falls back to Skip's own default. Not gated on any deployment's plugin state.
- **Per-widget `convertUnitTo` overrides untouched** — a stored value still wins; the server only supplies the default. Their retirement (making the server apply live everywhere), formula eval, and the electrical-family temperature seam are **Phase 2 (#347)**.

## Verification

- Full CI gate green (`lint`→`snc`→`test:headless`→`test:plugin`→`test:mcp-schema`).
- New `UnitsService` tests: server target preempts the group default; direct-match (`mbar`) and alias (`kn`→`knots`, `C`→`celsius`) mapping; the `C`→`celsius` alias does **not** mis-apply on a Charge path (group guard); fallback when absent/unmappable; and the label-matches-conversion invariant (resolved base has both a symbol and a working conversion).
- Grounded against the live boat's `displayUnits` vocabulary (15 categories).

### DIFF
--- a/src/app/core/interfaces/signalk-interfaces.ts
+++ b/src/app/core/interfaces/signalk-interfaces.ts
@@ -186,11 +186,27 @@ export interface ISkMetadata {
   properties: object; // Not defined by Kip. Used by GPS and Ship details and other complex data types
   method?: TMethod[];
   displayScale?: ISkDisplayScale
+  /**
+   * Server-supplied display-unit preference (Signal K unit-preferences plugin). Present per path
+   * when the plugin is installed; absent otherwise. `targetUnit`/`symbol` are the plugin's unit key
+   * and abbreviation; `formula`/`inverseFormula` are math.js expressions (SI <-> target) reserved for
+   * a later phase — Skip currently maps `targetUnit` onto its own conversion engine.
+   */
+  displayUnits?: ISkDisplayUnits;
   alertMethod?: TMethod[];
   warnMethod?: TMethod[];
   alarmMethod?: TMethod[];
   emergencyMethod?: TMethod[];
   zones?: ISkZone[];
+}
+
+export interface ISkDisplayUnits {
+  category?: string;
+  targetUnit?: string;
+  symbol?: string;
+  formula?: string;
+  inverseFormula?: string;
+  displayFormat?: string;
 }
 
 export interface ISkPossibleValue {
--- a/src/app/core/services/data.service.ts
+++ b/src/app/core/services/data.service.ts
@@ -1,7 +1,7 @@
 import { DestroyRef, inject, Injectable, OnDestroy } from '@angular/core';
 import { Observable, BehaviorSubject, ReplaySubject, Subject, map, combineLatest, interval, filter, Subscription } from 'rxjs';
 import { ISkPathData, IPathValueData, IPathMetaData, IMeta, IPathUpdateEvent } from "../interfaces/app-interfaces";
-import { ISignalKDataValueUpdate, ISkMetadata, ISignalKNotification, States, TState } from '../interfaces/signalk-interfaces'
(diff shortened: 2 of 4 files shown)


mairas:
## Independent review synthesis (3-lens, adversarially verified)

Ran correctness / adversarial / testing lenses with an adversarial verify pass. One finding was CONFIRMED by the verifier but is **incorrect in its conclusion**; its valid underlying concerns are addressed. Amended in the latest push.

### The CONFIRMED finding — conclusion refuted by live re-verification
The finding claims the alias map is keyed on the wrong strings, so temperature/time silently no-op. The verifier did deep work — it read the plugin's **internal** config (`units-preferences.json`) and source (`ConversionEngine.ts`), which use keys like `°C`/`F`/`second`/`HH.xx`. But those are the plugin's *internal* representation, **not** the `displayUnits.targetUnit` field it emits. I re-verified against the live server's emitted meta:

```
environment/water/temperature -> targetUnit='C'   symbol='°C'
.../capacity/timeRemaining     -> targetUnit='hour' symbol='hour'
tanks/fuel/0/capacity          -> targetUnit='liter'
navigation/speedOverGround     -> targetUnit='kn'
```

The emitted `targetUnit` is exactly `C`/`hour`/`liter`/`kn` — the map's keys. So temperature/time **do** resolve on the boat; they do not no-op. And even a hypothetical string mismatch degrades gracefully to Skip's own default (never a wrong value), so the severity would be overstated regardless.

### Valid underlying concerns — addressed
- **Documented** that the alias keys are the plugin's **emitted** `targetUnit` strings (verified from live meta), not its internal keys — to prevent exactly the confusion above.
- **Broadened tests**: the previously-untested `naut-mile→nm`, `degree→deg`, `hour→Hours` aliases; a parametrized label-matches-conversion sweep across all aliased categories (asserting each resolved base has both a symbol and a finite conversion), with empty defaults so the server value is the sole base source.
- **Table-integrity test**: asserts every conversion-list measure has a working conversion — so a group-valid server base can never be value-less (this underpins the group-membership guard; I considered a runtime converter-check but all 80 measures already convert, so a runtime guard for an impossible state is over-engineering — the test pins the invariant at the data instead).

7 further findings were refuted (mostly the C→celsius alias/group-guard ordering — the group guard already prevents any cross-group mismap — and test-coverage nits without a concrete defect).

Full CI gate green after the amend.

Issue #351, opened 2026-07-18
Electrical widgets: solar-charger & BMS diverge from the quad on 3 shared helpers (severity ordering, number coercion, device parsing)


mairas:
## Summary

While verifying byte-identity of the electrical widgets' private helpers for the #4 extraction, three helpers turned out **not** to be identical across the family: the `charger`/`alternator`/`inverter`/`ac` "uniform quad" share one implementation, but **`widget-solar-charger`** and **`widget-bms`** carry variants that behave differently on some inputs. This is precisely the "undetected divergence between widgets could produce inconsistent behavior across the electrical dashboard" risk #4 names — now made concrete.

These are **not yet classified as bugs.** Whether each divergence is an intentional domain difference or latent copy-paste drift depends on what data shapes each widget actually receives from Signal K. This issue records the divergences so they can be triaged deliberately, separately from the #4 refactor (which stays strictly behavior-preserving and does **not** merge these variants).

## Divergence 1 — `resolveMostSevereState` severity ordering

- **Quad (`charger`, `alternator`, `inverter`, `ac`):** short-circuit priority **Alert > Alarm > Warn > Normal**. Returns the first matching state in that order.
- **`solar-charger`:** a rank-based `Record<TState, number>` fold with ordering **Normal(0) < Nominal(1) < Alert(2) < Warn(3) < Alarm(4) < Emergency(5)**, returning the highest-ranked. It also handles `Nominal` and `Emergency`, which the quad's version does not.

**Observable inconsistency:** given inputs containing both `Alert` and `Alarm`, the quad surfaces **Alert**, solar surfaces **Alarm** — opposite conclusions about which state is "most severe." The two orderings disagree wherever more than one non-Normal state is present.

## Divergence 2 — `toNumber` coercion & conversion fallback

- **Quad:** rejects `null`/`boolean`; coerces strings via `Number(value)`; `Number.isFinite` guard; `convertToUnit(unitHint, raw)`; returns the converted value only if finite, else `null`.
- **`solar-charger`:** `typeof value !== 'number'` → `null` (rejects stringified numbers outright, no coercion); returns `convertToUnit(unit, value) ?? value` — i.e. falls back to the **raw, unconverted** value when conversion yields nullish, with **no finite check**.
- **`bms`:** no unit parameter at all — a pure number passthrough with no unit conversion.

[...]

PR #352, opened 2026-07-18
refactor(electrical): extract shared config/device normalizers (#4 Stage 1a)

mairas:
## Why

The six electrical widgets (bms, solar-charger, charger, alternator, inverter, ac) share a well-designed model layer but never had their behavioral layer extracted — each carries verbatim copies of the same config/device-tracking normalizers, maintained in lockstep. This is the first surgical slice of the #4 extraction (Stage 1a): the genuinely byte-identical config normalizers.

## What

Adds `src/app/widgets/shared/electrical-config.util.ts` (with a focused spec) exporting four pure functions, and migrates each widget to import them, deleting its local copy:

- `normalizeOptionalString` (was in all 6)
- `normalizeStringList` (5)
- `normalizeTrackedDevices` (the charger/alternator/inverter triplet)
- `buildIdToDeviceKeysMap` (the charger/alternator/inverter/ac quad; now takes the devices array as a parameter instead of reading component state)

## What is deliberately *not* merged

Verifying byte-identity surfaced that the "identical across all 6" framing was inflated: **solar-charger** and **bms** carry behaviorally-distinct `normalizeTrackedDevices` variants (string-item tolerance, `Partial`-cast tracking), and **ac** has `normalizeAcTrackedDevices` (a reserved-aggregate-id guard). These are left untouched — merging them would change behavior. The divergences (severity ordering, number coercion, device parsing) are documented in #351 for separate triage; this PR stays strictly behavior-preserving.

## Verification

`npm run snc` + `npm run lint` clean; the six electrical widget specs (87 tests) and the new util spec (11 tests) all pass locally. Net ~240 lines removed.

Stage 1a of #4. Follow-ups: Stage 1b (apply-path helpers), Stage 2 (ingest engine), Stage 3 (d3 draw).

### DIFF
--- a/src/app/widgets/shared/electrical-config.util.spec.ts
+++ b/src/app/widgets/shared/electrical-config.util.spec.ts
@@ -1,0 +1,88 @@
+import {
+  buildIdToDeviceKeysMap,
+  normalizeOptionalString,
+  normalizeStringList,
+  normalizeTrackedDevices
+} from './electrical-config.util';
+
+describe('electrical-config.util', () => {
+  describe('normalizeOptionalString', () => {
+    it('trims a non-empty string', () => {
+      expect(normalizeOptionalString('  house  ')).toBe('house');
+    });
+
+    it('returns null for empty or whitespace-only strings', () => {
+      expect(normalizeOptionalString('')).toBeNull();
+      expect(normalizeOptionalString('   ')).toBeNull();
+    });
+
+    it('returns null for non-string values', () => {
+      expect(normalizeOptionalString(42)).toBeNull();
+      expect(normalizeOptionalString(null)).toBeNull();
+      expect(normalizeOptionalString(undefined)).toBeNull();
+      expect(normalizeOptionalString({})).toBeNull();
+    });
+  });
+
+  describe('normalizeStringList', () => {
+    it('returns an empty array for non-arrays', () => {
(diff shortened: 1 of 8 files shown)


mairas:

Two independent lenses (findings verified against `main`):

**Behavior-identity: zero findings.** Every extracted function is semantically identical to the private method it replaced in every adopter (diffed each pre-extraction body via `git show main:…`, ignoring only formatting/param-names). The `buildIdToDeviceKeysMap` parameter refactor is equivalent to the old no-arg method. The divergent variants that had to survive did: ac's `normalizeAcTrackedDevices` (reserved-aggregate-id guard), solar-charger's and bms's own `normalizeTrackedDevices` (empty main-vs-branch diff) — none deleted. No dangling references.

**Standards / completeness: 2 findings.**
- *[nit]* docstring omitted ac's divergent variant → **fixed** (commit amended).
- *[low]* `buildIdToDeviceKeysMap`'s id→keys logic still appears inline in the reproject/flush paths (ac reproject builds it cleanly; solar's apply-path flush guards it; charger's reproject intermixes it with `devicesByKey`). **Deferred to Stage 2**: these sites are part of the stateful reproject/flush machinery the ingest engine consolidates — adopting the helper piecemeal now (some sites clean, charger's not) would create churn Stage 2 rewrites.

PR #353, opened 2026-07-18
refactor(electrical): extract shared apply-path helpers (#4 Stage 1b)

mairas:
## Why

Stage 1b of the #4 electrical extraction — the second surgical slice: the value-apply helpers the charger/alternator/inverter/ac quad carried verbatim. **Stacked on #352 (Stage 1a); review/merge #352 first.**

## What

Adds `src/app/widgets/shared/electrical-apply.util.ts` (with a spec) and migrates the call sites:

- `setValue` / `setMetricValue` — the `Object.is` dirty-check snapshot writers, now generic over the snapshot type (`<T, K extends keyof T>`), so all five adopters share one implementation.
- `toStringValue`, `toBoolean` — pure value coercers.
- `resolveMostSevereState` — the Alert > Alarm > Warn > Normal severity fold.

## What is deliberately *not* merged

- **solar-charger** adopts only `setValue`. Its `toStringValue` **stringifies** non-strings (quad returns `null`) and its `resolveMostSevereState` uses a **rank-based** ordering with extra states — both behaviorally distinct, kept private, tracked in #351 (divergence #4 added there).
- **bms** uses a different write path and is untouched.
- `toNumber` (units-dependent) and `escapeRegex` (path-regex construction) stay in the widgets — they belong with Stage 2's ingest engine, not this stateless slice.

## Verification

`npm run snc` + `npm run lint` clean; the six electrical widget specs + the new apply-util spec (99 tests) pass locally; solar's divergent bodies confirmed byte-unchanged. Net ~183 lines removed (~420 across Stage 1).

### DIFF
--- a/src/app/widgets/shared/electrical-apply.util.spec.ts
+++ b/src/app/widgets/shared/electrical-apply.util.spec.ts
@@ -1,0 +1,115 @@
+import { States } from '../../core/interfaces/signalk-interfaces';
+import {
+  resolveMostSevereState,
+  setMetricValue,
+  setValue,
+  toBoolean,
+  toStringValue
+} from './electrical-apply.util';
+
+interface TestSnapshot {
+  voltage: number | null;
+  voltageState: States | null;
+}
+
+function snapshot(): TestSnapshot {
+  return { voltage: null, voltageState: null };
+}
+
+describe('electrical-apply.util', () => {
+  describe('setValue', () => {
+    it('writes and returns true when the value changes', () => {
+      const s = snapshot();
+      expect(setValue(s, 'voltage', 12.4)).toBe(true);
+      expect(s.voltage).toBe(12.4);
+    });
+
+    it('returns false and does not write when the value is unchanged (Object.is)', () => {
+      const s = snapshot();
+      s.voltage = 12.4;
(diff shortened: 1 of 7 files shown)


mairas:

**Behavior-identity: zero findings.** Every extracted helper is behavior-identical to the private method it replaced (diffed each pre-extraction body via `git show main:…`). The generic change `setValue<K extends keyof XSnapshot>` → `setValue<T, K extends keyof T>` is confirmed **compile-time-only** — generics erase at runtime, and each call site passes the widget's snapshot so `T` infers to exactly the type the old signature hard-coded; the `state as T[S]` cast matches the old `state as XSnapshot[S]`. solar-charger's divergent `toStringValue` (stringifies) and `resolveMostSevereState` (rank-based) are present and unmodified; bms's write path untouched.

**Completeness / correctness: all pass.** Private copies fully removed from the quad; zero dangling `this.<fn>(`; import hygiene correct (`States` dropped where only `resolveMostSevereState` used it, `TState` kept for the `applyValue` state param); no `any`/`ts-ignore`; generics correct.

- *[nit]* `setMetricValue`'s "only value changes → true" branch wasn't explicitly pinned → **added** (spec now 13 cases; commit amended).

Issue #354, opened 2026-07-19
Electrical widgets: stale device-key card persists after untracking a device, duplicates on next delta (family-wide)


mairas:
## Summary

The four source-blind electrical widgets (`widget-alternator`, `widget-inverter`, `widget-ac`, `widget-solar-charger`) leave a **stale device-key snapshot** in their by-key store when a previously-tracked device is **untracked**, and then render a **duplicate card** once a fresh delta arrives for the same id. Masked while the device stays tracked (the visible-key computed returns only tracked keys), so it only surfaces after a track → untrack config edit.

Found via an adversarial investigation while re-checking the (false) "solar weaker reproject guard" claim. This is **pre-existing and family-wide** — not introduced by, and faithfully preserved by, the #4 Stage 2 refactor. Filing so it can be fixed deliberately.

## Repro

1. Config tracks `{id: '278', source: 'victron.0'}` → store holds `278||victron.0`.
2. Config is edited to **untrack** (trackedDevices → `[]`). `applyConfig([])` sets the tracked signal to `[]` and calls `reprojectSnapshotsToDeviceKeys([])`, which **early-returns on empty devices** (`if (!devices.length) return;`) — so `278||victron.0` **survives** in the store, now with no tracked device pointing at it.
3. With `trackedDevices().length === 0`, `visibleKeys` takes the **discovered branch**: `Object.keys(store).filter(k => discoveredIds.has(store[k].id))`. `278||victron.0` has `id === '278'` (discovered) → it is **rendered** as a frozen card.
4. A later live delta for id `278` (now `hasTracked === false`) routes to `[update.id]` → creates a plain-id `278` snapshot → also discovered → **two cards for one physical device** (one live `278`, one stale-frozen `278||victron.0`).

## Root cause

Two gaps, both family-wide (byte-identical across the four widgets):
- **`reproject` does not prune** device-key entries whose tracked device was removed; its empty-devices early-return skips all cleanup.
- **The discovered-branch `visibleKeys` does not dedupe by `snapshot.id`**, so multiple keys for the same id all render.

A related variant (same shared-`changed`-flag reproject) leaves a plain-id entry for an id whose device-key already exists when it is processed **before** another id first sets `changed = true` in the same `store.update` pass — the delete-guard `if (changed && …)` reads `changed === false` for the earlier id and never revisits it.

## Suggested direction

[...]

PR #355, opened 2026-07-19
refactor(electrical): extract ingest scheduler + topology store, adopt trio (#4 Stage 2a)

mairas:
## Why

Stage 2a of the #4 electrical extraction — the stateful core. The six electrical widgets each carried a verbatim copy of the same data pipeline (buffer + coalescing timer + reproject + rAF + subscription, ~130 lines each). This extracts the genuinely-shared machinery into two composable layers and adopts them in the source-blind trio. **Follows the layered plan (v2)** posted on #4 after document-review corrected the v1 design.

## What

Two new per-widget-owned collaborators under `widgets/shared/` (composition, not a base class — matches the host-directive idiom, no spec-provider churn):

- **`ElectricalIngestScheduler`** (Layer 0) — the universal *timing* machinery: de-duping pending buffer, 500 ms coalescing timer with the initial-paint synchronous drain + first-live-immediate-flush fast path, single-flight rAF scheduler, and the path-tree subscription lifecycle + teardown. Parse / store-mutation / snapshot-resolution / draw are injected seams, so it's behavior-neutral.
- **`ElectricalTopologyStore`** (Layer 1) — the source-blind keyed-snapshot store + discovered-id/tracked-device signals, the batch-apply flush body, reproject, and the tracked-else-discovered visible selection. `applyValue` and the per-changed-key `derive` are injected seams.

`alternator`, `inverter`, `ac` adopt both layers, keeping thin **delegating aliases** (`visibleX = store.visibleSnapshots`, etc.) so their templates and specs read unchanged. `solar-charger` and `bms` are untouched here — they compose their own paths in 2c/2d.

## Strictly behavior-preserving

Zero intentional behavior changes. The derive hook fires **per-changed-key** (matching the old inline flush cadence, not once-per-batch). The trio's reproject guard is carried over exactly. (The separate solar-leak investigation confirmed there is no solar-specific leak; it did surface a pre-existing family-wide untrack-duplicate bug, filed as **#354**, which this PR faithfully preserves.)

## Verification

- New layer specs pin the timing invariants with fake timers + rAF stub (coalescing, initial-drain, first-live-immediate, rAF single-flight, teardown) and the store invariants (reproject-on-late-config, tracked-else-discovered, derive-sees-full-batch).
- `npm run snc` + `npm run lint` clean; **the three migrated widgets' full existing specs stay green unchanged** (the behavior-preservation net) — 98 tests total across the six widget specs + two layer specs.
- Net ~480 lines of duplicated pipeline removed from the trio.

Stage 2a of #4. Next: 2b bms (primitive only), 2c solar (layer via hooks), 2d charger (source-aware, highest-risk).

### DIFF
--- a/src/app/widgets/shared/electrical-ingest-scheduler.spec.ts
+++ b/src/app/widgets/shared/electrical-ingest-scheduler.spec.ts
@@ -1,0 +1,170 @@
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { Subject } from 'rxjs';
+import type { DestroyRef } from '@angular/core';
+import { States } from '../../core/interfaces/signalk-interfaces';
+import { ElectricalIngestScheduler } from './electrical-ingest-scheduler';
+import type { DataService, IPathUpdateWithPath } from '../../core/services/data.service';
+
+interface TestEntry {
+  id: string;
+  key: string;
+  value: unknown;
+}
+
+interface TestSnapshot {
+  tag: string;
+}
+
+const makeUpdate = (path: string, value: unknown): IPathUpdateWithPath => ({
+  path,
+  update: {
+    data: { value, timestamp: new Date('2026-01-01T00:00:00.000Z') },
+    state: States.Normal
+  }
+});
+
+const parseUpdate = (update: IPathUpdateWithPath): { key: string; entry: TestEntry } | null => {
+  const match = update.path.match(/\.([^.]+)\.([^.]+)$/);
+  if (!match) return null;
+  return {
+    key: `${match[1]}::${match[2]}`,
+    entry: { id: match[1], key: match[2], value: update.update?.data?.value ?? null }
+  };
+};
+
+const makeDestroyRef = (): { ref: DestroyRef; destroy: () => void } => {
+  const callbacks: (() => void)[] = [];
+  const ref = { onDestroy: (cb: () => void) => { callbacks.push(cb); return () => undefined; } } as unknown as DestroyRef;
+  return { ref, destroy: () => callbacks.slice().forEach(cb => cb()) };
+};
+
+interface Harness {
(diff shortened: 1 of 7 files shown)


mairas:

Two lenses (behavior-identity vs `main`; layer correctness):

**Behavior-identity: zero findings — "BEHAVIOR-IDENTICAL".** All six flagged risk areas verified equivalent between the pre-refactor trio and the layered wiring:
- The **construction reorder is inert** — the scheduler's field-init initial drain writes only store signals, `_trackedDevices` is empty in both old and new (the config effect is deferred; Angular effects never run synchronously at registration), `live$` is a plain rxjs `Subject` (no synchronous emission on subscribe), and `store` is declared before `scheduler` so `onFlush → store.processBatch` resolves.
- `parseUpdate` value/state extraction + `id::key` dedup key byte-identical; `resolveRenderSnapshot` reproduces the old `requestRender` readiness guard + explicit-or-default shape exactly; the derive seam fires per-changed-key after `applyValue` (alternator power, inverter dcPower, ac none); the delegating aliases preserve reactivity; reproject byte-identical.

**Layer correctness: no new production defect** — the named edge cases (first-live-immediate-while-timer-armed unreachable via monotonic `initialPaintDone`; requestRender-during-draw; teardown races; the shared copy-on-write `changed` flag; generic `TSnapshot` spreads) all resolve safely. Two **spec-quality** findings, both addressed:
- *[low]* teardown test was a **false-green** for render-frame cancellation (`pendingRender=null` masked a missing `cancelAnimationFrame`) → **hardened** to spy on `clearTimeout`/`cancelAnimationFrame` and assert both fire on teardown.
- *[nit]* store spec omitted fan-out + seed fidelity (the edges 2b/2d build on) → **added** a one-id→multiple-device-keys test asserting `source`/`deviceKey` seeding.

Both folded into the commit; specs re-run green.

PR #357, opened 2026-07-19
refactor(electrical): solar-charger adopts the shared ingest scheduler (#4 Stage 2c)

mairas:
## Why

Stage 2c of #4. Per-body verification showed **solar-charger is a Layer-0-only consumer, not a Layer-1 one** (correcting the v2 plan's optimism): its flush routing `hasTracked ? idToTracked.get(id) ?? [] : [update.id]` is *stricter* than the shared Layer-1 store's `[entry.id]` fallback (the trio creates an invisible plain-id for untracked ids when config exists; solar drops them), and its dual-power derive commits differently. Forcing solar onto Layer 1 would change its store-state behavior — so it keeps its own store/reproject/flush and adopts only the timing primitive, exactly like bms.

## What

solar's duplicated timing machinery (pending buffer, 500 ms coalescing timer, initial-paint drain, rAF single-flight, subscription + teardown) → an `ElectricalIngestScheduler`. Everything else stays in-component:

- `flushPendingPathUpdates()` → `processBatch(entries)` (scheduler owns draining); the body — `trackDiscoveredSolar`, the `chargersByKey.update` with the `?? []` routing, the battery+panel dual-power derive with the `derivedChanged` gate — is byte-for-byte unchanged, wired as `onFlush`.
- `parseSolarPath` (with its Renogy metric-key remap) drives the `parseUpdate` seam — the remap is preserved.
- `reprojectSnapshotsToDeviceKeys`, `applyChargerValue`, `resolvePowerValue`, `buildRenderSnapshot`, `render`, and `applyConfig` (solar keeps its own reproject) are untouched.
- Deleted: the ingest fields, `enqueuePathUpdate`, `requestRender`, the constructor subscription block, `ngOnDestroy`.

## One spec change (called out)

The solar spec spied on the private `flushPendingPathUpdates` by name; renaming to `processBatch` retargets that one spy — all behavioral assertions unchanged (matches the bms precedent in #356).

## Verification

`npm run snc` + `npm run lint` clean; the six electrical widget specs + two layer specs (99 tests) pass. Net ~56 lines removed from solar.

Stage 2c of #4. Next: **2d charger** — the source-aware seam, the highest-risk stage, with a dedicated adversarial review.

### DIFF
--- a/src/app/widgets/widget-solar-charger/widget-solar-charger.component.spec.ts
+++ b/src/app/widgets/widget-solar-charger/widget-solar-charger.component.spec.ts
@@ -101,7 +101,7 @@ describe('WidgetSolarChargerComponent', () => {
       expect(dataServiceMock.subscribePathTreeWithInitial).toHaveBeenCalledWith('self.electrical.solar.*');
       expect((component as unknown as { discoveredSolarIds: () => string[] }).discoveredSolarIds()).toEqual(['sc1', 'sc2']);
 
-      const flushSpy = vi.spyOn(component as unknown as { flushPendingPathUpdates: () => void }, 'flushPendingPathUpdates');
+      const flushSpy = vi.spyOn(component as unknown as { processBatch: () => void }, 'processBatch');
       flushSpy.mockClear();
 
       liveSubject.next(makeUpdate('self.electrical.solar.sc3.voltage', 14.5));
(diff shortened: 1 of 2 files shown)


mairas:
## Adversarial behavior-identity review — clean

**Zero findings — behavior-identical.** All checkpoints pass: (1) `parseUpdate` reproduces the old `enqueuePathUpdate` extraction; `parseSolarPath` + the Renogy metric-key remap (controller.* strip; solar.*→panel*; charge.status→controllerMode) are byte-identical and untouched. (2) `processBatch` = old `flushPendingPathUpdates` minus the size-guard/drain (lifted into the scheduler, preserving insertion-order + latest-value-per-key), with the `?? []` routing, dual-power derive (battery + panel via `resolvePowerValue` + `derivedChanged`), and the `!changed && !derivedChanged` gate byte-for-byte. (3) reproject, `applyChargerValue`, `resolvePowerValue`, the visible computeds, render, applyConfig all untouched. (4) `resolveRenderSnapshot` mirrors the old `requestRender` readiness (`!svg`→null, else `explicit ?? buildRenderSnapshot()`). (5) Construction order correct; the field-init-before-effect-registration reorder is inert (effects don't run at registration). Mergeable.

PR #366, opened 2026-07-19
feat(units): server-driven display units — the flip (#347 Phase 2 Step 2b)

mairas:
## What & why

**Step 2b of #347** — the units flip. Display-path units stop being widget-owned and follow the SK server's per-path `displayUnits` preference (→ Skip group default → raw), with the per-path unit picker retired. Structural paths (`showConvertUnitTo===false` — dials, coordinates, durations, percent) are unchanged. Builds on the Step 1 resolver, the Step 2a structural markers, and the Step 2b-1 `convertBetweenMeasures` primitive.

The one invariant defended throughout: **the measure used to convert a value and the measure used to label it are always the same**.

## How

- **Streams directive** (`b9d23f42`) resolves each numeric path's measure by the structural/display split. Structural → `convertUnitTo` verbatim (byte-identical to before). Display → `resolvePathMeasure`, applied reactively: `combineLatest([sampledValue$, getPathMetaObservable tick])` re-emits the last value when `displayUnits` meta lands *after* subscribe, and tears down with the pipeline (no separate meta-subscription bookkeeping). Each update is tagged with `IPathData.measure` so widgets derive their symbol from the **same source as the value**.
- **Widgets** (`b08b8e19`) read the tagged measure for their unit symbol: widget-numeric (symbol + suppression + bitmap cache; lat/lon, duration and percent **format** branches also follow the measure), minichart, windsteer, racesteer (its three hard-coded ` kts` SVG labels became bound units; dead unit signals retired), racer-line (feet→′ kept; dts colour thresholds converted to the resolved unit), and the self-converting charts (data-chart pulled forward, history dialog mirrored). A boot placeholder (empty measure → no symbol) avoids painting a wrong unit before meta arrives.
- **Gauges**: ng-radial, ng-linear, steel and simple-linear move value, symbol, zones, `displayScale` bounds and the value-clamp to the resolved measure in lock-step; bounds are re-expressed from the stored unit via `convertBetweenMeasures` (identity until the measure resolves → no boot scale jump).
- **Picker** (`69292ced`): the per-path "Display Format" picker is removed (display units are server-driven; the `convertUnitTo` field stays for structural paths and as the gauges' bounds reference).

## Verification

- `npm run snc` · `npm run lint` · **full suite 1440 tests** · `test:mcp-schema` (no schema drift) · config specs — all green.
- **Adversarial review** of the diff (invariant / gauge-scale / structural-safety lenses) confirmed the value-side plumbing, `convertBetweenMeasures` direction, boot no-jump, and that structural paths are untouched — and caught three real defects the unit suite missed, fixed in `ddfb377c`: a widget-numeric format/label divergence (incl. a `toFixed`-on-string TypeError), a steel-gauge zone-band desync (high), and a simple-linear boot-band gap.

## Scope / follow-ups (not this PR)

Per the design review, this is **full retirement** for per-path display paths (the override model was considered and rejected). Deferred, recorded on #347: the ~14 remaining top-level `convertUnitTo` pickers (classify per widget), Step 3 (electrical temperature → resolver), Step 4 (retire `unitDefaults` + the Units screen), Step 5 (dead-field sweep). Known accepted limitations: an untyped path (no SK `meta.units`) shows raw (Skip can't know its unit); a runtime server-pref change converges on reload (boot convergence is guaranteed).

**On-device visual verification pending before merge** (set a non-default server unit preference; confirm value + symbol + gauge needle/scale move together with no flash).

Refs #347.


mairas:
## Code review — 8-persona pass (correctness, maintainability, testing, adversarial, api-contract, project-standards)

The core flip (invariant, gauge-scale direction, boot no-jump, structural-path safety) was already adversarially verified and is confirmed sound again here. project-standards: clean. Findings to address before merge (per our merge-findings gate — addressing all valid, non-nitpick ones):

**Correctness**
1. **[MED, corroborated ×2] data-chart late-meta reactivity gap** — `pathSignature` bakes a non-reactive `resolvePathMeasure(datachartPath)` and never rebuilds on a meta delta, so the history backfill and the live tail (+ title) can settle in different units, or stay `unitless` on a cold-boot race. Fix: fold `getPathMetaObservable(datachartPath)` into the rebuild trigger so the chart follows the resolved measure like the tiles do.
2. **[LOW, corroborated ×2] racesteer shared `speedUnit`** — targetVMG/VMG/vmgToWaypoint all write one symbol signal (last-writer-wins); if they resolve to different measures the symbol mislabels. Fix: per-path measure signals, matching `widget-windsteer`.
3. **[LOW] gauge first-value-before-meta clamp** — a `'unitless'`-tagged pre-meta value is clamped against real-unit bounds, pegging the needle for one frame. Fix: treat `'unitless'` as unresolved (fall back to the stored unit) for the clamp.

**Maintainability**
4. **[MED] dead code** — `dataset-chart-options` `unitList` (+ its `UnitsService`/imports) is write-only after the picker removal; `path-control-config` `MatOptgroup` import is now unused. Remove both.
5. **[MED] duplication** — history dialog `findPathConfig` reimplements `resolveConvertUnitTo`'s traversal, diverging on the falsy-`convertUnitTo` case. Express one in terms of the other.
6. **[MED] coupling** — the structural-vs-display rule (`showConvertUnitTo === false`) is now encoded in both the streams directive and the history dialog; extract a shared resolver so a tile and its history popup can't diverge.

**Testing**
7. **[MED] the three adversarial-fix defects (numeric format-branch crash, gauge-steel band rebuild, simple-linear boot fallback) + the new gauge `convertBetweenMeasures` logic shipped with no regression tests.** Add reproducing specs — especially the numeric format-measure crash guard and the gauge-steel `ngOnChanges` rebuild.

Addressing these now; will push and note here.

mairas:
## Findings addressed (commit 5832791f)

All valid findings resolved; full suite green (159 files / 1455 tests) + snc + lint + mcp-schema.

- **data-chart late-meta reactivity** — folded the path's meta subject into a `pathMeasure` signal that feeds `pathSignature`, so the chart rebuilds on a measure change like the tiles do.
- **racesteer shared symbol** — split into per-path measure signals + two SVG unit inputs, matching `widget-windsteer`.
- **gauge boot peg / raw-SI flash** — fixed at the source instead of the (inert) widget guard the review suggested: the **streams directive falls back to the widget's stored unit until `resolvePathMeasure` yields a real measure**, so a pre-meta display value is converted in a unit matching its scale + label. This also resolves the earlier raw-SI boot-flash concern. (Credit to the fix agent for proving the widget-side guard was numerically inert against `convertBetweenMeasures`' cross-group identity.)
- **dead code** — removed the orphaned `unitList`/`UnitsService` in dataset-chart-options and the unused `MatOptgroup` import.
- **duplication** — `resolveConvertUnitTo` now delegates to `findPathConfig`.
- **testing** — added regression specs for the three earlier adversarial-fix defects that had none (numeric format-branch crash guard, gauge-steel section rebuild, ng-radial reinterpretation) + the streams stored-unit fallback.

**Deferred (tracked, not a defect):** extracting the structural-vs-display rule into one shared helper used by both the streams directive and the history dialog (finding #6, coupling) — both encode it correctly today; doing it as a fast-follow to avoid churning the streams-directive linchpin under this PR.

PR #369, opened 2026-07-20
Retire client-side unit ownership; server is the sole unit source (#347)

### LINKED ISSUE #347: Units API Phase 2: retire client-side unit ownership in favour of the Signal K unit-preferences API

mairas:
## What & why

Completes #347's remaining Phase-2 scope: **all unit management now belongs to the Signal K server; Skip no longer owns, configures, or falls back for units.** The server's per-path `meta.displayUnits` is the single source of every path's display unit and symbol. This is the settled product decision recorded in #347/#246.

Phase 2a-b (per-widget display paths) already landed in #366. This PR removes the last client-side unit machinery.

## Approach (5 atomic commits)

1. **Electrical temperature → server** — the 5 electrical widgets (charger, alternator, inverter, bms, solar-charger) converted and labelled their internal temperature off the group-global `unitDefaults.Temperature`, ignoring the server's per-path preference. Now they resolve via `resolvePathMeasure(<temperature path>)` like every other display path, single-sourcing value and symbol from one resolve. Store raw Kelvin at ingest; resolve + convert + symbol at format time. Solar threads the real SK path (its keys are non-reconstructable).
2. **Retire client-side unit ownership** — delete `unitDefaults` / `DefaultUnitsConfig` / `getDefaults()` / `_defaultUnits`, the `options/units` settings screen, the `IAppConfig.unitDefaults` field, and the `Array<IUnitDefaults>` storage path. `getConversionsForPath` now returns the server measure when covered, else `'unitless'`. **No config-version bump** — the orphaned key self-heals on the next whole-app write, avoiding the bump cascade.
3. **Remove the last top-level unit pickers** — the widget-config "Display Units" mat-select and the two widgets that surfaced it (racer-line, data-chart), plus the vestigial `convertUnitTo` wiring in the chart-options children. `IWidgetSvcConfig.convertUnitTo` is kept (history-dialog fallback). Schema regenerated.
4. **Reactivity fix** — the electrical temperature measure now repaints when a `displayUnits` meta lands late or changes at runtime, via a `watchMeta`/`onMetaChange` seam on the shared ingest scheduler (mirrors how the streams directive folds `resolvePathMeasure` through `getPathMetaObservable`). Includes hardened specs: faithful Kelvin→measure conversion and a solar path-threading assertion (previously test-invisible).
5. **Docs** — correct two stale unit-fallback comments.

The conversion engine (`convertToUnit`, `convertBetweenMeasures`, `resolvePathMeasure`, `_conversionList`, `getUnitDisplaySymbol`) is kept purely as the mechanism that applies the server's conversion. **No server-covered path changes behaviour** (`serverDefault ?? …` short-circuits). On a server with no units-preference plugin, values render in SI base — Skip substitutes nothing, by design.

## Verification

- `npm run snc` + `npm run lint` + `npm run test:mcp-schema` green; full suite **1454 passed**.
- Multi-persona adversarial review (5 personas → per-finding refutation): 7 findings confirmed and all addressed in this PR (the reactivity gap, the K-conversion / solar-path test-invisibility, charger single-source coverage, two stale docs); 1 finding refuted.
- **On-device check pending:** solar/electrical temperature rendering against the live boat (the path-threading is now unit-tested, but the live REST-meta wiring is the final visual confirmation).

Net: 47 files, ~+180 / −480.

Closes #347
Refs #246


mairas:
## Code review synthesis (multi-persona, adversarially verified)

5 persona reviewers (correctness, adversarial, testing, maintainability, project-standards) over the diff → each finding independently re-verified by a refutation pass. **7 confirmed, 1 refuted.** All 7 confirmed are addressed in this PR (commits 4–5); recorded here for traceability.

| # | Sev | Finding | Resolution |
|---|-----|---------|------------|
| 1 | P2 | Electrical temperature resolves the measure inside a value-only computed, so a late/changed `displayUnits` meta doesn't repaint until the next value tick (diverges from the streams-directive/data-chart `getPathMetaObservable` fold). | Fixed — `watchMeta`/`onMetaChange` seam on the shared scheduler + per-widget `metaVersion` signal (commit 4). |
| 2 | P2 | Solar temperature path threading (the highest-risk change) is test-invisible — mocks stub `resolvePathMeasure: () => 'celsius'`, ignoring the path. | Fixed — solar spec records the path arg and asserts the real charger-temp vs panel-temp paths (commit 4). |
| 3 | P2/P3 | Kelvin→display conversion is test-invisible — identity `convertToUnit` mock; asserted `0.0 °C` is a value production never renders (0 K = −273 °C). | Fixed — faithful K→measure mock + Kelvin-fed values in charger/solar specs (commit 4). |
| 4 | P3 | Charger's new `temperatureUnit`/`temperatureText` single-source fields unasserted. | Fixed — charger spec asserts both from a Kelvin feed (commit 4). |
| 5 | P3 | `units.service.ts:776` comment still says "Skip's group default"; contradicts its updated sibling. | Fixed (commit 5). |
| 6 | P3 | `getConversions()` JSDoc describes the deleted Units picker UI. | Fixed (commit 5). |
| 7 | — | **Refuted:** removing `setPathUnits()` leaves a stale `convertUnitTo` a future consumer might mis-apply. | Self-admittedly inert; no current consumer reads it; contradicts the intended "keep field / self-heal" decision. Dropped. |

Full suite 1454 passed; `snc` + `lint` + `test:mcp-schema` green.

Issue #11, opened 2026-06-30
[ARCH-03] TypeScript strict mode disabled project-wide (duplicate of TS-01)


mairas:
**Severity:** medium | **Confidence:** confirmed | **Area:** architecture | **Source:** codebase review 2026-06-30, finding ARCH-03

This finding describes the same root cause as TS-01: `tsconfig.json` has no `strict` flag (nor `strictNullChecks`, `noImplicitAny`, `strictFunctionTypes`, etc.), leaving core TypeScript null- and type-safety checking disabled across the entire ~299-file codebase. The architectural dimension noted here is that there is also only one tsconfig (no app/spec split) and `module: es2020` is dated relative to the Angular 21 toolchain — but the underlying issue and fix are identical to TS-01.

Tracked together with TS-01 - see #6.

## Affected code

- `tsconfig.json:1-25`

mairas:
Closing as duplicate of #6 (TS-01). Both describe TypeScript strict mode being off project-wide; #6 is the canonical tracking issue and carries the strictNullChecks ratchet plan.

The two notes unique to this issue don't stand alone: the 'single tsconfig, no app/spec split' claim is false at HEAD (`src/tsconfig.app.json` and `src/tsconfig.spec.json` both exist), and 'module: es2020 is dated' is not a type-safety defect. Folded into #6.

PR #370, opened 2026-07-20
Hotkeys: bare keys with focus guards, and Esc to cancel edit

### LINKED ISSUE #191: Hotkeys: adopt bare keys with focus guards (post-navigation-chrome keymap)
Skip's hotkeys were inherited wholesale from upstream Kip, built around swipe-in Actions/Notification sidenavs. The `navigation-chrome` redesign (#183) removed those sidenavs. This issue audited the resulting keymap and decides the scheme we actually want.

## Status: decided

The audit is complete (full write-up in the K1 comment below). **The code, iframe injector, and docs on `main` are already mutually consistent** — #183 consolidated registration into `AppComponent`, deleted `MenuActionsComponent`, and reconciled the injector and the two doc tables. So this issue no longer tracks a reconciliation of drifted layers (the original tables here described the pre-#183 state and were stale). It tracks a **decided change of scheme**.

## Decision

Adopt **bare keys guarded by focus** (the audit's Candidate B), bare arrows included, replacing the current `Ctrl` / `Ctrl+Shift` scheme.

Why bare keys: they collide with **no** browser-chrome shortcut, which is the current scheme's only *unfixable* defect. `Ctrl+Shift+N` (night mode) is consumed by the browser as *New Incognito Window* before the page sees it and **cannot** be reclaimed with `preventDefault`; `Ctrl+Shift+E` opens Firefox's Network panel. Bare keys sidestep the entire `Ctrl+Shift+letter` minefield. The MFD also runs kiosk/fullscreen where a modifier buys nothing.

Per-action **semantics are preserved** — only the key scheme changes and focus guards are added.

## Target keymap

| Key | Action | Semantics |
|---|---|---|
| `←` / `→` | Previous / next dashboard page | Unchanged gating: fires only on a **visible, locked** dashboard, not mid-drag |
| `e` | Enter edit / layout mode | **Enter-only** (unchanged); exit via the on-screen Save / Cancel controls |
| `f` | Toggle fullscreen | Unchanged (no-op when embedded in a host iframe or unsupported) |
| `n` | Toggle night mode | Unchanged |
| `Esc` | Cancel an in-progress layout edit | Discards changes and re-locks; yields to an open dialog/menu (its Esc wins). No-op when not editing. |

**Edit is enter-only, `Esc` is the exit.** `e` enters edit and does not toggle — a bare-key toggle-off is ambiguous (save vs. discard). Instead `Esc` provides the keyboard exit (discard + re-lock), the counterpart of the on-screen **Cancel** button; **Save** stays on-screen only, so no single key can ambiguously commit-or-discard. No other new actions are added; Candidate A (dropping E/F/N entirely) was not taken because keyboard parity is wanted.

## Focus guard — the load-bearing part

Bare keys fire on every keydown, so a guard is mandatory. A single predicate — *"is this key event targeting a control that legitimately consumes keystrokes?"* — suppresses the hotkey when focus is on an editable/interactive element: text `input`, `textarea`, `select`, `contenteditable`, and ARIA / Material widgets that navigate with keys (listbox, menu, combobox, slider, spinbutton, radiogroup, tab; an open `mat-select`). One predicate covers both cases — letters (stop `n` firing while typing a widget label) and arrows (yield arrows to a focused slider/select or embedded app). Without it, bare keys are unsafe.

`preventDefault` is called **only when a hotkey is actually consumed** (guard passed, key matched, action taken), so bare arrows don't also scroll — while arrows on a non-dashboard route keep their native behavior.

## The iframe boundary

The injector (`iframe-inputs-inject.utils.ts`) forwards keydowns out of embed widgets (freeboardsk, anchor-alarm) to the parent, which re-dispatches them on `document`. Because the parent then sees `document` as the target (always safe), **the in-iframe guard is the only protection for the embed path** — it must forward the bare set *and* apply the same focus guard against the iframe's own focus, or arrows meant to pan an embedded plotter get stolen for page-nav. The guard therefore exists in two places by necessity (the injector is a serialized string and cannot import the parent predicate); they are co-located and tested against the same cases.

## Implementation units

- **Focus-guard predicate** (new small util): the interactive-target rule; consumed by the app handler.
- **`app.component.ts`**: register the bare key set (drop the `ctrlKey` filter); the handler drops modifier gating, applies the guard, and calls `preventDefault` on consume; pageNav gating and per-action semantics preserved.
- **`iframe-inputs-inject.utils.ts`**: forward the bare set, gated by an in-iframe mirror of the predicate, `preventDefault` on forward. Derive the key set from **one shared constant** so handler and injector can't drift (the audit's "single source of truth" fix).
- **Re-dispatch widget handlers** (`widget-iframe`, `widget-freeboardsk`, `widget-anchor-alarm`, `widget-hoekens-anchor-alarm`): they re-dispatch whatever the injector forwards, so no per-key change is expected — the injector filter is the contract. Verify only.
- **`dashboard.component.ts`**: `Esc` (`@HostListener('document:keydown.escape')`) cancels an in-progress edit via the existing `cancelLayoutChanges()`. Lives here, not in the global listener, because cancel is component-local (reloads the gridstack instance). Guarded on edit-mode and on no `.cdk-overlay-pane` being open (so an overlay's own `Esc` wins). `Esc` is intentionally kept out of the shared `HOTKEY_KEYS` / injector set — an embedded app's `Esc` must not cancel Skip's edit.
- **Docs**: the `welcome.md` shortcuts table and the `upgrade-config.component.html` input-modes table → bare keys. `embedwidget.md` is generic; no change.

## Verification criteria

- Bare `f` / `n` toggle fullscreen / night; bare `e` enters edit; bare `←` / `→` page-nav only on a visible, locked dashboard; bare `Esc` cancels an edit, is a no-op when locked, and yields when an overlay is open.
- Every hotkey is suppressed when an `input` / `textarea` / `select` / `contenteditable` / listbox / slider is focused; `preventDefault` fires on consume and not on suppression.
- The injector forwards the bare set and is suppressed when the iframe's own control is focused.
- Handler and injector key sets originate from a single constant (no drift).
- Both doc tables match the keymap.

## Out of scope

The full browser-conflict analysis, macOS/kiosk caveats, and per-action wiring references are in the K1 audit comment below. `Enter` / `Space` activation on focusable controls is accessibility, not part of this keymap.

mairas:
## Why

The hotkeys inherited from upstream Kip were built around swipe-in sidenavs that the `navigation-chrome` redesign (#183) removed. #183 already reconciled the code, iframe injector, and docs (the original #191 findings were stale) — but it left the `Ctrl` / `Ctrl+Shift` scheme, whose one *unfixable* defect is browser-chrome collision: `Ctrl+Shift+N` is consumed by the browser as **New Incognito Window** before the page sees it and cannot be reclaimed with `preventDefault`; `Ctrl+Shift+E` opens Firefox's Network panel.

## What

Switch the keymap to **bare keys guarded by focus**, which collide with no browser shortcut. Per-action semantics are preserved; only the key scheme changes and focus guards are added.

| Key | Action |
|---|---|
| `←` / `→` | Prev / next dashboard page (unchanged gating: visible + locked + not dragging) |
| `e` | Enter edit mode (enter-only) |
| `f` | Toggle fullscreen |
| `n` | Toggle night mode |
| `Esc` | Cancel an in-progress layout edit (discard + re-lock) — the keyboard exit for enter-only `e` |

Highlights:

- **Single source for the keymap** — `HOTKEY_KEYS` + a shared `INTERACTIVE_TARGET_SELECTOR` in a new `hotkey-target.util.ts`; the app listener lowercases the set and the iframe injector JSON-interpolates both, so handler / injector / guard can't drift.
- **Focus guard** (`isInteractiveKeyTarget`) suppresses on text entry and arrow-navigating ARIA/Material widgets, so typing "n" never toggles night mode and arrows adjust a focused slider instead of paging. Mirrored inside the iframe injector — the *only* protection for the embed path, since a re-dispatched key reaches the parent with a safe (`document`) target.
- **Bare-only + preventDefault on consume** — any modifier bails (so `Ctrl+E` etc. still reach the browser); arrows only `preventDefault` when a page actually changes, so they still scroll off-dashboard.
- **`Esc` cancels an edit** via `@HostListener` in `DashboardComponent` (cancel is component-local — it reloads the gridstack instance). Guarded so an open dialog/menu (`.cdk-overlay-pane`) keeps its own `Esc`; kept out of the shared/injector key set so an embedded app's `Esc` never cancels Skip's edit.
- Docs (`welcome.md`, `upgrade-config.component.html`) reconciled to the bare keymap.

## Verification

`npm test` (1466 tests / 160 files), `npm run snc`, `npm run lint` all green. TDD throughout. **Not yet driven end-to-end in a real browser** — the live keypress behavior (and overlay-Esc precedence) is best confirmed on the boat MFD; happy to add a headless-browser check if wanted.

Closes #191.

### DIFF
--- a/src/app/app.component.spec.ts
+++ b/src/app/app.component.spec.ts
@@ -19,10 +19,20 @@ import { ToolbarComponent } from './core/components/toolbar/toolbar.component';
 import { IConfig } from './core/interfaces/app-settings.interfaces';
 import { LATEST_APP_CONFIG_VERSION } from './core/constants/config-versions.const';
 
+// A minimal stand-in for the KeyboardEvent fields the handler reads.
+interface HotkeyTestEvent {
+  ctrlKey?: boolean;
+  altKey?: boolean;
+  metaKey?: boolean;
+  shiftKey?: boolean;
+  target?: EventTarget | null;
+  preventDefault?: ReturnType<typeof vi.fn>;
+}
+
 // Access the component's private surface without widening the class API.
 interface AppComponentHotkeyApi {
   dashboardVisible: { set: (v: boolean) => void };
-  handleKeyDown: (key: string, event: { shiftKey: boolean }) => void;
+  handleKeyDown: (key: string, event: HotkeyTestEvent) => void;
   onShellPointerDown: (event: { target: { closest: (sel: string) => unknown } }) => void;
 }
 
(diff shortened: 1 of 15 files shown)


mairas:
# Code review — bare-key hotkeys + Esc-cancel (PR #370)

**Scope:** `949bfcc8..HEAD`, 10 files (~120 non-test LOC). **Intent:** replace `Ctrl`/`Ctrl+Shift` global hotkeys with bare keys guarded by focus; add `Esc`-cancels-edit.
**Team (7):** correctness, testing, maintainability, project-standards (always-on) + security, api-contract, adversarial.

The core mechanics are sound (key-filter matching, modifier bail, focus guard, `pageNav` boolean→`preventDefault`, `as const`/typing, null-safety). **Security is clean** — the bare-key/relaxed-payload change is security-neutral; the unforgeable `event.source === iframe.contentWindow` + `instanceId` gate (unchanged) already confines the synthetic-keydown path to a widget's own embedded content, which could reach the same parent actions under the old `Ctrl+Shift` contract. **api-contract is clean** — all four internal contract changes (`pageNav` return, payload trim, listener options, new exports) degrade safely in every traced consumer.

The findings below cluster around one theme the review surfaced: **the two new keyboard entry points (global `handleKeyDown` and the `Esc` `@HostListener`) disagree about *when* a bare key should fire** — one over-guards, the other under-guards — plus a real regression on the embed path.

## P1 — High

| 1 | `iframe-inputs-inject.utils.ts:68` | **Embed key hijack.** The injected forwarder now forwards + `preventDefault`s **bare** `←`/`→`/`e`/`f`/`n` from inside embeds. The old script forwarded only `Ctrl+Arrow` / `Ctrl+Shift+EFN`, leaving bare keys to the embed. So an embedded app that uses bare arrows (freeboard-sk / OpenLayers `KeyboardPan`, focused map viewport — not matched by the interactive selector) loses arrow-pan and the dashboard slides instead. Asymmetric: `ArrowUp`/`ArrowDown` aren't in the set, so vertical pan still works. **Needs a design decision (see below).** | adversarial, correctness | 0.72 | manual |

## P2 — Moderate

| 2 | `app.component.ts` `handleKeyDown` | Bare `e`/`f`/`n` fire on **non-dashboard routes**. `e`→`setStaticDashboard(false)` has no `dashboardVisible()` gate (arrows do, via `pageNav`); navigation never resets `isDashboardStatic`, so a stray `e` on Settings/Help lands you silently in edit mode on return. Bare keys make the pre-existing latent case easy to trigger. | adversarial, correctness | 0.80 | gated_auto |
| 3 | `app.component.ts` `handleKeyDown` | Hotkeys **leak behind an open modal**. Frame dialogs open with `autoFocus:false`, leaving focus on a `<button>` (not in the interactive selector), so bare `n` toggles night / arrows slide the dashboard *behind* the dialog. The new `Esc` handler guards overlays; this handler doesn't — inconsistent. | adversarial | 0.74 | gated_auto |
| 4 | `dashboard.component.ts:383` | `Esc`-cancel **over-yields**: it bails on *any* `.cdk-overlay-pane`, but snackbars/toasts/tooltips render a pane and don't close on `Esc`. During edit-mode add-widget flows Skip shows duration-0 toasts → while one is up, `Esc` finds a pane, returns early, and neither cancels nor dismisses anything (key looks dead). | correctness | 0.72 | gated_auto |
| 5 | `iframe-inputs-inject.utils.spec.ts` | The injected guard (bare filter, focus suppression, `preventDefault`) is verified **only by source-substring matching**, never executed. A logic error inside the injected string passes green. Add a runtime spec: eval the IIFE in jsdom, spy `postMessage`, dispatch keydowns, assert forward/suppress/`preventDefault`. | testing | 0.85 | test_addition |

## P3 — Low

| 6 | `iframe-inputs-inject.utils.ts:72` | **Caps-Lock divergence.** Injector matches raw `event.key` case-sensitively; the app lowercases. Caps-Lock `E`/`F`/`N` (no Shift) fires in the app but not inside an embed. Fix: lowercase both sides. | correctness, api-contract | 0.80 | gated_auto |
| 7 | `dashboard.component.ts:380` | `Esc` cancels a layout edit **mid gridstack drag** (no `isDragging` guard, which `pageNav` has). `loadDashboard` replaces the dragged nodes; `dragstop` may not fire → `isDragging` stuck true → arrow nav silently disabled for the session. | adversarial | 0.60 | gated_auto |
| 8 | `dashboard.component.spec.ts`, `app.component.spec.ts` | `Esc` `@HostListener` binding and the changed `addHotkeyListener` registration are bypassed by direct method calls → a wiring regression (removed decorator, stale `ctrlKey:true`, bad lowercasing) passes green. Add a real `dispatchEvent` test each. | testing | 0.70 | test_addition |
| 9 | `welcome.md`, `upgrade-config.component.html` | New `Esc` shortcut is missing from both keyboard-doc tables, despite the new "single keys with no modifier" note that now implicitly over-claims completeness. | project-standards | 0.62 | mechanical |
| 10 | `hotkey-target.util.ts:1` | "single source of truth for the keymap" comment overclaims — `Esc` is registered separately in `dashboard.component.ts` with no cross-reference. Narrow the comment + point at the `Esc` handler. | maintainability | 0.62 | mechanical |
| 11 | 4 widget receivers | Injector payload trimmed to `{key,instanceId}`, but `widget-iframe`/`-freeboardsk`/`-anchor-alarm`/`-hoekens-anchor-alarm` still destructure `ctrlKey`/`shiftKey` (now permanently `undefined`). Works (`undefined`→`false`) but reads as a stale contract. | maintainability, api-contract | 0.70 | manual |

## Coverage

- **Security:** no findings — change is security-neutral (see above).
- **Testing gaps (pre-existing):** no receiver-side spec asserts the `event.source`/`instanceId` trust boundary or the `undefined`→`false` payload defaulting at the postMessage seam.
- **Residual (maintainability):** the guard *predicate* is mirrored in the injector string; only the selector/key *data* is JSON-single-sourced + test-pinned. A third condition added to `isInteractiveKeyTarget` won't auto-propagate. `INTERACTIVE_TARGET_SELECTOR` is a hand-maintained ARIA list (acceptable — no standard "is-interactive" API).

## Verdict: **Ready with fixes** (one decision blocks)

- **#1 needs a design decision** before merge (embed keyboard forwarding vs. embeds owning their keys). This is the bare-key scheme's one genuine tension — the original `Ctrl` requirement existed *specifically* to avoid clashing with the embedded app.
- **#2–#4** are real behavior gaps and should be fixed (the fix for #3 and #4 is the same: a shared, correct "is a blocking modal open" predicate used by both handlers — `.cdk-overlay-backdrop` excludes toasts/tooltips and includes dialogs/menus/selects).
- **#6–#11** are cheap fixes / test additions.
- All fixes to be applied as follow-up commits on this branch, then the changed scope re-reviewed and the suite re-run.

mairas:
## Review addressed

All findings resolved. History was recomposed into a clean 5-commit chain (the original "forward bare keys from embeds" work is folded away rather than added-then-removed), so the diff now tells the net story directly.

**Decision (finding #1):** embeds keep their own keyboard. The injector no longer forwards *any* keydowns to the parent (removed the forwarder + the four receivers' re-dispatch); bare keys belong to the embedded app (freeboard-sk arrow-pan is preserved), and swipe-to-page still works from a fullscreen embed. This also moots #5, #6, and #11.

| # | Finding | Resolution |
|---|---------|-----------|
| 1 | Embed key hijack | **Dropped embed keyboard forwarding entirely** (Option A). `embedwidget.md` updated. |
| 2 | Bare `e` fires off-route | `e` now gated on `dashboardVisible()`; `preventDefault` only when it acts. |
| 3 | Hotkeys leak behind a modal | New shared `isBlockingOverlayOpen()` (`.cdk-overlay-backdrop`) — `handleKeyDown` bails when a dialog/menu/select is open. |
| 4 | `Esc` over-yields to toasts | Same predicate: `Esc` now yields on `.cdk-overlay-backdrop` (dialogs/menus/selects) not `.cdk-overlay-pane`, so snackbars/tooltips no longer block it. |
| 5 | Injector guard only string-tested | Moot — no injector keydown path anymore. Spec now asserts swipe-only, no key forwarding. |
| 6 | Caps-Lock injector divergence | Moot — injector no longer matches keys. |
| 7 | `Esc` cancels mid-drag | `Esc` now bails on `_uiEvent.isDragging()` (matches `pageNav`). |
| 8 | Wiring untested | Added: real `document` `Escape` dispatch test (exercises the `@HostListener`), and an `addHotkeyListener` registration assertion. |
| 9 | `Esc` undocumented | Added a "Cancel page edit → Esc" row to both keyboard tables. |
| 10 | "single source" overclaim | Comment narrowed to the global shell hotkeys, with a pointer to the component-local `Esc` handler. |
| 11 | Dead `ctrlKey/shiftKey` reads | Removed from the four receivers along with the keydown path. |

**Security / api-contract:** no findings; unchanged by these fixes.

**Verification:** full suite green, `snc` + `lint` clean. Live on-device keypress verification on the boat MFD to follow before merge (per plan: after review, before merge).

Issue #386, opened 2026-07-21
Subscribe to remote (AIS/DSC) contexts only on widget demand


mairas:
Follow-up to the Connectivity-tab retirement (linked below), which makes AIS/DSC subscription **unconditional**.

## Problem

Unconditional `subscribe=all` makes SK push every remote vessel context even when nothing renders it, and Skip parses + buckets each delta in `DataService` for zero benefit. On the live boat that's ~10 targets (negligible), but AIS target count is **location-dependent** — hundreds in a busy strait/harbour. Subscribe to remote contexts only when a widget actually consumes them.

## Constraint (the crux)

Subscribe scope is a **WebSocket-handshake decision** — `?subscribe=all` vs `?subscribe=self` (`signalk-delta.service.ts:78-108`), set from `endpointStatus.subscribeAll`, chosen at boot in `initNetworkServices()` (`app-initNetwork.service.ts`) from the localStorage connection config. Dashboards load later from SK `applicationData` (post-auth), so **widget demand is not known at connect time.** That timing is the entire design tension.

## Demand set

Only Skip-side consumer of remote contexts is **`widget-ais-radar`** (+ its target dialog). Other "ais" grep hits (`video`, `solar-charger`, `boolean-control-layout`) are substring false positives; the Freeboard-SK embed runs its own SK connection and does not ride Skip's `subscribeAll`. The demand predicate is: "does any saved dashboard contain a remote-consuming widget type?"

## Two shapes (pick during planning)

- **Cheap, one-boot-stale:** recompute the demand boolean whenever dashboards are saved, persist it into the connection/bootstrap config, and consume it through the existing `subscribeAll` plumbing unchanged. New code is the demand predicate + a hook on the config-save path. Wart: adding an AIS widget takes effect on next reload — **no worse than the pre-retirement UX**, which already required enabling a setting and reconnecting.
- **Correct, no staleness:** boot `subscribe=self`; after dashboards load, if demand exists, reconnect the WS with `subscribe=all` using the existing `ConnectionStateMachine`/retry machinery. New bit: triggering a reconnect for a **scope change** rather than a failure. Small-to-moderate.

Neither needs a new service or a data-pipeline change.

## Verification

- Dashboards with no remote-consuming widget connect `subscribe=self` and receive no `vessels.*` contexts other than self (assert on `DataService` contents).
- Adding an AIS widget causes remote contexts to flow (after reload for the cheap shape; live for the correct shape).
- `npm run ci` green.

[...]

PR #387, opened 2026-07-21
Retire the Connectivity settings tab

### LINKED ISSUE #385: Retire the Connectivity settings tab (connection config is derivable/diagnostic)
## Problem

The **Connectivity** settings tab (`settings/signalk/`) holds no true user setting anymore. Grounded against the live boat server (`halpi.hurma`, SK 2.30.0):

- **Server URL** auto-derives from `window.location.origin` (`app-initNetwork.service.ts:396`, `settings.service.ts:576`). Skip is *served by* the SK server, so the origin is correct by construction. The editable field is vestigial from upstream KIP's connect-to-any-remote-server support, and pointing at a different origin would break the same-origin session cookie anyway.
- **"Signal K server is behind a Proxy"** is dead weight. SK's own endpoint discovery already returns the public Traefik host:
  ```
  signalk-http: https://halpi.hurma:4430/signalk/v1/api/
  signalk-ws:   wss://halpi.hurma:4430/signalk/v1/stream
  ```
  i.e. already same-origin. The `proxyEnabled=false` path (trust SK's URLs) is correct; the `true` path would produce an identical result. Always trust what SK reports.
- **AIS/DSC subscribe** (`signalKSubscribeAll`) is the only functional toggle, but it's a data-*demand* concern, not a connection setting. Measured live: 11 vessel contexts (~10 AIS targets) — negligible. The UI's "may generate heavy traffic" warning is unsupported at this scale; the real cost is that `SignalKDeltaService` + `DataService` parse and bucket remote deltas even when nothing renders them.
- **SSO identity block** just echoes SK's session — `AuthenticationService.refreshLoginStatus()` does `GET /skServer/loginStatus` `withCredentials`. Skip has no identity of its own; the `"HaLOS SSO"` name is SK's `oidcProviderName`. This is session *status* + an edge *action* (Sign in), not a setting.
- **Activity graph + version/status panel** are diagnostics, not settings.

## Decision

Retire the Connectivity tab as a settings tab. Its content is either **auto-derivable** (URL, proxy, AIS) or **status/diagnostics** (identity, graph, versions) — the latter moves to the **Actions** surface (`core/components/actions/`), which is already a tile hub (Settings / Remote Control / Help).

This is one logical change: removing the tab strands the sign-in action and read-only feedback unless the diagnostics move with it, so they ship together.

## Scope (v1)

- [ ] Remove the **Connectivity** tab from the settings tab group (`settings/tabs/tabs.component.html`). Remaining: Display, Notifications & Audio, Configurations.
- [ ] **AIS/DSC subscription becomes unconditional** — always connect with `subscribe=all`. Drop the checkbox. (Conditional-on-demand behavior is the follow-up below.)
- [ ] **Proxy**: always trust SK's discovered endpoints; drop the checkbox and stop honoring `proxyEnabled`.
- [ ] **Server URL**: keep auto-derivation from `window.location.origin`; remove the editable URL field + in-tab Connect flow. Verify no bootstrap/no-connection recovery path depends on manual URL entry (there was a stale no-connection redirect in `app-initNetwork`) — if one does, that recovery affordance must survive elsewhere or be deemed unnecessary given same-origin serving.
- [ ] **Relocate the connection status/diagnostics to Actions**: activity graph (delta/sec), Skip/browser/OS versions, server description, stream status, sign-in state, the read-only ("changes disabled on the server") warning, and internet reachability. As a new Actions tile routing to a connection/status view, or an inline panel — implementation's call.
- [ ] Remove the now-obsolete AIS-radar warning toast that tells users to enable the setting and reconnect (`widget-ais-radar.component.ts:265-270`), since AIS is always on.
- [ ] Update/prune specs referencing `SettingsSignalkComponent` and the tab.

## Config-schema decision

Keep the vestigial `IConnectionConfig` fields (`proxyEnabled`, `signalKSubscribeAll`) in place for v1 and simply stop reading them (hardcode behavior). Rationale: removing config fields triggers a `LATEST_APP_CONFIG_VERSION` bump + upgrade-service migration + perf-harness seed + spec churn (see the version-bump coupling we've hit before), and the follow-up **repurposes** `signalKSubscribeAll` to hold computed demand anyway. Field removal, if ever wanted, is a separate cleanup. This keeps v1 migration-free.

## Verification

- Settings shows 3 tabs; no Connectivity.
- With **no** AIS widget on any dashboard, the WS still connects `subscribe=all` and AIS targets populate (unconditional).
- AIS radar shows targets with no settings step and no toast.
- Sign-in state + read-only warning still reachable (now on Actions); Sign in still works.
- `npm run ci` green; a prod build.
- On-boat smoke: load Skip, confirm the connection/status view renders on Actions and the AIS radar populates.

## Follow-up

Conditional (widget-demand-driven) subscription is tracked separately — see the follow-up issue linked below.

Conditional subscription: #386.

mairas:
Closes #385.

## Why

The Signal K **Connectivity** settings tab held no true user setting anymore — verified against the live boat server (SK 2.30.0):

- **Server URL** auto-derives from `window.location.origin`; Skip is served *by* the SK server, so the origin is correct by construction.
- **"behind a Proxy"** is not a user choice: Skip is always served same-origin by SK and authenticates via the same-origin session cookie, so the connection must be same-origin. We hardcode the same-origin routing (see below) rather than expose a toggle.
- **AIS/DSC subscribe** is a data-*demand* concern, not a connection setting. Measured live: ~10 targets — negligible; the "heavy traffic" warning was unsupported.
- **SSO identity** and the **activity graph / versions** are session *status* and *diagnostics*, not settings.

## What

1. **Connection routing and subscription are now fixed, not user-configurable** — always serve the server's discovered API path from the app's own origin (`window.location.origin`), and always subscribe to all contexts. Forcing same-origin guarantees the session cookie applies and can't yield a cross-origin/mixed-content endpoint; on a correctly proxied deployment it's identical to trusting the reported URLs. The obsolete "AIS data is disabled" radar toast is removed.
2. **Connectivity tab removed** from Settings (remaining: Display, Notifications & Audio, Configurations).
3. **Connection status relocated to Actions** — the diagnostics component (SSO identity, server/stream state, versions, internet reachability, delta-throughput chart) moves to its own `connection-status` component behind a new **Connection** tile on the Actions page, mirroring Settings / Remote Control / Help.

The `IConnectionConfig` fields (`proxyEnabled`, `signalKSubscribeAll`) are retained but unread to avoid a config-version migration; #386 repurposes the subscription field for widget-demand gating.

## Review

Multi-persona review posted inline. Addressed on-branch: added tests locking the fixed routing/subscription args, the `/connection` route resolution, and the sign-in wiring; removed dead connect-form helpers (`validateSignalKUrl`, `setConnectionConfig`); signposted the retained-but-inert config fields. The endpoint-routing design point was resolved in favor of **forcing same-origin** (the more robust reading of "Skip is always served same-origin by SK").

## Verification

- All local gates green: lint, snc, prod build, unit (1503/1503), plugin (7/7), mcp-schema (31/31).
- Removing the editable URL field is safe: no-connection recovery is in-place/reload-based (no redirect, #190) and `validateSignalKUrl` was used only by this tab.

Still to do before merge: on-boat smoke test (load Skip → Actions → Connection renders; AIS radar populates with no settings step).

## Follow-up

#386 — subscribe to remote (AIS/DSC) contexts only on widget demand.


mairas:
## Code review — multi-persona (correctness, testing, maintainability, project-standards, adversarial)

**Scope:** `main..HEAD` (3 commits). **Mode:** interactive. **Verdict: Ready with fixes** — all findings are consequences of this change and are being addressed on-branch; one design decision is left for the author.

Reviewers were dispatched as independent Opus sub-agents. Findings below are merged, deduped, and confidence-gated; cross-reviewer agreement is noted.

### P2 — Moderate

| 1 | `app-initNetwork.service.spec.ts` | **Core contract untested** — nothing asserts `initializeConnection` is called with `(url, false, true)`. A regression re-reading `this.config.proxyEnabled`/`signalKSubscribeAll`, or flipping `subscribeAll` back to `false` (silently killing AIS/DSC), stays green. This is the one behavioral guarantee the PR exists to make. | testing, correctness, maintainability, adversarial | 0.95 | fix-here |
| 2 | `connection-status.component.spec.ts` | The moved component is now the **sole sign-in entry point**, yet its spec asserts only creation + the two status-line updates. Every SSO branch (signed-in / read-only / no-auth-required / not-signed-in + Sign-in button) and `signIn()→manualSignIn()` are uncovered. | testing, correctness, maintainability, project-standards, adversarial | 0.90 | fix-here |
| 3 | `app.routes.spec.ts` | New `/connection` route has **no resolution test**, unlike its siblings `/actions` and `/settings` (which assert they don't fall through to the `**`→`/page/0` wildcard). Sign-in reachability rests on this untested lazy route + guard. | testing, correctness | 0.85 | fix-here |
| 4 | `signalk-connection.service.ts:96` | `validateSignalKUrl` is **dead code** — its only caller (`connectToServer`) was deleted. Zero callers in src or specs. | maintainability, correctness | 0.92 | fix-here (remove) |
| 5 | `settings.service.ts:283` | `setConnectionConfig` is **dead code** — only caller was the removed connect flow. Leaving this public mutator invites re-introducing user-driven connection writes the design just retired. | maintainability | 0.85 | fix-here (remove) |
| 6 | `settings.service.ts:53`, `app-settings.interfaces.ts:7` | Retained `proxyEnabled`/`signalKSubscribeAll` are **inert but unmarked** — a dev sees two live public booleans with no signal that connection setup no longer reads them. The migration-avoidance tradeoff is fine; the missing signpost is the cost. | maintainability | 0.72 | fix-here (comment) |
| 7 | `app-initNetwork.service.ts:130` | **Design decision:** hardcoding `proxyEnabled=false` removes the same-origin rescue. A TLS-terminating proxy that reports `http://` endpoints (doesn't reflect the external origin) → mixed-content-blocked config writes + History API, with no UI recourse now that the checkbox is gone. **The live boat is NOT affected** (Traefik reflects `https://halpi.hurma:4430/…`); risk is other self-hosted proxy setups. | adversarial | 0.62 | author decision |

### P3 — Low

| 8 | `connection-state-machine.service.ts:237` | `shutdown()` now has zero callers (was called only by the removed restart-on-reconfigure step). | maintainability | 0.62 | remove |
| 9 | `settings.service.ts:279` | `getConnectionConfig` is production-orphaned but still exercised by the credential-purge spec — **keeping it** (a test relies on it). | maintainability, correctness | 0.65 | keep |

### Coverage / residual risks (report-only)

- **`subscribeAll=true` is now forced with no opt-out.** Default was `false`, so most existing users were on `?subscribe=self`; they now open `?subscribe=all`. On a Pi-class device in an AIS-dense area (hundreds of contexts) this is a broad CPU/bandwidth increase. This is the intended behavior for v1; **#386** gates it on widget demand. A perf-harness AIS before/after on the 10× target is advisable there.
- Removing the URL field ends pointing Skip at a different server (dev/`demo.signalk.org`). Intended (same-origin serving) and recoverable — the Configurations → Connection reset re-derives `signalKUrl` to `window.location.origin`.
- `processEndpointResponse` throws on a discovery body missing v1 endpoints (pre-existing); the Connection page is now the only place that error surfaces — worth confirming it renders legibly.

### Plan (this branch)

Addressing #1–#6 and #8 on-branch (3 tests, remove 2 dead methods + `shutdown()`, mark the inert fields). #7 is a design call surfaced to the author.

mairas:
### Review fixes applied (commits 40e93fad, 8ed65b63)

- **#1 core-contract test** — `app-initNetwork.service.spec.ts` now asserts `initializeConnection` is called with `({url}, false, true)` using stored flags set to the *opposite* values, so a regression that re-reads config or flips `subscribeAll` fails.
- **#3 route test** — `app.routes.spec.ts` asserts `/connection` resolves rather than hitting the `**`→`/page/0` wildcard.
- **#2 sign-in wiring** — `connection-status.component.spec.ts` asserts `signIn()` routes through `SsoRedirectService.manualSignIn`. (Rendering the full SSO branch would require faking `AuthenticationService`, which shares the root injector with the real `SignalKDeltaService` the component pulls in and would break its construction — per Skip's real-service test model — so the wiring is verified directly instead.)
- **#4/#5 dead code** — removed `validateSignalKUrl` and `setConnectionConfig` (only callers were in the deleted connect flow).
- **#6 signposting** — `proxyEnabled`/`signalKSubscribeAll` fields marked persisted-for-compatibility-only.
- **#8** `ConnectionStateMachine.shutdown()` — **kept**: it's a general connection-lifecycle method, not connect-form-specific; removing core infra is outside this PR's scope. **#9** `getConnectionConfig` — kept (a credential-purge spec relies on it).

Gates green after fixes: lint, snc, prod build, unit **1503/1503**.

**#7 (proxy hardcode) — awaiting author decision**, tracked in the discussion; not blocking the boat (Traefik reflects the external origin).

mairas:
Resolved review finding #7 (endpoint routing): adopted **force same-origin** — commit ee8ae8f5 now serves the server's discovered API path from `window.location.origin` (`proxyEnabled=true` behavior) instead of trusting the reported URLs. This guarantees the same-origin session cookie applies and can't produce a mixed-content endpoint; identical result on the boat, but robust against a server that reports a non-same-origin/`http://` endpoint. Test assertion flipped to `(url, true, true)` with stored flags set opposite. All findings now addressed on-branch.

Issue #390, opened 2026-07-21
Actions redesign · Phase 2 — toolbar hamburger menu; retire the /actions page


mairas:
Phase 2 of the Actions-surface redesign. Lands **after** Phase 1 (page management must already live in the edit-mode sheet before `/actions` can go).

## Scope

- **Hamburger app menu** off the toolbar `apps` button: a compact menu listing **Settings · Connection · Remote control · Help** (existing routes) plus a version/host footer. Repoint `openActions()` to open the menu instead of navigating to `/actions`.
- **Retire `/actions`**: remove `ActionsComponent` and the `/actions` route; drop the in-page `dashboards-editor` usage (now only in the Phase 1 sheet). Add an explicit `/actions → /page` redirect if bookmarks/deep-links matter (the `**` wildcard already catches unknown routes).
- **Docs**: update help content that references the Actions page or its tiles (welcome.md toolbar section; any "Actions" navigation paths in the help docs).

## Verification

- `apps` opens the menu; no `/actions` route resolves (or it redirects to `/page`).
- Each menu item routes correctly; Connection still reachable (moved from tile to menu item).
- No `ActionsComponent` / in-page `dashboards-editor` references remain; help docs match the new navigation.
- `npm run ci` green; on-boat smoke — menu opens, all four destinations reachable, page management via Edit intact.

Part of #388. Depends on Phase 1 (#389).

mairas:
Done — merged in #392 (toolbar hamburger menu, /actions retired, docs updated).

Issue #388, opened 2026-07-21
Redesign the Actions surface: hamburger menu + edit-mode page management


mairas:
## Problem

The `/actions` page (`core/components/actions/`) fuses two unrelated jobs behind one screen:

- an **app-navigation tile strip** — Settings · Connection · Remote control · Help, and
- a **page-layout tool** — `dashboards-editor` (add / drag-reorder / rename / duplicate / delete pages).

They share no task, so it reads as thrown-together. And page *navigation* and widget *editing* already live in the toolbar (`toolbar.component`: `page-nav-control` in the centre, the edit entry at the end), so the page manager is stranded away from where you actually work with pages. Page management also isn't Settings — it's layout.

Direction below was validated with an interactive prototype (design settled with the requester).

## Design (decisions)

Dissolve `/actions`; split its two halves by function.

**1. App navigation → a hamburger menu off the toolbar.** The toolbar `apps` button (today `openActions()` → `/actions`) instead opens a compact menu listing the existing routed destinations — **Settings · Connection · Remote control · Help** — with a version/host footer. The tile strip is retired. The Connection page added in #385/#387 survives unchanged — it moves from a tile to a menu item.

> **Superseded (2026-07-22):** page management is **not** in edit mode — it's a normal-mode **Manage pages** wrench next to the page dots. Edit mode is **Cancel · Done** only, with **no page dots** (page navigation is normal-mode only, so a mid-edit switch can't discard the unsaved layout). See the comments below and #389.

**2. Page management → Edit mode.** Skip's single toolbar already auto-hides (peek/reveal) and already gates editing through `DashboardService.isDashboardStatic` (the toolbar edit entry calls `setStaticDashboard(false)`). Recast Edit as an explicit mode: while the dashboard is non-static, the **same** toolbar renders edit contents instead of the normal nav controls —

- an **"Editing layout"** label (start),
- the page **dots** (centre, retained — you still switch pages while editing),
- **Manage pages**, **Cancel**, and **Done** (end), with **Done occupying the former edit/pen slot** — you tap that slot to enter, you tap it to leave; Cancel sits adjacent.
- The **peek strip renders in the accent colour while editing** — a persistent, non-obstructing "you're editing → reveal for Done/Cancel" cue.

Because it is the same auto-hiding toolbar, editing inherits the existing hide/reveal — no chrome sits pinned over the instruments. (This is what the prototype's second iteration fixed: a separate pinned edit banner was the wrong structure.)

[...]

PR #391, opened 2026-07-21
Actions redesign · Phase 1 — page management in Edit mode

mairas:
Phase 1 of the Actions-surface redesign (#388). **Additive** — `/actions` and its full-page `dashboards-editor` stay in place this phase; Phase 2 (#390/#392) removes them.

## What changed

**Page management is a normal-mode "Manage pages" wrench** next to the page dots (small spacer sets it apart) → opens a new `page-manager-bottom-sheet` hosting `dashboards-editor` in a new **compact** single-row layout (add / drag-reorder / rename / duplicate / delete). Page ops persist immediately and self-confirm — a separate, non-transactional surface.

**Edit mode is widget-layout only, on the auto-hiding toolbar.** While non-static the toolbar swaps to an accent **"Editing layout"** label + **Cancel · Done** (Done in the former pen slot). Page navigation (dots) and management (wrench) are **hidden while editing** — normal-mode only. With `pageNav()`'s existing keyboard/swipe guard and the new remote-nav guard (a local edit pre-empts a remote page change), this closes the mid-edit page-switch data loss for every **in-app** path. Browser Back still navigates: page-level Back intentionally abandons the edit; Back-should-close-the-widget-settings-dialog is tracked in #393. The pinned in-dashboard Done/Cancel FABs are gone; the peek strip is neutral normally and **accent while editing**.

> **Design note:** an earlier iteration put "Manage pages" *inside* edit mode. Review showed that gave edit mode a Cancel that couldn't undo the immediate page ops, and let a page op corrupt an in-flight widget edit. Splitting them (wrench in normal mode; edit mode widget-only) fixes both — see #388.

**How Done/Cancel reach the grid:** the toolbar doesn't hold the gridstack, so `DashboardService` gains `requestLayoutEditSave/Cancel` counters; the mounted `DashboardComponent` runs its existing transactional `saveLayoutChanges()`/`cancelLayoutChanges()`. Counters are snapshotted at construction so a stale value can't auto-fire on a navigation remount.

**Active-page identity reload:** the grid reloads when the active page's *id* changes (not just its index), so deleting the active page doesn't leave the deleted page's widgets on screen.

## Verification

- `npm run ci` green; `build:prod` succeeds.
- Specs: manage-pages wrench (present/opens in normal mode, absent in edit); edit-mode swap + Done/Cancel wiring; request-counter remount guard; active-page identity reload (reloads on delete-active, not on config saves); compact editor; page-manager sheet.
- On-boat smoke pending.

Part of #388.

### DIFF
--- a/src/app/core/components/dashboard/dashboard.component.html
+++ b/src/app/core/components/dashboard/dashboard.component.html
@@ -45,22 +45,3 @@
 }
 </div>
 <action-menu #actionMenu [items]="emptyAreaActions()" (selected)="onEmptyAreaAction($event)" />
-@if (!dashboardStatic) {
-  <div class="edit-layout-close-icon">
-    <button
-      mat-fab
-      class="cancel"
-      style="margin-right: 10px"
-      (click)="cancelLayoutChanges()"
-    >
-      <mat-icon>close</mat-icon>
-    </button>
-    <button
-      mat-fab
-      class="save"
-      (click)="saveLayoutChanges()"
-    >
-      <mat-icon>done</mat-icon>
-    </button>
-  </div>
-}
--- a/src/app/core/components/dashboard/dashboard.component.scss
+++ b/src/app/core/components/dashboard/dashboard.component.scss
@@ -35,28 +35,6 @@
   }
 }
 
-.edit-layout-close-icon {
-  position: absolute;
-  bottom: 20px;
-  right: 20px;
-  cursor: pointer;
-  z-index: 100;
-}
-.cancel {
-  @include mat.fab-overrides((
-    container-color: var(--mat-sys-primary),
-    foreground-color: var(--mat-sys-background),
-  ));
-}
-
-.save {
-  @include mat.fab-overrides((
-    container-color: var(--mat-sys-primary-fixed),
-    foreground-color: var(--mat-sys-background),
-
-  ));
-}
-
 /* Empty Dashboard State Styles */
 .dashboard-empty-state-container {
   position: absolute; /* detach from normal flow so it cannot change grid height */
--- a/src/app/core/components/dashboard/dashboard.component.spec.ts
+++ b/src/app/core/components/dashboard/dashboard.component.spec.ts
@@ -6,7 +6,7 @@ import { signal } from '@angular/core';
 import { provideRouter } from '@angular/router';
 
 import { DashboardComponent } from './dashboard.component';
-import { DashboardService } from '../../services/dashboard.service';
+import { Dashboard, DashboardService } from '../../services/dashboard.service';
 import { ToastService } from '../../services/toast.service';
 import { PluginConfigClientService } from '../../services/plugin-config-client.service';
 import { DialogService } from '../../services/dialog.service';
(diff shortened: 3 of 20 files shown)


mairas:
## Code review — Phase 1 (multi-persona)

Scope: `feat/actions-redesign-phase1` vs `main` (18 files). Intent: relocate layout-edit Done/Cancel to the toolbar edit-mode swap via a DashboardService request-counter channel; add the page-manager bottom sheet + compact `dashboards-editor`; accent peek strip while editing. Additive — `/actions` stays until Phase 2.

Team: correctness, testing, maintainability, project-standards (always-on) + adversarial (diff >50 LOC, config-persistence state). All 5 returned.

Fixes in this review were applied to the branch (commit follows this comment).

### P1 — Remount replays a stale request counter → spurious save/cancel  ✅ fixed
`correctness, testing, project-standards, adversarial` (4 independent reviewers)

`layoutEditSaveRequested`/`layoutEditCancelRequested` are root-scoped, monotonic counters that never reset; `DashboardComponent` is recreated on navigation (`page/:id` is eager, other routes are lazy, no `RouteReuseStrategy`). The effects guarded only `if (!tick) return`, so after the first Done/Cancel in a session the counter is ≥1 and a fresh mount replayed it — a spurious `saveLayoutChanges()`/`cancelLayoutChanges()` on mount (redundant config write + a `notifyLayoutEditCanceled()` that cascades into every `widget-host2` reinit). Correctness noted the worst-case data-loss is currently blocked by effect ordering + `isEqual`, but the spurious widget reinit on every page re-entry is a real, visible regression.

**Fix:** snapshot each counter at construction; act only on a later increment (ignores the construction-time value whether 0 or N). Mirrors `widget-host2`'s `_hasInitialized` guard. Added a remount regression test (fresh mount with counters already at 2 → no fire) plus a repeated-tick test (1→2 re-fires).

### P2 — Cancel test asserted static+notify but not the discard  ✅ fixed
`testing`

`discards the edit when the toolbar requests a cancel` stubbed `loadDashboard` but never asserted it fired — the reload *is* the discard. **Fix:** assert `loadDashboard(0)` after the cancel tick.

### P2 — Page structural op during an unsaved widget edit  ⚠ OPEN (design decision)
`correctness, adversarial`

The page-manager sheet is now reachable while a widget-layout edit is unsaved. Two traps:
- **Reorder** the active page mid-drag → `dashboards-editor.drop()` mutates `activeDashboard` → the `activeDashboard` effect reloads the grid from persisted config → the unsaved drag is silently discarded.
- **Delete** the active/last page → `delete()` reseeds and `activeDashboard.set(0)` with the *same* index, so the signal doesn't emit and the grid never reloads → grid (old widgets) and model (blank) diverge → a subsequent **Done resurrects the deleted widgets**.

Both reviewers route this to a design decision — it's the sharp edge of the design's documented "page ops are immediate, outside the widget-layout transaction" boundary. **Not changed in this PR** (honoring the documented behavior); see the open question raised with the maintainer before merge. A wrong auto-fix here risks the delicate page-transition code.

### P3 — cleanups  ✅ fixed
- `saveLayoutChanges`/`cancelLayoutChanges` narrowed `protected` → `private` (no longer template-reachable after the FABs were removed). `maintainability`
- Reworded a comment that narrated the change ("now live on the toolbar…") to current-state per repo comment style. `maintainability, project-standards`
- Reworded a garbled peek-strip SCSS comment. `maintainability`

### Coverage
- Suppressed: findings below 0.60 confidence.
- Residual risks (not fixed): navigating off `/page` mid-edit leaves `isDashboardStatic` false with no mounted toolbar (pre-existing; returning re-enters edit mode); Done during an in-flight page slide serialises the outgoing grid into the new index (pre-existing race the removed FAB also had); an open Manage-pages sheet survives a route change (root overlay).
- Testing gaps closed: remount-with-stale-counter, repeated ticks, cancel reload assertion. Still open: page-op-during-edit scenarios (tied to the design decision above).

### Verdict
**Ready with fixes** — P1 + the test/cleanup items are fixed and CI is green. The P2 page-op-during-edit interaction is an intentional-boundary question for the maintainer; hold merge until answered (per the overnight instruction to surface questions before merging).

mairas:
## Review — design pivot (page management → normal-mode wrench + active-page identity reload)

Focused correctness + adversarial review of the two pivot commits (`git diff 922c8eac a5409d20`), fresh contexts.

**correctness: clean (0 findings).** Traced the identity-reload effect: no spurious reload during a widget edit (Done's `updateConfiguration` keeps the same id → guard suppresses it), embed-safe (toolbar unmounted under embed), null/out-of-range safe, and the page-transition/#194 gate intact.

### P2 — Sheet delete of a lower page desyncs the URL from the active index  ✅ fixed
`adversarial`

`delete()`/reorder reassign `activeDashboard` in memory without navigating, so the URL keeps its old index. Concrete: pages `[a,b,c]` at `/page/1` (active `b`); delete `a` → active becomes `0` (`b`), grid + dots correct, but URL stays `/page/1`. Close the sheet, tap the dot now at index 1 (`c`) → `navigateTo(1)` issues a same-URL navigate the router **ignores** → the tap is dead and a stale slide direction is left set. The old full-page `/actions` editor masked this by navigating to `/page/0` on close; the sheet didn't.

**Fix:** the toolbar re-syncs the route (`navigateToActive()`) when the page-manager sheet dismisses — a no-op when nothing changed, a URL correction after a delete/reorder. Test added.

### P2 — Tapping a page dot mid-edit discards unsaved widget moves  ⚠ OPEN (pre-existing, design decision)
`adversarial` (pre_existing), `correctness` (residual risk)

In edit mode the page dots stay live (by design — "switch pages while editing"). Tapping another page runs `loadDashboard` over the live grid, so the current page's unsaved moves are lost with no save/confirm. Both reviewers confirm this **predates** the redesign (the dots and the reload-on-active-change were always there); the pivot doesn't touch this path. It's the same class as the page-op/transaction question — an edit-mode navigation semantics decision (block / commit-first / warn). Flagged for your call; not changed here.

### Hardening
- Added a spec asserting the identity-reload does **not** fire while editing when the active page's config changes (same id) — the load-bearing "no reload over unsaved drags" invariant, previously only indirect.

### Residual (not fixed, low risk)
- A page op landing inside the ~360ms nav-slide window could leave a stale grid (settle logic compares index value, not identity). Near-unreachable — you can't tap dots while the sheet's backdrop is up, and a confirm-gated delete outlasts the slide. Pre-existing settle limitation.
- Reordering the active page in static mode does a redundant (byte-identical) grid reload — perf-only flicker.

### Verdict
**Ready** — P2a fixed + tested, correctness clean, CI green. P2b (dots mid-edit) is a pre-existing design decision surfaced for you.

PR #399, opened 2026-07-22
fix(action-menu): open the page/widget context menu at the tap point

mairas:
## Why

Opening the page-manager (wrench → bottom sheet) and tapping a page card opened its Edit/Duplicate/Delete context menu **far to the right** of the tap, instead of at the tapped card. Widget action menus on the main dashboard were unaffected.

## Root cause

`action-menu` opens its Material menu by placing a `position: fixed`, 0×0 anchor at the tap's `clientX/clientY`, then anchoring the menu overlay to it. A `fixed` element is viewport-relative only if no ancestor establishes a containing block.

Inside the bottom sheet, `.mat-bottom-sheet-container` keeps a **non-`none` residual transform**: its enter animation is `forwards`-filled and its keyframes end at `transform: none`, but the *computed* filled value is `matrix(1,0,0,1,0, 1.2e-06)` — not literal `none`. That residual makes the container the containing block for the fixed anchor, so `left/top` resolve against the **centered, non-full-width** sheet (left edge ≈672px at 1920px wide) rather than the viewport — pushing the menu right by the sheet's left offset. Main-dashboard widgets use the same component with no transformed ancestor, so they positioned correctly.

Static source-grep for `transform` misses this because it's an animation end-state, not a declared rule.

## Approach

In `open()`, after placing the anchor, measure `getBoundingClientRect()` and re-set `left/top` to `2·x − rect.left` / `2·y − rect.top`, cancelling whatever offset a transformed ancestor introduced. It's a **no-op when there is no such ancestor** (widget menus resolve identically).

## Verification

- **Headless-Chromium reproduction** faithfully replicating the Material structure (centered non-full-width container + `forwards`-filled transform animation + fixed anchor): the buggy case placed the anchor intended for (900,500) at (1572,1420); after the fix, all three cases — transformed container / no-transform / static transform — land exactly at the tap point.
- `npm run lint` ✓, `npm run snc` ✓, `action-menu` + `dashboards-editor` specs ✓ (15 tests).
- jsdom cannot reproduce this (no layout engine; `getBoundingClientRect` → 0), so it's verified in a real engine rather than a unit test.

https://claude.ai/code/session_01GwAeQKbZAExuKHT5Q4wgSk


mairas:
## Code review — multi-persona (mairas-workflow)

**Scope:** `fix/action-menu-anchor-transformed-ancestor` vs `main` — 1 commit, 1 file (`action-menu.component.ts`, 8 added lines).
**Intent:** Fix the page-manager context menu opening far right of the tapped page card, without regressing the other `action-menu` consumers (widget-host2, group-widget, dashboard) which have no transformed ancestor.
**Team (4 always-on; no conditional reviewers — the diff touches no auth/API/reliability/large-surface):** correctness, testing, maintainability, project-standards. Each ran independently on Opus in its own context.

**Correctness: no bugs.** The reviewer traced all four consumers, confirmed the `2·x − rect.left` cancellation is exact for a pure-translation containing block (`rect.left = C + x` → set `left = x − C` → renders at `x`; no-op when `C = 0`), and specifically hunted for a scaling ancestor — the only thing that would break the linear correction. None exists among current consumers (`dashboards-editor`'s `scale(0.75)` is on the add-button, a sibling of the anchor, not an ancestor; the dashboard page-slide is pure `translateX`).

### P3 — Low (all addressed in this PR)

| 1 | action-menu.component.ts:59 | `2 * x - rect.left` folds the offset algebra away — reads as a magic coefficient, survives only via the comment | maintainability | 0.66 | Fixed — named `offsetX/offsetY = rect.left − x`, then `x − offsetX` |
| 2 | action-menu.component.ts:51 | First `left/top = x/y` write is overwritten two lines later, so it reads as dead code a future editor would delete — but it is load-bearing (the measured rect depends on it) | maintainability | 0.70 | Fixed — the named-offset form makes the measure-then-cancel ordering explicit |
| 3 | action-menu.component.ts:61 | Desktop pop-over branch (incl. the new lines) runs in zero tests; spec covers only the phone drawer | testing | 0.70 | Fixed — added a branch test asserting the pop-over opens (no geometry assertion; see Coverage) |

### Investigated and dismissed

- **"Comment misattributes the cause; `.page-slide` is the real transformed ancestor" (maintainability, 0.62)** — refuted against the code. `.page-slide` animates `translateX` with `fill: 'forwards'`, but `animatePhase().then()` calls `setSlideOffset(el, 0)` (`transform = ''`) **and** `animation.cancel()` for the settled active page, so it carries **no** residual transform — unlike the bottom-sheet container, which never cancels its fill. The reviewer also read the phone-branch `return` as "the bottom sheet is never open here," but the page manager *is* a `MatBottomSheet` (open on the wide-screen path); its container is the actual containing block. Comment attribution stands.

## Applied fixes

- Refactored the compensation to named `offsetX`/`offsetY` (behavior-identical: `x − (rect.left − x) = 2·x − rect.left`) — removes the magic `2·`, and makes the load-bearing first write self-evident.
- Added `action-menu.component.spec.ts` case: non-phone branch → asserts `MatMenuTrigger.openMenu` is called and the bottom sheet is not. It guards the branch's outcome without asserting position.

## Coverage

- **Geometry is not unit-testable here, by design.** The runner is vitest + jsdom, which has no layout engine — `getBoundingClientRect()` returns a zero rect, so `2·x − rect.left` collapses to `2·x`. An assertion on the compensated `left/top` would be **green against a real regression and red against correct code** — a false-confidence trap, so it is deliberately omitted. The fix is verified in headless Chromium (real layout): the buggy case placed the anchor intended for (900,500) at (1572,1420); post-fix all cases land at the tap point.
- **Residual risk (latent, not present):** the linear cancellation is exact only for a pure-translation containing block. A future ancestor that *scales* or *rotates* a consumer would need a different correction. No such ancestor exists today; noted so a future zoom/scale wrapper prompts a re-check.
- Suppressed: 0 findings below threshold. No reviewer failed or timed out.

---

**Verdict: Ready to merge.** No correctness issues; the three P3 maintainability/testing items are addressed in-PR; the one cross-cutting claim was verified and dismissed against the code.

PR #406, opened 2026-07-24
feat(subscription): widget-demand-driven AIS/DSC subscribe scope (#386)

### LINKED ISSUE #386: Subscribe to remote (AIS/DSC) contexts only on widget demand

mairas:
## Why

The Connectivity-tab retirement (#385) left AIS/DSC subscription **unconditional** — Skip always opens the Signal K stream with `?subscribe=all`, so the server pushes every remote vessel context and `DataService` parses/buckets each delta even when nothing renders it. On the live boat that's ~10 targets (negligible), but AIS target count is location-dependent: hundreds in a busy strait or harbour. Subscribe to remote contexts only when a widget actually consumes them.

## How

Choose the WebSocket subscribe scope (`all` vs `self`) from **widget demand**:

- **Demand predicate** (`remote-context-demand.util.ts`): a dashboard needs remote contexts if any widget is a known remote-context type (`widget-ais-radar`) **or** any string in its config begins `vessels.`/`atons.` (mirroring `path-discovery`'s context classifier). The string scan covers every path-storage shape — `paths` objects/arrays, `datachartPath`, etc. — so it also catches a generic widget bound to a remote path via `filterSelfPaths` (see note below).
- **Compute + persist**: recomputed in the `DashboardService` save effect (fires on the initial seed and every edit; skipped in embed) and stored per-device in `connectionConfig.remoteContextDemand`.
- **Consume at boot**: read pre-auth in `app-initNetwork` as `remoteContextDemand ?? true` and passed as the `subscribeAll` arg.

Because the first socket opens post-auth but before dashboards load, a demand change takes effect on the **next reload** — no worse than the pre-retirement enable-and-reconnect UX (tracked as the accepted limitation in #386).

### Fail-open (safety)

Under-subscribing would silently hide collision-relevant AIS targets, so the design fails open to `subscribe=all` on **any** uncertainty: an absent/never-computed field, a fresh device, or an ambiguous config. The field is a new **optional** `IConnectionConfig` member (absent ⇒ `?? true`), so no config-version bump/migration is needed and existing devices default to `all` until demand is computed. Narrowing to `self` happens only on positive evidence of no remote consumer.

### Two corrections to the issue's framing (verified in code)

- The issue treats scope as decided *at connect time, pre-auth*. In fact the first socket opens via the `isLoggedIn$` reconnect **after** auth but **before** dashboards load — so a staleness-free variant would mean deferring the first open past dashboard load (touching the auth/WS boot race). This PR deliberately takes the cached-hint approach instead.
- The issue names `widget-ais-radar` as the *only* remote consumer. It isn't: the generic path picker's `filterSelfPaths` toggle lets any widget bind a remote `vessels.`/`atons.` path. The predicate covers both.

## Piggy-backed: #1118 (trivial fly-by)

`build(deps): promote sk-ais-status-plugin to a runtime dependency` — the AIS radar consumes the plugin's `sensors.ais.status` states (`confirmed`/`unconfirmed`/`lost`/`remove`) to filter unreliable targets and suppress CPA/guard alarms for lost tracks. As a `devDependency` it was omitted from an app-store install, so those safety filters silently no-op. Replicates upstream `mxtommy/Kip#1118` (its Kip version bump is N/A to Skip's independent release). Isolated in its own commit.

## Verification

- `npm run ci` green: lint, snc, `test:headless` 1548/1548, `test:plugin` 7/7, `test:mcp-schema` 31/31.
- New tests: predicate (AIS type, remote `paths`/`datachartPath`/array shapes, self-only, decoy host selector, malformed configs); boot fail-open (absent→all, false→self, true→all); dashboard recompute on change; settings round-trip + no-op guard.
- Lockfile diff for #1118 is surgical (only the `"dev": true` removal + block relocation, zero package churn).
- **Pending before merge:** on-boat smoke on halpi — no-AIS dashboard connects `self`; AIS dashboard populates.

Closes #386.


mairas:

Ran an adversarial review across correctness, safety, config-contract, maintainability, and testing lenses; each raised finding was independently verified against the code. **11 raised → 7 confirmed, 4 refuted.** Every confirmed finding was a *fail-open hole* — a path where a remote consumer exists yet the scope could narrow to `self`, silently hiding AIS targets. All fixed in `3c39c74e`.

### Confirmed & fixed

| # | Sev | Finding | Fix |
|---|-----|---------|-----|
| 1 | high | Predicate ignored `group-widget` nested children (`subGridOpts.children`) — an AIS/remote-path widget *inside a group* computed demand=`false`. | `widgetHostNeedsRemoteContext` now recurses into `subGridOpts.children`. |
| 2/4 | high/med | Non-embed ephemeral `?profile` viewing wrote the **device** flag → poisoned the real profile's next boot. | Save effect gates the demand write on `!_embedMode.profile()`. |
| 5 | med | Degraded/empty boot seeded `DefaultDashboard` → overwrote a stored `true` with `false`. | Write also gated on `_storage.isRemoteContextBootstrapped()` (authoritative profile loaded). |
| 3 | med | Boot *read* trusted the device flag in embed/ephemeral sessions (regression from #385's always-`all`). | Embed/ephemeral sessions now force `subscribe=all` at boot. |
| 6 | med | Embed no-write guard was untested. | Added embed / ephemeral / not-bootstrapped no-write tests. |
| 7 | nit | Self-referential decoy assertion tested the fixture, not the SUT. | Replaced with a prefix-not-substring behavioral case. |

Net effect: the per-device demand flag is now derived **only** from the device's own authoritative profile, and every non-authoritative context (embed, ephemeral `?profile`, degraded seed) fails open to `all`.

### Refuted (correctly, on verification)

- *Dead `signalKSubscribeAll` beside `remoteContextDemand`* — pre-existing inert field, not introduced here.
- *Recursive any-string scan over-fires* — accurate observation, but fail-safe by design (over-match only widens scope).
- *`true→false` recompute untested* — unreachable; the recompute is a single shared effect line.
- *Unrelated connection write drops the demand* — invariant already satisfied by the single serialization chokepoint.

### Verification
`npm run ci` green after fixes: 1554 unit + 7 plugin + 31 mcp-schema. Still owed before merge: on-boat smoke on halpi.

mairas:
## Independent re-attack review (post-fix) — 1 HIGH found & fixed

Ran a fresh 5-facet re-attack over the current state (hardest on the fix logic from the first review round, which hadn't been independently reviewed). **6 raised → 4 confirmed (1 root cause at HIGH/MED/LOW + 1 test gap), 2 refuted.**

### 🔴 HIGH — profile switch under-subscribed (fixed in `40d1dbec`)

`remoteContextDemand` was a single **per-device** value, but demand is **per-profile**. Switching from a self-only profile (persisted `false`) to one with an AIS radar reloaded and booted reading the *outgoing* profile's stale `false` → `subscribe=self` → the AIS radar was **silently blank for the whole session** (self-healed only on a second, unprompted reload). A persisted switch is neither embed nor ephemeral, so the earlier guards didn't catch it — a fail-open violation on a first-class user action.

**Fix:** model demand as a map keyed by profile (`Record<profileName, boolean>`). The boot read looks up the active profile's entry; a missing entry — never computed, or a switched-to profile whose demand belongs to a sibling — **fails open to `all`**. This guards at the *read* site, so it's correct regardless of which code path changed the active profile (rather than resetting a flag at every switch chokepoint — the exact class of omission that caused the bug). Regression test added: a sibling-only demand map → `subscribe=all`.

### 🟡 LOW — depth-≥2 nested groups untested (fixed)

The predicate recursed correctly, but the test only covered one nesting level; a non-recursive regression would have passed silently. Added a group-in-group (depth-2) lock.

### Refuted (2, correctly)
- Pinned-`?profile` kiosk "always subscribes all" — fail-open working as designed.
- `isRemoteContextBootstrapped` vs `remoteContextDemand` naming overlap — coincidence, not a defect.

### Focused re-verify of the fix — FIX-SOUND ×3

Three independent adversarial angles (key-consistency, loaded-vs-configured profile, residual fail-open) all returned **FIX-SOUND**, no blocking defect. Confirmed: write-key and read-key are serialized in the same blob (can't diverge); the map is only read in the non-embed/non-ephemeral branch where the loaded profile equals the read key; degraded/404 paths don't start the WS at all; legacy-boolean and prototype-named inputs both degrade fail-open.

### Non-blocking notes (no action needed)
- **Cross-device staleness:** editing profile *A* on device B to add AIS, then booting *A* on device X, under-subscribes for one boot then self-heals — inherent to the pre-auth local-cache shape (chosen deliberately over a runtime reconnect), identical to the single-device next-reload lag, and present in the pre-fix design too.
- **`__proto__`-named profile:** a profile literally named `__proto__` would churn localStorage (never persists) but stays fail-open (`all`). Pathological; noted, not fixed.

### Verification
`npm run ci` green after the fix (unit + plugin + mcp-schema); PR CI green on `40d1dbec`; 166 tests pass across the four touched specs incl. the new regression + depth-2 locks.

Issue #408, opened 2026-07-26
Label & unit text illegible on small tiles: legibility floor + background-color halo (canvas widgets)


mairas:
## Problem

On small dashboard tiles — and in bright daylight or at a distance — the **label** and **unit** text in canvas widgets shrink below legibility while the main value stays readable. Reported on the Numeric widget; the cause is shared, so it affects the other canvas widgets too.

## Root cause

All canvas text is auto-fit to a bounding box whose **height is a fixed fraction of the tile height**, via `CanvasService.calculateOptimalFontSize` (it seeds `maxFontSize = maxHeight` and only shrinks further for width). The fractions are hardcoded and lopsided:

- value ≈ **0.70 H**
- unit ≈ **0.15 H** (numeric; racer-line similar)
- label = **0.10 H** (`titleFraction` default — every label-drawing canvas widget)

There is **no minimum size** (the floor is 1px) and no contrast/daylight consideration — the fit is purely geometric. So label (0.10 H) and unit (0.15 H) cross the legibility threshold long before the value (0.70 H) as the tile shrinks. The label is additionally drawn in a dimmed color, compounding it in sunlight.

## Scope (v1)

Fix the **CanvasService text substrate only.** This reaches 8 canvas widgets — numeric (flagship), text, datetime, position, racer-line, racer-timer, race-timer, label — and centralizes the currently per-widget unit placement (numeric and racer-line are the only canvas unit-drawers).

Label/unit rendered through **other substrates are out of scope for v1** — each renders through its own library/markup and needs a distinct technique (see Follow-ups): canvas-gauges lib (gauge-ng-linear/radial/compass), steelseries (gauge-steel), SVG templates (simple-linear, heel, windsteer, racesteer), chart.js (data-chart, minichart, windtrends), and the electrical SVG panels. A single CanvasService change does **not** reach any of these — "global" here means global across the canvas substrate.

## Design

Three coupled parts, in `CanvasService` plus the two canvas unit call sites:

[...]

PR #409, opened 2026-07-26
feat: legibility floor + background-color halo for widget labels & units

mairas:
## Why

On small dashboard tiles — and in bright daylight or at a distance — a widget's **label** and **unit** text shrink below legibility while the main value stays readable. They're auto-fit to a box whose height is a fixed fraction of the tile (label ~10%, unit ~15%, value ~70%) with no lower bound, so they cross the legibility threshold long before the number does.

Addresses the **canvas-substrate scope** of #408 (the library/SVG/chart widgets and the reflow fallback remain tracked there as follow-ups).

## How

Three coupled parts, all opt-in via new optional args on `CanvasService` (existing callers unchanged):

1. **Legibility floor (absolute CSS px).** `calculateOptimalFontSize` gains an optional `floorPx`, applied on read so floored and unfloored callers share one memo entry. Floored text drops its `maxWidth` cap so it overflows rather than horizontally squishing. `MIN_LABEL_PX = 16`, `MIN_UNIT_PX = 12`.
2. **Always-on background-color halo (knockout).** A new `strokeText` pass under the fill, stroked with the widget card background (`theme.cardColor`). Because it's the background color it's **invisible over the empty card** and only carves a clean channel where the text overlaps the value — self-nullifying, so it's always-on at zero cost until overlap happens.
3. **Composite label/unit above the value.** The knockout only works if the text is painted after the value, so each widget's cached label/unit bitmap moves from a background layer to a foreground layer blitted last. The bitmap stays cached (rebuilt only on size/text/theme change), so per-frame cost is unchanged.

## Scope

All six **canvas-text** widgets: `numeric` (label + unit), `text`, `datetime`, `position`, `racer-timer`, `racer-line` (label + DTS unit). Widgets that render label/unit through libraries (canvas-gauges, steelseries), SVG templates, or chart.js are untouched and remain follow-ups in #408.

## Assumption

The knockout relies on a flat `theme.cardColor` widget background (true today: `widget-host2` fills a solid `--skip-widget-card-background-color`, canvas transparent over it). A future gradient/image widget background would need revisiting.

## Verification

- `snc` clean, lint clean, full unit suite **1558/1558** (one run hit the known order-dependent `minichart` flake — passes isolated and on re-run; untouched by this change).
- Built and deployed to a live boat Signal K (`./run deploy-halos local`), tuned and confirmed on the Numeric widget across tile sizes; floor values settled interactively (label 16, unit 12).
- Perf-neutral by construction: same per-frame draw-call count (one cached-bitmap blit + value draw), blit merely reordered.

## Follow-ups (tracked in #408)

Reflow fallback for the pathological long-label-on-tiny-tile case; per-substrate treatment for the library/SVG/chart widgets.

### DIFF
--- a/src/app/core/services/canvas.service.spec.ts
+++ b/src/app/core/services/canvas.service.spec.ts
@@ -122,3 +122,55 @@ describe('CanvasService text measurement (#321 font-race)', () => {
     await expect(service.whenFontsReady()).resolves.toBeUndefined();
   });
 });
+
+/**
+ * Auxiliary text (widget labels and unit symbols) is sized as a fraction of the
+ * tile height and shrinks below legibility on small tiles. A pixel floor keeps it
+ * readable; the fit is capped only from below, and the floor is applied on read so
+ * floored and unfloored callers share one memo entry.
+ */
+describe('CanvasService font-size floor (label/unit legibility)', () => {
+  let service: CanvasService;
+  beforeEach(() => {
+    TestBed.configureTestingModule({});
+    service = TestBed.inject(CanvasService);
+  });
+
+  // measured width scales with font px and text length, so the binary search is deterministic.
+  function fakeCtx(onMeasure?: () => void) {
+    return {
+      font: '',
+      measureText(text: string): TextMetrics {
+        onMeasure?.();
+        const px = parseInt(this.font, 10) || 10;
+        return { width: text.length * px * 0.5 } as TextMetrics;
+      }
+    } as unknown as CanvasRenderingContext2D;
+  }
+
+  it('floors the result at floorPx when the box would force a smaller size', () => {
+    const ctx = fakeCtx();
+    const unfloored = service.calculateOptimalFontSize(ctx, 'Label', 200, 6, 'normal');
+    expect(unfloored).toBeLessThanOrEqual(6);            // height-bound to the tiny box
+    const floored = service.calculateOptimalFontSize(ctx, 'Label', 200, 6, 'normal', 12);
+    expect(floored).toBe(12);                            // never below the floor
+  });
+
+  it('leaves the size unchanged when the fitted size already clears the floor', () => {
+    const ctx = fakeCtx();
+    const fitted = service.calculateOptimalFontSize(ctx, 'X', 200, 40, 'normal');
+    expect(fitted).toBeGreaterThan(12);
+    expect(service.calculateOptimalFontSize(ctx, 'X', 200, 40, 'normal', 12)).toBe(fitted);
+  });
+
+  it('applies the floor on read without re-running the search (shared memo entry)', () => {
+    let measureCalls = 0;
+    const ctx = fakeCtx(() => { measureCalls++; });
+    const unfloored = service.calculateOptimalFontSize(ctx, 'Label', 200, 6, 'normal');
+    const afterSearch = measureCalls;
+    const floored = service.calculateOptimalFontSize(ctx, 'Label', 200, 6, 'normal', 12);
+    expect(measureCalls).toBe(afterSearch);              // cache hit: no new search
+    expect(unfloored).toBeLessThan(12);
+    expect(floored).toBe(12);                            // floor applied on the cached value
+  });
+});
--- a/src/app/core/services/canvas.service.spec.ts
+++ b/src/app/core/services/canvas.service.spec.ts
@@ -136,13 +136,16 @@ describe('CanvasService font-size floor (label/unit legibility)', () => {
     service = TestBed.inject(CanvasService);
   });
 
-  // measured width scales with font px and text length, so the binary search is deterministic.
+  // measured width scales with the font's px size (parsed from the `<weight> <n>px <family>`
+  // shorthand) and text length, so the binary search — including its width-bound branch — is
+  // deterministic. A naive parseInt(font) reads the leading weight word as NaN and collapses width
+  // to a constant, hiding the width path; extract the px number explicitly instead.
   function fakeCtx(onMeasure?: () => void) {
     return {
       font: '',
       measureText(text: string): TextMetrics {
         onMeasure?.();
-        const px = parseInt(this.font, 10) || 10;
+        const px = Number(/(\d+(?:\.\d+)?)px/.exec(this.font)?.[1]) || 10;
         return { width: text.length * px * 0.5 } as TextMetrics;
       }
     } as unknown as CanvasRenderingContext2D;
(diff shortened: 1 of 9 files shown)


mairas:

**Scope:** PR #409 vs `main` (`92aea87`) — 8 files, +182/−60. **Intent:** absolute-px font-size floor + always-on background-color halo for widget labels/units, adopted across 6 canvas widgets by compositing the label/unit bitmap after the value.

**Reviewers (5, each independent Opus context):** correctness, testing, maintainability, project-standards (always-on) + adversarial (diff >50 LOC, shared service consumed by ~8 widgets).

**Verdict: Ready with fixes.** No P0/P1. No correctness defect found — the draw-order flip, floor-on-read cache (idempotent `Math.max`, `floorPx` excluded from the key), `widthArg` overflow, and `theme()`-after-guard safety all verified correct. Two P2s worth closing before merge (one real cold-boot regression, one test-coverage gap on the headline feature), plus a few cheap P3s.

### P2 — Moderate

| # | File | Issue | Reviewer(s) | Conf |
|---|---|---|---|---|
| 1 | all 6 widget draw()s | **Cold-boot font race defeats the label-on-top composite.** `drawText` defers the value paint to `fontsReadyPromise` when fonts aren't loaded, but the `drawImage(titleBitmap)` blit is synchronous — so on a cold load the value paints *after* the label, reverting to value-on-top with no halo. Live-updating widgets self-heal on the next tick; **static-value widgets (text with a static string, position showing `--`) stay wrong** until a resize/theme/value change. | adversarial | 0.65 |
| 2 | canvas.service.ts (`strokeTextHalo`/`drawText`) | **Halo path has zero test assertions** despite being the headline feature, and it's cheaply testable at the public `drawText` seam (synchronous under the fonts shim). Nothing pins that `strokeText` fires with `cardColor` + `lineWidth`, is gated on `haloColor`, or precedes the fill. | testing (0.85) + maintainability, correctness, project-standards | 0.85 |

### P3 — Low

| # | File | Issue | Reviewer(s) | Conf |
|---|---|---|---|---|
| 3 | canvas.service.spec.ts | The floor tests' `fakeCtx` does `parseInt(this.font)` → **NaN → constant 10px**, so measured width is font-size-independent; the **width-bound branch of the fit is never exercised**. Floor tests only cover the height-bound path (pre-existing helper weakness inherited by the new tests). | adversarial | 0.70 |
| 4 | widget-numeric.component.spec.ts | Docstring "the required theme input is never read" is **now stale** — `drawWidget` reads `theme()`; the spec only passes because of the `if(!ctx)return` guard. A future `ngOnInit`/`detectChanges` in that spec would hit `theme()` and throw NG0950; the protective invariant is undocumented where it now matters. | correctness | 0.75 |
| 5 | 5 non-numeric widgets | **Halo-color staleness:** the title-bitmap cache key omits `cardColor`, so a theme that changes `cardColor` without changing the label's dim color would keep a stale halo. Low confidence (Skip's shipped themes change card + text together; numeric is immune — `setColors` nulls its bitmap on any theme change). | correctness, adversarial | 0.60 |
| 6 | canvas.service.ts | **Overflow-without-halo coupling:** `floorPx` always drops the `maxWidth` cap, but the halo is gated on `cardColor`; if `cardColor` ever resolves falsy the floored text overflows the value with no knockout. Low reachability (`--skip-widget-card-background-color` is defined at `:root`), but the empty-string branch is deliberate and reachable in a FOUC window / partial theme. | correctness, adversarial | 0.60 |
| 7 | canvas.service.ts | `drawText` now has 12 positional params; opting into the halo forces spelling out the `0.1` titleFraction default at 5 sites, and the adjacent trailing `(string?, number)` args are transposition-prone. Consider an options object for the tuning tail. Not blocking — extends a pre-existing long-positional pattern. | maintainability | 0.62 |
| 8 | canvas.service.ts:558 | `strokeTextHalo` hardcodes `miterLimit = 2`, `lineWidth = max(2, fontSize/5)` — unnamed. Matches the file's existing inline-literal style; optional to hoist to named constants. | project-standards | 0.60 |

### Notes / residual

- **racer-line** `TTL`/`TTB` captions get neither the floor nor the halo the main label/unit now get — likely intentional (dense fixed layout) but reads as an unexplained inconsistency; a one-line comment would resolve it.
- On very small tiles the floored 16px label + opaque halo can knock out most of the (unfloored) value — a legibility tradeoff baked into the floor; worth a product call on a minimum tile size, not a defect.
- **Adversarial explicitly cleared:** `strokeTextHalo` save/restore is balanced (no ctx-state leak into non-halo draws); 1px-tile `maxHeight→0` returns the floor sanely and overflow is clipped to the tile; no reactivity loop from `theme()` reads; the racer-line title blit-last does **not** paint over `TTL`/`TTB` or the SVG overlays.

### Coverage

- Testing gaps: halo invocation/gating/color/order; `widthArg` overflow branch; the draw-order flip (jsdom stub is a no-op — genuinely needs the deploy/visual check that was done); width-bound fit path (finding #3).
- Perf-harness not run — per-frame cost is unchanged by construction (halo baked once into the cached bitmap; still one `drawImage` + value draw per frame), so low risk, but unmeasured.

**Recommended fix order before merge:** #1 (cold-boot redraw hook) → #2 + #3 (halo test + fix the fake-ctx width path) → #4 (docstring). #5/#6 optional hardening; #7/#8 defer.

mairas:
## Review findings addressed

**Fixed** (commits `bf5d8bd`, `bbfa824`):
- **P2 — cold-boot font race.** The value paint defers until fonts load; the label/unit blit was synchronous, so it could land under a late value. New `CanvasService.drawTextBitmap` routes the blit through the same font-readiness gate, so within each `draw()` it registers *after* the value's paint and still composites on top. All 6 widgets now use it.
- **P2 — halo untested.** Added assertions at the `drawText` seam: the halo strokes in the passed color, `lineWidth` scales with font size, it's drawn under the fill, it's skipped when no `haloColor` is given, and the width cap drops when floored.
- **P3 — fake-ctx width path.** The floor test's fake ctx read the weight word as `NaN` and collapsed width to a constant; fixed to parse the px size, and added a width-bound floor case.
- **P3 — stale docstring.** The numeric spec docstring now documents that `drawWidget` reads `theme()` but bails at the ctx guard.

Gates: `snc` + lint clean, full suite **1563/1563**. Rebuilt and redeployed to the dev boat; verified the served bundle is the fixed build.

**Consciously deferred** (low-reachability hardening / nits, not blocking):
- Halo-color cache-key staleness on the 5 non-numeric widgets, and the overflow-without-halo coupling when `cardColor` resolves falsy — both require a theme that changes `cardColor` independently / leaves the CSS var empty, which the three shipped themes don't.
- The 12-positional-param growth (options-object refactor) and naming the halo stroke constants.
- A comment explaining why racer-line's `TTL`/`TTB` captions are out of the label/unit scope.

PR #410, opened 2026-07-26
fix(charts): round line joins to stop miter-spike flicker on narrow viewports

mairas:
## Problem

On mobile phones and any narrow viewport, Skip's time-series graph widgets flicker/jitter as they scroll. Instead of the line simply scrolling left, the thick polyline's corners are drawn as sharp, tall **spikes** that appear and disappear frame-to-frame.

## Root cause

Chart.js strokes line datasets with its element-default `lineJoin: 'miter'` and never sets `ctx.miterLimit`, so the canvas default (`10`) applies. At a sharp corner a miter juts out to a point whose length grows as the angle sharpens; when it crosses the miter limit the browser **discretely** switches the join from a drawn spike to a cut-off bevel. A realtime/streaming chart shifts every vertex's sub-pixel geometry each frame, so vertices sitting near the threshold flip spike↔bevel independently — the flicker.

It's a narrow-viewport problem specifically because the same point count packs into far fewer horizontal pixels, so segments get steeper and corners sharper — many more vertices land near the threshold. The worst offender is `widget-data-chart`'s Value line (`tension: 0` straight-segment polyline, `borderWidth: 3`).

## Fix

Set the line-element default join to `round` once, in the shared `chart-registration.util` hub. Round joins have no miter point — nothing spikes and nothing flips, so the line is stable under sub-pixel motion. One property, in the one place every chart widget (data-chart, minichart, windtrends) already routes through. Caps are left at the default — the artifact is joins (corners), not line ends.

## Verification

- Rendered the **real chart.js** with `widget-data-chart`'s exact Value-line options at phone width (DPR 2), before (miter) vs after (round): the spikes are present, then gone.
- `npm run snc` + `npm run lint` green; all four chart-widget specs pass (25 tests), plus a new round-join regression assertion in the registration-util spec.
- Deployed the prod build to the live boat (`halpi`) and confirmed the served bundle carries `borderJoinStyle="round"` (fetched from Signal K over HTTPS, not just an HTTP-200).

## Test-mock note

The hub now *writes* `Chart.defaults.elements.line`, so all four self-contained `chart.js` spec mocks were completed with that shape (else the register body throws under the mock).

## Out of scope

Chart.js's `fastPathSegment` collapses points sharing an integer-x column into a min/max vertical stroke; that envelope still shifts as the chart scrolls, but that's the real data being drawn, not spikes. Round joins are the correct, complete fix for the reported flicker.

### DIFF
--- a/src/app/core/utils/chart-registration.util.spec.ts
+++ b/src/app/core/utils/chart-registration.util.spec.ts
@@ -6,7 +6,8 @@ import { describe, it, expect, vi } from 'vitest';
 // registry inside the test and re-import fresh so the guard starts clean regardless of run order.
 // Each mocked component is a distinct sentinel string so the assertion can name the exact set.
 vi.mock('chart.js', () => ({
-  Chart: { register: vi.fn() },
+  // defaults.elements.line: registerChartComponents() writes the round join default onto it.
+  Chart: { register: vi.fn(), defaults: { elements: { line: {} } } },
   LineController: 'LineController',
   LineElement: 'LineElement',
   PointElement: 'PointElement',
@@ -46,5 +47,15 @@ describe('registerChartComponents', () => {
       'annotationPlugin',
       'ChartStreaming'
     );
+  });
+
+  it('sets the line-element default join to round to prevent miter-spike flicker', async () => {
+    vi.resetModules();
+    const { Chart } = await import('chart.js');
+    const { registerChartComponents } = await import('./chart-registration.util');
+
+    registerChartComponents();
+
+    expect(Chart.defaults.elements.line.borderJoinStyle).toBe('round');
   });
 });
--- a/src/app/core/utils/chart-registration.util.ts
+++ b/src/app/core/utils/chart-registration.util.ts
@@ -24,6 +24,8 @@ let registered = false;
  * set is present. Registering the explicit union rather than every registerable lets the
  * bundler drop the unused controllers/scales on this Pi-class target. The `realtime`
  * scale used by the live-tail charts is contributed by the streaming plugin registration.
+ * It also sets the app-wide line-join default to round here, once alongside registration,
+ * so thick chart lines never render the miter spikes that flicker as a chart scrolls.
  */
 export function registerChartComponents(): void {
   if (registered) {
@@ -44,4 +46,12 @@ export function registerChartComponents(): void {
     annotationPlugin,
     ChartStreaming
   );
+
+  // Round line joins instead of Chart.js's default miter. At a sharp corner a miter juts to a
+  // point until its length crosses the canvas miterLimit, at which point the browser flips it to
+  // a bevel. As a realtime chart scrolls, each vertex's sub-pixel geometry shifts and vertices
+  // near that threshold flicker spike<->bevel frame to frame — most visibly on narrow viewports,
+  // where points pack into fewer pixels and corners are sharpest. Round joins have no miter point,
+  // so nothing spikes and nothing flips.
+  Chart.defaults.elements.line.borderJoinStyle = 'round';
 }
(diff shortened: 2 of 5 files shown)


mairas:

**Scope:** `fix/chart-miter-spike-flicker` vs `main` — 5 files (1 production, 4 spec). Merge-base `5e7b56a0`.
**Intent:** Flip the chart.js line-element default join from `miter` to `round` in the shared `chart-registration` hub to stop miter-join spike flicker on scrolling time-series charts; complete the four self-contained chart.js spec mocks for the hub's new `Chart.defaults.elements.line` write and add a regression assertion.
**Mode:** interactive. **Reviewers:** correctness, testing, maintainability, project-standards (4 always-on; no security/api-contract/reliability/adversarial — the change touches none of those surfaces and non-test churn is well under 50 lines).

### P3 — Low

| # | File | Issue | Reviewer(s) | Confidence | Route |
|---|------|-------|-------------|------------|-------|
| 1 | `src/app/core/utils/chart-registration.util.ts:19` | JSDoc enumerated only the registration behavior and omitted the new global side effect (setting `Chart.defaults.elements.line.borderJoinStyle`); the `registerChartComponents` name reinforces the incomplete picture. | maintainability | 0.66 | **applied** |

### Applied fixes

- **#1** — Added a sentence to the function JSDoc noting it also sets the app-wide round line-join default once, alongside registration. Comment-only; no runtime/type surface. (commit `c0a801e9`)

### Coverage

- **Reviewers:** 4/4 returned; no failures or timeouts. No untracked files in scope.

**Residual risks (report-only, all benign):**
- **Blast radius is wider than "three widgets."** `registerChartComponents()` has five callers — `widget-data-chart`, `minichart`, `widget-windtrends-chart`, plus `widget-history-chart-dialog` and `connection-status` — and the global `Chart.defaults` write rounds joins for all of them. Round joins are visually benign everywhere, so this is expected reach, just broader than the PR framing.
- **Silently overridable.** Any future per-dataset/per-widget `borderJoinStyle` wins over this global default (chart.js precedence); nothing at either site documents that a global default exists, so a later per-chart join could reintroduce flicker unknowingly.
- **Set once, never re-applied.** The write is inside the register-once guard. If some future code path reset `Chart.defaults.elements.line` after registration, the round join would be lost silently. No such reset exists today.

**Testing gaps (acknowledged, no action):**
- The regression assertion checks `borderJoinStyle === 'round'` against a mock whose `defaults.elements.line` is test-supplied, so it proves the function sets the property but not that chart.js honors it. The property name was independently confirmed against the installed chart.js (`chart.js:1868`, `:6468`). A future chart.js rename/deprecation would still pass the unit test.
- The mocks pre-create `defaults.elements.line` unconditionally, so the unit test does **not** defend the register-before-write ordering real chart.js requires (write-before-register throws in real chart.js but passes the mock). Mitigation: the two non-mocking callers (`connection-status`, `widget-history-chart-dialog`) import against real chart.js at module load, so a reorder would fail *those* specs in CI — the ordering invariant is guarded there.
- The user-visible outcome (spikes gone, no spike↔bevel flicker) is verified out-of-band by a headless-browser before/after screenshot and a live-boat deploy, not by a committed pixel-level test. Inherent to a canvas rendering default; acceptable.

Project-standards found no violations: no `any`, no null-safety regression, all four self-contained chart.js mocks completed consistently (the known CI-sharding mock-completeness hazard is addressed), and no widget-config/design-system files touched (no `gen:mcp-schema` needed).

---

No blocking findings across four independent Opus reviewers. The single P3 (JSDoc completeness) is applied. Residual risks are benign and expected for a global rendering default; the noted ordering/rendering test gaps are covered respectively by the real-chart.js caller specs and the out-of-band visual verification.

PR #411, opened 2026-07-26
fix(charts): thin trend-chart points to ~1/pixel to stop horizontal scroll jitter

mairas:
## Problem

On mobile phones and other narrow viewports, the time-series trend charts jitter/boil horizontally as they scroll — the polyline's features shimmer and jump left/right instead of gliding smoothly. This is the second half of the scroll-jitter report; the vertical miter spikes were fixed separately in #410, and this is what remained.

## Root cause

`deriveDataSourceInfo` (`chart-window.util.ts`) targets a **fixed ~500 samples per window** (`TARGET_POINTS_PER_WINDOW`), independent of how wide the widget actually renders. On a phone the plotting area is ~150–250px, so that's **~2–3 data points per horizontal pixel**. You can't draw more than one distinct feature per pixel, so that sub-pixel structure aliases and *boils* as the chart scrolls sub-pixel. chart.js's `fastPathSegment` (used for `tension:0` lines) amplifies it by collapsing points into shifting integer-pixel columns, but over-sampling is the disease — wide desktop charts (≤1 pt/px) are already smooth.

I confirmed this with a headless scrolling repro on the real chart.js, measuring a scroll-invariant jitter metric (mean |2nd-difference| of the line's per-column position across frames):

| Render | jitter (px) |
|---|---|
| current (~2.6 pt/px, fastPath) | 4.25 |
| tension `1e-6` (bypasses fastPath, full density) | 3.13 |
| **evenly-spaced ~1 pt/px** | **0.75** |

Two findings drove the fix: (1) reducing to ~1 pt/px is what actually matters (5.6× smoother); (2) it must be **evenly-spaced** — peak-preserving min-max and chart.js's own decimation still boil because they re-select shifting points as the window slides.

## Fix

Thin the **drawn** points to about one per pixel along the time axis, bucketing by **timestamp** so the kept set stays evenly spaced and stable under scroll. The bucket width is recomputed from the live plot size on each emission, so it adapts across a resize with **no reload**, and wide charts (samples already fit) keep every point. Rolling stats / SMA / min / max still track full-resolution upstream data — only rendering is thinned.

Deliberately *not* used: chart.js's built-in decimation. It bails on the streaming `realtime` axis (`chart.js` guards to `linear`/`time` only), and its min-max/lttb algorithms re-select shifting points, so they'd keep boiling anyway.

## Tradeoff

Inherent to any ~1 pt/px approach: a brief spike between kept samples isn't drawn (today it only surfaces *as* the jitter this removes). Preserving spikes needs a min/max band, which — per the metric above — reintroduces boiling. This chooses smooth over spike-preserving.

## Verification

- Scroll-invariant repro metric above (evenly-spaced thinning = the 0.75 result).
- Unit tests: a dense 91-sample batch thins to ~one point per 30s bucket on a measured 20px plot; an unmeasured plot draws every sample (fallback). `npm run snc` + `npm run lint` green; full data-chart spec passes.
- Deployed a combined build (this + #410) to the live boat for on-phone confirmation of the scrolling behavior.

Independent of #410 (branched from main; #410 touches `chart-registration.util`, this touches `widget-data-chart`) — they compose cleanly and can merge in either order.

### DIFF
--- a/src/app/widgets/widget-data-chart/widget-data-chart.component.spec.ts
+++ b/src/app/widgets/widget-data-chart/widget-data-chart.component.spec.ts
@@ -182,4 +182,42 @@ describe('WidgetDataChartComponent', () => {
     expect(title).toContain('knots');
     expect(title).not.toContain('celsius');
   });
+
+  // A 91-sample batch spanning 90s of the default 600s window. Reused by the thinning tests below.
+  const denseBatch = (): IDatasetServiceDatapoint[] => {
+    const batch: IDatasetServiceDatapoint[] = [];
+    for (let t = 0; t <= 90_000; t += 1_000) batch.push({ timestamp: t, data: { value: t / 1_000 } });
+    return batch;
+  };
+  const drawnCount = (): number => fixture.componentInstance.lineChartData.datasets[0].data.length;
+
+  it('thins a dense batch to about one point per pixel on a narrow plot', async () => {
+    const emissions$ = new Subject<IDatasetServiceDatapoint[]>();
+    historyMock.getBackfillThenLive.mockReturnValue(emissions$);
+    await setup(makeConfig({ showAverageData: false }));
+
+    // A measured 20px plot over the 600s window buckets to 30s, so the 91-sample / 90s batch
+    // collapses to one evenly spaced point per bucket instead of shimmering at ~2.6 samples/px.
+    (fixture.componentInstance as unknown as { chart: { chartArea: unknown } }).chart.chartArea =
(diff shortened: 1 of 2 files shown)


mairas:
Closing as superseded. This density-thinning approach was aimed at the wrong lever: the horizontal scroll jitter is Chart.js's fast line path re-bucketing points into integer-pixel columns under the streaming plugin's per-frame re-path, not point density — thinning to ~1/px cleans up the line spatially but does nothing for the scroll shimmer. The actual fix is #412 (draw the value line off the fast path via a hair of tension), verified smooth on the live boat.

Issue #428, opened 2026-07-27
Effort B — Slim path config: source + choice controls (windsteer/racesteer/windtrends)


mairas:
Effort B of the specialized-widget config cleanup epic #418. **Depends on Effort A (#427) merged** — B inherits A's single cadence control, so B's dialog work is purely the picker/Source/choice.

## Goal

On the three wind-family specialized widgets, remove the redundant path autocomplete: keep **Data Source** per path, add a **choice** control where a path has genuine alternatives.

**Requirements:** R1 (fixed path, Source still editable), R2 (choice not autocomplete), R3 (per-path Source), R7 (generic widgets unchanged — the non-regression guard lives in B1).

## Choice cases (real SK alternatives)

| Slot | Default | Alternatives (`pathOptions`) |
|---|---|---|
| True wind angle | `environment.wind.angleTrueWater` | `angleTrueWater` / `angleTrueGround` |
| Heading | `navigation.headingTrue` | `headingTrue` / `headingMagnetic` |
| Course over ground | `navigation.courseOverGroundTrue` | `courseOverGroundTrue` / `courseOverGroundMagnetic` |
| Wind direction (windtrends) | `environment.wind.directionTrue` | `directionTrue` / `directionMagnetic` |

## Units

- [ ] **B1: Dialog decoupling + choice control** *(the load-bearing P1 from the review)* — Files: `path-control-config` (html+ts), `root-modal-widget-config.ts` (`generateFormGroups`, `hasConfigurablePaths`), `paths-options` if needed (+ specs). Must:
  - **Drive Source off the resolved stored path, not `path.valid`.** A fixed path renders disabled → `.valid` false → today runs `disablePathFields()` which resets+disables Source. Populate `availableSources` + enable Source from the stored path for fixed/choice rows.
  - **Reconcile the required-Source validator:** default Source to `'default'`/'Any' so the form is **saveable without live data**; define the offline-path state (disabled + hint, not empty-required-invalid).
  - **Choice paths never disable the group** (the select writes `path`) — branch `generateFormGroups` disable + `hasConfigurablePaths` on `pathOptions`, not solely `isPathConfigurable`.
  - **On choice flip:** recompute `availableSources`; reset a now-invalid pinned Source to 'Any'.
  - Keep generic widgets (no `pathOptions`, `isPathConfigurable:true`) byte-for-byte unchanged.
[...]

Issue #418, opened 2026-07-27
Epic — Specialized-widget config cleanup (cadence model + slim path config)


mairas:
**Epic.** Continues the redundant-path-removal thread (#415 position, #417 heel-gauge/horizon). Restructured after a six-persona document review into **three cleanly-separated efforts**, each its own PR.

## Efforts

- [x] **#427 — Effort A: Cadence model** — replace per-path `sampleTime` (a display-refresh throttle, misnamed) with one widget-level `updateInterval`; delete the per-path field. All path-bearing widgets. **Do first.**
- [x] **#428 — Effort B: Slim path config** — remove the redundant path autocomplete on windsteer/racesteer/windtrends; keep per-path Data Source, add a choice control for genuine alternatives. Depends on A.
- [x] **#429 — Effort C: Autopilot** — re-scoped; reuse autopilot's existing heading toggle instead of a path choice. Independent PR.
- [x] **#433 — Effort D (discovered): relocate updateInterval to the always-shown Display tab (it and the data-timeout controls were hidden by Effort C's suppressed Paths tab); keep the enableTimeout toggle, fix the TTL value at 5s. PR #434.**

**Sequencing:** A → B (B inherits A's single cadence control). C independent.

## Why it matters

Four specialized widgets expose a full SK path autocomplete for paths that are semantically fixed by the widget's purpose — a user should never retype `self.environment.wind.angleApparent`. `isPathConfigurable` is also one flag doing three jobs (hide picker + disable the whole group incl. Source + suppress the tab), so "fix the path but keep Source editable" is impossible today. And per-path `sampleTime` is a misnamed display-cadence throttle that should be a single widget-level value.

## Review outcome (2026-07-27)

Six personas (coherence, feasibility, product-lens, design-lens, scope-guardian, adversarial) reviewed the original single-effort plan against the live code. Key results folded into the re-cut:

- **Split into A/B/C** — the cadence remodel and autopilot were riding on a picker-cleanup goal (scope-guardian, product-lens).
- **Cadence = relocation, not rename** — one widget-level `updateInterval`, per-path `sampleTime` deleted. No two-name state, no deferred 393-ref rename; **dissolves** the "dialog re-adds stripped sampleTime" P1 by construction.
- **Autopilot pulled out** — its true-wind path is never observed (dead UI) and heading is toggle-switched, not a path choice (feasibility, adversarial).
[...]

PR #417, opened 2026-07-27
Heel-gauge & horizon: repair stranded attitude path settings (v15→v16)

mairas:
## Problem

Widgets whose paths are all fixed showed a widget-settings **Paths tab** with nothing useful to edit. `widget-heel-gauge` / `widget-horizon` additionally had a dead `pathType: 'number'` picker on the `navigation.attitude` object leaf (the position bug #414 class) — the boat's heel-gauge was stranded at v15 (`isPathConfigurable: true`) after #415, which rewrote only position.

## Fix

- **Config dialog** — suppress the Paths tab when a widget has **no user-configurable path** (`hasConfigurablePaths`). Array-form widgets (`multiChildCtrls`: `widget-boolean-switch`, `widget-zones-state-panel`) build their paths *through* that tab, so they're explicitly kept.
- **Default TTL on** — the Paths tab also hosted the data-timeout control, so the four all-fixed-path widgets (`heel-gauge`, `horizon`, `racer-timer`, `racer-line`) now default `enableTimeout: true` / `dataTimeout: 5`; stale data blanks the gauge/readout. Attitude widgets keep `pathType: 'number'` so the pipeline extracts the roll/pitch sub-field then converts rad→deg (documented at the defaults).
- **Migration v15 → v16** — enables the timeout on all four types, and for the two attitude widgets resets every stored path to the fixed default (`self.navigation.attitude`, `pathType: 'number'`, `isPathConfigurable: false`, `convertUnitTo: 'deg'`), discarding stale overrides. Racer paths are already scalar/fixed, so only their timeout changes. Bumps `LATEST_APP_CONFIG_VERSION → 16` with the usual couplings (boot gate `liveVersion === 15`, dispatch `case 15`, `version === 15` branch, perf-harness seed, regenerated schema, fixed `V16_MIGRATION_OUTPUT_VERSION`).
- **Seed** — heel-gauge/horizon set to `enableTimeout: true` (racers aren't seeded); seed guard test extended per-type.

## Review (two rounds, findings fixed)

- **P1 (fixed):** the first cut of `hasConfigurablePaths` hid the tab for array-form widgets (`paths: []`), breaking control creation for fresh boolean-switch/zones-panel instances — now excluded via `multiChildCtrls`, with a dialog test.
- **P2 (resolved):** the rule also caught the racer widgets; per the maintainer's decision they're included in the "hide the tab" rule and get the TTL-on default (not left stuck off).
[...]

PR #430, opened 2026-07-27
refactor(config): one widget-level updateInterval, drop per-path sampleTime (Effort A, #418)

mairas:
## Why

Per-path `sampleTime` on `IWidgetPath` is a misnomer and a false generality. In `widget-streams.directive.ts` it feeds an RxJS `sampleTime()` — an **output** throttle that bounds how often the widget re-emits and repaints; it does **not** change how often Signal K sends data. And it's a widget-level concern: verified across all 19 path-bearing widgets, **18 use a uniform per-path value** (only `widget-racer-line` diverges, 1000/500). So it should be one value in one place.

This is **Effort A of #418** (the foundation for the specialized-widget config cleanup) — #428 (slim path config) inherits the single cadence control from here.

## How

- **Interfaces** — add `IWidgetSvcConfig.updateInterval` (ms); remove `IWidgetPath.sampleTime`.
- **Runtime** (`widget-streams.directive.ts`) — read `cfg.updateInterval` with the existing coercion preserved (`Number()`, `!isFinite || <=0 → 1000`; not a literal `??`). Cadence moves into the widget-level (root) signature, so a change rebuilds every path's pipeline while reusing the cached base observable.
- **Dialog** — one "Display update interval" control (in the Paths tab, beside the data-TTL block) replaces the per-path Sample Time field. The `path-control-config` auto-add block that would otherwise re-inject a per-path `sampleTime` on save (via `getRawValue()`) is removed.
- **Widgets** — every path-bearing DEFAULT_CONFIG swept: per-path `sampleTime` removed, widget-level `updateInterval` added at its uniform value. `widget-racer-line` (the only divergent one) collapses to **500** (min). Five widgets that read `pathCfg.sampleTime` at runtime now read `cfg.updateInterval`. `dashboard-history-series-sync` reads the widget cadence for auto history series.
- **Migration `v16 → v17`** — generic collapse: for every widget, `updateInterval = min(per-path sampleTimes)` (fallback 1000), per-path `sampleTime` deleted. Full coupling checklist wired (version const, boot gate, `runUpgrade` branch, `migrateOneAppVersion` case, perf-harness seed, regenerated `skip-dashboard-schema.json`, version-hardcoding specs). Both seeded widgets (`widget-wind-steer`, `widget-autopilot`) in `config.blank.dashboard.ts` carry the new shape; wind-steer's stale seed cadence (8×100) is normalized to 1000.

## Testing

- `npm run ci` green locally: lint, snc, **1598 unit** + **7 plugin** + **31 mcp-schema** tests.
- New coverage: directive three-way cadence resolution (updateInterval / absent→1000 / 0→1000) + rebuild-on-change; v16→v17 migration (min-collapse, default/array/missing tolerance, skip-non-v16); v16 boot-gate; seed guard asserts `updateInterval`.

## Not yet run (pre-merge)

- `./run deploy-halos local` + served-bundle check on the boat.
- Confirm no deployed boat relies on a non-standard published path on the swept widgets (escape-hatch note — relevant to Effort B, not this PR).

Client-only change (config model + dialog); no server/runtime service impact.
- **What to watch:** after deploy, open a dashboard with a pre-v17 stored config and confirm the upgrade toast runs once and reloads; widgets render and update at their expected cadence.
- **Healthy signal:** browser console shows `[Upgrade] Collapsed per-path sampleTime ... on N widget(s)` then a single reload; no repeated upgrade loop.
- **Failure signal / rollback trigger:** an upgrade→reload loop (config not stamping v17), or widgets stuck/blank. Rollback = redeploy the previous build (stored configs already at v17 are forward-compatible: the runtime ignores absent per-path cadence and falls back to 1000ms).
- **Validation window/owner:** first boat reload after deploy; me.

Part of #418. Effort A.


mairas:
## Code review — Effort A (widget-level `updateInterval`)

**Scope:** `refactor/widget-update-interval` vs `main` (base `327bebcf`), 50 files, 1 commit.
**Reviewers:** correctness, testing, maintainability, project-standards (always-on) + api-contract (type/config-version/MCP-schema contracts) + adversarial (50+ line data-migration). All 6 returned.
**Verdict: Not ready — 2 P1s must fix first.** Both are clear, contained fixes. CI is green locally, but two of these are CI-invisible.

### P1 — Must fix

| 1 | `widget-boolean-switch` / `widget-zones-state-panel` `.component.ts` | **Config dialog crashes for a newly-added instance.** Both array-form widgets (`multiChildCtrls: []`) render the Paths tab, which now binds the new **required** `updateInterval` control — but their DEFAULT_CONFIG has no `updateInterval`, so the control resolves to `null` and `[formControl]="null"` throws. Migrated instances survive (the migration stamps it); fresh ones can't be configured. | correctness, adversarial, api-contract | 0.92 |
| 2 | `tools/schematics/create-host2-widget/**` | **`npm run generate:widget` scaffold is broken.** The template still emits per-path `sampleTime: 1000` inside an `IWidgetSvcConfig` literal and no `updateInterval`, so the next generated widget fails `snc` (excess property on `IWidgetPath`) and ships no cadence. No CI job exercises the schematic → invisible here. | project-standards | 0.85 |

### P2 — Should fix

| 3 | `root-modal-widget-config` / `paths-options` specs | **New `updateInterval` dialog control has no test** — the `case "updateInterval"` (required validator) + getter + save round-trip never execute; `paths-options` spec only asserts `toBeTruthy()`. This is why the P1 crash passed CI. | testing | 0.85 |
| 4 | `paths-options.component.html` + `root-modal…ts` | **`min="100"` is an HTML hint only** — the reactive control carries `Validators.required` but no `Validators.min`, and the runtime floors only `<=0` → 1000. A sub-100 cadence (e.g. 10 ms) can be typed, saved, and drives heavy repaints on the Pi target. | adversarial, maintainability, project-standards | 0.75 |
| 5 | `config.blank.dashboard.ts` vs `configuration-upgrade.service.ts` | **Seed↔migration cadence divergence for wind-steer.** The seed now declares `updateInterval: 1000` (normalized), but the migration min-collapses a stored default-dashboard wind-steer (stale `100`×8) to `100`. So a fresh install runs at 1000 ms and an upgraded default-dashboard user at 100 ms (10× load). A **decision**, not a code bug — see below. | adversarial, api-contract | 0.85 |
| 6 | `root-modal-widget-config.component.html:761` | **Widget-level cadence is trapped in the path-gated Paths tab.** `updateInterval` is a widget-level setting, but its only editor renders when `hasConfigurablePaths` is true — suppressed for all-fixed-path widgets (heel-gauge/horizon) since #417. They ship an uneditable cadence. (Effort B re-exposes their tab, so this self-resolves there.) | maintainability | 0.62 |

### P3 — Discretion

| # | File | Issue | Reviewers |
|---|------|-------|-----------|
| 7 | `widget-racer-line.component.ts:62` | Comment narrates the change ("Previous per-path sample times … collapsed") — violates CLAUDE.md "no narrating the change / current state, not dev history". | maintainability, project-standards |
| 8 | `widget-streams.directive.ts` / `configuration-upgrade.service.ts` | The `1000` default cadence is a bare literal in three places (directive, migration, doc); the gauges fall back to `500`. Extract a shared named constant. | maintainability |
| 9 | `gauge-animation.util.ts:18` | `gaugeAnimationDurationMs(sampleTimeMs)` param still named `sampleTime` though callers now pass `updateInterval` (pre-existing). | maintainability |

### Cleared by review (worth recording)

- **MCP schema is NOT a regression** — it never modeled widget-level settables (`enableTimeout`/`dataTimeout`/`color` live only in `defaultConfig`); dropping the per-path slot field and carrying `updateInterval` in `defaultConfig` keeps it self-consistent. (api-contract)
- **No autopilot nested-config migration gap** — autopilot's embedded configs are component-rebuilt signals (`updateInterval: 500`), not stored or migrated, so no stale nested `sampleTime` survives. (api-contract)
- **Root-signature rebuild is correct** — changing only `updateInterval` sets `rootChanged`, rebuilds each path pipeline, and does **not** re-acquire/release the cached base (no leak/over-release). (correctness, adversarial)
- **Full config-version coupling checklist applied** — all couplings verified present. (project-standards)

### Testing gaps

- No spec renders the config dialog for an array-form widget (would have caught P1 #1).
- Migration `>0` filter not tested with a present-but-zero/negative `sampleTime`; idempotency (re-run v17 on v17) only implicit.
- Seed↔migration cadence parity untested (would have caught #5).
- Directive runtime coercion (absent/non-finite/≤0 → 1000) — *actually covered* by two new directive tests, contrary to one reviewer's note.

### Decision for the author (finding #5)

The min-collapse is the right general policy (18/19 widgets are uniform, so min = the value). Only wind-steer diverges, and only for users still on the **default/starter dashboard** (real users who added wind-steer via the picker migrate to 1000 correctly). Options: (a) accept — 100 ms is more responsive, not broken, and self-corrects when the user touches settings; (b) add a seed↔migration parity test and normalize. Leaning (a) + a note.

### Fix plan (author will address before merge)

Must-fix #1 (add `updateInterval` to the two array-form widgets + a dialog-render regression test), #2 (schematic template + index.js + README + regen-verify). Should-fix #3 (dialog control tests), #4 (`Validators.min(100)`), #7 (drop the narrating comment), #8 (shared default constant). #5 pending the author's call; #6 deferred to Effort B; #9 pre-existing, skipped.

mairas:
## Review fixes applied (folded into `26af1efc`)

All CI green after fixes (**1601** unit + 7 plugin + 31 mcp-schema).

**P1 — fixed**
- **Dialog crash on array-form widgets** — added `updateInterval: 1000` to `widget-boolean-switch` and `widget-zones-state-panel` DEFAULT_CONFIG (matches what the migration stamps). Regression test added in `root-modal-widget-config.spec` that drives the **real shipped configs** through the form-builder and asserts the control is non-null.
- **Broken `generate:widget` scaffold** — template now emits `updateInterval: 1000` and no per-path `sampleTime` (also `index.js` + `README.md.template`). Verified by generating a throwaway widget: the `sampleTime` excess-property error is gone. *(Note: the scaffold still has two pre-existing strictNull errors — `runtime.options()` possibly-undefined — unrelated to this change; worth a separate follow-up.)*

**P2 — fixed**
- Added a `Validators.min(MIN_UPDATE_INTERVAL_MS)` (100) to the `updateInterval` control so a sub-100 cadence can't be saved; new dialog test covers required + min.
- Added a migration test for present-but-zero/negative `sampleTime` (excluded from the min).

**P3 — fixed**
- Removed the "narrates the change" comment in `widget-racer-line`.
- Extracted `DEFAULT_WIDGET_UPDATE_INTERVAL_MS` / `MIN_UPDATE_INTERVAL_MS` constants (referenced by the directive, migration, and dialog) to stop the fallback-cadence drift.

**Deferred / decisions**
- **#5 (seed↔migration wind-steer divergence)** — left as-is pending the author's call. The min-collapse is correct for the 18/19 uniform widgets; only a default-dashboard wind-steer that carried the stale `100` migrates to 100 while a fresh install gets the normalized 1000. Recommendation: accept (more responsive, not broken; self-corrects once Effort B exposes the control), or add a seed↔migration parity test.
- **#6 (Paths-tab cadence trap)** — deferred to Effort B (#428), which re-exposes the fixed-path widgets' tab.
- **#9 (`gaugeAnimationDurationMs` param name)** — pre-existing, out of scope.

Issue #429, opened 2026-07-27
Effort C — Autopilot path config (re-scoped: reuse existing toggle)


mairas:
Effort C of the specialized-widget config cleanup epic #418. **Independent PR**, pulled out of the wind-family sweep after the document review.

## Why separate

feasibility + adversarial confirmed autopilot does NOT fit the wind-family choice pattern (verified in `widget-autopilot.component.ts`):
- Heading is switched by the existing `autopilot.headingDirectionTrue` toggle — `observe('headingTrue')` vs `observe('headingMag')` — not by a path picker. A `pathOptions` true/magnetic choice would duplicate or contradict the toggle.
- `angleTrueWater` is configured but **never observed** — exposing it as a choice is dead UI.
- Autopilot has its own `select-autopilot` tab plus nested `paths` blocks — a materially different config surface.

## Goal

Remove autopilot's redundant top-level path pickers without duplicating its existing toggle.

## Unit

- [ ] **C1: Autopilot path config, re-scoped**
  - Heading/COG: reuse the existing `headingDirectionTrue`/`courseDirectionTrue` toggles; do not add a contradicting `pathOptions` choice.
  - Drop the dead `angleTrueWater` path rather than expose it.
  - Verify each top-level path is actually consumed before removing/exposing.
  - Own migration + full couplings; **autopilot IS seeded in `config.blank.dashboard.ts`** — update the seed + guard.

## Verification

`npm run ci` green; the dedicated autopilot tab AND the Paths controls both work in the running app; deploy-halos check on the boat.

mairas:
**Implemented in PR #432** (branch `feat/autopilot-path-config`). Investigation confirmed the re-scope: heading is toggle-switched (`headingDirectionTrue`), `windAngleApparent` is observed+displayed, and `windAngleTrueWater` is never observed. So headingMag/headingTrue/windAngleApparent become fixed and windAngleTrueWater is dropped. No configurable COG path exists, so the `courseDirectionTrue` mention had nothing to remove.

[...]

PR #431, opened 2026-07-27
feat(widget-config): slim wind-family path config — source + choice controls (Effort B, #418)

mairas:
## Why

**Effort B of #418** (#428). On the three wind-family specialized widgets — `windsteer` (type `widget-wind-steer`), `racesteer`, `windtrends-chart` — the Signal K path autocomplete is redundant: the paths are fixed by the widget's purpose. This removes the picker where it's pointless, keeps per-path **Data Source** selectable, and offers a small **choice** control where a path has genuine alternatives. Depends on Effort A (merged) — the widget-level cadence control is already in place.

## How

- **Decouple Data Source from the path picker** (the load-bearing change the Effort-A review traced). A path can now be **fixed** (no autocomplete) or a **choice** (`pathOptions` select) while its Source stays editable. Source is populated off the *resolved stored path* (not `path.valid`) and defaults to `'Any'`, so a fixed/choice path's Source is valid and **saveable without live data**. `generateFormGroups` disables only the `path` control (not the whole group) for a fixed path; `hasConfigurablePaths` surfaces the tab for choice paths too.
- **`pathOptions` field** (new, `IWidgetPath`): a closed set of `{ label, path }` alternatives → renders a Material `<select>` instead of the autocomplete; the chosen path isn't live-validated (the alternatives are known-good).
- **Choice slots** (per widget): true-water/true-ground wind angle, true/magnetic heading, true/magnetic COG (windsteer/racesteer); true/magnetic wind direction (windtrends). Descriptions neutralized ("Heading", not "True Heading") so the label doesn't assert one option. Every other configurable path → `isPathConfigurable: false`, Source-only.
- **`v17 → v18` migration**: resets each wind-family widget's swept paths to their fixed/choice default (path, editability, neutralized choice description), keyed by runtime type; hard-reset (escape-hatch loss accepted). The wind-steer seed is normalized to match (fixed slots, canonical paths, choice labels). Full coupling checklist wired (version const, boot gate, runUpgrade branch, migrateOneAppVersion case, `V18_MIGRATION_OUTPUT_VERSION`, perf-harness seed, regenerated schema, version-hardcoding specs). No `VERSION` bump — this is the 2nd PR in the 1.0.5 cycle.

**Paths tab identity:** kept "Paths" (it's shared with generic widgets that still pick paths there); a per-widget rename was judged more complexity than value.

## Testing

- `npm run ci` green: lint, snc, **1614 unit** + 7 plugin + 31 mcp-schema.
- New: Source-decoupling tests (fixed/choice path Source enabled+valid+saveable offline; pinned-source reset on choice flip); generic-widget non-regression; `generateFormGroups`/`hasConfigurablePaths` branch tests; v17→v18 migration (choice reset, description neutralize, fixed-slot flip, stale-path correction, non-wind untouched, skip-non-v17); v17 boot-gate.

## Not yet run (pre-merge)
- `./run deploy-halos local` + on-boat check of the three widgets' settings dialogs.

Client-only (config model + dialog); no server/runtime service impact.
- **Watch:** open a windsteer/racesteer/windtrends settings dialog after deploy — fixed paths show a Source dropdown (no autocomplete), choice slots show a labelled select, form saves.
- **Healthy:** console shows `[Upgrade] Reset N wind-family path(s)…` then a single reload on a pre-v18 config; widgets render.
- **Failure/rollback:** upgrade→reload loop, or a settings dialog that won't save. Rollback = redeploy previous build (v18 configs are forward-compatible: an older build reads the fixed paths and just doesn't offer the choice/Source UI).
- **Window/owner:** first boat reload after deploy; me.

Part of #418. Effort B.


mairas:
## Code review synthesis — 6 personas (correctness, adversarial, testing, maintainability, project-standards, api-contract)

Scope: `feat/slim-path-config` vs `main` (base `8660315b`), 18 files. All six ran in independent fresh contexts.

**No P0 or P1 findings.** The core design holds: fixed paths disable only the `path` control (value retained via `getRawValue()`), `setupSourceFor` keeps Source valid/saveable offline, choice paths keep `path` enabled for the select, the generic autocomplete branch is byte-unchanged, and the migration relies correctly on the runtime `merge(base, user)` to source `pathOptions` from `DEFAULT_CONFIG`. Adversarial explicitly disproved the feared Save-block, the `isPathConfigurable` merge-resurrection, and the windtrends `speedMeasure` break. Migration `type` keys verified against the three component selectors; VERSION correctly untouched at 1.0.5 (2nd PR in the cycle).

### Confirmed findings + dispositions

**1. Decoupling is not wind-scoped — autopilot's fixed paths now expose editable Source (correctness, P2, conf 0.78).**
Verified: `widget-autopilot` DEFAULT_CONFIG mixes `isPathConfigurable:true` (heading/wind) with `:false` (8 internal state/mode/engaged/target/rudder/XTE paths), so `hasConfigurablePaths` is true and its fixed paths now render a persisted Data Source picker they never had pre-PR. Autopilot is deliberately deferred to Effort C (#429). **Disposition: needs the maintainer's scope call** — accept the general decoupling (+ add an autopilot root-modal spec) vs. guard autopilot out of Effort B. Flagged to the user.

**2. Migration keeps the stored `source` pin while resetting the path (adversarial, P2, conf 0.63).**
Verified: `upgradeConfigV17toV18` rewrites `path`/`isPathConfigurable`/`description` but never `source`. A user who pinned e.g. `trueWindAngle`'s source to the device publishing the old `angleTrueGround` gets the path flipped to `angleTrueWater` while the pin survives → `computePathSignature` bakes source into the key → subscribes to a bucket the new path never fills → true-wind silently blank after auto-upgrade. The migration's own comment says it discards overrides; `source` is the one it doesn't. **Disposition: fix** — reset `source` to `default` for each swept slot, matching the stated intent.

**3. Opening config while a path is offline downgrades a valid pinned source to 'Any' and persists on Save (adversarial, P3, conf 0.62).**
Verified in `setupSourceFor`: `getPathObject()===null` (path offline / boat at dock) collapses `availableSources` to `['default']`, and the guard overwrites a still-valid pin. Before this PR the fixed-path group was disabled, so its source was never mutated. **Disposition: fix** — distinguish "offline/unknown" from "source removed"; keep the stored pin selectable when the path has no live data yet.

**4. MCP `pathSlots` summary omits `pathOptions` (api-contract, P2 advisory, conf 0.70).**
The four choice slots report `isPathConfigurable:true` with no options in the model-facing `pathSlots` summary, while `defaultConfig` carries the closed set. A model driving off the summary sees them as free paths. **Disposition: fix** — add `pathOptions` to the `PathSlot` type + extractor (cheap, correct).

**5. Wind-family path shape now lives in three hand-maintained places with no cross-check (maintainability, P2 advisory, conf 0.72).**
Widget `DEFAULT_CONFIG`, the `V18_WIND_PATH_DEFAULTS` migration snapshot, and the DefaultDashboard seed must agree; the seed had *already* drifted (`angleTrueGround`/`speedOverGround`) and this PR silently corrected it — no test caught the drift. **Disposition: fix** — add a spec asserting the seed's wind-family entries equal the widget `DEFAULT_CONFIG` (migration map stays a point-in-time literal).

### Test-coverage gaps (convergent across reviewers) — all to be added

- **Migration coverage for racesteer + windtrends** — flagged by 4 reviewers; only `widget-wind-steer` is exercised, so a typo in either other map key/path migrates to a no-op while stamping v18.
- **DEFAULT_CONFIG shape guard** — no test asserts the three choice slots carry `pathOptions` + the rest are `isPathConfigurable:false`; `widget-racesteer` has no spec file at all.
- **Choice-select wiring** — `onPathChoiceChange` + the `ngOnInit` choice branch + the template render are untested; only `setupSourceFor` is called directly.
- **`getRawValue()` retention** — no test proves a disabled fixed path's value survives Save (the whole fixed-path design rests on it).
- **Autopilot root-modal spec** — pending the scope decision above.

### Residual risks → release notes (intended, not defects)

- **Escape-hatch loss:** v17→v18 hard-resets wind-family path overrides with no backup slot. User-approved (reset-like-#417), but silently rewrites customized dashboards on upgrade.
- **Default true-wind reference change:** migrated + fresh dashboards move `angleTrueGround→angleTrueWater` and `speedOverGround→speedTrue`. This is the seed-drift correction (now matches each widget's `DEFAULT_CONFIG`), but it is a user-visible display change on upgrade.

mairas:
## Review findings addressed (4 commits, CI green locally: 1626 unit + 7 plugin + 32 mcp-schema)

- **`a86cef1a` — migration source pin (adversarial P2):** v17→v18 now resets `source` to `default` on each swept path, so a leftover pin can't starve the reset path. + racesteer & windtrends-chart migration coverage (was wind-steer only).
- **`0852e9ff` — offline source clobber (adversarial P3):** `setupSourceFor` now distinguishes "path offline / no live data yet" from "pinned source removed" and keeps a valid pin (surfacing it so the select still shows it) instead of overwriting it to "Any" and persisting the loss on Save. + choice-select wiring test (init routing + `onPathChoiceChange` recompute).
- **`f2c68a90` — MCP pathSlots (api-contract P2):** `pathOptions` now carried onto `PathSlot` + the extractor, so a choice slot's closed set surfaces in the model-facing summary the same way it does in `defaultConfig`. Schema regenerated.
- **`55eb24b5` — shape/seed drift (maintainability P2 + testing):** DEFAULT_CONFIG shape guard for all three wind widgets (racesteer had no spec at all), a seed-vs-default cross-check for wind-steer (the seed had silently drifted), and a `getRawValue` retention guard for disabled fixed paths.

### Still open — one scope decision for the maintainer

**Correctness P2 (autopilot):** the Source/path decoupling is general, so `widget-autopilot` (Effort C, #429) now renders editable Data Source pickers on its 8 fixed internal paths — an out-of-scope, unreviewed change to a safety widget under a wind-only PR. Deferred to a maintainer call: accept the general decoupling (+ add an autopilot root-modal spec) vs. guard autopilot out of Effort B. Not merging until resolved.

### Release-note items (intended, not defects)
- v17→v18 hard-resets wind-family path **and source** overrides (escape-hatch loss, reset-like-#417) — silently rewrites customized dashboards on upgrade.
- Default true-wind reference moves `angleTrueGround→angleTrueWater` / `speedOverGround→speedTrue` (seed-drift correction; now matches each widget's DEFAULT_CONFIG) — a user-visible display change on upgrade.

mairas:
**Autopilot scope resolved (maintainer): accept the general decoupling.** `isPathConfigurable` now means "path fixed", not "hide Source" — so autopilot's fixed internal paths keep an editable Data Source (only the path control disabled). Pinned by a root-modal non-regression spec (`3f5437de`) so Effort C (#429) revisits autopilot's config shape deliberately rather than a regression flipping it back. All findings addressed; CI green locally (1627 unit + 7 plugin + 32 mcp-schema). Ready for your merge call.

PR #432, opened 2026-07-28
Effort C — Slim autopilot path config (reuse toggle, drop dead true-wind slot)

### LINKED ISSUE #429: Effort C — Autopilot path config (re-scoped: reuse existing toggle)

mairas:
## Effort C — Autopilot path config (epic #418, closes #429)

The last piece of the specialized-widget config cleanup. Autopilot exposed a full Signal K path autocomplete for paths that are semantically fixed by the widget's purpose. Investigation of `widget-autopilot.component.ts` (which paths are actually `observe()`d) confirmed the re-scoped plan:

- **`headingMag` / `headingTrue`** — heading is switched by the existing `autopilot.headingDirectionTrue` toggle (`observe('headingTrue')` vs `observe('headingMag')`), *not* a picker → made **fixed**. No `pathOptions` choice (would duplicate the toggle).
- **`windAngleApparent`** — observed and displayed on the gauge, no meaningful alternative → made **fixed**.
- **`windAngleTrueWater`** — **never observed anywhere** (dead UI) → **dropped**.
- No configurable COG path exists, so the plan's `courseDirectionTrue` mention had nothing to remove.

### Outcome: autopilot's Paths tab is suppressed (maintainer decision)

With every path now fixed and no choice, `hasConfigurablePaths` is false, so the **#417 rule hides autopilot's entire Paths tab** — it joins heel-gauge/horizon as a no-Paths-tab widget, with config living on its purpose-built `select-autopilot` tab. This reverses the visible effect of Effort B's general-decoupling decision for autopilot (the fixed-path Source pickers go with the tab). **Confirmed by the maintainer** as the intended landing; the negligible trade-off is losing Source-pinning on autopilot's single-source standard SK paths.

### Migration (v18 → v19)

Fixes the three pickers (path unchanged, so a pinned source stays valid and is **kept** — unlike the wind sweep, no path/source mismatch to guard against) and **deletes** `windAngleTrueWater` outright (a lodash base⊕user merge would otherwise resurrect the orphan key). Full coupling checklist: `LATEST_APP_CONFIG_VERSION 18→19`, boot gate, `runUpgrade` branch, dispatch case, `V19_MIGRATION_OUTPUT_VERSION`, perf-harness seed, regenerated `skip-dashboard-schema.json`, DefaultDashboard seed (which carried its own `angleTrueGround`-under-"True Water" drift, now removed).

No VERSION bump — 3rd PR in the still-draft 1.0.5 cycle.

**Verification:** `npm run ci` green (1632 unit + 7 plugin + 33 mcp-schema). Living-plan note: updated the Effort-B autopilot spec (it asserted `hasConfigurablePaths === true`) to the new suppressed-tab reality.


mairas:
## Code review synthesis — 6 personas (PR #432, Effort C)

Scope: `feat/autopilot-path-config` vs `main` (base `2ea31013`). All six ran in independent fresh contexts.

**Five of six clean. No P0/P1.** Correctness verified `windAngleTrueWater` is genuinely dead (zero refs in component/html/svg-autopilot), fixed paths still stream (`observe()` never gates on `isPathConfigurable`), the migration is idempotent + guards null/partial configs + keeps the pinned source (path unchanged), and every coupling site is covered. Project-standards verified the full coupling checklist, a byte-identical schema regen, clean tsc/eslint, and VERSION correctly held at 1.0.5. Api-contract verified all three migration arrival paths reach v19 and the schema is internally consistent. Maintainability: zero findings (drift guards already in the CI gate).

### The one real finding (adversarial, low, conf 0.72) — a decision, not a defect

**Suppressing autopilot's Paths tab also removes the only UI for the widget-level controls, not just the path pickers.** `paths-options` hosts `updateInterval`, `enableTimeout`/`dataTimeout`, and `filterSelfPaths`, and it renders only inside the `hasConfigurablePaths`-gated Paths tab. With every autopilot path now fixed, that tab is suppressed, so those controls become uneditable in-app (the `select-autopilot` tab doesn't host them). Concretely: a user can no longer enable/retune a stale-data timeout on this **steering** widget, nor change its `updateInterval` (the control Effort A introduced). The migration preserves any stored `enableTimeout`, so runtime behavior is retained — but it's no longer user-adjustable. This is the maintainer-accepted #417 pattern (autopilot joins heel-gauge/horizon, which lost the same), surfaced because autopilot is a control widget where the data-timeout carries more weight.

**Status: reopened with the maintainer** — the original tab-suppression decision was made on an under-stated description of the loss (Source-only). Awaiting the call: ratify the full loss (+ regression test + a follow-up issue to decouple widget-level controls from the path-picker gate) vs. keep autopilot's Paths tab.

### Testing finding (low) — to be added regardless
The v18→v19 autopilot transform is exercised via `runUpgrade(18)` but not through the `migrateImportedConfig` import route (that matrix uses empty dashboards). Add one import-route case with an autopilot payload asserting fix+delete+source-keep+stamp-19.

### Residual risks / release notes (intended, not defects)
- Old (v18) build reading a v19 config is a graceful no-op (standard forward-compat); if it writes back it down-stamps to 18 — inherent to any bump, note for staged rollouts.
- Pinned-source recovery: once the tab is gone, a pinned autopilot source that later stops emitting has no in-app fix (edit raw config / recreate widget). Narrow.
- At release time, VERSION must bump so the narrowed MCP schema carries a new `skipVersion` (currently versioned only via `configVersion`).

PR #434, opened 2026-07-28
Effort D — updateInterval editable on every widget; data-timeout fixed at 5s

### LINKED ISSUE #433: Effort D — Relocate updateInterval to General tab (always editable); fix data-timeout at 5s
Part of the specialized-widget config cleanup epic #418. Surfaced while implementing Effort C (#429/PR #432): making every autopilot path fixed suppresses its Paths tab (#417), which also hides the widget-level controls that live in `paths-options` — the `updateInterval` (Effort A's display cadence) and the data-timeout controls, not just the path pickers.

## Decisions (maintainer)
- **`updateInterval` must be editable on every widget that uses it.** It's a `WidgetStreamsDirective` (+ a few gauge animation) concept; all 21 widgets that use it already carry it, but it's currently rendered only inside the `hasConfigurablePaths`-gated Paths tab, so the all-fixed widgets (autopilot, heel-gauge, horizon) can't reach it. The 15 widgets without `updateInterval` don't use the mechanism (electrical/ais via their own services; iframe/video/label static) — **relocate only, no retrofit**.
- **Data-timeout is not worth configuring.** `dataTimeout` is already universally `5` — hard-code it to a `5s` constant and drop its UI. `enableTimeout` stays a baked per-widget default (attitude gauges blank, most don't); its UI goes too.

## Scope
- Move the `updateInterval` control out of `paths-options` into the always-shown General settings tab (`root-modal-widget-config`), gated on `widgetConfig?.updateInterval !== undefined`.
- Remove `updateInterval` / `enableTimeout` / `dataTimeout` controls from `paths-options` (keep `filterSelfPaths` + path controls).
- `widget-streams.directive`: `dataTimeout` → `5s` constant (stop reading `cfg.dataTimeout`); keep reading `cfg.enableTimeout`.
- No config-version bump (no shape change / migration). Specs updated.

## Sequencing
Prerequisite for Effort C — lands first, then PR #432 rebases on top so autopilot keeps `updateInterval` editable in the General tab despite its suppressed Paths tab.

mairas:
## Effort D — updateInterval editable on every widget; data-timeout fixed at 5s (closes #433)

Part of epic #418, surfaced while implementing Effort C (#432): making every autopilot path fixed suppresses its Paths tab (#417), which also hid the widget-level controls that live in `paths-options` — not just the path pickers.

### updateInterval relocation
The display-update-interval control lived in `paths-options`, rendered only inside the `hasConfigurablePaths`-gated Paths tab. So the all-fixed widgets (autopilot, heel-gauge, horizon) couldn't reach it once their tab was suppressed. It now lives in the **always-shown Display tab, before the branch split** (gated on `widgetConfig?.updateInterval !== undefined`), so it also covers widgets with a custom Display body — notably autopilot, whose Display tab is `<select-autopilot>`, not the general grid. Verified consumer-tracing: `updateInterval` is a `WidgetStreamsDirective` (+ a few gauge-animation) concept; every widget that uses it already carries it, so this is a **relocation, not an addition** — the 15 widgets without it (electrical/ais/embedded) don't use the mechanism (maintainer decision, #433).

### Data-timeout hard-coded
`dataTimeout` was already universally `5` across every widget, so it's now a `5s` constant in `widget-streams` and its UI is gone. `enableTimeout` stays a baked per-widget default (attitude gauges blank on stale data, most widgets don't); `dataTimeout` leaves the root signature since it can no longer vary. No behavior change — the value was 5 everywhere already.

No config-version bump (no config-shape change, no migration). `dataTimeout`/`enableTimeout` stay in stored configs (the former inert, the latter still read).

**Verification:** `npm run ci` green (1628 unit + 7 plugin + 32 mcp-schema). New guard: `updateIntervalToControl` resolves for an all-fixed-path widget whose Paths tab is suppressed.

### Sequencing
Prerequisite for Effort C — **lands first**, then PR #432 (autopilot) rebases onto it so autopilot keeps `updateInterval` editable in the Display tab despite its suppressed Paths tab.

### DIFF
--- a/src/app/core/directives/widget-streams.directive.spec.ts
+++ b/src/app/core/directives/widget-streams.directive.spec.ts
@@ -471,25 +471,24 @@ describe('WidgetStreamsDirective', () => {
         vi.useFakeTimers();
         // Silence noisy console logs from timeout/retry handling to keep test output clean
         vi.spyOn(console, 'log');
-        // Configure a very short timeout (seconds) so the test runs fast
+        // The TTL is a fixed 5 s (no longer configurable); a stored dataTimeout is ignored.
         const cfg = makeCfg({ path: 'env.to', source: null, pathType: 'string', updateInterval: 100, displayName: 'Test', enableTimeout: true, dataTimeout: 0.02 });
         directive.setStreamsConfig(cfg);
 
         const hits: string[] = [];
         directive.observe('p', u => hits.push(String(u?.data?.value)));
 
-        // Do not emit anything; advance virtual time beyond 20ms to trigger timeout
-        await vi.advanceTimersByTimeAsync(30);
+        // Do not emit anything; advance virtual time beyond the fixed 5 s window to trigger timeout.
+        await vi.advanceTimersByTimeAsync(5100);
         expect(dataSvc.timeoutCalls.length).toBe(1);
-        // dataTimeout (0.02 s) is threaded through as the ms window; compute it the same way the
-        // directive does so the assertion is exact regardless of float representation.
-        expect(dataSvc.timeoutCalls[0]).toEqual({ path: 'env.to', source: 'default', pathType: 'string', dataTimeoutMs: 0.02 * 1000 });
+        // The window is the fixed constant, not the stored (0.02 s) dataTimeout.
+        expect(dataSvc.timeoutCalls[0]).toEqual({ path: 'env.to', source: 'default', pathType: 'string', dataTimeoutMs: 5000 });
     });
 
     it('forwards a configured non-default source into timeoutPathObservable', async () => {
         vi.useFakeTimers();
         vi.spyOn(console, 'log');
-        const cfg = makeCfg({ path: 'env.to', source: 'n2k-1', pathType: 'string', updateInterval: 100, displayName: 'Test', enableTimeout: true, dataTimeout: 0.02 });
+        const cfg = makeCfg({ path: 'env.to', source: 'n2k-1', pathType: 'string', updateInterval: 100, displayName: 'Test', enableTimeout: true });
         directive.setStreamsConfig(cfg);
 
         const hits: string[] = [];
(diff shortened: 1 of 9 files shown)


mairas:
## Code review synthesis — 5 personas (PR #434, Effort D)

Scope: `feat/update-interval-general-tab` vs `main`. **No P0/P1.** Correctness verified the relocation renders for autopilot's custom Display branch and is hidden for data-chart (no null binding); the hard-coded 5s TTL is behavior-neutral for every shipped default (all `dataTimeout: 5`); lint/snc clean; no version bump needed (no config-shape change); no dangling refs to the removed getters/inputs; `paths-options` is used only by root-modal.

### One decision (adversarial, medium, conf 0.72) — reopened with the maintainer

**Hard-coding the TTL to 5s + removing the enableTimeout toggle strands slow-path widgets.** The 5s TTL is neutral only for fast paths. A widget a user bound to a *slow* path (tank level, temperature, battery SoC updating every 15–60s) with `enableTimeout:true` now blanks to "--" for most of each update cycle — and with both TTL controls removed there's **no in-app recovery** (can't lengthen the TTL, can't turn the timeout off). Default configs are safe (streaming widgets ship `enableTimeout:false`; the 4 that ship `true` bind to fast IMU/timer paths), so this only bites a user who deliberately enabled a long TTL on a slow path. The directive was "hard-code the *value*" — removing the enable *toggle* too was my overreach. Resolution pending: keep the enableTimeout toggle (relocated to the Display tab) as the escape hatch vs. accept the loss.

### Clean fixes (batching regardless of the decision)

- **Inert `dataTimeout` field** (maintainability, medium): nothing reads `cfg.dataTimeout` anymore, but the interface comment still says it sets the timeout (and its unit note is wrong). Annotate it inert; keep it in configs to avoid a shape-change/migration.
- **Orphaned required validator** (maintainability medium + adversarial/testing residual): `generateFormGroups` still builds `dataTimeout` with `Validators.required` but nothing renders it — a config carrying `dataTimeout:null` would silently disable Save. Drop the validator.
- **Relocation not render-tested** (testing, medium): the new test proves the form control exists, not that it renders in the Display tab. Add a DOM render test with an autopilot config (`detectChanges` + query `input[name=updateInterval]`) — the relocation's whole point.
- **Stale JSDoc** (correctness P3 + maintainability low): `computeRootSignature` / class docs still list `dataTimeout`; remove it.
- **Comment-narration nits** (project-standards ×2): reword the const + spec comments to present-tense state (drop "No longer configurable" / "Effort D: … hid it before").

### Residual / note
`enableTimeout` remains honored at runtime (not dead) — a legitimate per-widget author default; keep it read.

mairas:
**Review findings addressed** (CI green locally: 1629 unit + 7 plugin + 32 mcp-schema).

- `12d7a121` — dropped the orphaned `dataTimeout` required validator, marked the field inert in the interface, fixed the stale streams JSDoc, and added a render-level guard that `updateInterval` shows on the Display tab when the Paths tab is suppressed.
- `4d41543e` — **restored the stale-data-timeout toggle** (maintainer decision): only the TTL *value* is fixed at 5s; the `enableTimeout` checkbox moves to the Display tab beside `updateInterval`, so a slow-path widget can still turn the timeout off (the escape hatch the adversarial review flagged). Comment-style nits reworded to present tense.

No P0/P1; the substantive finding (slow-path TTL with no escape hatch) is resolved by keeping the toggle. Ready for merge review.

PR #436, opened 2026-07-28
Fix: toolbar tap ignored after a flick-to-reveal (browser fling suppresses the click)

mairas:
### Symptom
On touch devices, after flicking to reveal the auto-hiding toolbar, the **first** tap on the lock button was often ignored; a second tap worked. Reproduced on a Samsung tablet and a Pixel 10 XL (Brave), and Hermit.

### Root cause
The reveal-swipe, when done as a **fast flick**, is recognised by Chromium as a **fling** — even though the gesture shell has `touch-action: none`. `touch-action: none` stops the *scroll*, but not the *fling recognition*. Chromium then applies its tap-suppression-after-fling: the tap that "stops" the fling has its synthesized `click` swallowed. Since the toolbar buttons use a native `(click)`, that tap does nothing. A slow pull (no momentum) generates no fling, so it always worked — which is exactly what distinguished the failing cases.

### How it was found
Instrumented the live browser over ADB + CDP: every dead tap showed a clean `pointerdown`+`pointerup` on the button with **no `click`**, and the failure correlated with a velocity-dependent window after the reveal-swipe (dead ≲ ~800ms, click after) that only appeared for a real hardware finger (synthetic/CDP touch never reproduced it — it doesn't drive the browser's fling recogniser). A user-confirmed test settled it: **flick suppresses, slow pull doesn't.** A `preventDefault`-on-touchmove patch injected live eliminated it.

### The fix
`gesture.directive` registers a **non-passive `touchmove`** on its host that `preventDefault`s while a gesture is being tracked — so the browser recognises no scroll/fling and never arms tap-suppression. Scoped:
- gated on `this.tracking` (the active-gesture window only), **and**
- **yields (does not `preventDefault`) when the touch is inside a scrollable descendant**, so routed full-page views (settings / remote / connection / help) and lists keep touch-scrolling. This mirrors the wheel-yield already in `scroll-nav.directive`. The claim-vs-yield decision is made once per gesture (the touch target is stable for the sequence) to avoid per-move layout reads.

Pointer-event–based swipe detection is untouched (it doesn't rely on touch events).

### Verification
- `npm run ci` green (1636 unit + 7 plugin + 33 mcp-schema); added two directive unit tests: touch-move is claimed only while tracking, and is **not** claimed when the touch targets a scrollable child.
- Deployed to the boat and confirmed on a Samsung tablet: flick-to-reveal + immediate lock tap now registers on the first tap, **and** Settings/Remote/Help still touch-scroll (no scroll/swipe/page-nav regressions).

No VERSION bump (later PR in the 1.0.5 cycle).

### DIFF
--- a/src/app/core/directives/gesture.directive.spec.ts
+++ b/src/app/core/directives/gesture.directive.spec.ts
@@ -186,6 +186,46 @@ describe('GestureDirective', () => {
         expect(swipeLeftCount).toBe(0);
     });
 
+    it('claims touch-move only while tracking a gesture (stops the fling that swallows the next tap)', () => {
+        component.mode = 'all';
+        syncFixture();
+        const touchMovePrevented = () => {
+            const e = new Event('touchmove', { bubbles: true, cancelable: true });
+            hostEl.dispatchEvent(e);
+            return e.defaultPrevented;
+        };
+
+        // Idle (not tracking): touch-move passes through, so scrolling in dialogs/sheets is unaffected.
+        expect(touchMovePrevented()).toBe(false);
+
+        // A gesture starts on pointerdown -> subsequent touch-moves are claimed so the browser
+        // recognises no scroll/fling (a fling would make Chromium suppress the click of the tap that
+        // stops it, losing a flick-to-reveal-then-tap).
+        dispatchPointerEvent(hostEl, 'pointerdown', { clientX: 50, clientY: 20, pointerId: POINTER_ID, pointerType: 'touch' });
+        expect(touchMovePrevented()).toBe(true);
+
+        // After release, tracking ends -> touch-move passes through again.
+        dispatchPointerEvent(hostEl, 'pointerup', { clientX: 50, clientY: 20, pointerId: POINTER_ID, pointerType: 'touch' });
+        expect(touchMovePrevented()).toBe(false);
+    });
+
+    it('yields touch-move to a scrollable descendant so routed pages/lists keep touch-scrolling', () => {
+        component.mode = 'all';
+        syncFixture();
+        // A scrollable child with content overflowing (jsdom has no layout, so stub the metrics).
+        const scroller = document.createElement('div');
+        scroller.style.overflowY = 'auto';
+        Object.defineProperty(scroller, 'scrollHeight', { value: 400, configurable: true });
+        Object.defineProperty(scroller, 'clientHeight', { value: 100, configurable: true });
+        hostEl.appendChild(scroller);
+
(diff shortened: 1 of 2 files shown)


mairas:
## Review synthesis (six personas: correctness, adversarial, testing, maintainability, project-standards, api-contract)

The independent review caught a **P1 regression in the first version of this fix** and it has been addressed. Posting the synthesis for the record.

### P1 (fixed) — `preventDefault`-while-tracking blocked touch-scrolling on routed pages
- **Found by:** correctness (0.65) + adversarial (0.75), converging.
- **Claim:** the original scoping rationale ("gesture hosts are `touch-action:none`, so nothing there touch-scrolls") was false. An ancestor's `touch-action:none` does **not** stop a descendant auto-scroller — proven by config dialogs scrolling under `body{touch-action:none}`. The shell gesture host (`mode='all'`) wraps `<router-outlet>`, so the routed full-page views (Settings / Remote / Connection / Help), which have their own `overflow-y:auto`, are descendants of the host and had their touch-scroll killed by the unconditional `preventDefault`.
- **Confirmed on-device:** with the first build deployed, Settings and Help would not touch-scroll.
- **Fix:** `onTouchMove` now walks `ev.target` → host and **yields** (no `preventDefault`) when any ancestor is a scroll container with overflowing content — reusing the same scroll-container test as `scroll-nav.directive`. Dialogs/sheets were never affected (CDK overlays live outside the gesture hosts). Re-verified on-device: flick-fix holds **and** Settings/Help scroll.

### P2 (addressed) — perf of a non-passive listener on the Pi-class target
- **Found by:** adversarial. A per-move `getComputedStyle` walk on every tracked touchmove is avoidable. The claim-vs-yield decision is now computed **once per gesture** and cached (`touchClaimDecided`/`claimTouchMoves`); the target is stable for a gesture, so subsequent moves are a single boolean check.

### Testing
- Added a spec asserting a touchmove targeting a scrollable child is **not** prevented (the yield path), alongside the existing "claimed only while tracking" spec. jsdom 26.1 enforces the passive flag, so the non-passive registration is genuinely exercised.

### Not a regression (verified, out of scope)
- **Help scrolls only a few pixels.** This is **pre-existing** and independent of this PR: Chromium promotes `.markdown-content`'s `overflow-y` to `auto` (it sets `overflow-x:scroll`), so the directive detects it as scrollable and **yields** — Help touch-scroll is unchanged from before this fix. The real cause is the year-old `.markdown-content { height: calc(100% - 75px) }` (upstream #548) whose percentage height doesn't resolve against the routed-page host chain. Tracked separately, not in this PR (one logical change per PR).

Issue #439, opened 2026-07-28
Theme selector: add "Follow system" (three-way system / light / dark)


mairas:
## Request

The theme selector should offer **three** options — **Follow system**, **Light**, **Dark** — instead of today's single light/dark boolean.

## Current state

- Theme is stored as `SettingsService._themeName` (`src/app/core/services/settings.service.ts:33`), default `''` (dark).
- The picker is a single boolean slide-toggle `isLightTheme` in `src/app/core/components/settings/display/display.component.ts:47` (template `display.component.html:52-55`); on save it maps to `'light-theme'` or `''` (`display.component.ts:104-108`).
- Applied via an effect that toggles `body.light-theme` (`app-service.ts:92-99`); CSS blocks + `mat.theme()` are keyed on that class (`src/styles.scss:16-197`).
- There is **no** `prefers-color-scheme` / OS-following anywhere (only `matchMedia` use is DPR detection in `canvas.service.ts:209`). The existing `autoNightMode` follows a Signal K `environment.mode` path, **not** the OS — orthogonal to this.

## Direction (decisions, not code)

- Replace the boolean toggle with a tri-state selector (Follow system / Light / Dark). Store the *user choice* as a small enum, distinct from the *resolved* theme actually applied.
- For "Follow system": resolve via `window.matchMedia('(prefers-color-scheme: dark)')` and **subscribe to its `change` event** so the app re-resolves live when the OS flips — then apply through the same body-class path.
- Preserve the night-mode overlay (`night-theme`) as an independent concern; the tri-state selector picks the *base* theme only.
- Migration: `themeName` currently stores `''`/`'light-theme'`. Decide whether existing stored values map to explicit Light/Dark or are re-interpreted; a stored explicit value should NOT silently become "follow system".

## Decision: default is Light

New users default to **Light**. Today's stored default is `''` (= dark), so this changes the new-user default — the shipped default config / `SettingsService` default (`settings.service.ts:22` `defaultTheme`) must be set to Light explicitly, and any first-run/unset path must resolve to Light rather than dark.

## Open questions

- UI control: a 3-button segmented toggle vs a select — match existing settings patterns.

## Dependencies

Live OS-theme following only recolors widgets correctly once the theme-reload bug (companion issue) is fixed — widgets otherwise keep a stale snapshot until reload.

## Verification

All three modes apply live on the boat; toggling the OS theme with "Follow system" active flips Skip live; `npm run ci` green.

mairas:
[...]

PR #444, opened 2026-07-28
Fix theme change requiring a reload for widget colors (#438)

### LINKED ISSUE #438: Theme change requires a page reload for widget colors to update
## Symptom

After changing the theme (light ⇄ dark), Material chrome and CSS-driven surfaces recolor immediately, but **widget colors don't update until the page is reloaded** — gauges, the D3 wind radar, Chart.js trend charts and the SVG steer widgets keep the previous theme's colors.

## Root cause

Widgets don't read CSS live. `AppService.readThemeCssRoleVariables()` (`src/app/core/services/app-service.ts:142-182`) snapshots every `--skip-*` / `--mat-sys-background` custom property from `getComputedStyle(document.body)` into a plain JS `ITheme` object and pushes it onto the `cssThemeColorRoles$` BehaviorSubject (`:180`). Widgets consume that snapshot as a signal (`widget-host2.component.ts:73`, fed to each widget via `inputBinding('theme', …)` at `:156`) and draw from it.

That snapshot is (re)published from only two places:
- `app-service.ts:130` — once, in the constructor at startup.
- `app-service.ts:226` — inside `toggleDayNightMode()` (night-mode only).

The light/dark theme-switch effect at **`app-service.ts:92-99`** toggles the `body.light-theme` class (so pure-CSS surfaces re-cascade live) but **never calls `readThemeCssRoleVariables()` and never re-emits `cssThemeColorRoles$`**. So widgets keep the stale startup snapshot until a reload re-runs the constructor.

## Suggested fix

In the theme-change effect (`app-service.ts:92-99`), after toggling the body class, re-read and re-publish the theme snapshot — exactly as `toggleDayNightMode()` already does at `:226`. The class toggle must complete before the `getComputedStyle` re-read so the new values are current.

## Verification

Switch light ⇄ dark on the boat with at least one gauge, one trend chart, and a windsteer widget on the dashboard; all widget colors update **without** a reload. `npm run ci` green.

Note: this is effectively a prerequisite for a live "follow system" theme option (companion issue) to recolor widgets when the OS theme changes.

mairas:
Fixes #438.

## Symptom
After switching the theme (light ⇄ dark), Material chrome and CSS-driven surfaces recolored immediately, but **widget colors stayed on the old theme until a full page reload** — gauges, the D3 wind radar, Chart.js trend charts and the SVG steer widgets.

## Root cause
Widgets don't read theme CSS live. `AppService` snapshots the `--skip-*` / `--mat-sys-background` custom properties into a plain JS `ITheme` object and publishes it on `cssThemeColorRoles$`; widgets consume that snapshot (via `widget-host2`'s `inputBinding('theme', …)`) and draw from it.

That snapshot was only (re)read in the constructor (startup) and inside `toggleDayNightMode()` (night mode). The **light/dark theme-switch effect toggled the `body.light-theme` class but never re-published the snapshot**, so widgets kept the stale startup values until a reload re-ran the constructor.

## Fix
Re-read and re-publish the snapshot inside the theme-change effect, right after the class toggle — exactly what `toggleDayNightMode()` already does. `getComputedStyle` forces a synchronous style recalc, so the just-applied class is reflected in the new snapshot. No new signal dependencies are read, so the effect doesn't self-trigger.

## Verification
- `npm run ci` green (1637 unit + 7 plugin + 33 mcp-schema). Added a spec asserting the theme-change effect publishes a **fresh** snapshot (mirrors the existing night-toggle snapshot test); it fails before the fix (same object reference) and passes after.
- Deployed to the boat and confirmed light ⇄ dark recolors gauges, trend charts, and the windsteer widget **without** a reload.

This is the basis for the theme cluster: it's also a prerequisite for a live "Follow system" option (#439) recoloring widgets on OS-theme changes.

### DIFF
--- a/src/app/core/services/app-service.spec.ts
+++ b/src/app/core/services/app-service.spec.ts
@@ -98,6 +98,16 @@ describe('AppService', () => {
       TestBed.tick();
       expect(document.body.classList.contains('light-theme')).toBe(false);
     });
+
+    it('publishes a fresh theme color snapshot when the theme changes', () => {
+      const service = createService();
+      const before = service.cssThemeColorRoles$.getValue();
+      settings.themeName.set('light-theme');
+      TestBed.tick();
+      const after = service.cssThemeColorRoles$.getValue();
+      expect(after).not.toBe(before);
+      expect(service.cssThemeColors).toBe(after);
+    });
   });
 
   describe('setBrightness', () => {
--- a/src/app/core/services/app-service.ts
+++ b/src/app/core/services/app-service.ts
@@ -96,6 +96,10 @@ export class AppService {
         // Remove the light theme class if it exists
         document.body.classList.remove('light-theme');
       }
+      // Re-snapshot the theme CSS custom properties so widgets (which draw from the
+      // JS snapshot, not live CSS) recolor without a reload. getComputedStyle here
+      // forces a synchronous style recalc, so the class change above is reflected.
+      this._cssThemeColorRoles = this.readThemeCssRoleVariables();
     });
 
     effect(() => {


mairas:
## Review synthesis (correctness · testing · maintainability · project-standards)

Four always-on personas, each in its own fresh context. Diff is ~6 non-test lines — under the adversarial/size threshold, no auth/API/reliability surface, so no conditional reviewers. **No P0/P1/P2 findings.** The three concerns flagged at dispatch (effect re-entrancy, startup double-publish, night-mode interaction) were each traced and cleared.

### Concerns cleared
- **Re-entrancy / infinite effect** — `readThemeCssRoleVariables()` reads **no** signals (only `document.body` / `getComputedStyle` / `getPropertyValue` / `cssThemeColorRoles$.next`). The effect's only tracked dep stays `this._theme()`. No self-retrigger. *(correctness)*
- **Startup double-publish** — the constructor's synchronous publish (`:130`) still runs before any widget injects the root service, so `requireSync` consumers (widget-host2 et al.) never see the `null` seed. The effect's initial run adds one content-equal republish → one redundant boot recolor. Harmless. *(correctness, maintainability)*
- **Night-mode interaction** — the diff changes only the snapshot re-read; the light-theme class toggling is unchanged, so DOM state after a base-theme change under night mode is identical to before — the snapshot now just reflects current DOM instead of staying stale. *(correctness)*

### P3 — Low
| 1 | `app-service.ts:142-182` (+ `:130`, `:~102`, `:226`) | `readThemeCssRoleVariables` is a mixed command/query and the field-cache assignment is triplicated across 3 sites; a future 4th trigger that forgets the field write would leave the `cssThemeColors` getter stale while subscribers stay current | maintainability | 0.62 | Filed as follow-up **#445** (reviewer explicitly recommended keeping this PR minimal) |

### Coverage
- **Suppressed:** 0 findings below the 0.60 gate.
- **Testing gap (inherent jsdom limit, converged across all 4 reviewers):** the added spec proves a **fresh snapshot object is republished** on theme change (object-identity is the only viable proxy — jsdom's `getComputedStyle` returns `''` for every `--skip-*` custom property regardless of the body class), but it **cannot** verify the colors actually differ between light/dark, nor the `getComputedStyle`-forces-recalc ordering the fix relies on. Both are real-browser-only → covered by **on-device verification** (deployed to the boat; light ⇄ dark recolors gauges/charts/windsteer without reload).
- **Residual (pre-existing, not introduced):** red-night-mode leaves both `night-theme` + `light-theme` classes on a light→night transition; the pre-existing always-true `classList.toggle` in the same effect. Neither touched by this PR.

### Verdict: **Ready to merge** (pending on-device confirmation of live recolor)
No blocking findings. Unit coverage proves the re-publish mechanism; the user-visible recolor is being confirmed on the boat. The one P3 is tracked in #445.

PR #447, opened 2026-07-28
Night mode: force a dark base regardless of selected theme (#443)

### LINKED ISSUE #443: Night mode should always use a dark base (not dim the light theme)
## Request

Night mode should **always render on a dark base**, in addition to its existing adjustments (red overlay, or dimming + sepia/hue). Right now, if the user's selected theme is Light, night mode sits on top of the light base instead of forcing dark.

## Root cause

`AppService.toggleDayNightMode()` (`src/app/core/services/app-service.ts:204-227`) has two night-mode branches, and the non-red one keeps the light base:

- **Dimming branch** (`redNightMode` off, `:210-216`): applies `setBrightness(nightBrightness, true)` (dim + sepia/hue) but then **re-applies `light-theme`** if the user's theme is light (`:212-214`). Result: night mode over a light theme is just a *dimmed light* theme — never dark.
- **Red branch** (`redNightMode` on, `:207-208`): adds the `night-theme` class but does **not remove `light-theme`**, so both classes coexist and the result depends on CSS cascade order rather than a deliberate dark base.

The separate light/dark effect at `:92-99` only reacts to the theme signal, so it doesn't re-assert anything when night mode toggles — class management during night mode is entirely inside `toggleDayNightMode()`.

## Direction (decisions, not code)

When `isNightMode()` is true, force the dark base in **both** branches — i.e. ensure `light-theme` is removed regardless of the user's selected theme — then layer the night adjustment on top:
- Red branch: dark base + `night-theme` (red) overlay.
- Dimming branch: dark base + dimming + sepia/hue filter.

When night mode turns off, restore the user's selected theme (the existing `:220-224` restore path already does this). Verify the `light-theme`-vs-`night-theme` class interaction and cascade so the intended dark tokens win.

Note: `toggleDayNightMode()` already re-publishes the widget theme snapshot (`:226`), so widgets recolor live on night-mode changes (this is why night mode already updates widgets while a plain light/dark switch does not — see #438).

## Verification

On the boat, with the base theme set to **Light**: enabling night mode (both red and dimming variants) yields a dark base plus the night adjustment, not a dimmed-light look; disabling night mode returns to Light. Repeat with base Dark (should be unchanged). `npm run ci` green.

mairas:
Closes #443.

Night mode is meant to layer its adjustments (red `night-theme`, or dimming + sepia) over a **dark base**, but it kept the light base when the user's selected theme was light.

## Root cause & fix
Two managers of the `light-theme` class both had to respect night mode:
1. `toggleDayNightMode()` — the non-red **dim** branch re-added `light-theme` when the theme was light (so night mode just *dimmed* light); the **red** branch never removed it. → Now a single `document.body.classList.remove('light-theme')` is hoisted to the top of the night-mode branch, forcing a dark base in both sub-branches.
2. The **theme-change effect** — toggled `light-theme` purely from the theme signal, so switching the base theme to light *while night mode was active* re-added it. → Now gated: `applyLight = theme === 'light-theme' && !isNightMode()`.

Both keep `light-theme` off whenever night mode is active, and restore it (idempotently, order-independent) when night ends. The #438 live-recolor snapshot re-read is preserved in the effect. Reading `isNightMode()` in the effect adds no infinite-loop risk (`readThemeCssRoleVariables` reads no signals).

## Verification
- `npm run ci` green (1638 unit + 7 plugin + 33 mcp-schema).
- Tests: rewrote the night-mode test to assert the dark base (was asserting the old kept-light behavior), added a `light-theme`-removed assertion to the red-night test, and **added an interaction test** (switch base theme to light mid-night → stays dark; night off → restored). All three were empirically confirmed to fail on the old code.
- Deployed to the boat and confirmed: with base theme **Light**, night mode (dimming and red) renders on a dark base; turning night mode off returns to Light.

### DIFF
--- a/src/app/core/services/app-service.spec.ts
+++ b/src/app/core/services/app-service.spec.ts
@@ -145,22 +145,39 @@ describe('AppService', () => {
       expect(document.body.style.getPropertyValue('--skip-nightModeFilters')).toContain('sepia(0.5)');
     });
 
-    it('keeps the light-theme class in night mode when the theme is light', () => {
+    it('forces a dark base (removes light-theme) in night mode even when the theme is light', () => {
       settings.themeName.set('light-theme');
       const service = createService();
       service.isNightMode.set(true);
       service.toggleDayNightMode();
-      expect(document.body.classList.contains('light-theme')).toBe(true);
+      expect(document.body.classList.contains('light-theme')).toBe(false);
     });
 
     it('applies the red night theme at full brightness when red night mode is on', () => {
+      settings.themeName.set('light-theme');
       settings.redNightMode.set(true);
       const service = createService();
       service.isNightMode.set(true);
       service.toggleDayNightMode();
       expect(document.body.classList.contains('night-theme')).toBe(true);
+      expect(document.body.classList.contains('light-theme')).toBe(false);
       expect(document.body.style.getPropertyValue('--skip-nightModeBrightness')).toBe('1');
       expect(document.body.style.getPropertyValue('--skip-nightModeFilters')).toBe('');
+    });
+
+    it('keeps the dark base when the theme is switched to light while night mode is active', () => {
+      const service = createService();
+      service.isNightMode.set(true);
+      service.toggleDayNightMode();
+      expect(document.body.classList.contains('light-theme')).toBe(false);
+
+      settings.themeName.set('light-theme');
+      TestBed.tick();
+      expect(document.body.classList.contains('light-theme')).toBe(false);
+
+      service.isNightMode.set(false);
+      service.toggleDayNightMode();
+      expect(document.body.classList.contains('light-theme')).toBe(true);
     });
 
     it('publishes a fresh theme color snapshot on each toggle', () => {
--- a/src/app/core/services/app-service.ts
+++ b/src/app/core/services/app-service.ts
@@ -90,12 +90,9 @@ export class AppService {
     );
 
     effect(() => {
-      if (this._theme() === 'light-theme') {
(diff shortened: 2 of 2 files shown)


mairas:
## Adversarial verification (3 diverse lenses, all passed)

Implemented and verified via a parallel ultracode workflow; each lens ran in its own fresh context. **No must-fix findings.**

- **Correctness + effect↔toggleDayNightMode interaction** ✅ — `light-theme` is removed at the top of the night branch (both red + dim → dark base); the effect's `applyLight = theme==='light-theme' && !isNightMode()` blocks re-adding it mid-night; when night ends both the day branch and the effect converge to the same class state in **every interleaving** (order-independent), and the last handler re-snapshots so colors match the final DOM; no infinite loop (`readThemeCssRoleVariables` reads no signals); the day path is byte-identical.
- **Test-correctness (empirically proven)** ✅ — the lens reverted `app-service.ts` to HEAD and ran the suite: **exactly the 3 changed tests fail (30/33)**, and all 33 pass on the fix. The rewritten night test, the red-night `light-theme`-removed assertion, and the new interaction test each pin a specific guard; no surviving test asserts the old kept-light behavior. (The tree was restored after the experiment — I re-verified `app-service.ts` on disk holds the fix.)
- **Regression** ✅ — night-mode-off light/dark switching is unchanged; the #438 live-recolor snapshot re-read is preserved; `toggleDayNightMode` still republishes; auto-night and red-night effects still work. `ng test` → 33/33.

**Verdict: Ready to merge** (pending your on-device confirmation). CI green (1638 unit + 7 plugin + 33 mcp-schema).

PR #449, opened 2026-07-28
Three-way theme selector: Follow system / Light / Dark, default Light (#439)

### LINKED ISSUE #439: Theme selector: add "Follow system" (three-way system / light / dark)

mairas:
Closes #439.

Adds a third theme option — **Follow system** — alongside Light and Dark, and makes **Light** the default for new users. Builds on #438 (live recolor) and #443 (night mode dark).

## Model
`SettingsService.themeName` becomes a tri-state choice: `'system' | 'light-theme' | 'dark-theme'` (legacy `''` still means dark). A single resolver in `AppService` maps the choice to a light/dark base:
- `'light-theme'` → light
- `'system'` → follows a **live** `prefers-color-scheme` signal (`matchMedia` + `change` listener, cleaned up via `DestroyRef`)
- legacy `''` / `'dark-theme'` → dark

The resolver is shared by the theme-change effect **and** `toggleDayNightMode`, so: night mode still forces dark regardless of choice (#443), the effect re-resolves **live** when the OS flips (only while `'system'` is selected — the `prefersDark` signal is read lazily), and the #438 snapshot re-read still fires so widgets recolor without a reload.

## UI
`Settings > Display` swaps the light on/off slide-toggle for a segmented `mat-button-toggle-group` (System / Light / Dark). Load maps legacy `''` → Dark; save writes the chosen mode verbatim.

## Default & migration
Default flips to `'light-theme'` (`settings.service` `defaultTheme` + `DefaultThemeConfig`). **No config migration needed**: `themeName` lives in the *unversioned* `IThemeConfig` blob (passed through every migration untouched), and the resolver treats legacy `''` as dark — so existing users are unaffected (`''` stays dark, `'light-theme'` stays light); only new-user seeds change.

## Verification
- `npm run ci` green (1644 unit + 7 plugin + 33 mcp-schema; lint + snc clean). TDD: 4 new `AppService` `system` tests (OS-light→light, OS-dark→dark, **live re-resolve on OS change**, night-forces-dark) written red-first; 2 new display-component mapping tests; 3 `settings.service` default assertions updated to Light.
- Deployed to the boat and confirmed all three modes apply live, Follow-system tracks the device, and existing dark/light users are unchanged.

### DIFF
--- a/src/app/core/components/settings/display/display.component.html
+++ b/src/app/core/components/settings/display/display.component.html
@@ -45,15 +45,19 @@
           </mat-panel-title>
           @if (!isPhonePortrait().matches) {
             <mat-panel-description>
-              Choose your preferred high-contrast theme.
+              Choose light, dark, or follow the device theme.
             </mat-panel-description>
           }
         </mat-expansion-panel-header>
-        <mat-slide-toggle class="full-width"
-          [(ngModel)]="isLightTheme"
-          name="isLightTheme">
-            Enable light theme for brighter display settings.
-        </mat-slide-toggle>
+        <mat-button-toggle-group class="full-width theme-toggle"
+          [(ngModel)]="themeMode"
+          name="themeMode"
+          aria-label="Theme"
+          hideSingleSelectionIndicator>
+          @for (mode of themeModes; track mode.value) {
+            <mat-button-toggle [value]="mode.value">{{ mode.label }}</mat-button-toggle>
+          }
+        </mat-button-toggle-group>
       </mat-expansion-panel>
       <mat-expansion-panel>
         <mat-expansion-panel-header>
--- a/src/app/core/components/settings/display/display.component.scss
+++ b/src/app/core/components/settings/display/display.component.scss
@@ -5,6 +5,10 @@
   margin-top: 10px;
 }
 
+.theme-toggle mat-button-toggle {
+  flex: 1 1 0;
+}
+
 .sliders {
   margin: 5px 0px 5px 0px;
 }
--- a/src/app/core/components/settings/display/display.component.spec.ts
+++ b/src/app/core/components/settings/display/display.component.spec.ts
@@ -102,6 +102,29 @@ describe('SettingsNotificationsComponent', () => {
         expect(component).toBeTruthy();
     });
 
+    it('maps a legacy empty stored theme to Dark in the picker', () => {
+        // SettingsServiceMock.getThemeName() returns '' (legacy dark)
+        expect(component['themeMode']()).toBe('dark-theme');
+    });
+
+    it('loads a stored system theme into the picker unchanged', () => {
+        const settings = TestBed.inject(SettingsService) as unknown as SettingsServiceMock;
+        vi.spyOn(settings, 'getThemeName').mockReturnValue('system');
+        const fx = TestBed.createComponent(SettingsDisplayComponent);
+        fx.detectChanges();
+        expect(fx.componentInstance['themeMode']()).toBe('system');
+    });
+
+    it('saves the selected theme mode verbatim', async () => {
+        const settings = TestBed.inject(SettingsService) as unknown as SettingsServiceMock;
+        const setThemeName = vi.spyOn(settings, 'setThemeName');
+        component['themeMode'].set('system');
+        component['saveAllSettings']();
+        await flushPromises();
+        await flushPromises();
+        expect(setThemeName).toHaveBeenCalledWith('system');
+    });
+
     it('shows single warning prompt with Ok action when plugin is installed but not enabled/configured', async () => {
         // Toggle auto night mode on
         component['isAutoNightModeSupported']({ checked: true, source: {} as MatSlideToggle });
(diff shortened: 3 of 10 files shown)


mairas:
## Review synthesis (correctness · adversarial · api-contract · testing · maintainability · project-standards)

Six personas, each in a fresh context. The review found **one HIGH-severity crash** and strong test-gap convergence; both are now fixed in the branch.

### Fixed in this PR

| # | Sev | File | Issue | Reviewer(s) | Fix |
|---|-----|------|-------|-------------|-----|
| 1 | **P1** | `app-service.ts` matchMedia setup | On legacy Safari <14 / older Android WebView the returned `MediaQueryList` isn't an `EventTarget` (no `addEventListener`), so the call **threw in the AppService constructor → aborted app bootstrap (white screen)**. A regression (old code never touched matchMedia); the boat runs older WebViews (Hermit / OBP MAP 4). | adversarial (0.70) | Feature-detect `addEventListener`; fall back to deprecated `addListener` so legacy engines still get live updates; initial `matches` read runs regardless. + regression test (matchMedia without `addEventListener` → constructs, resolves). |
| 2 | P3 | `themes.const.ts` | Redundant `as const` after an explicit widening type annotation (no effect). | maintainability (0.65) | Removed. |

### Test gaps closed (correctness + testing + project-standards converged)
- **Switch away from `system`** → explicit theme then an OS change must NOT flip (the dynamic-dependency-drop premise). Added.
- **Legacy stored `''` preserved (dark), not flipped** to the new light default through hydration — the invariant the no-migration claim rests on. Added (settings.service + app-service body-class assertions).
- **Explicit `dark-theme` / legacy `''` → dark** body class. Added.
- **Display loads stored `system` into the picker unchanged.** Added.

CI after fixes: **1650 unit + 7 plugin + 33 mcp-schema**, lint + snc clean.

### Verified clean (no action)
- **Back-compat**: existing users unaffected — `pushSettings` hydrates a stored `''` verbatim (truthy-object guard), so dark users stay dark; the light default only lands where the `theme` blob is genuinely absent. Confirmed by correctness tracing + new tests.
- **No migration / version bump**: `themeName` is in the *unversioned* `IThemeConfig`; the migration chain passes `theme` through untouched (api-contract confirmed `LATEST_APP_CONFIG_VERSION` stays 19; the MCP `themeNames` schema lists CSS body-class names, unaffected).
- **Interpretation surface contained**: an exhaustive grep shows only `AppService.resolveIsLightTheme()` and the display picker interpret `themeName`; storage/settings pass it through opaquely; no component/guard/SCSS assumes the old two-value model.
- **Signal reactivity**: no stale-class glitch switching light→system while OS is dark (the effect re-reads `_prefersDark` fresh).

### Advisory (documented, not code changes)
- **Default flip** ('' → 'light-theme') changes new-install/reset first-render to Light — intended (the feature). Worth a **release-note** line so docs/screenshots match.
- **Legacy paths** `transformTheme` / fresh-start-after-failed-upgrade still emit `''` (dark), so v10-migrated / degraded-boot users default dark while brand-new users default light. **Intentional** — migrated users aren't "new"; both values render correctly.
- **Mixed-version fleet**: during a rolling update, an old build sharing the same user config renders `'system'` as dark and, if its display settings are saved, clobbers `'system'` back to `''`. Transient; resolves once all devices run this build (boat deploys are coordinated).

### Verdict: **Ready to merge** (pending on-device confirmation). All P1/P3 findings fixed; remaining items are advisory.

PR #451, opened 2026-07-28
Consolidate theme-snapshot cache + publish into one method (#445)

### LINKED ISSUE #445: Consolidate theme-snapshot field-cache + publish into one method (dedupe 3 call sites)
## Summary

`AppService.readThemeCssRoleVariables()` (`src/app/core/services/app-service.ts:142-182`) is a mixed command/query: its name and return type read like a pure query, but it also publishes a side effect internally (`this.cssThemeColorRoles$.next(...)`). Separately, **every** call site additionally assigns its return to the backing field:

- constructor bootstrap (`:130`)
- the light/dark theme-change effect (`:~102`, added in #438 / PR #444)
- `toggleDayNightMode()` (`:226`)

So the snapshot is pushed to two sinks (the `BehaviorSubject` inside the method, the `_cssThemeColorRoles` field at each caller) and the `this._cssThemeColorRoles = this.readThemeCssRoleVariables()` line is triplicated.

## Why it matters

A future edit that adds a fourth trigger and forgets the field assignment would leave the `cssThemeColors` getter (`:184`) stale while subscribers stay current — a silent divergence between the synchronous getter and the observable.

## Suggested fix

Consolidate into one private side-effecting method (e.g. `refreshThemeSnapshot()`) that both caches the field **and** publishes, called from all three sites; keep the pure read private and side-effect-free. Removes the triplication and the command/query mix.

## Provenance

Surfaced by the maintainability reviewer on PR #444 (fix for #438), which correctly kept that PR minimal (mirroring the established `toggleDayNightMode` pattern) and recommended this as a separate cleanup. Low priority, no behavior change.

mairas:
Closes #445.

A behavior-preserving refactor flagged by the maintainability reviewer on PR #444.

## Before
`readThemeCssRoleVariables()` was a mixed command/query — its name/return type read like a pure query, but it also published the snapshot internally (`cssThemeColorRoles$.next(...)`). Separately, all three call sites additionally cached the return: `this._cssThemeColorRoles = this.readThemeCssRoleVariables()` (constructor bootstrap, theme-change effect, `toggleDayNightMode`). The assignment was triplicated, and a future 4th caller that forgot the field write would leave the `cssThemeColors` getter stale while subscribers stayed current.

## After
- `readThemeCssRoleVariables()` is now a pure private query (builds and returns `ITheme`, no side effect).
- `refreshThemeSnapshot()` does both: caches the field **and** publishes.
- All three sites call `refreshThemeSnapshot()`.

Identical semantics: 3 sites × (1 cache + 1 publish), same as before.

## Verification
- `npm run ci` green (1650 unit + 7 plugin + 33 mcp-schema). No test changes — the existing publish-site tests are the guard ('publishes CSS theme color roles' at construction, '…when the theme changes', '…on each toggle'), all still pass.
- Independently verified behavior-preserving: `readThemeCssRoleVariables` has exactly one caller (`refreshThemeSnapshot`), no publish dropped/doubled, getter unchanged.

### DIFF
--- a/src/app/core/services/app-service.ts
+++ b/src/app/core/services/app-service.ts
@@ -114,7 +114,7 @@ export class AppService {
       // Re-snapshot the theme CSS custom properties so widgets (which draw from the
       // JS snapshot, not live CSS) recolor without a reload. getComputedStyle here
       // forces a synchronous style recalc, so the class change above is reflected.
-      this._cssThemeColorRoles = this.readThemeCssRoleVariables();
+      this.refreshThemeSnapshot();
     });
 
     effect(() => {
@@ -146,7 +146,7 @@ export class AppService {
       });
     });
 
-    this._cssThemeColorRoles = this.readThemeCssRoleVariables();
+    this.refreshThemeSnapshot();
 
     this.browserVersion.set(this.getBrowserVersion());
     this.osVersion.set(this.getOSVersion());
@@ -196,8 +196,12 @@ export class AppService {
       zoneAlarm: computedStyle.getPropertyValue('--skip-zone-alarm-color').trim(),
       zoneEmergency: computedStyle.getPropertyValue('--skip-zone-emergency-color').trim(),
     };
-    this.cssThemeColorRoles$.next(cssThemeRolesColor);
     return cssThemeRolesColor;
+  }
+
+  private refreshThemeSnapshot(): void {
+    this._cssThemeColorRoles = this.readThemeCssRoleVariables();
+    this.cssThemeColorRoles$.next(this._cssThemeColorRoles);
   }
 
   public get cssThemeColors() : ITheme {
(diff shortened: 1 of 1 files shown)


mairas:
## Adversarial verification (2 lenses, both passed)

Behavior-preserving refactor, verified in fresh contexts. **No must-fix findings.**

- **Behavior-preservation** ✅ — `readThemeCssRoleVariables()` is now a pure query (no `.next()` inside); `refreshThemeSnapshot()` does cache + publish; all three sites call it → 3 × (1 cache + 1 publish), identical to base. `readThemeCssRoleVariables` has exactly one caller; no publish dropped/doubled; the `cssThemeColors` getter is unchanged. The intra-call ordering flipped (was publish-then-cache, now cache-then-publish) but it's unobservable — the getter and the emitted value are the same reference after each refresh.
- **Test-coverage** ✅ — the lens ran the suite: all 41 `app-service.spec` tests pass, including the three publish-site guards (construction / theme change / toggle). No test referenced the removed internal `.next()`.

**Verdict: Ready to merge.** CI green.

Issue #450, opened 2026-07-28
Night-mode toasts/snackbars are dark-on-dark (inverse-surface half-overridden)


mairas:
## Symptom

In **night mode**, Material snackbars/toasts (`ToastService.show` → `ToastSnackbarComponent`) render dark text on a dark-red container, and the toast **close icon is near-invisible**.

## Root cause

Same class of bug as #448 (Cancel FAB), a different role pair. The `night-theme` `mat.theme-overrides` block (`src/styles.scss:187-196`) sets `inverse-surface: #831d1d` (dark red) but leaves `inverse-on-surface` at the generated dark value (`#3a2e2c`, neutral-20). Material maps the snackbar container to `inverse-surface` and its supporting-text to `inverse-on-surface` (`_m3-snack-bar.scss`), and the toast close icon uses `--mat-sys-inverse-on-surface` directly (`toast-snackbar.component.scss:79`). Result: ~1.35:1 — dark-on-dark.

## Why it's NOT a one-line palette fix

Unlike `on-primary` (which only feeds the Cancel FAB + notification badge), `--mat-sys-inverse-on-surface` is **also consumed as a standalone background fill** by ~4 components in night mode (widget-slider fill, remote-control, widget-list-card, tile-large-icon backgrounds). Simply brightening `inverse-on-surface` to pair with `inverse-surface` would turn those 4 surfaces bright in night mode — collateral damage. The pair is deliberately decoupled.

## Suggested fix (needs care)

Scope the fix to the snackbar/toast rather than the global token — e.g. override the toast component's text + close-icon color (and/or its container) within `ToastSnackbarComponent`'s own styles, so night toasts get a legible foreground without touching the 4 background consumers of `inverse-on-surface`. Verify on-device in night mode.

## Provenance

Surfaced by the night-palette completeness audit during PR review of #448 (Cancel-FAB contrast). Split out because #448 is scoped to the Cancel FAB and this needs a component-scoped approach to avoid the background-consumer collateral.

## Also noted (low / not user-visible)

`tertiary-fixed` / `tertiary-fixed-dim` are overridden to `#520000` with no matching `on-tertiary-fixed`, but no component renders text on those containers today — no visible defect. Mentioning for completeness only.

PR #452, opened 2026-07-28
Make the Cancel FAB icon legible in night mode (#448)

### LINKED ISSUE #448: Cancel FAB icon low-contrast in night mode (night palette overrides --mat-sys-primary but not --mat-sys-on-primary)
## Symptom

In **night mode**, the dashboard edit-mode **Cancel** FAB icon has low contrast — dark-red glyph on a dark-red container.

## Root cause

`dashboard.component.scss` `.cancel` uses `container-color: var(--mat-sys-primary)` with `foreground-color: var(--mat-sys-on-primary)` (the correct M3 pairing, set in #440/PR #446). But the `night-theme` block overrides `--mat-sys-primary` to a dark red (~`#831d1d`) **without** correspondingly overriding `--mat-sys-on-primary` (stays ~`#640d06`), so foreground and container are both dark red.

The `.save` FAB avoids this by using the **fixed** roles (`primary-fixed` / `on-primary-fixed`), which the night palette does regenerate as a pair.

## Options

1. Give `.cancel` the same fixed-role treatment as `.save` — but `on-primary-fixed` is only correct on a `primary-fixed` container, so this means switching `.cancel`'s container too (a visual change to the button color).
2. Override `--mat-sys-on-primary` (and any other half-overridden roles) in the `night-theme` block so the night palette is a complete, self-consistent M3 token set. This is the more general fix and likely helps other components too.

Option 2 is preferable — it treats the real defect (an incomplete night palette) rather than working around it per-component. Audit `src/styles.scss` `.night-theme` / `src/themes/_m3night.scss` for other roles overridden without their paired `on-*`.

## Provenance

Surfaced by the regression lens during PR #446 review; called out as out-of-scope there (a night-palette gap, not a light-theme issue). Related to the night-mode work in #443.

mairas:
Closes #448.

The Cancel FAB's close icon was dark-red on dark-red (~1.6:1) in night mode.

## Root cause (two layers)
1. The `night-theme` overrides `--mat-sys-primary` to a dark red (`#831d1d`) but left the paired `--mat-sys-on-primary` at the generated dark value (`#640d06`) — a half-overridden role pair.
2. Even with the pair completed, the night theme sets a **global** `--mat-icon-color` (dark red, via `mat.icon-overrides`) that colours the actual `<mat-icon>` glyph and **wins over** the FAB's `on-primary` foreground. So fixing only the palette token would leave the glyph unchanged (the "correct attribute overridden downstream still ships a bug" trap).

## Fix (both parts)
- Complete the pair: night `on-primary` → `#ffdad4` (also fixes the notification badge's text-on-primary).
- Pin the `.cancel` FAB's icon colour to `on-primary` (`mat.icon-overrides`) so the **glyph** is light-on-dark. No change in light/dark, where `--mat-icon-color` isn't set and the icon already inherits `on-primary`.

`.save` was unaffected (its container is the light `primary-fixed`, so the global dark icon colour already reads on it).

## Verification
- `npm run ci` green (SCSS compiles; 1650 unit + 7 plugin + 33 mcp-schema).
- **Glyph colour confirmed in real Chromium** (the review caught that the token-only fix left the glyph dark): before = `rgb(82,0,0)` on `rgb(131,29,29)` (~1.6:1); after = `rgb(255,218,212)` (~7.5:1).
- Deployed to the boat for on-device confirmation of the Cancel icon in night mode.

The audit also found a **separate** night dark-on-dark bug in toasts/snackbars (`inverse-surface` half-overridden) that needs a component-scoped fix (its paired token doubles as a background elsewhere) — filed as #450, out of scope here.

### DIFF
--- a/src/app/core/components/dashboard/dashboard.component.scss
+++ b/src/app/core/components/dashboard/dashboard.component.scss
@@ -49,6 +49,12 @@
     container-color: var(--mat-sys-primary),
     foreground-color: var(--mat-sys-on-primary),
   ));
+  // The night theme sets a global dark --mat-icon-color that would otherwise win over
+  // the FAB foreground for the glyph; pin the icon to on-primary so it stays legible on
+  // the (dark-red, in night) primary container. No-op in light/dark where they match.
+  @include mat.icon-overrides((
+    color: var(--mat-sys-on-primary),
+  ));
 }
 
 .save {
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -190,6 +190,7 @@ html {
     on-surface-variant: #831d1d,
     inverse-surface: #831d1d,
     primary: #831d1d,
+    on-primary: #ffdad4,
     outline: #831d1d,
     tertiary-fixed: var(--skip-contrast-color),
     tertiary-fixed-dim: var(--skip-contrast-color),


mairas:
## Adversarial verification (3 lenses) — caught an insufficient first fix

The review was decisive here. The initial fix (palette `on-primary` only) **did not actually fix the bug**, and two lenses flagged must-fix issues — both now addressed.

### Caught & fixed
- **P1 (contrast lens, verified in real Chromium)** — the palette-only fix left the Cancel **glyph** dark-red on dark-red (still ~1.6:1). Mechanism: `.night-theme` sets a global `--mat-icon-color` (dark red) via `mat.icon-overrides`, and `.mat-icon { color: var(--mat-icon-color, inherit) }` makes that win over the FAB's `on-primary` foreground — so `on-primary` reached the FAB element's `color` but never the glyph. **Fixed** by pinning `.cancel`'s icon colour to `on-primary` (`mat.icon-overrides` on the button). Re-confirmed in Chromium: glyph before `rgb(82,0,0)` → after `rgb(255,218,212)` (~7.5:1).
- **Medium (audit lens)** — `inverse-surface` is a *second* half-overridden pair (night toasts/snackbars go dark-on-dark). But `inverse-on-surface` doubles as a background fill in ~4 components, so a blunt global brighten would cause collateral damage. **Split out as #450** (needs a component-scoped fix) rather than forced into this PR.

### Clean
- **Regression-scope** ✅ — the only colour change is the added night `on-primary`; `:root`/light/dark blocks byte-identical; no correct role worsened; token names valid (build would hard-error otherwise).

**Verdict: Ready to merge** (pending your on-device confirmation of the Cancel icon in night mode). CI green; glyph contrast verified.

PR #453, opened 2026-07-29
Make night-mode toasts legible (dark-on-dark text + close icon) (#450)

### LINKED ISSUE #450: Night-mode toasts/snackbars are dark-on-dark (inverse-surface half-overridden)

mairas:
Closes #450.

In **red night mode** the snackbar container is `inverse-surface` (dark red, `#831d1d`) but the night palette leaves the paired `inverse-on-surface` dark (`#3a2e2c`) — so the toast **title, message, and close icon were dark-on-dark** (~1.35:1). `inverse-on-surface` also doubles as a *background fill* in ~4 other components, so it can't be brightened globally (why #448's blunt approach was split out here).

## Fix (component-scoped, red-night only)
Within `:host-context(.night-theme)`:
- Re-point `--mat-sys-inverse-on-surface` to a light tone (`#ffdad4`) for the toast subtree — fixes the **close icon** (which reads the token) without touching the 4 background consumers (they're outside the toast).
- Colour the **title** (inherited via `.toast`) and the **message** (a new `.toast-message` class that beats `matSnackBarLabel`'s own colour — custom-property substitution resolves the Material label token *above* this scope, so the token re-point alone doesn't reach it).

Only red night mode (`.night-theme` class) is affected; dim night mode uses the dark palette and was already fine.

## Verification
- `npm run ci` green (SCSS compiles; 1650 unit + 7 plugin + 33 mcp-schema).
- **Cascade verified in real Chromium** (two earlier reasoned cascades were wrong — the message label needed the explicit class): title / message / close-icon all resolve to `rgb(255,218,212)` (~7.5:1) on the `#831d1d` container.
- Deployed to the boat for on-device confirmation (trigger a night-mode toast).

## Noted (out of scope)
The severity **accent icon** (`--toast-color`) is dark for `error` (`#93000a`) and `success` (`#005500`) on the night toast — a separate accent-colour concern from the reported text/close-icon bug. The message text (the important part) is now legible; can tune the accent icons in a follow-up if wanted.

### DIFF
--- a/src/app/core/components/toast-snackbar/toast-snackbar.component.html
+++ b/src/app/core/components/toast-snackbar/toast-snackbar.component.html
@@ -6,7 +6,7 @@
      @if (severity() !== "message") {
       <div class="toast-title">{{ title() }}</div>
      }
-    <div matSnackBarLabel>{{ data.message }}</div>
+    <div matSnackBarLabel class="toast-message">{{ data.message }}</div>
     @if (hasAction()) {
       <div class="button-group">
         <button matButton="tonal" matSnackBarAction (click)="onActionClick()">{{ data.action }}</button>
--- a/src/app/core/components/toast-snackbar/toast-snackbar.component.scss
+++ b/src/app/core/components/toast-snackbar/toast-snackbar.component.scss
@@ -85,3 +85,14 @@
     height: 18px;
   }
 }
+
+// The night palette's inverse-on-surface is dark and doubles as a background fill in other
+// components, so it can't be brightened globally — re-point it for the toast subtree only.
+// .toast-message is needed because matSnackBarLabel carries its own colour resolved above this scope.
+:host-context(.night-theme) {
+  --mat-sys-inverse-on-surface: #ffdad4;
+  .toast,
+  .toast-message {
+    color: var(--mat-sys-inverse-on-surface);
+  }
+}
--- a/src/app/core/components/toast-snackbar/toast-snackbar.component.spec.ts
+++ b/src/app/core/components/toast-snackbar/toast-snackbar.component.spec.ts
@@ -36,6 +36,14 @@ describe('ToastSnackbarComponent', () => {
         expect((actionButton.nativeElement as HTMLButtonElement).textContent?.trim()).toBe('Enable Plugin');
     });
 
+    it('renders the message on the toast-message styling hook (anchors the night-mode override)', () => {
+        createComponent({ message: 'Saved', severity: 'success' });
+
+        const label = fixture.debugElement.query(By.css('div[matSnackBarLabel].toast-message'));
+        expect(label).toBeTruthy();
+        expect((label.nativeElement as HTMLElement).textContent?.trim()).toBe('Saved');
+    });
+
     it('does not render action button when action is missing', () => {
         createComponent({ message: 'Saved', severity: 'success' });
 


mairas:
## Review synthesis (correctness · testing · maintainability · project-standards)

Four always-on personas, fresh contexts (small CSS/template diff → no conditionals). **No correctness or standards findings.** Two beneficial refinements applied.

### Applied
- **Comment trim** (maintainability P3) — the two SCSS comment blocks duplicated a fact, narrated the selectors, and cited the bug number (against the "comments rare/short, no bug ref" rule). Consolidated to a short WHY block (the two load-bearing reasons: can't-brighten-globally, and why `.toast-message` is needed).
- **Regression guard** (correctness + testing + maintainability converged) — added a spec asserting the message div renders on the `div[matSnackBarLabel].toast-message` hook, so a future template edit dropping the class can't silently break the night override with a green suite.

CI after fixes: **1651 unit + 7 plugin + 33 mcp-schema**, lint + snc clean.

### Verified clean (correctness, independently)
- **Scope**: `.night-theme` is only ever on `document.body`, only in the red-night branch — so the override never leaks into light, dark, or DIM night mode.
- **CDK overlay + `:host-context`**: the custom `AppOverlayContainer` resolves to `#app-filter-wrapper`, which **is** `<body>`, so the overlay is always a body descendant → `:host-context(.night-theme)` matches.
- **Blast radius contained**: the `--mat-sys-inverse-on-surface` re-point lands on the toast host, cascading only to the toast subtree; the ~4 components using that token as a *background* are elsewhere in the DOM, unaffected; the dark-red container (an ancestor) keeps its colour.
- **Completeness**: title (inherited) + message (`.toast-message`) + close icon (token) covered; action buttons are `matButton="tonal"` (light label). Only red night mode affected.

### Residual (noted, not blocking)
- Implicit dependency on Emulated encapsulation (for the `.toast-message` specificity + the cascade boundary) — holds today, low likelihood of changing.
- The `info`/`warn` severity **accent icons** on the dark-red night container weren't explicitly measured; they're decorative (severity is also in the title text), and this PR doesn't worsen them. `error`/`success` accents are dark (noted out of scope in the PR).

### Verdict: **Ready to merge** (pending your on-device confirmation of a night-mode toast). CI green; cascade Chromium-verified.

Issue #460, opened 2026-07-29
Page-switch PoC — Unit 3: PageSwitchService evaluator + eager wiring


mairas:
Part of the event-driven page-switch PoC (see #457). **Revised after independent review — see the review comment on #457.**

## Goal
The evaluator: watch every configured trigger path and, on an edge transition into a page's value **that then holds for a short dwell**, navigate to that page **and show a brief cue** — respecting the edit/drag guards.

**Requirements:** R6, R7, R8, R9, R10, R11, R12 (on-switch cue), R13 (dwell).

**Dependencies:** Unit 1 (#458). Independent of Unit 2.

## Files
- Create: `src/app/core/services/page-switch.service.ts`
- Create: `src/app/core/services/page-switch.service.spec.ts`
- Modify: `src/app/app.component.ts` (inject for eager construction)

## Approach
- `@Injectable({ providedIn: 'root' })`. Constructor: if `EmbedModeService.embed()` → return (read-only; no auto-switch). Inject DashboardService, DataService, uiEventService, EmbedModeService, and `MatSnackBar` (cue).
- Keep `Map<path, PathWatch>` where `PathWatch = { value: unknown; release: () => void; sub: Subscription; dwellTimer?: ReturnType<setTimeout> }`.
  - **Teardown [fixes review #5]:** store BOTH the DataService `release` AND the RxJS `Subscription` from `data$.subscribe`. On path removal, `clearTimeout(dwellTimer)`, `sub.unsubscribe()`, then `release()`. `release()` alone only decrements the shared refcount — if any widget co-subscribes that path, the handler would otherwise keep firing after the page is gone.
- `effect(() => dashboard.dashboards())` → `untracked`: compute the distinct set of `trigger.path`; diff vs the map — `acquirePath(path,'default')` + `subscribe(handler)` for new paths, tear down removed ones.
- **Per-path handler(update):**
  1. `v = update.data.value`.
  2. if `Object.is(v, entry.value)` → return (edge dedupe; identical replays don't re-arm).
  3. `entry.value = v`.
  4. Does `v` match any page trigger for this path (`String(trigger.value) === String(v)`)? If **no** → `clearTimeout(entry.dwellTimer)` (the state moved off a target before dwell completed) and return.
  5. If **yes** → (re)start `entry.dwellTimer` for DWELL_MS with the candidate `v`.
- **Dwell fire [R13]:** re-check `entry.value === v` still holds (no newer change), then the guards, then navigate:
[...]

Issue #459, opened 2026-07-29
Page-switch PoC — Unit 2: Page-options dialog + editor wiring


mairas:
Part of the event-driven page-switch PoC (see #457). **Revised after independent review — see the review comment on #457.**

## Goal
Let the user set/clear a page's auto-show trigger in the existing page-options dialog, and forward it through the editor to `DashboardService.update`.

**Requirements:** R1, R2, R4, R5.

**Dependencies:** Unit 1 (#458).

## Files
- Modify: `src/app/core/interfaces/dialog-data.ts` (extend `DialogDashboardPageEditorData`)
- Modify: `src/app/core/components/dialog-dashboard-page-editor/dialog-dashboard-page-editor.component.ts` + `.html` (+ `.scss` if needed)
- Modify: `src/app/core/components/dashboards-editor/dashboards-editor.component.ts`
- Test: `src/app/core/components/dialog-dashboard-page-editor/dialog-dashboard-page-editor.component.spec.ts` (create if absent)
- Test: `src/app/core/components/dashboards-editor/dashboards-editor.component.spec.ts` (extend)

## Approach
- `DialogDashboardPageEditorData`: add `trigger?: IPageSwitchTrigger | null` that rides in (prefill) and back out (on save).
- Dialog (template-driven / `ngModel`, matching the existing Name + select-icon pattern): add an optional "Auto-show this page when…" section:
  - **Path** — `mat-autocomplete` over `DataService.getCachedPaths(true)` (self-only), free text allowed; bound to the trigger path.
  - **Value** — a **single `mat-autocomplete`** (not a mode-switching select): when `DataService.getPathMeta(path)?.possibleValues` is non-empty, offer those as options (**display `title ?? value`, store `String(value)`** [fixes review #4]); always allow free entry. This one control resolves the review's select↔free-text swap concern [#8]: the entered/stored string is **preserved across path changes** and shown even when it isn't in the current path's `possibleValues`. Matching is `String(liveValue) === trigger.value`, so the stored value MUST be the `String()`-coerced `possibleValue.value`, never the human `title`.
  - **Enable/clear** — a checkbox (or clear affordance) toggling the section; when off, or when path/value is empty, `save()` emits `trigger: null`.
  - **Accessibility [review #14]:** real `<mat-label>`/`aria-label` on the path and value fields and an accessible name on the section toggle (the "Auto-show…" heading is decorative).
  - Inject `DataService` for paths + meta.
[...]

Issue #458, opened 2026-07-29
Page-switch PoC — Unit 1: Dashboard model + trigger persistence


mairas:
Part of the event-driven page-switch PoC (see tracking issue).

## Goal
Add an optional auto-show trigger to the persisted page model and wire it through
the `DashboardService` mutators — with **no config-version bump** (the field is
back-compatible; absent = no trigger).

**Requirements:** R1, R3, R4 (storage side); enables R6–R9.

**Dependencies:** none (first unit).

## Files
- Modify: `src/app/core/services/dashboard.service.ts`
- Test: `src/app/core/services/dashboard.service.spec.ts`

## Approach
- Add and export `interface IPageSwitchTrigger { path: string; value: string }`
  (co-located with `Dashboard`).
- Add `trigger?: IPageSwitchTrigger` to the `Dashboard` interface.
- Extend `update(itemIndex, name, icon, trigger?)` to accept the trigger:
  - a valid `{ path, value }` (both non-empty) → set it,
  - `null` or an empty path/value → clear it (field becomes absent),
  - `undefined` (arg omitted) → leave any existing trigger unchanged.
  Keep the existing name/icon behavior intact.
- `duplicate()`: after the deep clone, drop the trigger from the copy
  (scope boundary — a duplicate must not silently compete for the same state).
- `add()`: unchanged — new pages start with no trigger.
- Persistence needs **no** extra code: `saveDashboards` / `getDashboardConfig`
  already round-trip the whole `Dashboard` verbatim. Confirm `LATEST_APP_CONFIG_VERSION`
  stays 19 and no migration is added.

## Patterns to follow
- Existing immutable `dashboards.update(...)` spread style in `update()`/`duplicate()`.

## Test scenarios
- Happy: `update(i, name, icon, {path:'self.navigation.state', value:'sailing'})`
  sets `dashboards()[i].trigger` to that object.
- Edge: `update(i, name, icon, null)` clears an existing trigger (field absent).
- Edge: `update(i, name, icon)` (trigger omitted) leaves an existing trigger intact.
- Edge: `update(i, name, icon, {path:'', value:''})` clears (empty = clear).
- Happy: `duplicate(i, name, icon)` on a source page **with** a trigger yields a
  copy with **no** trigger; other fields (config, new id/uuids) behave as before.
- Happy: `add(...)` yields a dashboard with no `trigger` field.
- Integration: a trigger set via `update` survives `getDashboardConfig()`
  (persistence writes/returns it verbatim).

[...]

PR #461, opened 2026-07-29
feat: Event-driven page switching (PoC)

### LINKED ISSUE #457: feat: Event-driven page switching (PoC)
## Overview

Let a dashboard page come up on its own when a Signal K value changes to signal a
new activity — e.g. when `navigation.state` becomes `sailing`, switch to the
Sailing page. Proof-of-concept scope: the simplest useful shape — one path, one
value, per page.

Origin (requirements): `_local/brainstorms/2026-07-29-event-driven-page-switch-requirements.md`

## Problem Frame

On the boat, different activities want different pages, but the only way to change
page today is manual (swipe, arrows, hotkeys, remote). The helm wants Skip to
follow the situation automatically. Skip already has both halves of the machinery
to model on: night-mode auto-switch (`app-service.ts`) reacts to a SK path value
with a previous-value dedupe, and the remote-control feature
(`remote-dashboards.service.ts`) switches the active page but refuses to while the
user is editing.

## Requirements Trace

- **R1** A page MAY carry an optional `{ path, value }` auto-show trigger; a page
  with no trigger behaves exactly as today.
- **R2** The trigger is edited in the existing page-options dialog, alongside
  Name and Icon.
- **R3** The trigger is stored on the page and travels with the page's stable
  `id`; deleting a page removes its trigger.
- **R4** Strict equality against the path's current value (PoC targets the
  enum-string case; compare as strings). No thresholds/ranges/multi-value.
- **R5** When the server publishes `possibleValues` meta for the path, the value
  field offers a dropdown; otherwise free text.
- **R6** Edge-triggered: fires only on a transition *into* the value (per-path
  dedupe), never continuously — this is what lets the user navigate away and stay.
- **R7** Startup: on the first known value, an already-matching value switches the
  page (first known value = transition from "unknown").
- **R8** Re-entry re-fires.
- **R9** If several pages match the same emission, the first in page order wins.
- **R10** No auto-switch while editing (`isDashboardStatic() === false`) or
  dragging (`uiEventService.isDragging()`).
- **R11** Drives the same router navigation the app already uses; must not clobber
  an in-flight page transition (reuse `navigateTo`, which already guards
  `isPageTransitioning`).
- **R12** On an auto-switch, show a brief snackbar cue naming the target page
  (added after review — no silent teleport).
- **R13** A trigger fires only after its value has held for a short dwell (≈2–3 s),
  not on the first instantaneous emission (added after review — debounce).

> **Independent review applied (2026-07-29).** Authority model = force + cue +
> dwell; master pause deferred; R7 startup race, subscription teardown, and the
> dropdown `.value`/coercion issues folded into the units. See the review-outcome
> comment on this issue for the full findings and accepted PoC limitations.

## Scope Boundaries

- Per-page triggers only — no global rules screen.
- No global master on/off toggle (clearing a page's trigger disables it).
- No compound conditions, thresholds, ranges, zone-state, or multiple values.
- Duplicating a page does NOT copy its trigger.
- Page switching only — no other auto-actions.

## Context & Research (patterns to follow)

- **Reactive path→action precedent:** `AppService` night-mode effect
  (`src/app/core/services/app-service.ts`) — `toSignal(subscribePath(...))` + an
  `effect` with a `previous`-value dedupe.
- **Page-switch precedent + edit guard:** `RemoteDashboardsService`
  (`src/app/core/services/remote-dashboards.service.ts`) — calls
  `dashboard.navigateTo(idx)` only when `isDashboardStatic()`, and no-ops in embed
  mode. Eagerly constructed by being injected in `src/app/app.component.ts`.
- **Persistence:** `DashboardService` (`src/app/core/services/dashboard.service.ts`)
  holds `dashboards = signal<Dashboard[]>`; a constructor `effect` calls
  `SettingsService.saveDashboards()`, which writes the array verbatim via
  `saveConfigSection('Dashboards', …)`. `getDashboardConfig()` returns it verbatim.
- **Path/value pickers:** `DataService.getCachedPaths(selfOnly)` → `string[]`
  (self-prefixed, e.g. `self.navigation.state`); `DataService.getPathMeta(path)?.possibleValues`
  → the enum options.

## Key Technical Decisions

- **No config-version bump / migration.** Adding an optional `trigger?` to the
  `Dashboard` shape is back-compatible: old stored configs simply lack the field
  (treated as no trigger), and the persistence path writes/reads the array
  verbatim with no schema that strips unknown fields. `LATEST_APP_CONFIG_VERSION`
  stays 19. (Resolves the main deferred question.)
- **Trigger lives on the page, keyed by page `id`.** No dangling references; delete
  removes it for free.
- **Store the full `self.`-prefixed path** (matches how widget paths are stored and
  how `subscribePath`/`getPathMeta` expect them).
- **Single evaluator service** with diff-managed per-path subscriptions and a
  per-path last-seen value — keeps the edge dedupe and the first-match rule
  deterministic. Modeled on the two precedents above.
- **String equality** for the match (`String(value) === trigger.value`) — covers
  the enum case; numeric/boolean paths stringify. Documented PoC limitation.

## Open Questions — resolved during planning

- *Version bump?* No — see Key Decisions.
- *Path prefix convention?* Full `self.`-prefixed path, verbatim to DataService.
- *Eager construction?* Inject the evaluator in `app.component.ts` like
  `RemoteDashboardsService`.

## Open Questions — deferred to implementation

- **Startup timing (R7):** the seed emission may arrive before the first page route
  resolves (`activeDashboard()` still null). Guard navigation on
  `activeDashboard() !== null` and verify live on the boat that the startup jump
  actually fires. Also prevents yanking away from a non-page route (e.g. settings)
  on boot.
- **Config-change re-eval:** changing a trigger's value to match the *current*
  state won't switch until the next emission (no re-eval on config edit).
  Acceptable for PoC; note it.
- **Dialog spec mock:** the page-editor dialog will newly inject `DataService` —
  add a local fake (`getCachedPaths()→[]`, `getPathMeta()→null`) and run that
  spec locally (hand-rolled mocks break when a component gains a service call).

## High-Level Technical Design

> Directional guidance for review, not implementation specification.

```mermaid
flowchart TB
  A[dashboards signal changes] --> B[Diff desired trigger paths vs subscribed]
  B --> C[acquirePath new / release removed]
  C --> D[Per-path subscription emits value]
  D --> E[record last-seen for this path]
  E --> F{value changed vs last-seen?}
  F -- no --> Z[ignore]
  F -- yes --> G{a page is active AND not editing AND not dragging?}
  G -- no --> Z
  G -- yes --> H{first page whose trigger matches path and value?}
  H -- none --> Z
  H -- found --> I{already the active page?}
  I -- yes --> Z
  I -- no --> J[dashboard.navigateTo index]
```

## Risks & Dependencies

| Risk | Mitigation |
|------|------------|
| Auto-switch fights the user | Edge-trigger + dwell (R13) + cue (R12) + `isDashboardStatic`/`isDragging`/already-active guards |
| Startup jump race (seed before route resolves) | Dwell delays the switch past route resolution + a `null→non-null` activeDashboard re-eval backstop |
| Silent teleport disorients the operator | Snackbar cue on switch (R12) |
| `possibleValues` meta absent for a path | Value field falls back to free text (typo risk accepted for PoC) |
| Leaked subscriptions on trigger churn | Diff-managed subscribe/release via `acquirePath` |

## Implementation Units

- [ ] #458 — Unit 1: Dashboard model + trigger persistence (update/duplicate wiring; no version bump)
- [ ] #459 — Unit 2: Page-options dialog + editor wiring (edit/clear the trigger)
- [ ] #460 — Unit 3: PageSwitchService evaluator + eager wiring

Dependency order: 458 → (459 ∥ 460). Units 2 and 3 both depend only on Unit 1 and
can proceed in parallel.

## Sources

- Origin: `_local/brainstorms/2026-07-29-event-driven-page-switch-requirements.md`
- Precedents: `src/app/core/services/app-service.ts`,
  `src/app/core/services/remote-dashboards.service.ts`,
  `src/app/core/services/dashboard.service.ts`

mairas:
## What & why

Lets a dashboard page come up on its own when a Signal K value changes to signal a
new activity — e.g. when `navigation.state` becomes `sailing`, Skip switches to the
Sailing page. Proof-of-concept: the simplest useful shape — one path, one value,
per page.

Closes #457. Implements #458, #459, #460.

Brainstorm → plan → **independent 6-persona plan review** → build. The review
reshaped the design before any code (see the [review-outcome comment](https://github.com/halos-org/skip/issues/457#issuecomment-5122093193)).

## How it works

- **Model (#458):** an optional `trigger { path, value }` on `Dashboard`. `update()`
  gains set/clear/leave semantics; `duplicate()` drops the trigger so a copy never
  competes for the same state. Round-trips through the existing verbatim persistence
  — **no `LATEST_APP_CONFIG_VERSION` bump** (an absent field = no trigger).
- **Editor (#459):** an "Auto-show this page on a Signal K value" section in the
  page-options dialog (edit flow only), with a path autocomplete and a value
  autocomplete seeded from the path's `possibleValues` meta (free text when the
  server publishes none). Stored value is the `String()`-coerced value, never the
  display title, so it matches the live path value.
- **Evaluator (#460):** `PageSwitchService` — one subscription per distinct trigger
  path (diff-managed, torn down with `unsubscribe()` **and** `release()`), edge
  dedupe, first-match-wins, guarded by `isDashboardStatic()` / `!isDragging()` / an
  active page. Two behaviours: a value **already true at startup** switches
  **immediately and silently** (one-time boot-into-the-right-page); a genuine
  **runtime change** switches only after a short (~2.5s) **dwell**, with a
  "Switched to …" snackbar cue. Eagerly constructed in the app shell; inert in
  embed mode.

## Decisions from the plan review

- **Authority = force + cue + dwell.** Zero-touch switch kept, but a snackbar names
  the page and a short dwell debounces a transient value (addresses the silent-teleport
  and jittery-signal findings). Operator confirms `navigation.state` is stable on the vessel.
- **No global master pause** — deferred (clear a page's trigger to disable).
- **Accepted PoC limitations:** last-writer-wins if the remote-control feature and a
  local trigger disagree; multi-path near-simultaneous emissions resolve last-wins.

## Testing

- New/updated unit specs: `dashboard.service` (trigger set/clear/leave, duplicate
  strips, persistence), `page-switch.service` (16 scenarios: dwell, flap, dedupe,
  re-entry, guards, conflict, already-active, startup backstop, teardown, embed),
  `dialog-dashboard-page-editor` (value coercion, prefill, clear, path-change
  preservation), `dashboards-editor` (trigger forwarding).
- `npm run snc` + `npm run lint` clean; full headless suite green (1688 tests).
- **Live on hurma:** deployed via `./run deploy-halos local` and confirmed on-device —
  opening the app while `navigation.state` is already `moored` lands on the trigger
  page immediately with no cue, and swiping away is not re-grabbed.

## On-device refinements (post first deploy)

- Startup now switches **immediately and silently**; the earlier build let the
  already-true value both arm the dwell and fire the backstop, which re-switched the
  page ~2.5s after the user had swiped away. Fixed by treating a path's first known
  value as a baseline (never arms the dwell); the dwell + cue apply only to genuine
  runtime changes. Two regression tests cover it.
- Dialog polish: padding between the path/value fields, examples moved to muted
  field **hints** (they read as placeholders before), and the path suggestions are
  now sorted.

- **What to watch:** open a page's options → the "Auto-show…" section renders; set a
  trigger to `self.navigation.state = <current state>` and reload → the app lands on
  that page with a "Switched to <page>" snackbar; after an auto-switch, swiping to
  another page is not undone while the state holds.
- **Healthy signal:** page switches only on a genuine state transition (held for the
  dwell) and never while editing the layout.
- **Failure signal / mitigation:** unexpected page jumps, or a switch that fights
  manual navigation → clear the offending page's trigger (Page Options) to disable it;
  the feature is inert with no triggers configured.
- **Validation window/owner:** the operator, on the vessel, over the next few sails.

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -21,6 +21,7 @@ import { EmbedModeService } from './core/services/embed-mode.service';
 import { NotificationsService } from './core/services/notifications.service';
 import { ConfigurationUpgradeService } from './core/services/configuration-upgrade.service';
 import { RemoteDashboardsService } from './core/services/remote-dashboards.service';
+import { PageSwitchService } from './core/services/page-switch.service';
 import { ToastService } from './core/services/toast.service';
 import { ReloadService } from './core/services/reload.service';
 import { AppNetworkInitService, IBootstrapIssue } from './core/services/app-initNetwork.service';
@@ -49,6 +50,7 @@ export class AppComponent implements AfterViewInit, OnDestroy {
 
   private readonly _dashboard = inject(DashboardService);
   private readonly _remoteControl = inject(RemoteDashboardsService);
+  private readonly _pageSwitch = inject(PageSwitchService);
   private readonly toast = inject(ToastService);
   private readonly _reload = inject(ReloadService);
   private readonly _notifications = inject(NotificationsService);
--- a/src/app/core/components/dashboards-editor/dashboards-editor.component.spec.ts
+++ b/src/app/core/components/dashboards-editor/dashboards-editor.component.spec.ts
@@ -30,7 +30,7 @@ describe('DashboardsEditorComponent', () => {
   };
   let confirm$: Subject<boolean>;
   /** Result the (mocked) page-editor dialog emits on close; null = cancelled. */
-  let editorResult: { name: string; icon: string } | null;
+  let editorResult: { name: string; icon: string; trigger?: { path: string; value: string } | null } | null;
   let open: ReturnType<typeof vi.spyOn>;
 
   beforeEach(async () => {
@@ -102,9 +102,27 @@ describe('DashboardsEditorComponent', () => {
     call('onTileTap', 0, clickAt({ x: 0, y: 0 }));
     call('onPageAction', 'edit');
     expect(dialog.openDashboardPageEditorDialog).toHaveBeenCalledWith(
-      expect.objectContaining({ title: 'Page Options', name: 'Sailing' })
+      expect.objectContaining({ title: 'Page Options', name: 'Sailing', enableTrigger: true })
     );
-    expect(dashboard.update).toHaveBeenCalledWith(0, 'Sailing (renamed)', 'dashboard-map');
+    // No trigger returned -> cleared (null), preserving the four-arg update contract.
+    expect(dashboard.update).toHaveBeenCalledWith(0, 'Sailing (renamed)', 'dashboard-map', null);
+  });
+
+  it('forwards an auto-show trigger returned by the editor to update()', () => {
+    const trigger = { path: 'self.navigation.state', value: 'sailing' };
+    editorResult = { name: 'Sailing', icon: 'dashboard-sailing', trigger };
+    call('onTileTap', 0, clickAt({ x: 0, y: 0 }));
+    call('onPageAction', 'edit');
+    expect(dashboard.update).toHaveBeenCalledWith(0, 'Sailing', 'dashboard-sailing', trigger);
+  });
+
+  it('passes the existing trigger into the editor dialog', () => {
+    dashboard.dashboards.set([{ ...pages[0], trigger: { path: 'self.navigation.state', value: 'sailing' } }, pages[1]]);
+    call('onTileTap', 0, clickAt({ x: 0, y: 0 }));
+    call('onPageAction', 'edit');
+    expect(dialog.openDashboardPageEditorDialog).toHaveBeenCalledWith(
+      expect.objectContaining({ trigger: { path: 'self.navigation.state', value: 'sailing' } })
+    );
   });
 
   it('routes the duplicate action and applies the result', () => {
--- a/src/app/core/components/dashboards-editor/dashboards-editor.component.ts
+++ b/src/app/core/components/dashboards-editor/dashboards-editor.component.ts
@@ -90,11 +90,13 @@ export class DashboardsEditorComponent {
       title: 'Page Options',
       name: dashboard.name ?? '',
       icon: dashboard.icon || 'dashboard-dashboard',
+      enableTrigger: true,
+      trigger: dashboard.trigger ?? null,
       confirmBtnText: 'Save',
       cancelBtnText: 'Cancel'
     }).afterClosed().subscribe(data => {
       if (!data) { return } //clicked cancel
-      this._dashboard.update(itemIndex, data.name, data.icon);
+      this._dashboard.update(itemIndex, data.name, data.icon, data.trigger ?? null);
     });
   }
 
--- a/src/app/core/components/dialog-dashboard-page-editor/dialog-dashboard-page-editor.component.html
+++ b/src/app/core/components/dialog-dashboard-page-editor/dialog-dashboard-page-editor.component.html
@@ -21,6 +21,41 @@
         <select-icon [iconFile]="'assets/svg/icons.svg'" [currentIcon]="data.icon" (selectedIcon)="onIconSelected($event)"></select-icon>
       </div>
     </div>
+    @if (data.enableTrigger) {
+      <div class="field-margin trigger-section">
+        <mat-checkbox name="triggerEnabled" [(ngModel)]="triggerEnabled">
+          Auto-show this page on a Signal K value
+        </mat-checkbox>
+        @if (triggerEnabled) {
+          <div class="field-wrapper">
+            <mat-form-field class="field-flex">
+              <mat-label>Signal K path</mat-label>
+              <input matInput name="triggerPath" [(ngModel)]="triggerPath" (ngModelChange)="onPathChange()"
+                     [matAutocomplete]="pathAuto" aria-label="Auto-show Signal K path"/>
+              <mat-autocomplete #pathAuto="matAutocomplete">
+                @for (path of filteredPaths; track path) {
+                  <mat-option [value]="path">{{ path }}</mat-option>
+                }
+              </mat-autocomplete>
+              <mat-hint>e.g. self.navigation.state</mat-hint>
+            </mat-form-field>
+          </div>
+          <div class="field-wrapper">
+            <mat-form-field class="field-flex">
+              <mat-label>Value</mat-label>
+              <input matInput name="triggerValue" [(ngModel)]="triggerValue"
+                     [matAutocomplete]="valueAuto" aria-label="Auto-show trigger value"/>
+              <mat-autocomplete #valueAuto="matAutocomplete">
+                @for (option of filteredValueOptions; track option.value) {
+                  <mat-option [value]="option.value">{{ option.label }}</mat-option>
+                }
+              </mat-autocomplete>
+              <mat-hint>e.g. sailing</mat-hint>
+            </mat-form-field>
+          </div>
+        }
+      </div>
+    }
   </mat-dialog-content>
   <mat-dialog-actions align="end">
     <button mat-flat-button [mat-dialog-close]="false">{{data.cancelBtnText}}</button>
--- a/src/app/core/components/dialog-dashboard-page-editor/dialog-dashboard-page-editor.component.scss
+++ b/src/app/core/components/dialog-dashboard-page-editor/dialog-dashboard-page-editor.component.scss
@@ -43,3 +43,14 @@ max-height: max-content;
     font-weight: 400;
   }
 }
+
+.trigger-section {
+  mat-checkbox {
+    display: block;
+    margin-bottom: 8px;
+  }
+
+  .field-wrapper + .field-wrapper {
+    margin-top: 12px;
+  }
+}
--- a/src/app/core/components/dialog-dashboard-page-editor/dialog-dashboard-page-editor.component.spec.ts
+++ b/src/app/core/components/dialog-dashboard-page-editor/dialog-dashboard-page-editor.component.spec.ts
@@ -1,23 +1,159 @@
 import { ComponentFixture, TestBed } from '@angular/core/testing';
-import { beforeEach, describe, expect, it } from 'vitest';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { MAT_DIALOG_DATA, MatDialogRef } from '@angular/material/dialog';
 import { DialogDashboardPageEditorComponent } from './dialog-dashboard-page-editor.component';
+import { DataService } from '../../services/data.service';
+import { DialogDashboardPageEditorData } from '../../interfaces/dialog-data';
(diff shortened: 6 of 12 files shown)


mairas:
## Code review — PR #461 (event-driven page switching)

**Scope:** `feat/event-driven-page-switch` vs `main` (a6108946), 12 files / ~864 insertions.
**Reviewers (6, independent Opus, fresh context):** correctness, adversarial, reliability, testing, maintainability, project-standards.
**Verdict: Not ready to merge** — 2 P2 correctness findings to fix, plus cheap P3 fixes and test strengthening. No P0/P1.

This complements the pre-implementation 6-persona *plan* review ([#457 comment](https://github.com/halos-org/skip/issues/457#issuecomment-5122093193)); it caught two bugs that survived the startup rework.

### P2 — Should fix before merge

| # | File:line | Issue | Reviewers | Conf |
|---|-----------|-------|-----------|------|
| 1 | `page-switch.service.ts:114` | **Late-first-value silent wrong-page switch.** `_startupHandled` guards only the startup *effect*, not the per-path first-value branch. After startup, a trigger path's first value calls the **global** `evaluateCurrentValues()`, which switches to the first *already-satisfied* page — can be unrelated to the firing path, and can yank the user off a page they manually swiped to during boot. On a boat, SK trigger paths frequently deliver their first delta a beat after connect, squarely in that window. | correctness, adversarial | 0.90 |
| 2 | `page-switch.service.ts:105` | **`null` emissions mishandled.** `if (value == null) return` runs before the dedupe/timer logic, so a `null` neither cancels an armed dwell nor updates `watch.value`. Two consequences: (a) a value that flaps to a trigger then to `null` mid-dwell still **fires** after the dwell (it never actually held); (b) `X → null → X` **swallows** a genuine re-entry (deduped as unchanged), so a re-asserted state never re-shows. Reachable for `navigation.state` via DataService's data-timeout cross-clear to `{value:null}`. | correctness, adversarial | 0.76 |

### P3 — Fix if straightforward

| # | File:line | Issue | Reviewers | Conf | Route |
|---|-----------|-------|-----------|------|-------|
| 3 | `page-switch.service.ts:158` | `switchTo` shows the "Switched to …" cue unconditionally, but `navigateTo` silently no-ops when `isPageTransitioning()`. If the dwell fires during a manual transition: false cue + the switch is dropped with no retry (edge consumed). | adversarial, reliability | 0.78 | fix-here |
| 4 | `page-switch.service.ts:165` | `hasMatchingPage` and `matchingPageIndex` encode the identical predicate; `hasMatchingPage === matchingPageIndex >= 0`. A future match-rule change must touch both. | maintainability | 0.68 | fix-here |
| 5 | `page-switch.service.ts:127` | Two different trigger paths matching in one delta batch → double `navigateTo` + double cue; final page by dashboard-array order (no arbitration). | adversarial, testing | 0.68 | accept (documented multi-path last-wins) — will suppress the double *cue* |
| 6 | `page-switch.service.ts:176` | `String()` equality: object/boolean-valued paths match nothing (or everything for `'[object Object]'`); free-text value has no validation, so a dead trigger is silent. | adversarial | 0.62 | accept (documented: string-equality enum case, PoC) |
| 7 | `dashboard.service.ts:239` | `update()` three-state (set/clear/leave) contract exceeds its sole caller (always passes `?? null`); the `undefined`=leave and empty-clear branches are test-only. | maintainability | 0.63 | accept (defensible boundary hardening) |

### Test gaps (addressing with the fixes)

- **No test** for a path's first *matching* value arriving **after** startup (finding 1's path) — every runtime test seeds a non-matching baseline first.
- The **swipe-away** test (`spec:235`) advances timers but never `TestBed.tick()`s after the `activeDashboard` change, so it proves "no dwell armed", **not** that the `_startupHandled` guard prevents a re-switch (finding at `:70`).
- Teardown **clearing a pending dwell** is never exercised (no armed timer at removal).
- `update()` **partial-empty** clear (path set / value empty, and vice-versa) untested (only both-empty).
- **Multi-path** concurrent dwells untested.

### Coverage

- Suppressed: 0 findings below 0.60. project-standards: no findings (no `any` introduced, `@if`/`@for` used, `import type` correct, mirrors `RemoteDashboardsService`).
- Residual risks noted: a startup silent auto-jump changes `activeDashboard`, which `RemoteDashboardsService` then republishes as the remote highlight index (controller reflects a page the user didn't pick) — acceptable, noted.

### Plan of action (author addressing via receiving-review)

Fix 1 by treating a **post-startup first value as a normal (scoped, dwelled, cued) change** rather than a global silent evaluate — the startup effect remains the single one-time global jump. Fix 2 by cancelling the dwell + resetting the baseline on `null`. Fix 3 by guarding `switchTo` on `isPageTransitioning()`. Fix 4 by the one-line dedup. Add the missing tests (late-first-value, `TestBed.tick` in swipe-away, teardown-cancels-dwell, null re-entry, `update()` partial-empty). Accept 5–7 as documented PoC limitations (will silence the double *cue* on 5).

mairas:
## Review findings addressed (36b0f93a)

**P2 #1 — late-first-value silent wrong-page switch:** removed the global `evaluateCurrentValues()` from the per-path first-value branch. A first value arriving *after* startup now takes the normal per-path **dwell + cue** path (scoped to the firing path); the startup effect remains the single one-time silent jump. New tests: a late first matching value is dwelled+cued, and a late (even non-matching) first value on one path never silently switches to another path's still-matching page.

**P2 #2 — null handling:** a `null` (cleared) value now cancels any pending dwell and resets the path baseline. Fixes both directions — a transient null mid-dwell no longer fires (it never held), and a genuine re-entry after a clear (`X → null → X`) re-fires instead of being deduped away. New tests for both.

**P3 #3 — false cue:** `switchTo` now bails while `isPageTransitioning()`, so the "Switched to …" cue can't announce a navigation that `navigateTo` would have dropped. New test.

**P3 #4 — duplication:** `hasMatchingPage` is now `matchingPageIndex(...) >= 0` (single predicate).

**Test gaps closed:** the swipe-away test now `TestBed.tick()`s after the `activeDashboard` change so it actually exercises the `_startupHandled` once-guard; added teardown-cancels-pending-dwell and `update()` half-formed-clear tests.

**Accepted as documented PoC limitations (not changed):** #5 multi-path last-writer-wins (incl. double cue when two different paths match in one delta batch), #6 `String()`-equality/enum-only matching with no path/value validation, #7 `update()`'s three-state contract (kept as boundary hardening).

Gates: `npm run snc` + `npm run lint` clean; full headless suite **1696 tests** green (evaluator spec now 21 scenarios). Verdict now **ready to merge** pending your go-ahead.

Issue #467, opened 2026-07-30
Audit all widgets for hardcoded time constants that filter independently of updateInterval


mairas:
## Motivation

The windsteer needle-animation bug (see companion issue) is an instance of a **class** of
defect: a hardcoded time constant baked into a widget (or a shared util/directive) that
silently filters, throttles, or caps responsiveness **independently of the user's
`updateInterval` setting** — or, worse, inverts it (there, a smaller interval produced
*less* needle motion).

Now that `updateInterval` is a single per-widget knob applied as
`sampleTime(updateInterval)` in `WidgetStreamsDirective`, any *other* time constant in the
render/animation path can quietly defeat it. We need to sweep the whole widget set for
these before trusting `updateInterval` end-to-end.

## Task

Audit **all ~46 widgets** plus shared rendering utils/directives for hardcoded time
values that affect how fast a value reaches the screen. For each, classify:

- **legitimate** (cosmetic transition that completes within a sample period, genuinely
  intended smoothing, unrelated timing), or
- **bug** (filters/caps the data faster than or independent of `updateInterval`, or
  overrides it), and note whether it should track `updateInterval`, become configurable,
  or be removed.

## Where to look (grep starting points)

- Fixed animation durations: `ANIMATION_DURATION`, literal ms in `animateRotation` /
  `animateAngleTransition` / `animateSectorTransition` / `animateRudderWidth`
  (`core/utils/svg-animate.util.ts` and all callers).
- RxJS rate operators inside widgets: `throttleTime(`, `debounceTime(`, `auditTime(`,
  `sampleTime(` (any `sampleTime` outside the one in `WidgetStreamsDirective` is suspect),
  `bufferTime(`.
- Raw timers: `setInterval(`, `setTimeout(` with literal ms, `requestAnimationFrame`
  tween loops with a fixed `duration`.
- Dead-bands / change gates that can hide motion: `*_EPSILON`, `DEG_EPSILON`,
  `SPEED_EPSILON`, `distinctUntilChanged` with rounding/quantization upstream.
- Any hardcoded "refresh"/"render"/"redraw" cap (canvas gauges, charts, radar).

## Deliverable

- A table: widget/util · file:line · constant · classification · recommended action.
- Split out follow-up fix issues for the confirmed bugs (or fold small ones into the
  windsteer fix if they share the same util).
- If a shared util is the common culprit (likely `svg-animate.util.ts`), recommend fixing
  it once at the source.

## Reference

[...]

PR #468, opened 2026-07-30
fix: dial animations track updateInterval (and interpolate linearly)

### LINKED ISSUE #466: Windsteer needle animation is a hardcoded 900 ms low-pass that overrides (and inverts) updateInterval
## Summary

On the windsteer widget, the AWA/TWA (and every other) needle barely moves while the
underlying values swing, and — counterintuitively — **lowering `updateInterval` makes the
needle move *less*, not more**. The cause is a hardcoded 900 ms needle-rotation animation
that behaves as a low-pass filter whose time constant is decoupled from, and dominates,
the user-configurable `updateInterval`.

Found during on-boat investigation (halpi.hurma, 2026-07-30) after a user set
`updateInterval: 100` on their configured windsteer expecting a ~10 Hz needle.

## Reproduce

1. Add a windsteer widget, feed it a live/swinging wind angle (real masthead, or any
   source that moves several degrees per 100 ms).
2. Set the widget's **Display → update interval** to 100 ms.
3. Watch the AWA/TWA needle vs the numeric wind readouts elsewhere.

**Observed:** needle looks nearly stationary; numeric wind values swing normally.
**Expected:** at 100 ms the needle tracks the wind at ~10 Hz.

## Root cause

The needle rotation is a **fixed 900 ms `ease-in-out-cubic` tween that is re-triggered on
every sample**:

- `src/app/widgets/svg-windsteer/svg-windsteer.component.ts:90` — `ANIMATION_DURATION = 900`
- `:219` / `:248` — AWA / TWA effects call `animateRotation(..., ANIMATION_DURATION, ...)`
  directly on every new signal value (same for compass `:130`, COG `:161`, waypoint `:194`,
  drift-set `:282`).
- `src/app/core/utils/svg-animate.util.ts:89–92` — a new call **cancels the in-flight
  tween** via the WeakMap.
- `:104` — the new tween starts `from = getCurrentAngle()` (the needle's *current*,
  mid-transit position).
- `:114–128` — fixed `duration`, `easeInOutCubic`.

`easeInOutCubic` advances only `4·t³` early on: at `t = 100 ms / 900 ms ≈ 0.11` it has
covered **~0.5 %** of the gap. So when a new sample arrives every 100 ms, each tween is
cancelled after moving ~0.5 % of the way and restarted from there — the needle never
escapes the slow-start region of the curve. The effective needle speed is set by the
900 ms constant and is **inversely** related to the sample rate.

Feeding path (why samples arrive at the interval): `updateInterval` →
`WidgetStreamsDirective` `sampleTime(updateInterval)` (`widget-streams.directive.ts:158,189`)
→ `onAppWindAngle`/`onTrueWindAngle` (`widget-windsteer.component.ts:267,291`, 1° dead-band)
→ the SVG effect above. The dead-band and 1 Hz-default sampling are *secondary*; the
dominant filter is the animation.

## Quantified

Simulation of the exact re-triggered-ease semantics against a representative swinging
wind signal (±20° @ 4 s + ±8° @ 1.3 s jitter):

| updateInterval | needle shows |
|---|---|
| 1000 ms | 100 % of real motion |
| 500 ms  | 65 % |
| 200 ms  | 22 % |
| **100 ms** | **11 %** |

The animation happens to "work" only at the historical 1000 ms default, because 900 ms ≈
the sample period so each tween nearly completes before the next sample. Any faster
cadence is silently throttled.

## Scope

All needles in the widget share `animateRotation` with the same fixed 900 ms, so the
whole gauge is affected — not just AWA/TWA. Laylines/sectors use the same
`ANIMATION_DURATION` via `animateAngleTransition`/`animateSectorTransition`.

## Fix direction (not prescriptive)

The animation must not act as a hidden filter that overrides a user setting. Options to
weigh in planning:

- Make the tween **duration track `updateInterval`** (e.g. clamp to it) so a tween
  completes within one sample period — preserves smoothing at slow cadence, restores
  responsiveness at fast cadence.
- Replace the re-triggered fixed-duration ease with **frame-rate-independent exponential /
  critically-damped smoothing** toward the live target, with an explicit (and ideally
  configurable) time constant — an honest, declared filter rather than an accidental one.
- Decide whether any needle smoothing should exist at short intervals at all.

Whatever is chosen, the acceptance test is behavioural: at `updateInterval = 100 ms` the
needle must track a swinging input at ~100 % of its motion (allowing a small, bounded,
*declared* smoothing lag), and responsiveness must be **monotonic** in the cadence
(smaller interval ⇒ same-or-more needle motion, never less).

---

### Decisions

- **Approach:** the needle glide **duration is derived from the widget's `updateInterval`**
  so the tween completes within one sample period (the needle reaches each value about as
  the next sample arrives). `updateInterval` becomes the single honest control of needle
  responsiveness. Chosen over a fixed smoothing time-constant (bigger util rework, and it
  would decouple the knob the user is reaching for) and over snap-to-value (no smoothing).
- **Scope:** fix **all three** animation-driven widgets that share `svg-animate.util.ts` —
  `svg-windsteer`, `svg-racesteer`, `svg-autopilot`. Note racesteer and autopilot already
  ship `ANIMATION_DURATION = 1000` against a default `updateInterval: 500`, so they are
  over-animated even at their defaults (same defect, milder). Non-animation time gates in
  other widgets remain the job of the audit (#467).
- **Policy:** `duration = min(updateInterval, cap)` with a small floor; `cap ≈
  DEFAULT_WIDGET_UPDATE_INTERVAL_MS (1000 ms)` to preserve today's feel at/above the
  default. This guarantees `duration ≤ updateInterval` (completes within a period) and is
  monotonic in cadence. Exact cap/floor constants are finalized in implementation; the
  **invariant** (monotonic, `duration ≤ updateInterval`) is the contract.
- The 900/1000 ms is a **caller** constant, not baked into the util — `animateRotation`
  et al. already accept a `duration`. So the util's animation functions are unchanged; the
  only shared new piece is a pure duration-policy helper.

### Mechanism / key files

- New pure helper + tests: `src/app/core/utils/svg-animate.util.ts` (+ create
  `src/app/core/utils/svg-animate.util.spec.ts`).
- `updateInterval` reaches each SVG child as a **new `input`**, bound in the parent
  template from `runtime.options()?.updateInterval ?? DEFAULT_WIDGET_UPDATE_INTERVAL_MS`.
- `DEFAULT_WIDGET_UPDATE_INTERVAL_MS` lives in
  `src/app/core/interfaces/widgets-interface.ts`.

### Implementation units

- [ ] **U1 — Shared duration-policy helper (the honest core).**
  Add a pure `effectiveAnimationDuration(updateInterval)` to `svg-animate.util.ts`.
  Tests (`svg-animate.util.spec.ts`, create): monotonic (smaller interval ⇒ ≤ duration);
  `duration ≤ updateInterval` across the range; cap applied above the default; floor
  applied at tiny values; non-finite / ≤ 0 falls back to the default.
- [ ] **U2 — `svg-windsteer`.**
  Add `updateInterval` input; bind it in `widget-windsteer.component.html`; replace the six
  needle `ANIMATION_DURATION` uses (compass, COG, waypoint, AWA, TWA, drift-set) and the
  layline/sector transitions with the derived duration.
  Spec (`svg-windsteer.component.spec.ts`, exists): assert `animate*` receives the derived
  duration for a given `updateInterval`; assert input wiring.
- [ ] **U3 — `svg-racesteer`.**
  Same wiring; replace the five needle uses (compass, tack, waypoint, TWA, set) plus the
  layline/sector transitions. `ANIMATION_DURATION = 1000` today; rotation centre `[600,620]`
  stays unchanged. No spec today → verify via build (`strictTemplates`) + on-device.
- [ ] **U4 — `svg-autopilot`.**
  Same wiring; replace the dial + AWA rotations. Also review the `animateRudderWidth`
  rudder-bar calls and derive their duration too if hardcoded. Parent has several
  `updateInterval: 500` config entries — use the widget-level effective cadence
  (`runtime.options()?.updateInterval`). No spec today → build + on-device.
- [ ] **U5 — Behavioural verification (on-boat).**
  Deploy to halpi (`./run deploy-halos halpi.hurma`), set a windsteer widget's
  **Display → update interval** to 100 ms against live/swinging wind, and confirm the needle
  tracks (not the ~11 % attenuation) and that motion is monotonic in cadence
  (100 ms ≥ 500 ms ≥ 1000 ms). This is the user-visible acceptance test.

### Verification contract

At `updateInterval = 100 ms` the needle tracks a swinging input at ~100 % of its motion
(minus a small, bounded, **declared** glide lag), and needle motion is **monotonic** in
cadence — a smaller interval must never yield less motion. This fix also resolves the
animation-util slice of #467 (those three widgets); the audit retains only non-animation
time gates.

mairas:
## Motivation

The windsteer dial looked heavily filtered — the AWA/TWA needle sat nearly still while the numeric wind values swung. Root cause: the needle animation was a fixed 900/1000&nbsp;ms re-triggered `ease-in-out-cubic` tween, which behaves as a low-pass filter whose lag is **decoupled from — and inverts —** the widget's `updateInterval`. At `updateInterval = 100 ms` the needle showed only **~11%** of the real motion; the *shorter* the interval, the *less* the needle moved. Full root-cause writeup and the attenuation table are in #466.

## Approach

- **`effectiveAnimationDuration(updateInterval)`** (new pure helper): caps a tween at one sample period — `min(updateInterval, DEFAULT_WIDGET_UPDATE_INTERVAL_MS)` — so it completes before the next value arrives. Monotonic in the interval (a shorter cadence never yields less motion). Fully unit-tested.
- **Thread `updateInterval` into `svg-windsteer`, `svg-racesteer`, `svg-autopilot`** and size every needle / rudder-bar / layline / sector tween from it instead of the hardcoded constant. (racesteer and autopilot shipped a 1000&nbsp;ms tween against a 500&nbsp;ms default cadence, so they were over-animated even at their defaults.)
- **Linear interpolation instead of `ease-in-out-cubic`** in the shared animation helpers. Cubic zeros velocity at both ends of every tween, so a continuously re-triggered tween dead-stops at each sample (~0→64&nbsp;deg/s velocity pulsing) and reads as *stepping*. These helpers only ever drive continuously-updating values, so linear gives constant-velocity, smooth tracking with no added lag.

## Verification

- **Unit:** `effectiveAnimationDuration` (monotonic, `≤ interval`, cap, non-finite fallback) + a windsteer duration-wiring test. Full suite **1723/1723** green.
- **Static:** ESLint, `strictNullChecks` (`npm run snc`), and the `strictTemplates` build all clean.
- **On-boat:** deployed to halpi and confirmed on the MFD — the needle tracks at 100&nbsp;ms, motion is monotonic in cadence, and (after switching to linear) glides smoothly instead of hopping.

## Scope note

Only the three animation-driven widgets are touched here. The broader audit for *other* hardcoded time constants across widgets is tracked separately in #467.

Closes #466.

### DIFF
--- a/src/app/core/utils/svg-animate.util.spec.ts
+++ b/src/app/core/utils/svg-animate.util.spec.ts
@@ -1,0 +1,42 @@
+import { describe, expect, it } from 'vitest';
+import { effectiveAnimationDuration } from './svg-animate.util';
+import { DEFAULT_WIDGET_UPDATE_INTERVAL_MS } from '../interfaces/widgets-interface';
+
+describe('effectiveAnimationDuration', () => {
+  it('returns the update interval unchanged when it is below the cap', () => {
+    expect(effectiveAnimationDuration(100)).toBe(100);
+    expect(effectiveAnimationDuration(500)).toBe(500);
+  });
+
+  it('caps the duration at the default update interval', () => {
+    expect(effectiveAnimationDuration(DEFAULT_WIDGET_UPDATE_INTERVAL_MS + 1)).toBe(DEFAULT_WIDGET_UPDATE_INTERVAL_MS);
+    expect(effectiveAnimationDuration(5000)).toBe(DEFAULT_WIDGET_UPDATE_INTERVAL_MS);
+  });
+
+  it('returns the cap at exactly the default interval', () => {
+    expect(effectiveAnimationDuration(DEFAULT_WIDGET_UPDATE_INTERVAL_MS)).toBe(DEFAULT_WIDGET_UPDATE_INTERVAL_MS);
+  });
+
+  it('never animates longer than one update interval, so a tween completes within the sample period', () => {
+    for (const interval of [50, 100, 200, 333, 500, 750, 1000, 1500, 3000]) {
+      expect(effectiveAnimationDuration(interval)).toBeLessThanOrEqual(interval);
+    }
+  });
+
+  it('is monotonic non-decreasing in the update interval (smaller cadence never yields more motion)', () => {
+    const intervals = [50, 100, 200, 333, 500, 750, 1000, 1500, 3000];
+    let previous = -Infinity;
+    for (const interval of intervals) {
+      const duration = effectiveAnimationDuration(interval);
+      expect(duration).toBeGreaterThanOrEqual(previous);
+      previous = duration;
+    }
+  });
+
+  it('falls back to the default for non-positive or non-finite input', () => {
+    expect(effectiveAnimationDuration(0)).toBe(DEFAULT_WIDGET_UPDATE_INTERVAL_MS);
+    expect(effectiveAnimationDuration(-100)).toBe(DEFAULT_WIDGET_UPDATE_INTERVAL_MS);
+    expect(effectiveAnimationDuration(Number.NaN)).toBe(DEFAULT_WIDGET_UPDATE_INTERVAL_MS);
+    expect(effectiveAnimationDuration(Number.POSITIVE_INFINITY)).toBe(DEFAULT_WIDGET_UPDATE_INTERVAL_MS);
+  });
+});
--- a/src/app/core/utils/svg-animate.util.spec.ts
+++ b/src/app/core/utils/svg-animate.util.spec.ts
@@ -39,4 +39,11 @@ describe('effectiveAnimationDuration', () => {
     expect(effectiveAnimationDuration(Number.NaN)).toBe(DEFAULT_WIDGET_UPDATE_INTERVAL_MS);
     expect(effectiveAnimationDuration(Number.POSITIVE_INFINITY)).toBe(DEFAULT_WIDGET_UPDATE_INTERVAL_MS);
   });
+
+  it('coerces string and undefined updateInterval from a persisted config', () => {
+    expect(effectiveAnimationDuration(undefined)).toBe(DEFAULT_WIDGET_UPDATE_INTERVAL_MS);
+    expect(effectiveAnimationDuration('500' as unknown as number)).toBe(500);
+    expect(effectiveAnimationDuration('5000' as unknown as number)).toBe(DEFAULT_WIDGET_UPDATE_INTERVAL_MS);
+    expect(effectiveAnimationDuration('abc' as unknown as number)).toBe(DEFAULT_WIDGET_UPDATE_INTERVAL_MS);
+  });
 });
--- a/src/app/core/utils/svg-animate.util.spec.ts
+++ b/src/app/core/utils/svg-animate.util.spec.ts
@@ -1,5 +1,5 @@
-import { describe, expect, it } from 'vitest';
-import { effectiveAnimationDuration } from './svg-animate.util';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { animateRotation, effectiveAnimationDuration } from './svg-animate.util';
 import { DEFAULT_WIDGET_UPDATE_INTERVAL_MS } from '../interfaces/widgets-interface';
 
 describe('effectiveAnimationDuration', () => {
@@ -45,5 +45,29 @@ describe('effectiveAnimationDuration', () => {
     expect(effectiveAnimationDuration('500' as unknown as number)).toBe(500);
     expect(effectiveAnimationDuration('5000' as unknown as number)).toBe(DEFAULT_WIDGET_UPDATE_INTERVAL_MS);
     expect(effectiveAnimationDuration('abc' as unknown as number)).toBe(DEFAULT_WIDGET_UPDATE_INTERVAL_MS);
+  });
+});
+
+describe('animateRotation interpolation', () => {
+  afterEach(() => vi.restoreAllMocks());
+
+  it('interpolates linearly between angles, not with an ease curve', () => {
+    let transform: string | null = null;
+    const el = {
+      getAttribute: (name: string) => (name === 'transform' ? transform : null),
+      setAttribute: (name: string, value: string) => { if (name === 'transform') transform = value; }
+    } as unknown as SVGGElement;
+
+    const frames: FrameRequestCallback[] = [];
+    vi.spyOn(globalThis, 'requestAnimationFrame').mockImplementation(cb => { frames.push(cb); return 1; });
+    vi.spyOn(performance, 'now').mockReturnValue(1000);
+
+    animateRotation(el, 0, 100, 1000);
+    expect(frames.length).toBeGreaterThan(0);
+
+    frames[0](1250); // 25% through a 1000ms tween
+
+    // Linear: 0 + 100 * 0.25 = 25. An ease-in-out-cubic curve would give 6.25.
+    expect(transform).toBe('rotate(25 500 500)');
   });
 });
--- a/src/app/core/utils/svg-animate.util.ts
+++ b/src/app/core/utils/svg-animate.util.ts
@@ -1,4 +1,5 @@
 import type { NgZone } from '@angular/core';
+import { DEFAULT_WIDGET_UPDATE_INTERVAL_MS } from '../interfaces/widgets-interface';
 
 /**
  * SVG Animation Utilities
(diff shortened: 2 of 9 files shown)


mairas:
## Code review — PR #468

**Scope:** 9 files (~98 non-test LOC) vs `origin/main` @ `651c5184`.
**Intent:** fix #466 — dial animations were a fixed re-triggered cubic tween decoupled from `updateInterval`; derive duration from `updateInterval` and interpolate linearly.
**Reviewers:** correctness, testing, maintainability, project-standards, adversarial (5; diff >50 non-test LOC touching a shared util used by 3 widgets).

No P0/P1. The fix is sound — every reviewer independently confirmed: all prior `ANIMATION_DURATION` call sites converted, the linear swap's blast radius is contained to the 3 svg-steer widgets, first-render/equal-value paths still use `setRotationImmediate`, shortest-path wrap math untouched, the `computed`-in-`untracked` reads fresh values, and no VERSION bump is required (1.0.5 cycle already open).

### P2 — Moderate

| 1 | `svg-animate.util.spec.ts` | The cubic→linear swap **is the load-bearing #466 fix but has no test** — a revert of any `const eased = progress` back to cubic passes all 1723 tests green. It's assertable in jsdom (mock rAF, drive `performance.now()` to `start+duration*0.25`, assert interpolated value = `0.25·delta` (linear) not `0.0625·delta` (cubic)). | testing (echoed by all 4 others as a testing gap) | 0.85 | add test |

### P3 — Low

| 2 | `svg-animate.util.ts` (57,160,247,306) | Four per-function JSDoc lines still say "cubic easing" after the switch to linear (module note was updated, the function docs weren't). | correctness + maintainability | 0.97 | fix |
| 3 | `svg-animate.util.ts:48` | `effectiveAnimationDuration` guards with `Number.isFinite`, but the data sampler (`widget-streams.directive.ts:158`) coerces the same field with `Number()`. A persisted **numeric-string** `updateInterval` (`"500"`) would sample data at 500 ms while the needle caps at 1000 ms — re-arming the exact #466 lag. Only reachable via hand-edited/imported/legacy config (no current writer emits strings). | adversarial | 0.60 | coerce with `Number()` |
| 4 | `svg-*.component.ts` ×3 | `?? DEFAULT_WIDGET_UPDATE_INTERVAL_MS` duplicated at 3 call sites and overlaps the helper's own guard — two encodings of the same default. | maintainability | 0.63 | widen param to `number \| undefined`, drop the `??` |
| 5 | `svg-animate.util.ts` (140,216,282,337) | `const eased = progress;` — pointless alias left after removing easing; the name `eased` now misleads. | maintainability | 0.66 | inline `progress` |
| 6 | `svg-animate.util.ts:42` | `effectiveAnimationDuration` doc narrates the prior implementation ("the fixed durations these helpers used before…") — CLAUDE.md forbids dev-history narration in comments. | project-standards | 0.63 | reword to current-state |
| 7 | `svg-animate.util.ts:51` | Animation cap reuses `DEFAULT_WIDGET_UPDATE_INTERVAL_MS`, coupling max-tween-length to the default-cadence constant (coincidentally both 1000). | maintainability | 0.60 | accept (intentional) / optional named constant |
| 8 | `svg-windsteer.component.ts` | Compass-mode toggle is a one-shot ~180° re-centering that now reuses the per-sample linear duration — at a low `updateInterval` it crams the swing into ~100 ms with a hard stop (was 900 ms cubic). Motion-quality nuance on a rare interaction; no clean fix without re-introducing a one-shot/tracking distinction. | adversarial | 0.65 | accept |

| # | File | Issue | Reviewer | Route |
| 9 | `svg-racesteer.component.ts:195,228` | The tack & waypoint `animateRotation` calls omit `ngZone` (the other 3 in the file pass it), so those rAF loops run **inside** Angular's zone and fire change detection every frame. Not introduced here, but this PR's faster/more-frequent tweens make it worse. | adversarial | **fix-here** (trivial + in blast radius) |

### Coverage
- Testing gaps: the linear interpolation curve is unasserted end-to-end (#1); racesteer/autopilot have no component spec, so their `updateInterval`→duration wiring is verified only by code-identity with windsteer + compiler (deleted constant) + strictTemplates — low-value to duplicate.
- Verified live on-boat (needle tracks at 100 ms, smooth) — the P2 is about a regression *guard*, not unverified behavior.

---

Recommended before merge: add the linear-interpolation test (#1); fix the 4 stale docstrings (#2); coerce `updateInterval` with `Number()` and widen the param, which also removes the `??` duplication (#3+#4); inline `progress` (#5); reword the dev-history comment (#6); and fix the pre-existing racesteer `ngZone` omission (#9, trivial + directly in this change's perf path). Accept #7 (intentional coupling) and #8 (rare interaction, no clean fix).

<sub>🤖 Review orchestrated with Claude Code — 5 independent persona reviewers.</sub>

mairas:
## Review addressed

| Finding | Resolution | Commit |
|---|---|---|
| #1 (P2) linear swap untested | Added a regression test: drives a mocked rAF to 25% and asserts the linear value `25` (cubic would give `6.25`). | `4a428b7` |
| #2 stale cubic docstrings | Corrected the four function docstrings to say linear. | `9198879` |
| #3 `Number.isFinite` vs `Number()` coercion | `effectiveAnimationDuration` now coerces with `Number()` and accepts `number \| undefined`; added string/undefined coercion tests. | `9198879` |
| #4 `?? DEFAULT` duplicated ×3 | Dropped at all three call sites (the widened param + helper guard cover it). | `9198879` |
| #5 `const eased = progress` alias | Inlined `progress`. | `9198879` |
| #6 dev-history comment | Reworded to current-state rationale. | `9198879` |
| #9 pre-existing racesteer `ngZone` | tack & waypoint tweens now run outside the Angular zone, like the sibling calls. | `fd54f73` |
| #7 cap reuses `DEFAULT_WIDGET_UPDATE_INTERVAL_MS` | **Accepted** — intentional ("never smooth longer than the default cadence"). |
| #8 compass-toggle one-shot linear | **Accepted** — rare interaction; a clean fix needs a one-shot/tracking split not worth the complexity. |

**Gates:** ESLint + `snc` clean; full suite green (**1725/1725** on clean runs).

**Note:** an intermittent, **pre-existing** flake in `minichart.component.spec.ts` (`plots a backfill batch…`, `length` 0 vs 3) trips under loaded full-suite timing on the dev box. It passes 4/4 in isolation, imports none of the changed files, and is `isolate:true`-isolated from this change — it's the documented order-dependent chart.js-mock/stream determinism class, not a regression from this PR. Worth a separate follow-up.

PR #470, opened 2026-07-30
fix(help): Help page scrolls long docs (bounded flex layout)

### LINKED ISSUE #437: Help page scrolls only a few pixels — .markdown-content height doesn't resolve
## Symptom

On the Help full-page view, the content scrolls only a few pixels vertically — long help documents can't be read past the first screenful. Reproduced on touch (Samsung tablet) and independent of input method.

## Root cause

`.markdown-content` (`src/app/core/components/app-help/app-help.component.scss:50-64`) sets:

```
height: calc(100% - 75px);
overflow-x: scroll;   /* Chromium promotes overflow-y -> auto */
```

The `calc(100% - 75px)` percentage height only resolves if an ancestor establishes a **definite** height. The routed-page host chain — `app-help` component host → `<router-outlet>` → the app shell — does not set one, so `100%` resolves to `auto` and `.markdown-content` collapses to roughly content-height. With almost no difference between `scrollHeight` and `clientHeight`, there's nearly nothing to scroll. The `75px` is also a magic-number approximation of the header + `mat-divider` height.

For contrast, the Settings full-page view scrolls correctly because its scroll child sits in a proper height-bounded flex column.

## Provenance

The rule dates to upstream Kip's "Add Help markdown component" (#548, 2025-03-30) and has been untouched since. It is **not** related to the fling/touch fix in #436 — that directive detects `.markdown-content` as a scroll container (via the promoted `overflow-y:auto`) and yields to it, so Help touch-scroll is unchanged by that PR. Confirmed during #436 review.

## Suggested fix

Give the Help page a bounded flex layout instead of the `calc(100% - 75px)` magic height: make the routed host a `flex-direction: column` filling its height, let the header + divider size naturally, and give `.markdown-content` `flex: 1; min-height: 0; overflow: auto`. Drop the `75px` constant. Verify long docs scroll fully on the boat (touch) and desktop (wheel).

mairas:
## Motivation

On the Help full-page view, long docs scrolled only a few pixels — unreadable past the first screenful (reported on a Samsung tablet, #437). `.markdown-content` used `height: calc(100% - 75px)`, but the routed Help host had **no `:host` rule** setting a definite height, so `100%` resolved to `auto` and the content collapsed to ~content-height — almost nothing to scroll. (Settings scrolls fine only because *its* host sets `height: 100%`.)

## Approach

Give the Help host the bounded flex column the shell already provides for routed content (`.skip-app-shell > *:not(app-toolbar) { flex: 1 1 auto; min-height: 0 }`):

- `:host { display: flex; flex-direction: column; height: 100%; min-height: 0 }`
- header pinned with `flex: 0 0 auto`
- `.markdown-content { flex: 1 1 auto; min-height: 0; overflow: auto }` — dropping the `height: calc(100% - 75px)` magic number and the `overflow-x: scroll` hack (Chromium was only promoting `overflow-y → auto` off it).

Single-file SCSS change; mirrors the Settings page's working layout.

## Verification

Deployed to the boat and confirmed on the tablet: long help docs now scroll through the entire document via touch. Settings-parity layout, so desktop wheel scrolling follows.

Closes #437.

### DIFF
--- a/src/app/core/components/app-help/app-help.component.scss
+++ b/src/app/core/components/app-help/app-help.component.scss
@@ -33,10 +33,6 @@
   letter-spacing: var(--mat-dialog-subhead-tracking, var(--mat-sys-headline-small-tracking, 0.03125em));
 }
 
-.mat-mdc-dialog-content {
-  max-height: max-content;
-}
-
 .dialog-close-icon {
   display: flex;
   flex-direction: row;
@@ -55,7 +51,7 @@
   background: var(--mat-sys-surface-container-high);
 }
 
-::ng-deep .markdown-content {
+.markdown-content {
   display: block;
   flex: 1 1 auto;
   min-height: 0;
--- a/src/app/core/components/app-help/app-help.component.scss
+++ b/src/app/core/components/app-help/app-help.component.scss
@@ -1,7 +1,15 @@
+:host {
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  min-height: 0;
+}
+
 .fullpage-header {
   display: flex;
   justify-content: space-between;
   align-items: center;
+  flex: 0 0 auto;
 }
 
 .page-icon {
@@ -49,9 +57,10 @@
 
 ::ng-deep .markdown-content {
   display: block;
-  height: calc(100% - 75px);
+  flex: 1 1 auto;
+  min-height: 0;
   width: 100%;
-  overflow-x: scroll;
+  overflow: auto;
   padding: 25px;
   font-size: 1.1rem;
   line-height: 1.7;


mairas:
## Code review — PR #470

**Scope:** 1 file, ~12 LOC (`app-help.component.scss`) vs `origin/main` @ `6be2f09d`.
**Intent:** fix #437 — Help page scrolled only a few pixels; give the routed host a bounded flex column so `.markdown-content` can fill and scroll.
**Reviewers:** correctness, testing, maintainability, project-standards (4 always-on; CSS-only <50 LOC, no auth/API/reliability surface → no conditionals).

The fix is sound and independently confirmed: `:host { height: 100% }` resolves against the shell's definite-height flex column (`.skip-app-shell`), matching the working Settings `tabs.component` under the *same* shell; the three `.markdown-content` template variants are mutually exclusive; grep found **no other `.markdown-content`** in the repo; `overflow: auto` is a strict improvement over the old always-on `overflow-x: scroll`; **no VERSION bump required** (1.0.5 cycle already open, patch-level fix); existing app-help specs don't touch the changed styles (stay green).

### P3 — Low (both pre-existing, optional)

| 1 | `app-help.component.scss:~58` | The container's layout is applied through a **global** `::ng-deep .markdown-content` selector. But `.markdown-content` is app-help's *own* template element — it needs no encapsulation piercing (only the descendant rules that style ngx-markdown output do). Routing the scroll fix's `flex`/`overflow` through a global selector means any future `.markdown-content` elsewhere silently inherits this scroll behavior. Inert today (unique class), but this change widens the global rule's blast radius. | correctness + maintainability | ~0.82 | drop `::ng-deep` from the container rule (keep it on the `h1/pre/table/...` descendants) |
| 2 | `app-help.component.scss:~36` | Dead `.mat-mdc-dialog-content { max-height: max-content }` — a vestige from when Help was a dialog. Help is now a routed full-page component, so the class never appears in the template and the rule can never apply. | maintainability | 0.80 | delete the rule |

### Coverage
- Suppressed: 0.
- Testing: intentionally none added — CSS flex/scroll has no layout engine in jsdom, and a computed-style/text-match assertion would be false-confidence. Visual verification on the boat tablet (done) is the correct check for this class. Residual: the fix relies on an unbroken `height:100%` / `min-height:0` chain; a future ancestor dropping height would silently regress it (would break Settings identically — a shell-wide invariant).
- No shared page-layout mixin (settings/remote/connection/help): premature — only 2 instances and they diverge in technique (Settings still uses `calc(100% - 63px)`); YAGNI.

---

No functional issues. Two optional P3 cleanups, both pre-existing but in the touched file and worth folding in: scope the container rule (#1, removes the global-leak footgun this change touches) and delete the dead dialog rule (#2).

<sub>🤖 Review orchestrated with Claude Code — 4 independent persona reviewers.</sub>

mairas:
Review cleanups applied in `c365616a`: scoped the `.markdown-content` container rule (dropped the unnecessary top-level `::ng-deep`, kept on the descendant rules that pierce into ngx-markdown output) — finding #1; deleted the dead `.mat-mdc-dialog-content` rule — finding #2. `build:dev` green. No behavior change (same element, same styles, just scoped). Merging once CI passes.

PR #471, opened 2026-07-30
fix(pwa): don't lock the installed app to landscape orientation

mairas:
## Motivation

Installing Skip as a PWA in Brave ("Install and create shortcut" → "Install") forced the app into **landscape**. The web-app manifest set `orientation: "landscape"`, which installed PWAs obey as an orientation **lock**.

The only other orientation references are the `apple-touch-startup-image` splash media queries in `index.html` (which select a splash image *by* orientation — they don't lock), and there is no `screen.orientation.lock()` anywhere in the app, so the manifest field was the sole cause.

## Approach

`src/manifest.json`: `orientation: "landscape"` → `"any"` — the installed app now follows the device orientation. `display: fullscreen` and `id` / `scope` / `start_url` are unchanged.

## Verification

Deployed to the boat; confirmed the **served** manifest reports `orientation: "any"`, then re-installed the PWA in Brave and confirmed it no longer forces landscape (rotates with the device).

Note for existing installs: the manifest is read at install time, so an already-installed shortcut keeps the old lock until it is uninstalled and re-installed.

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -8,7 +8,7 @@
   "name": "Skip",
   "short_name": "Skip",
   "description": "An advanced and versatile marine instrumentation package to display Signal K data.",
-  "orientation": "landscape",
+  "orientation": "any",
   "icons": [
     {
       "src": "assets/manifest-icon-192.maskable.png",


mairas:
## Code review (light) — PR #471

**Scope:** 1 line (`src/manifest.json`, `orientation: landscape → any`) vs `origin/main` @ `c4235090`.
**Reviewers (right-sized for a one-value manifest change):** project-standards, correctness. (Skipped testing — no testable surface; maintainability — nothing in a value change.)

**No findings.**

- **project-standards:** ✅ No VERSION bump required (1.0.5 cycle already open; patch-level fix). Fork manifest invariants intact — `id` / `scope` / `start_url` all unchanged at `/@halos-org/skip/`.
- **correctness:** ✅ `"any"` is the correct W3C manifest value ("no lock / follow device") — better than `"natural"` (would lock to portrait on phones) or deleting the field (relies on UA default). No conflict with `display: fullscreen` (independent members). No `screen.orientation` / `.lock()` anywhere in `src/`. The only landscape CSS (`dashboard.component.scss:84`) is additive short-height padding that simply doesn't apply in portrait — no hard landscape assumption to break.

**Residual (non-blocking):** portrait rendering was confirmed only as "no longer forced landscape" in Brave, not visually swept across every breakpoint — phone/tablet *portrait* layout falls through to default styling and wasn't exhaustively checked. The app is portrait-aware, so any crowding at very narrow widths would be a separate cosmetic follow-up, not a reason to keep the orientation lock.

---

<sub>🤖 Light review — 2 focused persona reviewers.</sub>

Issue #476, opened 2026-07-31
WindSteer: course-up fallback to COG when heading is unavailable


mairas:
Follow-up to #475 (deferred deliberately).

When heading is lost but COG is available, orient the WindSteer card **course-up** (COG) instead of freezing/blanking. This is more than a label swap: `currentHeading` is the single reference frame for the entire compass-mode display, so a COG fallback cascades into:

- the dial rotation
- the **COG arrow** (`cogAngle − heading`) → becomes degenerate (points straight up) and should be hidden in fallback
- the **TWA needle** and `computeTrueWindBase`
- the **laylines** (`toDialLocal`)
- the **wind sectors** (`computeSectorPath`)

So the whole card becomes course-up. Considerations:
- **Compass mode only** — simple mode is already bow-up.
- **True vs Magnetic** — `headingPath` and `courseOverGround` are independently True/Magnetic; a Magnetic-heading → True-COG fallback silently injects magnetic variation as a card offset. Needs handling or a documented constraint.
- Likely warrants **its own toggle** (course-up vs freeze/blank) rather than being implicit.

Depends on #475 landing first (freeze/blank/hide-labels is the both-missing terminal state).

PR #477, opened 2026-07-31
WindSteer: freeze then hide on data loss instead of showing 0°

### LINKED ISSUE #475: WindSteer: absent/invalid data renders as 0° / 0 speed instead of freeze-then-hide
### Problem

The TWA needle on the WindSteer widget intermittently jumps to 0°. Root cause is a widget-side handling bug, not the data:

- Every angle/speed handler coerces an absent/invalid sample to `0` (`if (u.data.value == null) u.data.value = 0` in `onHeadingUpdate`/`onCOGUpdate`/`onSetUpdate`/`onAppWindAngle`/`onTrueWindAngle` and the speed handlers), then renders it.
- The needles aren't gated on data presence. The **laylines** already hide on data loss (`… && trueWindActive()`), but the **TWA needle** shows on `twaEnabled()` alone, and the **AWA needle** has no presence gate at all — so on a dropout they snap to 0° (bow) and stay shown.
- `lastRawTrueWindAngle` also caches the `0`, so a later config-change recompute re-asserts it.

TWA is the visible case because true wind legitimately drops out; heading/AWA rarely drop, so their identical bug is latent.

**Principle:** lack of data or invalid data must never be shown as a value (0°, 0 kn, north).

### Fix — freeze, then hide after the TTL

1. **Never write 0 on absent/invalid data.** Each handler skips the update on a null/non-finite sample, so the signal holds its last value (freeze).
2. **Hide after the TTL.** A per-path "fresh" flag, true only while a valid sample arrived within the last **5 s** (the widget's `dataTimeout`), re-armed on each valid sample. Needles/readouts gate on it: brief dropout → frozen but visible; sustained (>5 s) → hidden; recovery → shown again. Also fixes the *before-first-data* case (starts hidden, not at 0). Independent of the existing `enableTimeout` streams timeout (works whether that's on or off; no conflict).

**Scope (all angle + speed paths):**
- Needles — TWA, AWA, COG, drift-set: freeze, then **hide** after TTL.
- Speed readouts — AWS, TWS, drift: freeze, then show **"--"** after TTL (never 0 kn).
- Heading — freeze (never spin to north) + heading readout **"--"** + **hide the N/E/S/W + degree labels** once heading is stale past the TTL. (Simple mode is already bow-up, so only the readout/labels matter there.)

### Out of scope (separate issue)

Course-up fallback to COG when heading is unavailable — that re-references the whole card (COG arrow, TWA, laylines, sectors) and is a display-mode change, tracked separately.

mairas:
## What

Fixes the WindSteer widget rendering absent/invalid data as `0` — the reported symptom being the TWA needle intermittently jumping to 0° (bow). Closes #475.

## Root cause

Every angle/speed handler coerced a missing sample to zero (`if (u.data.value == null) u.data.value = 0`) and rendered it, and the needles weren't gated on data presence. The laylines already hid on data loss (`… && trueWindActive()`) but the TWA needle showed on `twaEnabled()` alone and the AWA needle had no presence gate at all — so on a dropout they snapped to 0° and stayed. TWA is the visible case because true wind legitimately drops out; the identical bug on heading/AWA/etc. was latent.

## Fix — freeze, then hide after the TTL

- **Never write 0 on absent/invalid data.** Each handler skips the update on a null/non-finite sample, so the value signal holds its last value (freeze). Value signals stay real numbers, so the widget's internal frame math (`computeTrueWindBase`, wind sectors, laylines) keeps working on the last value.
- **Hide after the TTL.** A per-path freshness flag, set true on each valid sample and re-armed with a 5 s (`dataTimeout`) hide-timer, gates the indicator: brief dropout → frozen but visible; sustained loss (>5 s) → hidden; recovery → shown again. Also hides *before the first sample* instead of showing 0. Independent of the streams-directive `enableTimeout` (works whether that's on or off; no conflict).

Applied to every angle path (heading, AWA, TWA, COG, drift-set) and the speed readouts (AWS/TWS/drift → **"--"**, never `0` kn). Heading additionally blanks its readout to "--" and hides the compass N/E/S/W + degree labels (`dialLabelsCompass`, nested inside the existing compass-mode `@if` so simple-mode labels aren't affected) once stale; the dial freezes rather than spinning to north.

## Out of scope

Course-up fallback to COG when heading is lost — that re-references the whole card (COG arrow, TWA, laylines, sectors) and is a display-mode change, tracked in #476.

## Testing

- `npm run snc`, `npm run lint`, full `npm test` (**1746**, +7) — all green. No `DEFAULT_CONFIG` / schema change.
- New coverage: freeze holds the last value on null (doesn't reset to 0); freshness flips false after the 5 s TTL and re-arms on later samples (fake timers); SVG gates hide the compass labels + blank the heading readout, hide a stale needle, and render "--" for a stale speed readout.
- Deployed to the boat (which publishes intermittent-null TWA) for a live look.


mairas:
## Code review — freeze-then-hide on data loss (#475)

**Scope:** commit `ca99e924` vs `origin/main`, 6 files. **Mode:** interactive, PR-first.
**Intent:** stop the WindSteer widget rendering absent/invalid data as 0 (TWA needle jumping to 0°); freeze on brief dropout, hide each indicator after a 5s TTL, across all angle + speed paths.

**Team (6):** correctness, testing, maintainability, project-standards (always-on) + reliability (adds per-path timers / data-loss handling) + adversarial (>50 non-test changed lines).

No P0/P1. Reviewers confirmed the freeze guards (null/NaN/±Inf all freeze, value + sector history untouched — and it actually removes a latent NaN-poisoning of the sector deque the old `value=0` allowed), the timer lifecycle (ngOnDestroy clears all; ≤1 live timer per key; zoneless signal-write from setTimeout reaches OnPush CD like the existing rAF/interval code), the no-double-hide interaction with the streams `enableTimeout`, and the nested-`@if` label matrix (compass+stale → neither label group; simple mode keeps port/stbd labels).

### P2 — High
| 1 | widget-windsteer.component.ts | **Shared `driftFresh` key masked a stalled sub-channel** — `onDriftUpdate` (speed) and `onSetUpdate` (direction) re-armed one flag, so if only one path died the other kept rendering a frozen value | adversarial, reliability, maintainability, testing, correctness | 0.90 | **fixed** — split into `driftFresh` (speed → readout) + `setFresh` (direction → arrow), gated independently; added an independence test |
| 2 | widget-windsteer.component.ts | `*Active` name overload — freshness `headingActive` read like the threshold `driftActive`/`sogActive` | maintainability | 0.68 | **fixed** — renamed the freshness set to `*Fresh` (headingFresh/courseFresh/appWindFresh/appWindSpeedFresh/trueWindSpeedFresh/trueWindFresh + drift/setFresh); threshold gates stay `*Active` |
| 3 | widget-windsteer.component.spec.ts | Test gaps: `driftFresh`/`setFresh`, freeze-on-null only heading+AWA, `appWindSpeed` freshness, resume-after-hide, COG/TWA needle-hide | testing | 0.90 | **fixed** — freeze test now covers all 7 paths; added split-independence, resume-after-hide, AWS readout, COG/TWA needle-hide, set-arrow-hide tests (+6) |

### P3 — Low
| 4 | widget-windsteer.component.ts | `dataTimeout ?? 5` didn't catch `0`/NaN (→ instant hide), and the TTL could be shorter than a slow `updateInterval` (no max) → flicker on healthy data | correctness, reliability, adversarial, project-standards | 0.72 | **fixed** — `dataTtlMs` treats non-finite/≤0 as the default and floors at `cadence·2`; extracted `DEFAULT_DATA_TIMEOUT_SECONDS` (mirrors the sector-window const, killing the duplicated `5`) |
| 5 | svg-windsteer.component.svg | Recovery "flash" — a hidden needle reappears at its frozen pre-dropout angle and sweeps to the new one over ~1s after a >5s dropout | correctness, adversarial | 0.68 | **known limitation** — deferred (see below) |

### Known limitation (finding #5, not fixed)
On recovery after a >5 s outage with a materially different angle, the needle re-appears at its stale angle and animates to the correct one over ~one animation duration (it never takes the snap branch because it's not the first sample). Final value is correct; it's a one-time visible sweep. The clean fix (snap on an inactive→active transition, tracked per needle) adds real state to the SVG animation effects, so it's deferred; happy to file a follow-up if you want it.

### Residual notes (accepted)
- Freeze-then-hide is now **always-on**, so the `enableTimeout` config checkbox no longer solely governs blanking for this widget (blanking is the fix's intent). Worth a later label tidy.
- A frozen **heading** during a heading-only dropout can drift the sector/layline placement (they read the stale heading); the compass labels + heading readout blank to signal the staleness, and the TWA needle itself stays correct (the stale heading cancels out). Strictly better than the old reset-to-0.

### Coverage
- No untracked files excluded. All 6 reviewers returned.

Zero P0/P1. All P2 and the actionable P3 are fixed in `ca99e924`; the one remaining P3 is a documented, deferred cosmetic. Local gates: `snc`, `lint`, `test:mcp-schema`, full `npm test` (**1752**, +6). No `DEFAULT_CONFIG`/schema value change. Deployed to the boat for a live look.

🤖 Reviewed via mairas-workflow (6 independent Opus persona agents)

PR #480, opened 2026-07-31
fix(audio): play alarms via Web Audio to drop the Android media card

### LINKED ISSUE #479: Alarm audio spawns an Android media-session notification (play/pause + seek + logo)
## Symptom

Installed as an Android PWA, Skip shows a media-player card in the notification drawer — play/pause, a seek slider, app logo behind it — like a music app.

## Root cause

Alarm sounds play through **looping `HTMLAudioElement`** (`new Audio(...)`, `loop = true`):

- `NotificationsService.getPlayer()` (`src/app/core/services/notifications.service.ts:404-406`) — alarm tracks 1001–1004 (`alert`/`warn`/`alarm`/`emergency`).
- `ToastService` (`src/app/core/services/toast.service.ts:61`) — one-shot `notification.mp3`.

Chrome on Android shows a Media Session notification for any `<audio>`/`<video>` that plays **≥ 5 s** and is audible. The looping alarm satisfies that; it lingers because clearing an alarm only calls `pause()` and `navigator.mediaSession` is never cleared, so a **paused** card stays in the drawer. Confirmed against Chrome docs: Web Audio playback produces no such card. Only the looping alarm triggers it; the sub-5 s toast chime does not.

## On-device verification (done)

The load-bearing risk was that the fix could trade a cosmetic card for a **silent alarm when the phone is backgrounded/locked** (AudioContext suspension). Verified on-device: an audible `AudioContext` oscillator loop **kept playing in the background with no media session at all**. Since a backgrounded browser tab is throttled at least as hard as an installed PWA, this is conservative evidence that Web Audio survives backgrounding in the real PWA. **The approach is viable.** The lighter `mediaSession`-clear alternative was only a fallback for a negative result and is not pursued.

## Approach

Move all notification/alarm audio off `HTMLAudioElement` onto the **Web Audio API** (`AudioContext` + `AudioBufferSourceNode`), which produces no media card.

- **New shared `SoundService`** (`providedIn: 'root'`) owns all Web Audio playback. `NotificationsService` and `ToastService` consume it.
- **Convert both** alarm (looping) and toast (one-shot). No `HTMLAudioElement` audio remains — a deliberate secondary goal (one audio path), owned beyond the card fix which the alarm alone requires.
- **No `navigator.mediaSession` management** — the card is removed by eliminating the audio element. In-app mute stays the way to silence an active alarm.

### `SoundService`

- **Lazy, feature-detected `AudioContext`** (`window.AudioContext || webkitAudioContext`). Never created at construction; absent (jsdom) ⇒ every method is an inert no-op (keeps the ~30 real-service specs safe).
- **Pre-decode alarm buffers** (`alert`/`warn`/`alarm`/`emergency`) as soon as the context is created (first user gesture), so a cold first alarm is not silent while `decodeAudioData` runs. Buffers cached by key; decode failures degrade to inert.
- **Looping playback** — one active source; a new key replaces the current; a monotonic token guards the async decode→start against a superseding stop/switch. The **same-key no-op guard lives in `NotificationsService.playAlarm`** so a reconnect never restarts the tone.
- **Start-while-suspended** — the looping source is started even if the context is suspended (Web Audio plays it once the context resumes), so the one-time gesture-unlock makes an already-active alarm audible without re-invoking `playAlarm` (which the same-key guard would reject).
- **One-shot playback** at a caller gain. Alarms play at **gain 1.0 (full volume)**; the toast at **0.3**.
- **Gesture unlock** — a one-time global user-gesture listener resumes the context. A `blocked` signal (set when a play is attempted against a suspended context, cleared on resume) lets a caller show the existing "audio blocked" prompt. `SoundService` takes no `ToastService` dependency.

### Consumers

- `NotificationsService` — `playAlarm` maps 1000→`stopLoop`, 1001–1004→`playLoop(name, 1.0)`; keeps `_activeAlarmSoundtrack` and the same-key early-return. Drops `_players`/`getPlayer` and the constructor pre-cache. `ngOnDestroy` calls `stopLoop()` only (never closes the shared context). Owns the **single** audio-blocked prompt via an effect on `SoundService.blocked`.
- **Mute stays severity-driven** — `mutePlayer` sets `_isMuted`; `getNotificationSeverity` already zeroes `audioSev`, so `updateNotificationsState` calls `playAlarm(1000)` and the source stops. No GainNode-mute (it would be redundant); the vestigial `.muted` line is removed. Unmute restarts the tone from the start, unchanged from today.
- `ToastService` — one-shot via `SoundService` at 0.3 gain; drops its own `HTMLAudioElement` and its separate blocked prompt (consolidated into the one owned by `NotificationsService`).

### Test plan

- All `activeTrack()` / `latestInfo()` assertions (track-id + mute + severity) survive unchanged.
- The two reconnect specs that spied `HTMLAudioElement.play`/`pause` are re-expressed against a **spec-local `SoundService` fake** whose `playLoop`/`stopLoop` are spied — a local fake **does** override a root service injected by a root service (the spec already does this for SettingsService/DataService; the earlier "can't override" rationale was wrong). Assert the looping track is not stopped/restarted across a reconnect.
- New `SoundService` spec: inert (no throw, no context) when `AudioContext` is absent; with a fake `AudioContext` + stubbed `fetch`/`decodeAudioData`: `playLoop` decodes and starts a looping source; same-key is a no-op; a new key replaces the previous; `start()` is called even when the context is suspended; `blocked` flips true on a suspended play and false on resume.

### Device acceptance criteria (manual, pre-merge)

Unit specs with a fake `AudioContext` cover refactor equivalence only, not platform audio. Before merge, on the Android PWA: (1) an emergency alarm sounds within ~1 s from cold with no prior alarm; (2) the alarm keeps sounding with the screen locked for ≥ 60 s; (3) the media-player card no longer appears; (4) muting from the in-app menu silences it.

mairas:
## Why

Installed as an Android PWA, Skip's looping alarm plays through a `HTMLAudioElement`, which makes Chrome show (and leave lingering) a media-player card in the notification drawer — play/pause, seek slider, app logo — so an alarm looks like a music app. Fixes #479.

## What

Route all notification audio through the Web Audio API, which never spawns a media-session card.

- **New `SoundService`** (`providedIn: 'root'`) owns a lazily created, feature-detected `AudioContext` (inert no-op when absent, so the ~30 real-service jsdom specs stay safe), a decoded-buffer cache with the four alarm tracks pre-decoded, looping + one-shot playback, and gesture-based unlock exposing a `blocked` signal.
- **`NotificationsService`** plays alarm loops via `SoundService`; the same-key guard keeps a reconnect from restarting the tone. Mute stays severity-driven (a muted alarm zeroes `audioSev` and stops via `playAlarm(1000)`); the vestigial per-player `.muted` is dropped. The autoplay-blocked prompt is consolidated here, driven off `blocked`.
- **`ToastService`** plays its chime one-shot via `SoundService` (0.3 gain); alarms play full volume.

## Verification

- `npm run ci` (lint + snc): clean. Full unit suite: all pass except one **pre-existing** failure (`app.component.spec.ts`, a local `localStorage` env artifact that reproduces on clean `main` and is unrelated to this change).
- New `SoundService` spec covers: inert-when-unsupported, decode+loop start, same-key no-op, key replacement, start-while-suspended, and the `blocked` transitions. The two reconnect specs now assert against a spec-local `SoundService` fake.
- **Premise verified on-device:** an audible Web Audio loop keeps playing when the Android PWA is backgrounded, so removing the card does not silence a backgrounded/locked alarm.

## Before merge — manual device acceptance (per #479)

Unit specs use a fake `AudioContext` and cannot cover platform audio. On the Android PWA, confirm: (1) a cold emergency alarm sounds within ~1 s; (2) it keeps sounding with the screen locked ≥ 60 s; (3) the media card no longer appears; (4) in-app mute silences it.

https://claude.ai/code/session_01CDtPDenpQ64FybWqpKiYoc

### DIFF
--- a/src/app/core/services/notifications.service.spec.ts
+++ b/src/app/core/services/notifications.service.spec.ts
@@ -1,5 +1,5 @@
 import { TestBed } from '@angular/core/testing';
-import { signal, WritableSignal } from '@angular/core';
+import { signal, Signal, WritableSignal } from '@angular/core';
 import { Subject } from 'rxjs';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 import { INotification, INotificationInfo, NotificationsService } from './notifications.service';
@@ -7,6 +7,7 @@ import { DataService } from './data.service';
 import { SettingsService } from './settings.service';
 import { SignalkRequestsService } from './signalk-requests.service';
 import { ConnectionState, ConnectionStateMachine } from './connection-state-machine.service';
+import { SoundService } from './sound.service';
 import { INotificationConfig } from '../interfaces/app-settings.interfaces';
 import { ISignalKDataValueUpdate, ISignalKNotification, ISkMetadata, Methods, States, TMethod, TState } from '../interfaces/signalk-interfaces';
 import { IMeta } from '../interfaces/app-interfaces';
@@ -54,6 +55,12 @@ describe('NotificationsService', () => {
   let connectionState$: Subject<ConnectionState>;
   let putRequest: ReturnType<typeof vi.fn>;
   let service: NotificationsService;
+  let sound: {
+    playLoop: ReturnType<typeof vi.fn>;
+    stopLoop: ReturnType<typeof vi.fn>;
+    playOnce: ReturnType<typeof vi.fn>;
+    blocked: Signal<boolean>;
+  };
 
   function setup(initialConfig: INotificationConfig = makeConfig()): void {
     configSignal = signal<INotificationConfig>(initialConfig);
@@ -73,6 +80,7 @@ describe('NotificationsService', () => {
     const requestsStub: Partial<SignalkRequestsService> = {
       putRequest: putRequest as SignalkRequestsService['putRequest'],
     };
+    sound = { playLoop: vi.fn(), stopLoop: vi.fn(), playOnce: vi.fn(), blocked: signal(false) };
 
     TestBed.configureTestingModule({
       providers: [
@@ -80,6 +88,7 @@ describe('NotificationsService', () => {
         { provide: DataService, useValue: dataStub },
         { provide: SignalkRequestsService, useValue: requestsStub },
         { provide: ConnectionStateMachine, useValue: { state$: connectionState$.asObservable() } },
+        { provide: SoundService, useValue: sound as unknown as SoundService },
       ],
     });
     service = TestBed.inject(NotificationsService);
@@ -99,10 +108,6 @@ describe('NotificationsService', () => {
 
   function activeTrack(): number | null {
     return (service as unknown as { _activeAlarmSoundtrack: number | null })._activeAlarmSoundtrack;
-  }
-
-  function alarmPlayer(track: number): HTMLAudioElement {
-    return (service as unknown as { _players: Map<number, HTMLAudioElement> })._players.get(track)!;
   }
 
   describe('alarm state aggregation from deltas', () => {
@@ -401,9 +406,8 @@ describe('NotificationsService', () => {
       expect(latestNotifications()).toHaveLength(1);
       expect(activeTrack()).toBe(1003);
 
-      const player = alarmPlayer(1003);
-      const playSpy = vi.spyOn(player, 'play');
-      const pauseSpy = vi.spyOn(player, 'pause');
+      sound.playLoop.mockClear();
+      sound.stopLoop.mockClear();
 
       vi.useFakeTimers();
       // A drop must not clear; the reconnect that follows must not blank or silence the alarm.
@@ -414,15 +418,15 @@ describe('NotificationsService', () => {
       expect(activeTrack()).toBe(1003);
 
       // The snapshot re-delivers the same still-active alarm: recognised as the existing entry, not
-      // cleared-and-re-added, so the looping player is never paused/restarted.
+      // cleared-and-re-added, so the looping track is never stopped/restarted.
       notificationMsg$.next(makeDelta('notifications.bilge', makeValue(States.Alarm, [Methods.Visual, Methods.Sound])));
       vi.runOnlyPendingTimers();
 
       expect(latestNotifications()).toHaveLength(1);
       expect(latestInfo().alarmCount).toBe(1);
       expect(activeTrack()).toBe(1003);
-      expect(playSpy).not.toHaveBeenCalled();
-      expect(pauseSpy).not.toHaveBeenCalled();
+      expect(sound.playLoop).not.toHaveBeenCalled();
+      expect(sound.stopLoop).not.toHaveBeenCalled();
     });
 
     it('sweeps a notification the server no longer reports once the snapshot settles', () => {
@@ -494,7 +498,7 @@ describe('NotificationsService', () => {
       expect(latestNotifications()).toHaveLength(1);
       expect(activeTrack()).toBe(1003);
 
-      const pauseSpy = vi.spyOn(alarmPlayer(1003), 'pause');
+      sound.stopLoop.mockClear();
 
       vi.useFakeTimers();
       connectionState$.next(ConnectionState.Disconnected);
@@ -508,7 +512,7 @@ describe('NotificationsService', () => {
       expect(latestNotifications().map(n => n.path)).toContain('notifications.bilge');
       expect(latestInfo().alarmCount).toBe(1);
       expect(activeTrack()).toBe(1003);
-      expect(pauseSpy).not.toHaveBeenCalled();
+      expect(sound.stopLoop).not.toHaveBeenCalled();
     });
 
     it('does not reconcile or clear notifications while the connection stays up', () => {
--- a/src/app/core/services/notifications.service.spec.ts
+++ b/src/app/core/services/notifications.service.spec.ts
@@ -1,5 +1,5 @@
 import { TestBed } from '@angular/core/testing';
-import { signal, Signal, WritableSignal } from '@angular/core';
+import { signal, WritableSignal } from '@angular/core';
 import { Subject } from 'rxjs';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 import { INotification, INotificationInfo, NotificationsService } from './notifications.service';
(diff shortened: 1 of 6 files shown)


mairas:
## Code review — Web Audio alarm migration

**Scope:** `feat/web-audio-alarm-playback` vs `main` (5 files) · **Intent:** move alarm/toast audio off `HTMLAudioElement` onto a root `SoundService` (Web Audio) to drop the Android media card, without regressing alarm audibility, mute, or the reconnect no-restart invariant.

**Reviewers:** correctness, testing, maintainability, project-standards (always-on) + reliability (async decode / context lifecycle) + adversarial (safety-relevant alarm path, >50 LOC).

Verdict: **Not ready** — one P1 to fix, then merge is gated on the manual device-acceptance in the PR body. Fixing the items below in this PR.

### P1 — High

| 1 | notifications.service.ts:412 / sound.service.ts:51 | **A transient decode/fetch failure permanently silences an active alarm.** `playAlarm` latches `_activeAlarmSoundtrack = trackId` *before* `sound.playLoop` confirms playback; if the decode's `.then` hits a null buffer it returns without setting `current`. Every later recompute re-calls `playAlarm(1003)` which short-circuits at the same-key guard, so `playLoop` is never re-invoked — the marine safety alarm stays silent for its whole episode. `SoundService` deletes the failed decode (would retry), but the caller-side guard blocks it. | adversarial, reliability, correctness | 0.85 |

### P2 — Moderate

| 2 | notifications.service.ts:112 | Autoplay-blocked prompt (`effect(blocked())` + `showAudioBlockedPrompt`) has **zero coverage** — the fake `blocked` signal is never flipped true. | testing | 0.85 |
| 3 | toast.service.ts:60 | `ToastService.playOnce('notification', 0.3)` delegation is **never exercised** (spec uses `disableSound:true` / `silent`), so the audio swap ships untested. | testing | 0.85 |
| 4 | sound.service.ts:123 | Decode/fetch **failure path untested** — `fetch`/`decodeAudioData` are stubbed to always succeed. | testing, reliability | 0.80 |
| 5 | sound.service.ts:91 | **Background/screen-lock suspension** could silence the alarm; no `visibilitychange` handling. Device-dependent. | adversarial | 0.60 |

### P3 — Low

| 6 | notifications.service.ts:112 | Blocked-audio toast (`duration:0`) **lingers after the context resumes** — nothing dismisses it on `blocked→false`; the flag only resets on manual dismiss. | correctness, adversarial | 0.68 |
| 7 | notifications.service.ts:20 | Dead/misleading `alarmTrack` entry `1000: 'notification'` — unreachable (guard stops at `!name || trackId===1000`) and implies 1000 plays a sound when it silences. | maintainability | 0.83 |
| 8 | notifications.service.spec.ts:322 | Track-selection tests assert only the private `_activeAlarmSoundtrack`; nothing asserts `playLoop` is called with the mapped name/gain (1003→'alarm', gain 1) or `stopLoop` on clear. | testing, project-standards | 0.72 |
| 9 | toast.service.ts:60 | Blocked-prompt ownership moved wholly to `NotificationsService`; undiscoverable from `ToastService`, and the "Alarm sounds blocked" wording is alarm-specific even for a toast-triggered block. | maintainability | 0.63 |
| 10 | sound.service.ts:68 | `playOnce` happy path (`loop=false`, `onended` disconnect) untested. | testing | 0.80 |
| 11 | sound.service.ts:119 | No timeout/abort on the asset `fetch`; a stalled Pi-class request never rejects. Low risk for a local asset. | reliability | 0.65 |

### Pre-existing (not this PR)

- **Mute suppresses an emergency/MOB escalation.** `getNotificationSeverity` zeroes `aSev` whenever `_isMuted`, so after muting, a later emergency computes `audioSev=0` and stays silent. Preserved verbatim from the old code — a real safety property worth a product decision (should emergency override mute?), not a regression here. Route: file a follow-up issue if the team wants to change it.

### Being fixed in this PR

- **#1** — retry a failed alarm decode inside `SoundService` (bounded) so a transient failure re-attempts instead of latching silent; keeps the reconnect no-restart invariant.
- **#6 / #9** — dismiss the blocked toast on `blocked→false` and generalize its wording.
- **#7** — drop the dead `alarmTrack` 1000 entry.
- **#2, #3, #4, #8, #10** — add the missing tests (blocked-prompt lifecycle, ToastService `playOnce`, decode-failure + retry, positive `playLoop` name/gain, `playOnce` cleanup).

### Left as residual / advisory

- **#5** background suspension — covered by the manual device-acceptance criteria in the PR body (and on-device probing already showed Web Audio keeps playing backgrounded). No code change.
- **#11** fetch timeout — low risk for a bundled local asset; the decode-retry (#1) already re-attempts. Noted, not changing.

### Coverage

- project-standards: **0 violations** — no `any`, null-safety clean, testing model respected, comments WHY-focused, and the patch-level no-bump is correct (cycle already open at 1.1.1).
- Residual risks: background-suspension is device-dependent (acceptance-gated); the version-bump *level* is an author judgment the CI check can't enforce (patch is defensible for a behavior-preserving reimplementation).

mairas:
### Review findings addressed (70ca26d)

- **P1 #1** — `SoundService.playLoop` now retries a failed decode (bounded, 3× / 500ms) so a transient fetch/decode failure no longer latches a marine alarm permanently silent. The caller-side same-key guard and reconnect invariant are unchanged.
- **P3 #6 / #9** — the "sounds blocked" toast is now dismissed when the context resumes (captured ref, dismissed on `blocked→false`); wording generalized to cover toast-triggered blocks; added a breadcrumb comment in `ToastService`.
- **P3 #7** — dropped the dead `alarmTrack` `1000→'notification'` entry.
- **Coverage (#2, #3, #4, #8, #10)** — added: decode-retry, `playOnce` (loop=false + `onended` disconnect), the blocked-prompt lifecycle (shown on block, dismissed on unblock), `ToastService.playOnce` firing/gated paths, and a positive `playLoop('alarm'|'emergency', 1)` / `stopLoop` mapping assertion. Service specs: 51 tests green; lint + snc clean.

**Left as noted:** #5 background suspension (device-acceptance gated; on-device probe already showed Web Audio keeps playing backgrounded) and #11 fetch timeout (low risk for a bundled asset; the decode-retry already re-attempts). The pre-existing "mute suppresses an emergency escalation" is a product decision — say the word and I'll file a follow-up issue.

Still gated on the manual device-acceptance checklist in the PR body before merge.

mairas:
### Device acceptance ✅ — passed on halpi.hurma

Deployed this branch's build (`main-SJTZO7K3.js`, verified serving) to halpi.hurma; fired an `emergency` notification via the data fiddler and ran the checklist on the installed Android PWA:

1. **Cold start** — alarm sounds within ~1 s ✅
2. **Backgrounded** — keeps sounding with the screen locked ≥ 60 s ✅
3. **The fix** — no media-player card in the notification drawer ✅
4. **Mute** — in-app mute silences it ✅

All four green. CI green (tests / version-check / version-bump-check), review findings addressed. Ready to merge.

Issue #214, opened 2026-07-06
feat: expose Skip instruments as Freeboard-SK plotter-extension widgets


mairas:
## Summary

Expose individual Skip instruments (SVG gauges, numeric displays, wind/steer, etc.) inside Freeboard-SK as **widgets** — movable, glanceable overlays on the chart — using the Freeboard-SK [plotter extension provider](https://github.com/SignalK/freeboard-sk/blob/master/docs/api/plotter_extension_provider_plugins.md) mechanism.

## Direction / how the mechanism works

Freeboard-SK is the **host** chartplotter; Skip is a **contributor**. Contributions are declared in a manifest registered as a read-only `plotterExtensions` resource; iframe assets are served by the plugin at `/plotterext/<id>`. A **widget** is a small chart-overlay iframe placed by the user in a grid area, sized `1x1` / `2x1` / `1x2` / `2x2`, with an optional separate `configPanel` iframe for its settings.

## Prerequisite: companion server plugin

Same prerequisite as the panel variant (#213): Skip is webapp-only since #176/#177, and only a Signal K server plugin can register the `plotterExtensions` resource and serve assets at `/plotterext/<id>`. Build that thin companion plugin once, shared by both issues.

## Why this is the expensive path

A widget must speak the plotterext **bus** for everything:

- config via `client.state.get()` (per-instance scope), not Skip's dashboard config store
- live data via `client.signalk.subscribe([paths], cb)`, not Skip's SK connection
- units via the `units` capability, not Skip's `UnitsService`
- settings UI via a separate `configPanel` iframe

Skip's widgets today get all of that from Angular DI instead — `WidgetStreamsDirective` → `DataService` for data, `UnitsService` for conversions, `WidgetRuntimeDirective` for config — all riding Skip's own SK connection and dashboard config. Reusing Skip's *visual* components while replacing their entire data/config/units substrate is the work.

## Approach

1. **Widget host bundle.** A separate lightweight build entry that boots a *single* Skip SVG component in an iframe — not the Skip app shell. Skip's directives already support "manual config injection in embedded/hardcoded contexts" (used by `widget-autopilot`, `widget-freeboardsk`), so the seam to run one widget standalone exists.
2. **Bus adapter layer.** Implement Skip's data/units/config surface on top of the bus:
   - `client.signalk.subscribe()` → the value stream `WidgetStreamsDirective.observe()` feeds components
   - `units` capability → `UnitsService`
[...]

PR #483, opened 2026-07-31
feat: expose Wind Steer as a Freeboard-SK plotter-extension widget

### LINKED ISSUE #400: [Feature]: Expose "Wind Steer" component as a Plotter Extension Widget
### Feature Request

The "Wind Steer" component is one of the unique and complex components supplied by Skip. Exposing it as a small (1x1) and large (2x2) widget using the plotter extensions API would allow a Freeboard user to embed it directly on their chart (e.g. in the lower left hand corner) for an "Always on" view.  This would complement the "Panel Extension" already supplied, and would serve as a POC for other complex Skip controls that may benefit from being exposed as an FSK first class widget

[Plotter extensions API docs](https://github.com/SignalK/freeboard-sk/blob/master/docs/api/plotter-extensions-api.md)

[Widget provider reference code](https://github.com/joelkoz/signalk-instrument-widgets)

mairas:
## What

Expose Skip's **Wind Steer** control as a Freeboard-SK plotter-extension **widget** — an always-on chart tile (1x1 and 2x2) that a user can **add, configure, and remove** from the chart, alongside the existing panel. This is the concrete POC (#214, closed) for surfacing complex Skip controls as first-class Freeboard-SK widgets.

Closes #400.

## How it works

A widget iframe is Skip booted chromeless on a dedicated single-widget route, reusing the app's own machinery rather than reimplementing anything:

- **Rendering + data:** `#/widget/:type` → `SingleWidgetHostComponent` mounts one `widget-host2` full-bleed against the app's own live Signal K session. `:type` is the widget's component selector (`widget-wind-steer`). Embed-only (`embedRequiredGuard`), so the dashboard is force-locked and read-only chrome is route-enforced.
- **Manifest** (`plugin/index.js`): the existing `plotterExtensions` resource gains a `widgets[]` section (both sizes) with a `configPanel`, plus a config `panels[]` entry. `widgets`/`state` are in `optional`, never `requires` — so a Freeboard-SK without widget support still gets the panel.
- **Removal + settings menu** (`WidgetHostBridge`): a placed widget has no host-side controls, so the bus carries exactly two things — a long-press relayed to `ui.openConfigPanel` (opens the host dialog's **Remove** button + settings), and the per-instance config in host `state`, applied live to the tile via a new minimal `widget-host2.reconfigure(cfg)`. Verified against the Freeboard-SK host source.
- **Settings panel** (`#/widget-config/:type` → `WidgetConfigPanelComponent`): reuses Skip's **real** widget-options dialog (`DialogService.openWidgetOptions` → `RootModalWidgetConfigComponent`) over the bus — no duplicated form. Seeded from host `state`; Save writes the edited config back to `state`, which the widget follows via `state.changed`.

The only new dependency is `signalk-plotterext-bus` (the zero-dep reference bus client). Data still comes from Skip's own session; the bus carries only removal + config UI state.

## Generality

The wrapper is deliberately generic on widget type (`:type` is any selector), and the settings panel drives the real options dialog with whatever config it's handed — so the same three routes/components should host other Skip widgets with only a manifest entry. Whether that holds for widgets with heavier interaction or per-widget config quirks is a specific question for review (see review comment).

## Verification

- `npm run ci` green except a known pre-existing local `app.component.spec.ts` `localStorage` artifact (reproduces on clean `main`, passes on CI Node 24).
- New unit coverage: manifest shape (widgets, configPanel↔panel resolution, requires/optional), the widget↔registry/route binding, the long-press detector, the config-result handler, and the bridge enable gating.
- **On-device (halpi.hurma, v3.1.0-beta.2 Freeboard-SK):** add, live data, long-press → settings + Remove, edit → Save → live update, persistence across reload, and removal all verified working.

## Follow-ups (not in this PR)

- Per-placement weight: each widget iframe loads the full Skip bundle (documented tradeoff of reuse-the-app; measure if many tiles).
- Nested dialog chrome in the settings panel (FSK dialog → iframe → Material dialog) is functional but redundant; candidate for streamlining.
- Upstream: `SignalK/freeboard-sk#653` (the remove dialog's "no settings" text is invisible in dark mode) — not our code.

https://claude.ai/code/session_01CDtPDenpQ64FybWqpKiYoc


mairas:
## Code review — PR #483 (Wind Steer plotter-extension widget)

**Scope:** `plugin/index.{js,spec.ts}`, `app.routes.ts`, `single-widget-host/*`, docs + version bump. **Intent:** expose Wind Steer as an always-on FSK chart widget by reusing `widget-host2` in a chromeless single-widget route; must not regress the panel; read-only.
**Reviewers (6, parallel, independent):** correctness, testing, maintainability, project-standards, api-contract, adversarial.

**Verdict: Ready with fixes.** All findings are P2/P3 — no P0/P1. The core mechanism was independently verified working: correctness traced the actual Freeboard-SK host source and confirmed (a) `widget-host2`'s directives pull `DataService`/`UnitsService` from the root injector with no gridstack/dashboard dependency, so the standalone widget receives live data, and (b) the host's `isCompatible`/`addableWidgets` render our widgets *and* preserve the panel with `widgets` omitted from `requires`. Adversarial verified several risks *safe* (UUID stability, no config collision, no XSS, bus-silent tile still renders). Version bump 1.1.1→1.2.0 (minor) confirmed correct by project-standards.

### P2 — Moderate
| 1 | single-widget-host.component.ts:39 | `known` computed is dead — the template gates on `widget()`, nothing consumes `known`; it also duplicates `widget()`'s registry lookup. Spec assertions on it prove nothing. | testing, maintainability | 0.85 | fix in PR |
| 2 | plugin/index.js:26 ↔ widget.service.ts | `WIND_STEER_TYPE`/manifest URL and the route target the selector `widget-wind-steer` by unchecked string across the JS-plugin ↔ TS-app boundary. A selector rename passes all tests (plugin asserts a literal, component spec fakes the registry) but ships an "Unknown widget" tile. **This is the feature's real silent-break mode.** | testing, api-contract, maintainability, adversarial | 0.85 | fix in PR |
| 3 | app.routes.ts / component doc | Read-only rests on the `?embed=1` URL flag + default lock, not the route (no guard, unlike settings/remote/connection). Skip is also a standalone webapp, so a non-embed load can reach the edit surface (unlock on `/page/0`, navigate to the widget route). Blast radius bounded — no persistence — but the "read-only by construction" doc is inaccurate, and the JSDoc's "history affordances are inert" is wrong (history long-press is *enabled* under lock; it's a no-op only because Wind Steer isn't history-eligible). | adversarial, maintainability, project-standards | 0.70 | fix in PR (add embed-required guard + correct docs) |
| 4 | plugin/index.js (lifecycle) | `whileEnabled` + self-connecting means each placed widget boots a full always-on Skip app with its own SK websocket; panel + two widgets = 3 full apps + 3 sockets on Pi-class hardware. FSK's own guidance flags this as a poor widget shape. | adversarial, correctness | 0.70 | advisory — documented tradeoff of the reuse-the-app POC; measure on-device |

### P3 — Low
| 5 | single-widget-host.component.html:3 | The unknown-type `@else` fallback branch is never rendered by any test (spec never calls `detectChanges`). Cheap to cover (no `widget-host2` in that branch). | testing | 0.80 | fix in PR |
| 6 | plugin/index.js | `widgets` is (correctly) absent from `requires`, but the spec's reference example lists it in `optional` when contributing widgets. A stricter future host gating on the declared capability would silently drop the widgets. | correctness | 0.60 | fix in PR (`optional: ['widgets']`) |

### Coverage
- Suppressed: 0. No P0/P1.
- **Verified safe (adversarial):** UUID-in-`computed` reference is stable (recomputes only on type change, which the fixed iframe URL never triggers); `getSavedConfigForSelf` can't pull wrong config (only fires on a cancel tick never emitted here; random uuid won't collide); no XSS via `{{ type() }}`; empty/multi-segment `:type` hits the `**` redirect, not the fallback; bus-silence doesn't blank the tile (host renders the iframe src directly).
- **Testing gap that remains (accepted boundary):** no automated test mounts `widget-host2` and asserts it receives a live SK path update — jsdom has no such layer; this is what on-device acceptance covers.
- **Considered alternative:** `widget-embedded` (structural read-only, no options/action-menu/history). Not adopted here because it doesn't auto-resolve a widget's `DEFAULT_CONFIG` (Host2 does), and Host2's live-data path is already verified. Noted as a candidate if history-eligible widgets are exposed later.

### Fixes applied in this PR
1. Delete dead `known()`; spec asserts known-ness via `widget()` and renders the unknown-type fallback.
2. Add a real-registry test binding the manifest's widget type to a live `WidgetService` selector, plus a route-config assertion — a selector or route rename now fails CI.
3. Add an `embedRequiredGuard` to the `widget/:type` route so read-only is route-enforced; correct the JSDoc + CLAUDE.md (history long-press is enabled-but-ineligible, not "inert").
4. `optional: ['widgets']` in the manifest (+ spec) for spec-idiom alignment.
5. Finding #4 (weight) stays a documented tradeoff; on-device measurement gates final sign-off.

**Still gating merge:** on-device acceptance on a `widgets`-capable Freeboard-SK build (its widget capability is draft/feature-branch).

mairas:
**Review fixes applied** (3 commits: `f315828c`, `a0119152`, `811f7e19`):
- **Read-only is now route-enforced** — added `embedRequiredGuard` to `widget/:type`, closing the non-embed edit-surface path (finding #3).
- **Dropped dead `known()`** and corrected the JSDoc (history long-press is enabled-but-ineligible under embed, not "inert") (findings #1, #3).
- **Bound the manifest/route widget type to the real `WidgetService` registry and the embed-guarded route** — a selector or route rename now fails CI instead of shipping an "Unknown widget" tile (finding #2, the 4-reviewer convergence).
- **Rendered-fallback test** for the unknown-type branch (finding #5).
- **`optional: ['widgets']`** in the manifest for spec-idiom alignment (finding #6).

Finding #4 (per-placement full-app weight) stays a documented tradeoff of the reuse-the-app POC — to be measured during on-device acceptance.

Verification: `npm run ci` green except the known pre-existing local `app.component.spec.ts` localStorage artifact (reproduces on clean `main`, passes on CI). Net +5 tests.

mairas:
## Widget removal fixed (`99a96a86`)

**Diagnosis (confirmed against the Freeboard-SK plotter-ext host source, `joelkoz/freeboard-sk`):** a placed widget has *no* host-side remove path. `widget-overlay.component.ts` explicitly ignores presses on placed widget cells (`.pe-cell`), and the only trigger for the remove dialog is the widget calling `ui.openConfigPanel` — `plotterext.service.ts` `openConfigPanel` opens the dialog with `panel: null` ("still gets a dialog so it can be removed"). Pointer events inside the sandboxed iframe are invisible to the host, so a bus-silent widget can never surface that dialog. That was the bug.

**Fix:** `WidgetRemovalBridge` connects over the plotter-extension bus (`signalk-plotterext-bus@^0.11.0`, zero-dep reference client) and relays a long-press to `ui.openConfigPanel`. That's the *only* use of the bus — data still comes from Skip's own session. Outside a host (direct URL) the handshake times out and the long-press is an inert no-op.

**Scope note:** this narrows the earlier "bus-silent" framing — the widget now speaks the minimal bus needed to be removable. Per-instance **config** is still deferred (that needs a config panel + `state.*`).

**Tests:** `installLongPress` unit tests (hold fires, early-release/move cancel, teardown); component asserts the bridge is enabled for a hosted widget and not for the unknown-type fallback. `npm run ci` green except the known pre-existing local `app.component.spec.ts` localStorage artifact.

**Deployed to halpi.hurma** (v1.2.0; `ui.openConfigPanel` + bus protocol confirmed in the bundle). To test: hard-reload Freeboard-SK (the browser caches the old widget chunks), then **long-press the Wind Steer tile** → the remove dialog should appear.

mairas:
## Per-instance settings added (`0b99e0ac`)

The widget now has settings, wired into the same host long-press menu as Remove.

- **Manifest:** both widget entries declare `configPanel: 'wind-steer-config'`; a `panels[]` config entry serves `#/widget-config/widget-wind-steer`; `optional` gains `state`. With a configPanel declared, the host long-press dialog shows the **settings iframe + Remove button** together (verified against the host source: `panel-dialog.component.ts` renders both).
- **Settings panel reuses Skip's real widget-options dialog** (`DialogService.openWidgetOptions` → `RootModalWidgetConfigComponent`) — no duplicated form, and generic for any future widget. Seeded from host `state`; Save writes the edited config back to `state`.
- **Widget side:** `WidgetRemovalBridge` → `WidgetHostBridge` also reads per-instance `state` on connect and follows `state.changed`, applying edits live via a new minimal `widget-host2.reconfigure(cfg)` (delegates to the existing runtime-config path the options dialog already uses — no behavior change for the dashboard).

**Tests/CI:** `npm run ci` green except the known local `app.component.spec.ts` localStorage artifact. New unit coverage for the config-result handler; existing bridge/manifest/route tests updated. The full interactive bus loop (long-press → settings render → save → live re-apply) is on-device verification.

**Deployed to halpi.hurma** (manifest + `#/widget-config` route confirmed in the bundle).

**Known caveat:** the settings dialog opens as a Material dialog *inside* the config-panel iframe, which is inside FSK's own panel dialog — nested chrome. Functional, but may be cramped in the ~440px host dialog; will refine styling if the UX is poor on-device.

mairas:
## Thorough review — PR #483 (Wind Steer plotter-extension widget, full lifecycle)

**Scope:** full 9-commit branch vs `main` (~16 files). **Reviewers (8, parallel, independent):** correctness, testing, maintainability, project-standards, api-contract, reliability, adversarial, plus a dedicated generalization analysis.

**Verdict: Ready — review fixes applied (`0be61963`).** No P0/P1. Correctness traced the core paths and confirmed them sound (no effect loop; the reconfigure-before-init race is benign because Host2 writes config back into `widgetProperties.config`; uuid stable; config shape consistent). Reliability confirmed the bus connect/timeout/teardown are safe (connect clears its timers, calls have timeouts, `reconfigure` doesn't churn subscriptions). project-standards is clean (version bump, the new dependency, conventions, naming). The findings were medium/low — mostly latent lifecycle and forward-compat risks with clean fixes, now applied.

### Findings and dispositions

| # | Sev | Finding | Reviewers | Disposition |
|---|-----|---------|-----------|-------------|
| 1 | P2 | `WidgetHostBridge` was `providedIn:'root'` but holds per-instance connection state; `disposed` never reset → permanently inert after one teardown (latent: each iframe is a fresh injector) | maintainability, reliability, correctness, adversarial | **Fixed** — component-scoped; fresh instance per mount |
| 2 | P2 | Save `state.set` fire-and-forget, no `.catch` → failed persist = unhandled rejection + silent data loss | reliability (0.8), api-contract | **Fixed** — `.catch` added (surfaced-error UI is a follow-up) |
| 3 | P2 | Config panel seeded `saved ?? default` (no merge) → fields added by a later Skip version become un-editable in the form | adversarial, api-contract | **Fixed** — seeds `merge(currentDefault, saved)`, matching how the tile applies it |
| 4 | P2 | Persisted host-state config is an unversioned `IWidgetSvcConfig` snapshot, applied on read with only a `typeof object` guard, bypassing `ConfigurationUpgradeService` | api-contract, adversarial, correctness | **Partly fixed** — `parseStoredConfig` validates (rejects non-objects/arrays) + merge onto current default; full versioned migration of renamed/removed keys **deferred** (see below) |
| 5 | P2 | The `getComponentType`-before-`getDefaultConfig` ordering (the empty-form bug that shipped once, `99d7c7ca`) had no regression test | testing (high) | **Fixed** — regression test pins the ordering |
| 6 | P2 | Bridge bus loop (state load / `state.changed` → config signal) and the reconfigure-apply effect were untested (spec stubbed `config()→null`) | testing, maintainability, correctness, adversarial | **Fixed** — bridge state-loop test (mocked client) + extracted `applySavedConfig` with unit tests |
| 7 | P3 | Manifest URL route segments not bound to the app routes → a route rename falls through `embedRequiredGuard` to `/page/0` silently | maintainability, api-contract | **Fixed** — test asserts the `widget-config/:type` route exists + is embed-guarded (the `widget/:type` + real-registry binding tests already existed) |
| 8 | P2 | Version-skew: an FSK host that handshakes but lacks `ui.openConfigPanel`/`ui.closePanel` → dead remove/settings affordance, swallowed | adversarial (0.62) | **Deferred/verified** — confirmed against the real FSK host source (`plotterext.service.ts` registers these); a diagnostic distinction (method-not-found vs no-host) is a nice-to-have |
| 9 | P3 | Config panel shows a fieldless form if the widget chunk fails to fetch (getComponentType→undefined) — no retry/error | adversarial, correctness | **Deferred** — flaky-network edge; folds into the generalization follow-up (widgets with no static DEFAULT_CONFIG) |

### Deferred follow-ups (documented, not blocking this POC)

- **Config migration across Skip upgrades** (#4 tail): the persisted blob still isn't version-stamped, so renamed/removed config keys won't migrate. `parseStoredConfig` + default-merge cover the common add-a-field case; a versioned `{v, config}` envelope run through the upgrade path is the real fix if this ships beyond a POC.
- **Fieldless form for widgets with no static `DEFAULT_CONFIG`** (#9): ~10 registered widgets expose only an instance-level default; `getDefaultConfig` won't see it. Wind Steer is safe. This is part of the generalization work below.
- **Diagnostic surfacing** of `ui.*` method-not-found vs no-host (#8); **panel-context reachability** via a hand-typed `#/widget-config/...` hash in the main panel (needs a hand-typed URL); **`state.changed` fan-out** O(tiles) churn and a **double-save race** — all host-dependent or narrow; noted as residual risks.
- **Two-tile instance-scoping** is delegated to host state scoping (matches the spec) and can only be integration-verified on a real host with two placements.

### Generalization — can the wrapper host other Skip widgets?

**Yes, and it's already ~95% type-agnostic.** The routes (`:type`), the single-widget host, the config panel, and `widget-host2` all take any registered selector; the only Wind-Steer coupling is the two manifest constants. Notably, the settings dialog's sub-components inject only root services (not `DashboardService`), so **configuration works generically** — not the blocker I'd feared.

- **Exposing another display-only, self-scaling, plugin-free gauge is a pure-manifest, zero-code change** (compass, horizon, heel-gauge, steel, simple-linear are ready candidates).
- **What blocks a blanket "expose everything":** (a) fixed `1x1`/`2x2` tiles vs. widgets built for large flexible grid cells; (b) a **history long-press gesture** that only Wind Steer accidentally dodges — a history-eligible widget would double-fire Host2's history dialog *and* the bridge's `ui.openConfigPanel` (a named code gap); (c) **no plugin-availability gating** in the static manifest; (d) **per-tile weight** — each tile boots a full Skip app + its own SK websocket.
- **Recommendation:** a curated allowlist plus a small manifest builder helper (a `WIDGET_CONTRIBUTIONS` array expanded into the `widgets[]`/`panels[]` entries — kills the ~3N manifest duplication), seeded with the safe gauges; gate history-eligible and plugin-gated widgets behind two one-time fixes (suppress Host2's history long-press in the single-widget host; consult plugin availability in `buildManifest`); **do not** auto-expose the whole registry. This is a follow-up, not part of this PR.

### Coverage
- `npm run ci` green except the known pre-existing local `app.component.spec.ts` `localStorage` artifact (passes on CI Node 24). +9 tests this pass.
- On-device (halpi.hurma, FSK v3.1.0-beta.2): add / live data / long-press → settings + Remove / edit → Save → live update / persistence / removal all verified working.

Issue #496, opened 2026-08-07
Toolbar auto-hide ignores keyboard focus, so it cannot be operated by keyboard


mairas:
The toolbar hides 4 seconds after it is revealed. The only thing that suppresses that timer is `mouseenter` on the bar (`toolbar.component.ts` `onPointerEnter`/`onPointerLeave`, the sole `suppressHide`/`allowHide` caller pair).

Keyboard focus does not suppress it. The buttons are focusable once `[inert]` lifts on reveal, so Tab and Enter work in principle — but the bar hides out from under you mid-Tab and `inert` snaps back, taking the focus with it. Four seconds is not enough to reach the Menu button, open it and pick Settings.

`ChromeVisibilityService`'s own class doc already describes the intended behavior: "Hiding can be suppressed (ref-counted) while a popup/dialog is open **or focus is inside the toolbar**". The focus half was never wired.

The help table records the resulting gap — "Show the toolbar | Keyboard | —" in `welcome.md`. Today it is partly masked because every page change reveals the bar for free; with #495's opt-out that is no longer guaranteed.

Fix shape: `focusin`/`focusout` on the toolbar host paired to `suppressHide`/`allowHide`, mirroring the existing pointer pair. A keyboard shortcut to reveal the toolbar is worth adding on top, but only once focus actually holds the bar open — on its own it would reveal a toolbar that vanishes before it can be used.

EdKok:
Much better now, thanks.
But why is there still a gray bar showing on each swipe for about a second? And after that still a thinner bar. Does it serve a purpose? For me it is only lost real-estate (on a "7 screen).
Getting into the bar with the swipe up/down would already suffice.

PR #497, opened 2026-08-07
Make the toolbar's automatic reveal optional

### LINKED ISSUE #495: [Feature]: Stop the top bar from showing (optionally)
### Feature Request

**My problem**
It gets annoying pretty fast to have to wait for the top bar time-out and to go away.

**Why**
I'm using an (out of reach) display at the helm station and use an (IR) remote to walk through several screens frequently to get an overview of all systems. It is just not pleasant to have to wait in each screen to get the full picture.

**solution**
Make it a configurable option to disable showing the top bar. Screens for purely looking at don't need the tools in that top bar.

_No response_

mairas:
Closes #495.

## Why

The toolbar reveals itself twice without being asked: once for the boot dwell, and again on every page change, where its page-icon strip doubles as the page-position indicator. On a display that is only watched — the reporter steps through screens at the helm with an IR remote — every screen costs four seconds of chrome that serves no purpose there.

## What

A profile setting, **Settings → Display → Toolbar → "Show the toolbar automatically"** (on by default, so nothing changes for existing users). With it off, neither automatic reveal fires.

Every explicit way in still works: swipe down from the top, scroll up, tap the peek strip, or dwell the cursor in the top edge band. That is deliberate — the toolbar is the only route to Settings, the menu, page management and edit mode, so a hard "never show it" would lock a user out of the setting that got them there.

Two details worth a reviewer's eye:

- The boot dwell is retracted by an effect that depends on the setting **alone**, never on the active page. Hiding on page change instead would snap the bar shut under someone who revealed it by hand and is tapping its page icons to navigate.
- `autoRevealToolbar` is an optional field on `IAppConfig` hydrated to `true` when absent, the same shape as `keepScreenAwake`, so no config-version bump and no migration.

## Notes

A keyboard shortcut to reveal the toolbar was considered and dropped. The toolbar's buttons are focusable once `inert` lifts, but the only thing that suppresses the 4-second idle-hide is `mouseenter` (`toolbar.component.ts`, the sole `suppressHide` caller), so the bar hides mid-Tab and focus is lost. Keyboard operation of the toolbar is already broken today, independently of this change; a shortcut would have been half a feature. Filed separately as #496.

## Verification

`npm run lint`, `npm run snc` and `npm test` all run locally. The suite is 1798 passed / 1 failed; that one failure (`AppComponent — embed boot performs zero server-config writes`) reproduces identically on the base commit (1790 passed / 1 failed) and is a local `localStorage` environment artifact, untouched by this branch. CI on Node 24 is the authoritative check.

New coverage: hydration default, stored `false`, and setter persistence in `settings.service.spec`; four gating cases in `app.component.spec` (no reveal on page change when off, boot dwell retracted, untouched when on, and no re-hide on later navigation); load/save round-trip in `display.component.spec`.


mairas:
## Code review — PR #497

**Scope:** `b92158d8..HEAD` (2 commits, 14 files) · **Mode:** interactive
**Intent:** Add a profile setting `autoRevealToolbar` (default on) that suppresses both automatic toolbar reveals — the boot dwell and the per-page-change reveal — while leaving explicit reveal paths working.
**Reviewers:** correctness, testing, maintainability, project-standards (always-on) · api-contract (`IAppConfig` is a persisted config contract) · adversarial (>50 changed non-test lines; alters global chrome-visibility behavior)

### P1 — High

| 1 | `src/app/app.component.ts:151` | **Lockout on the very device class the issue describes.** With the setting off, every surviving reveal path needs a pointer or touch: swipedown is a `pointerdown/move/up` sequence, scroll-up is a `wheel`-only listener, the peek strip is `tabindex=-1` until `pulsePeek()` fires (only from `pointerdown` and throttled `mousemove`), and hover dwell needs a moving cursor. A keyboard/IR-remote display therefore has none. Since the toolbar is the only route to Settings, enabling it there is a **one-way door** — it cannot be turned back off from that display. | correctness, adversarial | 0.95 | manual → human |

Fix: add a bare reveal hotkey (`t` → `chrome.reveal()`) in `HOTKEY_KEYS` + `handleKeyDown`, or make the peek strip permanently tabbable.

### P2 — Moderate

| 2 | `src/app/app.component.spec.ts:342` | `vi.spyOn(settings, 'autoRevealToolbar').mockReturnValue(x)` replaces the signal with a plain function, so neither effect registers a reactive dependency. With the value false the page-change effect short-circuits before reading `activeDashboard()` too — both effects end up with an empty dependency set. "does not re-hide on later page changes" is green by construction and no implementation could fail it. | testing, maintainability | 0.95 | gated_auto → review-fixer |
| 3 | `display.component.spec.ts:123` | `expect(component['autoRevealToolbar']()).toBe(true)` matches both the model's own initializer and the mock's return, so deleting the `ngOnInit` hydration line leaves it green. Only the save direction is verified. The sibling theme test gets this right by choosing a stored value that differs from the default. | testing | 0.85 | gated_auto → review-fixer |
| 4 | `app-settings.interfaces.ts:36` | **Profile scope vs. per-device copy.** `IAppConfig` is shared by every device on the profile; the per-device store is `connectionConfig` (where `isRemoteControl`/`instanceName` live for exactly this reason). The shipped copy — "turn this off on a display you only watch" — describes a per-device decision the storage scope cannot express. A helm display and a saloon tablet on one profile cannot differ, and whole-blob replace means either can revert the other. | api-contract, adversarial | 0.80 | advisory → human |
| 5 | `src/app/app.component.ts:150` | Reveal policy now has no single owner: the service opens a boot reveal its own file presents as unconditional, and AppComponent retracts it from a second effect. Correctness rests on undocumented construction ordering. The in-repo precedent is `keepScreenAwake`, which is *pushed into* the service that owns the behavior (`uiEvent.setKeepAwake`) rather than counteracted. | maintainability | 0.72 | manual → human |
| 6 | `widget-iframe.component.ts:126` | A page whose single widget is a full-bleed **cross-origin** iframe swallows every pointer reveal path: pointer events never surface in the parent, so `pulsePeek()` never fires and the peek strip stays `pointer-events:none`; the postMessage swipe relay is disabled for cross-origin frames (`contentDocument` access throws). What's left is a 4 px gutter arming an 8 px strip for 1.3 s. | adversarial | 0.65 | advisory → human |

### P3 — Low

| 7 | `settings.service.ts:114` | Degraded boot (kiosk starts before signalk-server is serving) returns before `pushSettings()`, so the flag keeps its in-memory default `true` and the boot dwell the user turned off happens anyway. The setting fails open. Also, per the correctness pass, this is currently the *only* automatic recovery from finding 1. | adversarial | 0.70 | advisory → human |
| 8 | `app.component.spec.ts:375` | Test name promises "a manual reveal survives navigation" but no reveal is performed and `chrome.revealed` stays false throughout — the state the title describes is never entered. | testing | 0.72 | manual → review-fixer |
| 9 | `src/app/app.component.ts:146` | The new comment restates the next line and then argues a road not taken ("retracting there would…"), the "we don't do Y because" shape the repo's own comment rules exclude. | maintainability | 0.65 | safe_auto → review-fixer |
| 10 | `display.component.html:156` | Three-sentence helper against one-sentence siblings, and its gesture list is near-verbatim from the `welcome.md` paragraph this same PR edits — three copies that must now drift together. | maintainability | 0.60 | advisory → human |

### Requirements completeness (plan = issue #495, `plan_source: explicit`)

- **"Make it a configurable option to disable showing the top bar"** — met for touch and mouse; **partially met** for the reporter's own device: the option exists but cannot be reversed from a remote-driven display (finding 1).

| File | Issue | Reviewer | Conf | Action |
|------|-------|----------|------|--------|
| `display.component.ts:114` | If `autoNightMode` is on and the derived-data plugin is missing, `validateAndSaveSettings` abandons the **entire** save — including unrelated preferences — while the toast talks only about night mode. This PR adds the toolbar preference to that swallowed set, and it bites hardest on the recovery leg from finding 1. | adversarial | 0.70 | **file-issue** |
| `display.component.spec.ts:33` | `SettingsServiceMock` has no type relationship to `SettingsService`, so signature drift is unenforced by `snc`. | testing | 0.62 | advisory |

### Coverage

- **project-standards returned zero findings.** It independently confirmed: the `1.2.0 → 1.3.0` minor bump is correct under the AGENTS.md decision procedure (VERSION equalled the released `v1.2.0+3` upstream version, this PR touches package-affecting files, one bump via `./run bumpversion`); no config-version bump or migration is needed (matches the `keepScreenAwake` precedent, `LATEST_APP_CONFIG_VERSION` stays 19); webapp-only invariant intact; TypeScript-only, strictNullChecks and no-`any` clean; commit and branch conventions followed.
- **Untracked and excluded from scope:** `perf-harness/mock-serve.mjs` (pre-existing, unrelated).
- **Residual risks:** no boot flash on a healthy boot — `SettingsService.startup()` reaches `pushSettings()` synchronously and `DashboardService` (which injects it) is constructed before `ChromeVisibilityService`, so the effect sees the hydrated value in the first CD pass. That holds only by construction order, not by contract. · `chrome.hide()` no-ops while `suppressHide` is held, so a pointer resting on the toolbar at boot skips the retraction until `pointerleave` re-arms the idle timer; self-healing. · `ChromeVisibilityService`'s docstring claims hiding is suppressed "while a popup/dialog is open" — no dialog, menu or sheet actually calls `suppressHide` (pre-existing, tracked as #496).
- **Testing gaps:** no test asserts any reveal path *survives* with the setting off — every new case asserts a negative against a mocked service; no runtime on→off transition (the only path a real user takes); no coverage of the degraded-hydration path.

---

**Verdict: not ready.** Finding 1 makes the feature self-locking on the device class in the originating issue, and findings 2–3 mean two of the new tests cannot fail. Fix order: 1 (decide the reveal path) → 2, 3, 8 (make the tests falsifiable) → 4 (decide scope, or fix the copy) → 9, 10 (prose).

mairas:
Review findings addressed. CI green.

**Findings 2, 3, 8 — unfalsifiable tests.** Both reviewers were right. `mockReturnValue` on a signal accessor left the effects with no reactive dependency. The stub is now backed by a real signal, and I mutation-checked it: restoring the old stub fails the runtime-transition test, and keying the retract on the active page fails the page-change test. The display-settings load assertion now stores a value that differs from the model default, as the sibling theme test does.

**Finding 5 — split reveal policy.** Taken. `ChromeVisibilityService` now owns it: `revealAuto()` covers the reveals the app makes on the user's behalf and consults the flag, `setAutoReveal()` carries the setting in and retracts a boot dwell still on screen. `reveal()` stays unconditional, so every user-initiated route is untouched. AppComponent drops to one wiring effect, and the behaviour is pinned in the service's dependency-free spec instead of against a component mock — mutation-checked there too (ungating `revealAuto` fails the no-op test).

**Finding 4 — profile scope.** Keeping profile scope; the copy was the defect. Both the settings helper and the help doc now say the setting applies to every display using that configuration profile. Moving it to per-device `connectionConfig` stays available if the shared value proves wrong in practice.

**Finding 1 — keyboard/remote reveal.** Not a regression on inspection: the toolbar has never been operable by keyboard, which `welcome.md` already documents (`Show the toolbar | Keyboard | —`). Out of scope here; the underlying gap — auto-hide ignores keyboard focus despite the service docstring claiming otherwise — is #496.

**Findings 9, 10 — prose.** Comment trimmed to the why; helper text cut to match its one-sentence siblings, with the gesture list left to the help doc as the single copy.

**Pre-existing** — the auto-night-mode validation swallowing unrelated Display saves is filed as #498. The untyped `SettingsServiceMock` is left as-is; it is a repo-wide spec pattern, not this PR's to change.

Deployed to a test device (halpi.hurma) and serving.

PR #499, opened 2026-08-07
Stop a failed night-mode dependency check from discarding Display settings

### LINKED ISSUE #498: A failed auto-night-mode plugin check discards the entire Display settings save
In `SettingsDisplayComponent.saveAllSettings`, a profile with `autoNightMode` enabled routes the save through `validateAndSaveSettings`, which first checks the Signal K derived-data plugin. When that check fails — plugin not installed, or the API call errors — the else branch resets the toggle and returns, so `applyAndSaveSettings()` never runs.

Every unrelated preference on that page is discarded with it: theme, browser tab title, night-mode brightness, remote-control identity, keep-screen-awake, and (as of #497) the toolbar auto-reveal setting. The user sees a red toast about a night-mode plugin dependency and no indication that the setting they actually came to change was not saved. The toggle still reads as changed in the UI.

Pressing Save a second time works, because the failed attempt has already set `autoNightMode` to false — but nothing tells the user that.

Repro: enable automatic night mode on a server without the derived-data plugin. Open Settings > Display, change any other setting, press Save. The error toast appears; the other setting is not persisted.

Fix shape: either persist the settings that have no plugin dependency even when the auto-night-mode validation fails, or make the failure toast state plainly that no settings were saved. Either removes the case where one feature's unmet precondition silently swallows an unrelated preference.

Found during the review of #497.

mairas:
Fixes #498.

## Why

A profile with Automatic Night Mode switched on routes *every* Display save through a Signal K plugin check. When that check fails — the Derived Data plugin is missing, the API errors, or the user declines the enable prompt — `validateAndSaveSettings` reset the toggle and returned without calling `applyAndSaveSettings` at all.

So a user who came to change their browser tab title, theme, brightness, keep-awake or (as of #497) the toolbar preference lost all of it, silently. The only feedback was an error toast about a night-mode plugin dependency, and the toggle still read as changed. Pressing Save a second time worked, because the failed attempt had already reset `autoNightMode` — but nothing said so.

## What

Refuse only what was actually refused. The rest of the page has no bearing on the plugin and now saves as asked.

Two deliberate details on the refusal path:

- **The stored night-mode value is left as it was**, rather than overwritten with the reset toggle. The same check fails on a transient API error, and persisting the reset there would quietly disable a working setup. The toggle still resets in the UI, so the refusal is visible; the stored intent survives.
- **The success toast is held back.** The refusal is a persistent snackbar and `MatSnackBar` shows one at a time, so announcing the save would dismiss the explanation and then disappear on its own 1s timer. The form going pristine is the save's feedback there.

## Verification

`npm run lint`, `npm run snc`, `npm test` — 1808 passed, plus the one pre-existing local `localStorage` failure that reproduces on main and passes in CI.

Four new tests, written first and confirmed red: two failed before the fix (unrelated settings not persisted, via both the plugin-missing and the declined-prompt paths). The other two guard against the naive fix rather than the bug — mutating `applyAndSaveSettings(true)` to `applyAndSaveSettings()` makes both fail, which is what pins the stored-value and toast decisions above.

No `VERSION` bump: the cycle is already open at 1.3.0 from #497, and a patch-level fix does not raise it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * The Connectivity settings show your Signal K session identity in same-origin mode (including a read-only-session indicator) instead of a credential form.
 * Security: the Signal K login password is no longer stored in the browser; it is used only in memory to obtain a session token.
 ## Fixes
+* Saving Display settings no longer discards your changes when Automatic Night Mode cannot be enabled. Previously, if that feature was switched on while the Signal K Derived Data plugin was missing or unreachable, the failed dependency check abandoned the whole save — theme, browser tab title, brightness and the rest were silently lost, and the error message mentioned only night mode. Everything unrelated to the plugin now saves as asked; only automatic night mode is held back, and its stored value is left as it was. Fixes #498
 * OIDC/SSO users could not sign in to KIP because it required a server-local password they do not have; same-origin SSO mode resolves this.
 ## Behavior changes
 * Cross-origin token sign-in: the Signal K server provides no token-refresh endpoint, so the stored password is no longer re-sent to renew a session. The session token still persists across browser reloads until it expires; when it expires you are prompted to sign in again.
--- a/src/app/core/components/settings/display/display.component.spec.ts
+++ b/src/app/core/components/settings/display/display.component.spec.ts
@@ -1,6 +1,6 @@
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
-import { EMPTY, of } from 'rxjs';
+import { EMPTY, Subject, of } from 'rxjs';
 import { signal } from '@angular/core';
 import { BreakpointObserver } from '@angular/cdk/layout';
 import { ToastService } from '../../../services/toast.service';
@@ -24,10 +24,22 @@ class AppServiceMock {
 }
 
 class ToastServiceMock {
+    // Prompts resolve as "dismissed without action" (the user declined) unless a test pushes to
+    // `action$` first. Without a live action stream the accept-the-prompt path is unreachable, and
+    // with it the only route that ever stores autoNightMode = true.
+    public readonly action$ = new Subject<void>();
     public show = vi.fn().mockReturnValue({
-        onAction: () => EMPTY,
+        onAction: () => this.action$,
         afterDismissed: () => of({ dismissedByAction: false })
     });
+
+    /** Answer the next prompt with its action button instead of a dismissal. */
+    public acceptNextPrompt(): void {
+        this.show.mockReturnValueOnce({
+            onAction: () => of(undefined),
+            afterDismissed: () => EMPTY
+        });
+    }
 }
 
 class SettingsServiceMock {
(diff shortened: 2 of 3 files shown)


mairas:
## Code review — PR #499 (light)

**Scope:** `d2e6b1e3..HEAD` · **Mode:** interactive
**Intent:** On a refused Automatic Night Mode request, still persist everything the plugin check has no bearing on.
**Reviewers:** correctness, testing, maintainability. Conditionals dropped (no auth, routes, or contracts touched; 19 changed non-test lines) and project-standards dropped (the only standards question, the version bump, is machine-checked and green).

### P2 — Moderate

| 1 | `display.component.ts:140` | **The fix deferred the loss by exactly one save.** The toggle was forced to `false` while the same call deliberately declined to store it, so the form, the stored config and `AppService`'s live night-mode effect held three different states. The next save then takes the `!autoNightMode()` fast path in `saveAllSettings` — no check, no prompt — and writes `setAutoNightMode(false)`. The working setup the docstring claimed to protect is disabled one save later, silently. Testing proved it with a probe: `{"toggleAfterRefusal":false,"setAutoNightModeCalls":[[false]]}`. | correctness, testing, maintainability | 0.95 | gated_auto → human |
| 2 | `display.component.spec.ts:28` | **The accept-the-prompt path is unreachable by the entire spec.** `ToastServiceMock.onAction()` returns `EMPTY`, which completes without emitting, so `enableAndConfigureAutoNight` runs in no test — including the only route that ever stores night mode *on*, and the plugin-save-failure branch that lands in the new partial-save path. Proven by mutation: a `throw` as its first statement left all 14 tests green. | testing | 0.90 | manual → review-fixer |
| 3 | `display.component.ts:202` | A pending prompt is dismissed by *any* other snackbar (`MatSnackBar` shows one at a time, and `_animateSnackBar` calls plain `dismiss()`, so `dismissedByAction` stays false). That now resolves the refusal path, which **writes**. The subscriptions have no teardown, so a destroyed `/settings` instance can resolve later and overwrite a newer save; and `_pluginCheckSeq` was only bumped on the night-mode branch, so a subsequent non-night-mode save did not invalidate a check in flight. | correctness | 0.70 | manual → human |

### P3 — Low

| 4 | `display.component.ts:126` | Declining the prompt closes its own snackbar, so nothing is on screen — yet the success toast was suppressed there too. The user cancels, sees the toggle flip, and gets no confirmation, while a full settings write did happen. The suppression rationale holds only for the two error branches. | correctness | 0.85 | gated_auto → review-fixer |
| 5 | `display.component.spec.ts:230` | The two negative-only tests pass in a world where `applyAndSaveSettings` is never called — which is the #498 bug itself. Proven by mutation: pushing the save past the assertions leaves them green. Their value depended on sibling tests they did not reference. | testing | 0.90 | gated_auto → review-fixer |
| 6 | `display.component.ts:95` | The `applyAndSaveSettings` doc comment restated the guard below it and narrated the fix ("The rest of the page is unrelated to that plugin, so it still saves (#498)") — the change-narration shape the repo's comment rules exclude. Two clauses earned their place; six lines did not. | maintainability | 0.66 | gated_auto → review-fixer |
| 7 | `display.component.spec.ts:252` | The declined-prompt test arranges nothing, relying on two ambient mock defaults; hoisting `withMissingPlugin()` or changing either default silently converts it into a duplicate of the first test that keeps passing while covering nothing. | maintainability | 0.63 | gated_auto → review-fixer |

### Resolution

All seven addressed in `2723fd07`.

**1 —** Taken; the rationale in the original commit message was wrong and the reviewers were right. The refusal path now writes the *stored* value back into the toggle rather than forcing it off. The form agrees with the config, a following save re-runs the check instead of writing an off state, and the special-cased write disappears entirely — the ordinary `setAutoNightMode` call carries the same value. Pinned by a new test that starts from a stored `true` and a transient (non-`not-found`) failure, asserting `setAutoNightMode` is never called with `false`; forcing the toggle off again fails it.

**2 —** `ToastServiceMock` gained a live action stream and an `acceptNextPrompt()` helper. Two new tests cover the accept path: plugin enabled and configured → night mode stored on; plugin save fails → the rest still saves, night mode held back, no confirmation. Re-ran the reviewer's own mutation (`throw` as the first statement of `enableAndConfigureAutoNight`) — it now fails both.

**3 —** Both prompt subscriptions are tied to the component with `takeUntilDestroyed`, so a destroyed instance can no longer resolve and write. `_pluginCheckSeq` is now bumped by *every* save, so a check in flight is superseded whichever branch the next save takes. The remaining case — an ambient toast dismissing the prompt on a live component — writes the state the user typed and pressed Save for, so it is early rather than wrong; left as is.

**4 —** The check now returns an outcome (`granted` / `explained` / `declined`) rather than a boolean, and the save is confirmed except where an explanation is already on screen. A declined prompt gets its confirmation.

**5, 7 —** Both negative tests gained a positive assertion in the same test that the save actually ran, so neither passes when the save is absent. **6 —** Comment cut to the two non-obvious rationales.

### Coverage

- **Not addressed:** the `await flushPromises()` idiom pins a microtask-tick count rather than a completion condition (testing, residual risk) — pre-existing throughout this spec, not worth diverging here.
- **Verified, no action:** the brightness side effect on the new path matches the ordinary save; `enableAndConfigureAutoNight`'s two resolve routes cannot both take effect.
- **Method note:** one mutation in this round produced a false negative — a regex whose `[^)]*` stopped inside a parameter's function type, so the mutation never applied and the suite passed for the wrong reason. Re-run with an exact string match, it failed the two tests as expected. A green suite under a mutation is only evidence once you have confirmed the mutation landed.

---

**Verdict: ready with fixes — applied.** `npm run lint`, `npm run snc`, `npm test` green (1812 passed; the one failure is the pre-existing local `localStorage` artifact that reproduces on main and passes in CI).

Issue #502, opened 2026-08-09
app.component spec: 'the same boot NOT under embed DOES fire the self-heal writes' fails on localStorage being undefined


mairas:
`src/app/app.component.spec.ts` > `AppComponent — embed boot performs zero server-config writes (#216 E6)` > *the same boot NOT under embed DOES fire the self-heal writes (the guard is embed-specific)* fails locally on a clean `main` tree:

```
TypeError: Cannot read properties of undefined (reading 'setItem')
 ❯ _SettingsService.saveConnectionConfigToLocalStorage src/app/core/services/settings.service.ts:581:18
 ❯ _SettingsService.setRemoteContextDemand src/app/core/services/settings.service.ts:423:10
 ❯ src/app/core/services/dashboard.service.ts:129:26
```

The run also emits `ExperimentalWarning: localStorage is not available because --localstorage-file was not provided`, so the worker has no `localStorage` binding at all and `saveConnectionConfigToLocalStorage` dereferences `undefined`.

Reproduces both in the full `npm run ci` suite and in isolation via `npx ng test --include='src/app/app.component.spec.ts'`. Found while working on #501; unrelated to that change, which was verified by reproducing on a stashed (clean) tree.

Either the spec needs a `localStorage` shim in `src/test.ts`, or `saveConnectionConfigToLocalStorage` needs to tolerate its absence — the latter matters beyond tests, since the same dereference would throw in any browser context where storage is unavailable (private mode with storage blocked).

mairas:
CI is green on this same suite (https://github.com/halos-org/skip/actions/runs/31303461301/job/93219893212, the `checks / tests` job on PR #503), so this reproduces locally only. The differentiator is the local Node build emitting `ExperimentalWarning: localStorage is not available because --localstorage-file was not provided` — CI's Node 24 apparently provides a `localStorage` binding where the local one does not. That narrows it to a missing environment shim rather than a code regression, but the underlying dereference in `saveConnectionConfigToLocalStorage` is still unguarded.

PR #494, opened 2026-08-01
Make @halos-org/skip webapp-only; extract Freeboard-SK plugin (Phase B)

mairas:
Makes `@halos-org/skip` a **webapp-only** package. The Freeboard-SK plotter-extension plugin was extracted to the standalone `@halos-org/skip-freeboard-panel` package (now live on npm at `1.0.0`); this PR removes it from here and wires Skip to depend on it. Phase B of halos-org/skip#484.

## What

- **#487 — webapp-only package.** Remove `main`, the `signalk-node-server-plugin` keyword, `signalk-plugin-enabled-by-default`, and `plugin/**` from `files`; delete `plugin/`. Add `@halos-org/skip-freeboard-panel: ^1.0.0` as a runtime dependency (lockfile regenerated). Drop `test:plugin` from scripts + the `ci` gate; remove the plugin lint globs.
- **#488 — CI enforcement.** Invert the `run-tests` tarball assertion: it required `plugin/index.js`; now it requires `public/` and **fails if any `plugin/` ships** — machine-enforcing the rule.
- **#489 — conditional deploy restart.** `deploy-halos` ships `public/` + `package.json` only and restarts Signal K **only on a first install** (an already-mounted webapp is served fresh from disk).
- **#490 — docs + rules.** `CLAUDE.md` gains a **webapp-only** rule and a **TypeScript-only** rule; the plotter-extension / serving-path / npm-publish notes now point at the separate package and describe the cross-package URL contract.

## Why the coupling is an npm dependency

Installing/updating/removing Skip from the appstore (a registry `npm install`) carries the panel automatically. On the `file:`-symlink deploy path (`deploy-halos`, and possibly the image) the dependency does **not** resolve, so the panel is shipped explicitly there — see the tracking issue for the full rationale and the deferred device-install question (#491).

## How the serving path became a cross-package contract

`SKIP_URL` (`/@halos-org/skip/`), the widget-type ids, and the `#/widget/:type` embed routes now live in the panel package's manifest. Renaming any of them here needs a coordinated panel release **and** bumping its caret floor. Documented in `CLAUDE.md`; the webapp-side embed routes / `WidgetHostBridge` are unchanged.

## Verification

- `npm run lint` ✅ · `npm run snc` ✅ · dependency resolves (panel `1.0.0` installed) · `npm pack --dry-run` ships **no `plugin/`**, keeps `public/`.
[...]

PR #503, opened 2026-08-09
Allow saving a widget config while one of its paths is unpublished

### LINKED ISSUE #501: Widget config cannot be saved while a configured path is unpublished
## Problem

A widget whose configured path is not currently published by the server cannot be saved at all — the Save button in the widget settings dialog is disabled, so *no* setting in that widget can be changed.

The path being absent does not mean it is wrong. Signal K publishes a path only once some source has actually sent it, so switching off the instrument behind it (an autopilot control unit off ⇒ no `steering.rudderAngle`) makes a perfectly correct path read as unknown. The widget then locks its own configuration until the gear is turned back on.

## Cause

`pathRequiredOrValidMatch` in `src/app/widget-config/path-control-config/path-control-config.component.ts` returns a `requireMatch` error whenever the value is absent from `DataService.getPathsAndMetaByType()`, the delta-derived live path list. `root-modal-widget-config.component.ts` drives `saveDisabled` straight off `formMaster.invalid`, so one unpublished path disables Save for the whole widget. `dataset-chart-options.component.ts` carries a byte-identical copy of the validator with the same effect on the datachart widget.

The only escape hatch today is the global **Allow saving widgets with invalid paths (testing only)** toggle in Display settings, which is labelled dev-only and is not persisted across restarts.

## Second defect: saving would lose configuration

On open, an invalid path takes the `disablePathFields()` branch, which does `source.reset('')` and `convertUnitTo.reset('')`. `submitConfig()` uses `getRawValue()`, so disabled controls still persist — the moment saving becomes possible, the widget's pinned data source and unit conversion get written back as empty strings. `convertUnitTo` has no UI of its own since the per-path Display Format picker was removed, so it is only ever derived from the picked path; clearing it is unrecoverable by the user.

`setupSourceFor()` already solves exactly this for fixed/choice paths ("instrument off / boat at dock", 0852e9ff). That reasoning simply never reached the free-path branch.

## Approach

1. An unpublished path stops being a validation error. Only an empty path in a required slot still blocks Save. The form surfaces the unpublished path as a warning hint under the field instead of a `mat-error`, so a typo is still visible without being fatal.
2. The two duplicated validators collapse into one shared helper under `core/utils`, alongside a predicate the components use to drive the warning.
3. For an unpublished path the stored `source` and `convertUnitTo` are preserved rather than reset — matching what `setupSourceFor` already does for fixed and choice paths. When the user actively changes the path to a different unpublished one, both are re-derived (source to Any, no unit conversion), since nothing is known about the new path.
4. The **Allow saving widgets with invalid paths (testing only)** toggle and its `SettingsService` accessors are removed. It becomes dead weight once an unknown path no longer blocks Save, and leaving a dev toggle that no longer changes anything is worse than removing it.

## Verification

Specs on `PathControlConfigComponent` and `DatasetChartOptionsComponent` covering: an unpublished stored path leaves the control valid; the stored source and unit conversion survive a dialog open on an unpublished path; an empty required path still invalidates; changing to a different unpublished path re-derives source and unit. Full gate via `npm run ci`.

mairas:
## Why

A widget whose configured path is not currently published by the server could not be saved at all — Save was disabled for the whole dialog, so *no* setting in that widget could be changed until the path came back.

The path being absent does not mean it is wrong. Signal K publishes a path only once some source has actually sent it, so switching off the gear behind it (an autopilot control unit off ⇒ no `steering.rudderAngle`) makes a perfectly correct path read as unknown. The widget then locked its own configuration.

A second defect sat behind the first: opening the dialog on such a path ran `disablePathFields()`, which resets `source` and `convertUnitTo`. `submitConfig()` saves with `getRawValue()`, so those empty strings would have been persisted the moment saving became possible. `convertUnitTo` has no UI of its own since the per-path Display Format picker was removed, so clearing it is not something the user can undo.

## How

An unrecognized path is no longer a validation error — it renders as a warning hint under the field, in tertiary rather than error colour, so a typo is still visible without being fatal. Only an empty path in a required slot still blocks Save.

For an unpublished path the stored `source` and `convertUnitTo` are preserved rather than reset, which is what `setupSourceFor()` already does for fixed and choice paths (0852e9ff, "instrument off / boat at dock"); that reasoning simply never reached the free-path branch. When the user actively changes the path to a different unpublished one, both are re-derived — Any source, no conversion — since nothing is known about the new path.

The datachart path field carried a byte-identical copy of the validator and gets the same treatment; both now share one helper in `core/utils`.

The **Allow saving widgets with invalid paths (testing only)** toggle in Display settings is removed along with its `SettingsService` accessors. It existed only to unblock Save on an unrecognized path, which no longer blocks it.

## Verification

`npm run ci` — lint and `snc` clean, 1830 of 1831 tests pass. The one failure, `app.component.spec.ts` > *the same boot NOT under embed DOES fire the self-heal writes*, reproduces on a stashed clean tree and is unrelated; filed as #502.

New coverage in `path-validators.util.spec.ts`, `path-control-config.component.spec.ts` and `dataset-chart-options.component.spec.ts`: an unpublished path leaves the control valid; the stored source and unit conversion survive a dialog open on one; an empty required path still invalidates; an optional blank path still clears both; changing to a different unpublished path re-derives them. The two central tests were confirmed to fail under the pre-fix validator behaviour.

Closes #501


mairas:
## Code review — PR #503

**Scope:** `e867e2d0..HEAD`, 19 files, +278/−121. Untracked `perf-harness/mock-serve.mjs` excluded (not staged, unrelated).

**Intent:** Make an unrecognized Signal K path a warning rather than a save-blocking validation error, preserve the stored source and unit conversion for such a path, apply the same to the datachart path field, and remove the redundant "Allow saving widgets with invalid paths (testing only)" toggle.

**Reviewers:** correctness, testing, maintainability, project-standards (always-on) + adversarial (157 net non-test lines, changes save-gating behaviour). Security, api-contract and reliability were not selected — no auth or permission surface, the only schema diff is a version string, and no error-handling paths are touched.

### P2 — High

| # | File | Issue | Reviewer | Conf. | Route |
|---|------|-------|----------|-------|-------|
| 1 | `path-control-config.component.ts:248` | **Touching then reverting the path field wipes source and unit.** Any keystroke marks the control dirty; the debounced subscriber then calls `enableFormFields(true)` whether or not the value actually changed. For an unpublished path that runs `source.setValue('default')` and `convertUnitTo.setValue('')`, and `setupSourceFor` then reads the just-overwritten source, so `gps.7` disappears from the value *and* the dropdown. Type one character, backspace, Save — and the exact loss this PR set out to prevent is persisted, which it could not be before because Save was disabled. | adversarial | 0.82 | fix |
| 2 | `dataset-chart-options.component.ts:61` | **Datachart Source select renders empty for an unpublished path.** `setPathSources` only runs when `getPathObject` resolves, so `pathSources` stays `[]`, while `setInitFormState` enables the control because it holds a stored value. The result is an enabled, `required` select with zero options and no matching option for its own value — in exactly the scenario this PR unblocks. `path-control-config` solves this at `setupSourceFor`; the datachart got the validator relaxation without the other half. | adversarial, correctness | 0.95 | fix |
| 3 | `dataset-chart-options.component.ts:45` | **Datachart keeps the previous path's source when the new path is typed.** An unpublished path produces no autocomplete option, so typing is the only way to enter one — and `changePath()` is wired solely to `(optionSelected)`, so it never fires. The stale concrete `$source` is saved against the new path, and `widget-data-chart` subscribes to a (path, source) pair that can never deliver, silently, even after the path starts publishing. | adversarial, correctness | 0.90 | fix |
| 4 | `path-control-config.component.ts:136` | **The hint claims "not published" for paths that are published but filtered out of the slot.** `getPaths()` filters by `pathType`, `supportsPut`, `zonesOnlyPaths` and `filterSelfPaths`, so a live path of the wrong type, without PUT support, or on another vessel is labelled "Not published by the server right now" and the user is advised to keep it if the instrument is off — advice that guarantees a broken widget. `enableFormFields` meanwhile branches on the *unfiltered* `getPathObject`, so the hint and the code disagree about the same value. Omitting the `self.` prefix lands here too, and now saves cleanly. | adversarial, maintainability, correctness | 0.88 | fix |
| 5 | `path-control-config.component.ts:99` | **The hint's recompute on path edit is untested.** All three `refreshPathUnpublished()` call sites outside `ngOnInit` could be deleted and every new test would still pass — the `startWith('')` makes the init path self-covering. The interactive loop the feature exists for (type an unknown path → warning appears; correct it → warning clears) has no coverage. | testing | 0.85 | fix |
| 6 | `CHANGELOG.md:10` | **The user-visible fix is missing from `# Unreleased`.** Every recent user-facing PR (#499, #497, #494, #483) updates the changelog; the preceding fix added "Fixes #498" under `## Fixes`. This cycle would otherwise ship both a fix for a reported issue and a removed settings toggle with no user-facing note. | project-standards | 0.75 | fix |

### P3 — Low

| # | File | Issue | Reviewer | Conf. | Route |
|---|------|-------|----------|-------|-------|
| 7 | `path-control-config.component.html:55` | The warning markup is never asserted to render — both suites read the signal, not the DOM, so losing the `@if` block or having Material suppress the hint under an error state passes CI. The new `@else if` also silently suppresses "Leave blank to disable this path." for an optional slot with an unpublished path. | testing (+3 as gaps) | 0.78 | fix |
| 8 | `path-control-config.component.ts:135` | The hint never clears when a path starts publishing mid-dialog. `pathUnpublished` refreshes only on control changes, but the published set is delta-driven. A dialog opened before the first tree lands flags every path. Not a regression — the old validator had the same staleness — but the hint is now the only signal. | adversarial | 0.72 | issue |
| 9 | `dataset-chart-options.component.scss:8` | The PR de-duplicates one validator and adds two new duplicates: a byte-identical `.pathUnpublishedHint` rule (comment included) and the identical hint sentence, in both components. | maintainability | 0.70 | fix |
| 10 | `dataset-chart-options.component.ts:89` | `getPaths()` repeats verbatim the call that seeds `numericPaths`, giving the component two sources for one list with different freshness — `numericPaths` is an init snapshot, `getPaths()` reads live. A path arriving mid-dialog clears the warning but never appears in the autocomplete, and toggling "Restrict to own vessel" makes the two disagree. | maintainability (+2 as risks) | 0.68 | fix |
| 11 | `path-control-config.component.ts:242` | The unpublished branch never re-enables `convertUnitTo`, so the three branches leave it in three different states. Blank the path then type an unpublished one and it keeps the disabled state `disablePathFields()` left, while `source` is re-enabled. Invisible today (nothing binds it, `getRawValue()` saves disabled controls); a data-loss path the moment a unit picker returns. | testing, maintainability (+2 as risks) | 0.76 | fix |
| 12 | `path-control-config.component.ts:86` | After the validator change `path.invalid` means exactly "empty and required", which `enableFormFields`'s third branch already handles — so both `else { disablePathFields(); }` sites now make the same decision twice, and a reader must prove three different "is there a path" expressions agree. | maintainability | 0.68 | issue |
| 13 | `path-validators.util.ts:14` | `pathRequiredValidator()` takes no parameters and always returns the same closure — a leftover factory shape from the version that closed over `getPaths`. It reads as configurable (`Validators.minLength(n)`) but behaves like `Validators.required`. | maintainability | 0.62 | fix |

| # | File | Issue | Conf. | Action |
|---|------|-------|-------|--------|
| 14 | `package-lock.json:3` | Root version is `1.2.0` against `package.json` `1.3.1`. `.bumpversion.cfg` doesn't list the lockfile and `check-versions` compares `VERSION` to `package.json` only, so the drift widens silently and surfaces as an unrelated lockfile diff in someone else's PR. | 0.68 | file-issue — `run bumpversion` tooling gap, not this PR's work |
| 15 | `CHANGELOG.md:5` | Two `# Unreleased` entries describing Skip's own SSO work name the product "KIP" (lines 5 and 12), so the next release notes credit a different project. Line 15 refers to KIP on purpose and stays. | 0.62 | fix-here — trivial, and this PR is editing the file anyway |

### Coverage

- Untracked and excluded from scope: `perf-harness/mock-serve.mjs`.
- Residual risks: a whitespace-only path now passes and `WidgetMetadataDirective.observe()` allocates a session-lifetime `BehaviorSubject` keyed on the junk string; `refreshPathUnpublished()` adds a second `getPathsAndMetaByType().sort()` pass per debounced keystroke on top of `filterPaths`, unmeasured on the Pi-class target; removing the dev toggle leaves no runtime override if a future validator blocks Save again.
- Testing gaps beyond finding 5: no test reaches `saveDisabled` in `root-modal-widget-config` — the issue's actual symptom — every new test stops one level below at control validity; the "re-derives source and unit conversion" test calls the private `enableFormFields(true)` directly, bypassing the dirty + 350 ms debounce route that finding 1 lives in; two of the new tests ("accepts an empty path in an optional slot", "clears source and unit conversion when an optional path is left blank") pass against the pre-fix validator too, so they guard unchanged behaviour rather than covering the fix.

---

**Verdict: not ready.** Finding 1 is a regression that undoes the PR's own stated purpose and is reachable by touching the path field at all. Findings 2 and 3 are defects the PR makes reachable for the first time by unblocking Save on the datachart without giving it the source handling its sibling got. Finding 4 makes the only remaining feedback signal actively misleading in the most common typo case.

Fix order: 1 → 2, 3 → 4 → 11, 13 → 5, 7 → 6, 15. Findings 8 and 12 to follow-up issues, 14 to a tooling issue.

mairas:
## Review findings addressed

All P2s fixed, plus the P3s that were cheap or shared a code path. Three went to follow-up issues. Re-pushed as a clean five-commit chain rather than fixups, so history carries no superseded version of the code the review flagged.

**Fixed**

| # | What changed |
|---|---|
| 1 | `enableFormFields` now re-derives only when the path value actually moved, tracked in `_derivedForPath`. Editing and undoing an edit keeps the source and unit. This also fixes the same loss on a *published* path, which predates this branch. |
| 2, 3 | `setPathSourcesFor` in the datachart mirrors `setupSourceFor`: an unsent path gets `['default', storedSource]` and an enabled select, and a typed path change resets the stale source instead of pinning it to a path that can never fill it. |
| 4 | Replaced `isPathUnpublished` with `pathSlotWarning`, which resolves the path against the *unfiltered* store and distinguishes six causes: not sent at all, sent no value yet, wrong value type (naming both types), read-only in a PUT slot, another vessel, and no alarm zones. The reviewer's claim that a missing `self.` prefix lands in the filtered bucket was wrong — nothing is stored under a bare key, so that typo is genuinely absent and gets the spelling hint. |
| 5, 7 | New specs drive the real dirty + 350 ms debounce route with fake timers, assert the rendered `.pathWarningHint` element and the absence of a `mat-error`, and cover a warning clearing when the edited path becomes one the slot offers. |
| 6, 15 | CHANGELOG entry under `## Fixes` and the toggle removal under `## Behavior changes`; the two Unreleased entries describing Skip's own SSO work no longer say "KIP". |
| 9 | The hint colour moved to `src/styles-common/pathWarningHint.scss`; the message text now has a single home in the shared helper. |
| 10 | The datachart derives `numericPaths` and the warning from one `getPaths()` refresh, so the autocomplete and the hint cannot disagree. |
| 11 | `convertUnitTo` is enabled in the unpublished branch, so its state depends only on whether a path is configured. |
| 13 | `pathRequiredValidator` is a plain `ValidatorFn` rather than a zero-argument factory. |

**Deferred**

| # | Issue |
|---|---|
| 8 | #504 — the warning goes stale while the dialog is open. Pre-dates this branch (validators re-run on control changes too); needs a live-derived signal. |
| 12 | #505 — three expressions decide "is there a path". No user-visible defect; a control-flow cleanup that deserves its own diff. |
| 14 | #506 — `package-lock.json` root version drift. A `run bumpversion` tooling gap, not this PR's work. |

**Verification.** Each of the three regression fixes was mutation-tested: reverting `path !== this._derivedForPath` to `true`, dropping the datachart typed-path re-derive, and restoring the empty `pathSources` each fail exactly the test written for it. `npm run ci` is clean except `app.component.spec.ts` (#502), which fails identically on a stashed tree and passes in CI.

One correction to the review itself: finding 4 counted five conditions, but the metadata-only case is real and separate — `DataService` builds a path entry from meta alone with `type` derived from units, so a path with no units and no value yet has `type === undefined` and fails a number slot despite being published. It gets its own message.

Issue #507, opened 2026-08-09
Filled-button label ink is inverted against the widget accent colours (3 of 4 theme/colour pairs fail WCAG AA)


mairas:
Four widgets set their filled-button label colour from the theme's surface-contrast ink:

```scss
@include mat.button-overrides((
  filled-container-color: var(--skip-blue-color),   // or --skip-port-color
  filled-label-text-color: var(--skip-contrast-color)
));
```

- `src/app/widgets/widget-racer-timer/widget-racer-timer.component.scss`
- `src/app/widgets/widget-race-timer/widget-race-timer.component.scss`
- `src/app/widgets/widget-racer-line/widget-racer-line.component.scss`
- `src/app/widgets/widget-autopilot/widget-autopilot.component.scss`

`--skip-contrast-color` is the ink that contrasts with the *widget surface* (`#000000` light, `#FFFFFF` dark). The button is not the widget surface — it is a fixed accent colour that does not track the theme in the same direction. So the pairing is wrong, and for the blue button it is exactly inverted.

Computed WCAG contrast ratios:

| theme | button | colour | current ink | ratio | correct ink | ratio |
|---|---|---|---|---|---|---|
| light | blue | `#005bb5` | black | **3.16** | white | 6.64 |
| light | port | `#b70000` | black | **3.01** | white | 6.97 |
| dark | blue | `#3298ff` | white | **2.97** | black | 7.07 |
| dark | port | `#8F0000` | white | 9.69 | white | 9.69 |

Three of four fail AA for normal text (4.5:1); dark-theme blue at 2.97 fails even the 3:1 large-text threshold. Light theme is wrong for both accent colours.

The reason it inverts: `blue` is dark in the light theme (`#005bb5`) and light in the dark theme (`#3298ff`), so the ink it needs flips the opposite way from `--skip-contrast-color`.

Fix direction: add per-accent on-colour tokens (`--skip-on-blue-color`, `--skip-on-port-color`) to `src/themes/_m3light.scss`, `_m3dark.scss` and `_m3night.scss`, wire them in `src/styles.scss` beside the existing `--skip-blue-color` / `--skip-port-color`, and use those for `filled-label-text-color` in all four widgets. Check the night theme's values too — it is not measured above.

Found while fixing the racer timer's control legibility; kept out of that PR because it spans four widgets and the theme files.

PR #508, opened 2026-08-09
Make the Racer Timer's controls legible and named

mairas:
## Why

Three problems with the Racer Timer's control row, all confirmed by rendering the component's compiled SCSS against the app's real stylesheet and measuring.

**Labels were scaled down by empty space, not by the button.** Every label was SVG `<text>` in a fixed viewBox 2–5x wider than the word, and `preserveAspectRatio="xMidYMid meet"` fits the *whole viewBox* — so the glyphs shrank to make room for emptiness beside them. On a 320x240 tile they rendered at 17px inside a 44px button; on a 160x120 tile `Start` measured 1px. The viewBox width also differed per mode (`0 0 80 14`, `0 0 50 14`, `0 0 100 14`), so the same `-1m` label changed size as you cycled modes.

**The absolute start-time field was black on black in the light theme.** `.set-start-at` set `background-color: var(--skip-contrast-color)`, which `_m3light.scss` defines as `#000000`. Computed style came out `background rgb(0,0,0)` / `color rgb(0,0,0)`.

**No control had an accessible name.** Each label sat inside `<mat-icon>`, which Angular Material marks `aria-hidden="true"` unless given an explicit value (`icon.mjs:99`). Every button was unlabelled to a screen reader.

Review then found two more defects in that same field, both predating this branch and both defeating the fix: it is fed `toLocaleTimeString`, which `input[type=time]` rejects outside 24-hour colon locales — **blank in fi-FI and en-US** — and clearing it built an Invalid Date whose `toISOString()` threw out of the change handler.

## How

Each control is a `container-type: size` container, and its label is ordinary text at `min(<h>cqh, <w>cqw)`. Bounding by both axes is what the SVG got right and what a first attempt at this dropped: a tile can be far taller than it is wide, and centred text clips symmetrically, so the leading `+`/`-` is the first thing lost. An `@supports not (container-type: size)` fallback covers the third-party chartplotter browsers Skip runs in through the Freeboard-SK panel, where `cq` units would otherwise resolve against the viewport.

The row's bottom gap becomes proportional (`3%`) rather than a fixed 14px that took most of the row on a small tile. Nothing sets `overflow` on the button, so Material's absolutely positioned 48px touch-target span is not clipped; `text-overflow: ellipsis` on the label makes any residual overflow visible instead of cutting both ends.

The time field sits on the widget's own surface with the theme's ink — the pairing every theme is built around — rather than an M3 surface pair the night theme does not override. `matInput` came off it: there is no `<mat-form-field>` here, so the directive contributed nothing.

## Result

Measured across 3 themes x 8 tile sizes (including portrait: 92x163, 163x92, 213x360, 240x320) x 5 modes: **no label clips anywhere**, and the time field's value fits at every size.

Field contrast against the real `--skip-widget-card-background-color`: light **21:1** (was 1:1, black on black), dark **21:1**, night **1.36:1** — the night theme is deliberately dim red on black and every label in it sits at 1.36:1, so the field now matches the theme rather than being a special case.

## Verification

`npm run ci` — lint and `snc` clean; 171 of 172 test files pass. The failure is `app.component.spec.ts` (#502), which reproduces on a clean tree and passes in CI.

Twelve tests on the widget: pronounceable accessible names across all five modes, no `mat-icon`/`svg text` in any mode, and nine table-driven cases clicking each label and asserting the request it sends. Both were mutation-checked — wrapping `Start` in an aria-hidden span fails the name test, and flipping `adjustStartTime(-1)` to `(+1)` fails `mode 3: -1s`.

`src/app/core/utils/contrast-token-usage.spec.ts` scans `src/**/*.scss` and fails on `--skip-contrast-color` used as a `background-color`, naming file and line. Verified by re-introducing the shipped regression as a scratch file.

## Out of scope

- #507 — filled-button label ink is inverted against the accent colours in 3 of 4 theme/colour pairs (down to 2.97:1). Spans four widgets and the theme files.
- #509 — `widget-race-timer` and `widget-racer-line` still use the viewBox-scaled labels and have no accessible names.
- #510 — four button classes with two rule bodies and a `-small` suffix that changes nothing.


mairas:
## Code review — PR #508

**Scope:** `origin/main...HEAD`, 5 files. **Reviewers:** correctness, testing, maintainability, adversarial. Security, api-contract and reliability were not selected — template and SCSS only, no auth, contract or error-handling surface.

Two reviewers rebuilt the widget in headless Chrome against the repo's compiled `styles.scss`, Angular Material 21.2.5's real button CSS and DOM, and widget-host2's card box model. That falsified two claims in the PR body and found four regressions this change introduces.

### P2 — High

| 1 | `…component.scss:51` | **`overflow: hidden` clips Material's 48px touch target, so the tap area shrank.** Material renders `<span class="mat-mdc-button-touch-target">` as an absolutely positioned 48px-tall child and sets `overflow: visible` on `.mdc-button` precisely so it can extend past a short button. Clipping it drops the tappable height to the visual button height — measured 48px → 13.6px on the widget's own default 4×4 tile at 1024×600. Verified both ways with `elementFromPoint`: pre-change a probe 6px above the button hits `mat-mdc-button-touch-target`, post-change it hits the canvas. **The PR table and the CHANGELOG say the opposite.** | adversarial, correctness | 0.90 |
| 2 | `…component.scss:41` | **Height-only sizing clips labels on any tile taller than it is wide.** `45cqh` scales type with row height while button width comes from tile width, and `justify-content: center` clips symmetrically — so the leading `+`/`−` goes first. Rendered at 92×163 (the default 4×4 tile on a portrait tablet): mode 1 reads `tar 1m 1m`, mode 3 reads `1n 1s 1s 1n`. On a start line you cannot tell add-a-minute from subtract-a-minute. The old SVG fit both axes, so it shrank instead of truncating — this change traded one failure mode for a worse one. | adversarial, correctness | 0.90 |
| 3 | `…component.scss:115` | **The time field truncates the value it displays.** Scaling the input's type off row height overflows its width; `::-webkit-datetime-edit` is clipped inside the shadow DOM, so `scrollWidth === clientWidth` and nothing signals it. Measured: 163×242 shows `10.35.0`, 163×300 shows `10.35`, 92×163 shows `10.3`. Pre-change the field was 13.3px and showed the full value. | adversarial | 0.85 |
| 4 | `…component.scss:120` | **The mode-0 readout wraps and overflows the row.** `.start-at` lost its SVG and gained no `white-space: nowrap`; `@ HH:MM:SS` is 10 characters sized as if it were `-1m`, so it breaks at the space. Measured at 163×300: 64px of text in a 63.52px row, spilling into the canvas and clipped by `:host { overflow: clip }`. | correctness, adversarial | 0.85 |
| 5 | `…component.scss:112` | **Night theme leaves the field at 1.27:1.** `.night-theme` overrides `on-surface` to `#831d1d` but not `surface-container-highest`, so the value renders dark red on `#3e3230`. Pre-existing (was 1.36:1 black on `#520000`) — but this PR rewrites that exact declaration and leaves the theme actually used at sea after dark unreadable. | adversarial | 0.85 |
| 6 | `…component.spec.ts:58` | **The accessible-name test cannot detect the defect it is named after.** `accessibleName` falls back to `textContent`, which includes text inside `mat-icon` and `svg text` — the very markup the a11y tree drops. Re-wrapping `Start`/`Sync`/`Reset` in a `mat-icon` leaves them nameless and the test still passes. The author's mutation check only covered the `⋮` button, whose glyph has no letters, so it confirmed one button rather than the rule. | testing, adversarial | 0.90 |
| 7 | `…component.spec.ts:16` | **No test clicks anything.** Ten buttons were re-authored by hand with sign-sensitive arguments (`-60`/`+60`/`-1`/`+1`) and two modes reuse the same `-1m`/`+1m` labels. `putRequest` is a `vi.fn()` no test ever reads, so a swapped handler ships green. | testing | 0.80 |
| 8 | `…component.ts:211` | **Clearing the time field throws.** `''.split(':').map(Number)` is `[0]` → `setHours(0, undefined, 0, 0)` → Invalid Date → `toISOString()` raises `RangeError` out of the change handler, after `mode.set(0)` has already run. Pre-existing, but this PR is what makes the field usable enough to reach it. | correctness | 0.72 |

### P3 — Low

| 9 | `…component.ts:138` | **The field is blank in 12-hour and dot locales.** `startAtTimeEdit` is filled from `toLocaleTimeString`, which gives `05:30:45 PM` under en-US and `17.30.45` under fi-FI; `input[type=time]` coerces both to `""`. Only ISO-like locales work. Pre-existing, and it defeats this PR's fix for anyone in those locales. | correctness | 0.76 |
| 10 | `…component.spec.ts:26` | The `CanvasService` stub declares `MIN: 0`; the widget reads `MIN_LABEL_PX` (real value 16). The comment claiming the stub covers every member the widget touches is already false, and `useValue` being `any` means nothing catches it. | correctness, testing, maintainability | 0.95 |
| 11 | `…component.spec.ts:82` | The `mat-icon` / `svg text` guard runs in mode 1 only; modes 0, 2 and 3 carry five more labels that had the same markup. | testing | 0.80 |
| 12 | `…component.spec.ts:62` | Comments narrate the change ("the labels used to live inside…") rather than the constraint, against the repo standard that a comment is written for a reader who never saw the diff. | maintainability | 0.68 |

| # | Issue | Action |
|---|-------|--------|
| 13 | `widget-race-timer` and `widget-racer-line` still use SVG-text-in-`mat-icon` labels, including the same missing accessible names. Nothing records it. | file-issue |
| 14 | `.main-button` / `.main-small-button` and `.other-button` / `.other-small-button` are byte-identical pairs; the `-small` suffix promises a size difference no declaration delivers. | file-issue |

### Corrections to the PR body

- The touch-target row is wrong in the direction that matters: the tap area **shrank**, it did not cross 44px.
- The measured heights omit widget-host2's 1px dashed card border. Real values are 20.3/9.14, 49.1/22.1, 77.9/35.06 px, not 20.8/9.4, 49.6/22.3, 78.4/35.3.
- The colour sentence has foreground and background swapped: light theme measures text `rgb(27,27,31)` on background `rgb(229,225,230)`.

### Coverage

- Suppressed: 0 below the 0.60 gate.
- **Container query units have no fallback.** On an engine without container-query support (pre-Chrome 105 / Safari 16) `45cqh` resolves against the viewport — measured 315px type at a 700px viewport. Skip runs inside third-party chartplotter browsers via the Freeboard-SK panel iframe, where the engine version is not controlled.
- `container-type: size` is safe only while the percentage-height chain resolves. Under an auto-height ancestor the row measures 8px (padding only) and every label drops to the 9px floor. Both real mount points give a definite height today.
- Neither reported defect has a regression guard, and jsdom cannot provide one — it performs no layout and `cssstyle` resolves neither `var()`, `max()` nor `cq*`. The token half is closable cheaply with a source-scanning spec in the style of `help-docs-image-paths.spec.ts`; the label-size half needs a rendering check.

---

**Verdict: not ready.** Findings 1–4 are regressions this change introduces, and 1 and 2 make the widget worse than before on a touchscreen and on any portrait tile — the two conditions a HaLOS device is most likely to be in. The sizing model needs to bound by both axes, as the SVG it replaced did, and the touch target must stop being clipped.

Fix order: 2 → 1 → 3, 4 → 5 → 9, 8 → 6, 7, 11, 10, 12. Findings 13 and 14 to issues.

mairas:
## Review findings addressed

All eight P2s and the P3s are fixed. Two pre-existing items went to issues. The PR description has been rewritten — three of its claims were wrong and are corrected below.

**The sizing model changed rather than being patched.** Each control is now its own `container-type: size` container and its label takes `min(<h>cqh, <w>cqw)`, so it is bounded by the narrow axis too. That is what `preserveAspectRatio="meet"` got right and what my first attempt dropped. Verified across 3 themes × 8 tile sizes (including 92×163, 163×92, 213×360 and 240×320) × 5 modes: **zero clipped labels**, where the previous revision clipped `Start` and every sign in mode 3.

| # | Fix |
|---|---|
| 1 | `overflow: hidden` is gone from `.button`, so Material's 48px touch-target span is no longer clipped. Correct sizing makes the clip unnecessary; `text-overflow: ellipsis` on the label makes any residual overflow visible instead of cutting both ends. |
| 2 | Two-axis bounding, above. |
| 3 | The time field is bounded by its own width too (`13cqw`), and its value fits at every size measured. |
| 4 | `.start-at` is a flex box with `white-space: nowrap` on the label and its own width bound (`11cqw`) — it cannot wrap out of the row. |
| 5 | The field takes the widget's own surface and `--skip-contrast-color` rather than the M3 pair the night theme does not override. Measured against the real `--skip-widget-card-background-color`: light **21:1**, dark **21:1**, night **1.36:1** — the night theme is dim red on black by design and every label in it sits at 1.36:1, so the field now matches the theme instead of being a special case. |
| 6 | `accessibleName` strips `[aria-hidden="true"]` subtrees before reading `textContent`. Mutation-checked by wrapping `Start` in an aria-hidden span the way `mat-icon` did — the test now fails, where before it passed. |
| 7 | Nine table-driven tests click each label and assert the request it sends. Mutation-checked by flipping `adjustStartTime(-1)` to `(+1)` — `mode 3: -1s` fails. |
| 8 | `setStartTime()` rejects a value that is not two or three finite numbers, before touching `mode`, so clearing the field leaves you in the edit mode instead of throwing. |
| 9 | `startAtTimeEdit` is built from local hours/minutes/seconds as `HH:MM:SS`. `toLocaleTimeString` yields `17.30.45` under fi-FI and `05:30:45 PM` under en-US, both of which `input[type=time]` silently blanks — the field showed nothing in those locales. The mode-0 readout keeps the locale formatting. |
| 10 | `canvasMock` is typed `Partial<CanvasService>` and declares `MIN_LABEL_PX: 16`. The typing immediately caught a second error: `createTitleBitmap` must return an `HTMLCanvasElement`, not `null`. |
| 11 | The `mat-icon` / `svg text` guard runs in all five modes. |
| 12 | Comments rewritten as the standing constraint rather than what changed. |

**New guard.** `src/app/core/utils/contrast-token-usage.spec.ts` scans `src/**/*.scss` and fails on `--skip-contrast-color` used as a `background-color`, naming file and line. Verified by re-introducing the shipped regression as a scratch file — it fails; removed, it passes. It also asserts the scan found files, so a cwd change cannot make it pass vacuously.

**Corrections to the original description**, all confirmed:
- The touch-target row was wrong in the direction that matters: the tap area **shrank** from 48px to the visual button height. It did not cross 44px.
- The heights omitted widget-host2's 1px dashed card border.
- The colour sentence had foreground and background swapped.

**Deferred:** #509 (`widget-race-timer` and `widget-racer-line` still carry the old technique and the same missing names), #510 (four button classes, two rule bodies). The accent-colour contrast inversion remains #507.

`npm run ci`: lint and `snc` clean, 171 of 172 test files pass — the failure is `app.component.spec.ts` (#502), which reproduces on a clean tree and passes in CI.

PR #511, opened 2026-08-09
Move the widget unit onto the label's row

mairas:
## Why

On the Numeric widget the label sat in the top-left corner and the unit in the bottom-right, so two bands of the tile went to chrome and the value was squeezed between them. The Linear gauge had the same shape for the same reason — canvas-gauges reserves a title band at the top and a units band at the bottom (15% of the canvas each on a horizontal gauge).

A survey of the other widgets found no further cases: Radial and Classic Steel draw label and unit inside the dial's dead centre, and Compact Linear already puts the unit on the value's row.

## What

**Numeric** — label and unit now share the top band, one in each corner, on a common baseline. The value box is the whole card below that band (derived from the row, replacing the `0.85w × 0.70h` constants), so the reading is drawn larger. A name too long for its share of the row is ellipsised: a label is never drawn below `MIN_LABEL_PX`, so one that does not fit at that size used to overflow harmlessly off the card and would now run into the unit.

This also drops the widget's reach into `CanvasService`'s private `drawTitleInternal` — the label goes through the public `drawText` like everything else.

**Linear** — `title` and `units` are left unset, so the library reserves neither band, and the component renders both as its own header row on the card. The bar and the scale grow into the reclaimed height. The gauge canvas is bottom-anchored rather than centred, because the library draws ticks and numbers above the bar and nothing below it — centring stacked its own bottom slack on top of the container's. The header also shows before the first datapoint arrives, where the library title only appeared once the gauge rendered.

`unitless` is suppressed rather than printed: `getUnitDisplaySymbol` falls back to the measure key, so the default config used to render the literal word "unitless" under the bar.

## Verification

Built and driven in a headless browser against the perf-harness mock Signal K server, screenshotting before/after in both themes across tile sizes: 4x4 and 2x3 numerics, a long name on a small tile, min/max enabled, a wide unit symbol (`ftm`), `unitless`, and vertical and horizontal gauges from 2x4 up to 6x4.

Five specs added for the layout invariants: unit in the top-right corner on the label's baseline, the label's width box narrowed by the unit's, ellipsis on overflow, no unit for a symbol-less measure, and the value box exactly filling the card below the label row.

`npm run ci` — lint and `snc` clean, 1849 tests pass. The one failure, `app.component.spec.ts > the same boot NOT under embed DOES fire the self-heal writes`, reproduces on a clean `main` in this environment (`localStorage` unavailable without `--localstorage-file`) and is unrelated to this change.


mairas:
## Code review — `fix/widget-unit-row`

**Scope:** `3238c316..a215f05b`, 8 files. **Intent:** move the unit out of its own band onto the label's row on the Numeric and Linear widgets, handing the reclaimed space to the reading.

**Reviewers:** correctness, testing, maintainability, project-standards (always-on) + adversarial (~250 changed lines across two widget render paths). All findings below were **fixed on the branch** before this comment; the branch was force-pushed with the fixes folded into their own commits.

### P1 — High

| # | File | Issue | Reviewer(s) | Fix |
|---|---|---|---|---|
| 1 | `widget-gauge-ng-linear.component.ts:onResized` | The header row leaves a one-row tile ~6px for the gauge; `height - 10` then went negative and canvas-gauges threw `IndexSizeError` from inside its own mutation observer — past the component's `try/catch` — leaving the canvas **blank** and spamming uncaught errors during a drag-resize. Reproduced in Chrome against the real library at 45×25; threshold is any dashboard shorter than ~892px, which covers 1024×600 panels. | adversarial (0.9) | `updateGaugeSize` sends a size only when both axes are positive |
| 2 | `widget-numeric.component.ts:drawLabelRow` | `labelWidth` was unclamped. On a 1-column tile (43px at 1280) it reaches ~0; `truncateLabel` always returned ≥1 glyph, and a floored draw carries no width cap, so the label was painted straight through the unit and the unit's halo erased it. Measured 8.5px of overlap at 43px wide — the exact collision the truncation was added to prevent. | adversarial (0.85), correctness (0.7), testing | Clamp; ellipsise to fit; drop the label when even one character will not fit |

### P2 — Moderate

| # | File | Issue | Reviewer(s) | Fix |
|---|---|---|---|---|
| 3 | `widget-numeric.component.ts:calculateMaxMinTextDimensions` | The enlarged value box ran to the card bottom, over the band `drawMinMax` writes into — which draws after the value with no halo, so they interleave. 151 (width, height) combinations overlap at DPR 2, none of which overlapped before. The new comment claimed "there is nothing else down there to avoid", which was wrong. | adversarial (0.7), correctness (0.65), maintainability (0.78) | `valueBoxBottom()` subtracts the min/max band and re-centres; recomputed per draw since a config edit can toggle it without a resize |
| 4 | `widget-numeric.component.ts` | The row is laid out from measured text, and a cold boot measures the **fallback** font — up to 8.7% off Roboto. Nothing re-ran it: the bitmap cache key is `displayName + '|' + unit` plus canvas pixel size, none of which change when the font loads. Self-heals only at fractional DPR, so it persists on exactly the integer-DPR panels HaLOS ships on. | adversarial (0.7), correctness (0.75) | `whenFontsReady()` drops the cached bitmap and redraws |
| 5 | both widgets | `getUnitDisplaySymbol` falls back to the measure key, so the `' '` ("No unit label") measure was truthy: it stole ~13px of the name's row and painted nothing, and the gauge rendered an empty `<span>` plus its flex gap. | adversarial (0.75), maintainability (0.62) | New `UnitsService.getRenderableUnitSymbol`, used by both |

### P3 — Low

| # | File | Issue | Reviewer(s) | Fix |
|---|---|---|---|---|
| 6 | `widget-numeric.component.ts:truncateLabel` | `slice(0, -1)` cut UTF-16 code units, so an emoji in a widget name truncated to a lone surrogate (12 tile widths reproduce it for `Engine 🔧🔧🔧 Temp`). | adversarial (0.8), correctness (0.6) | Slices code points |
| 7 | `widget-gauge-ng-linear.component.scss` | `.gaugeUnit` was `flex: 0 0 auto` with no overflow rule, so on a narrow tile the name collapsed to zero width and the unit was clipped mid-symbol 17px past the card edge. | adversarial (0.75) | Header clips; the unit may shrink and ellipsise, the name yields first |
| 8 | `widget-numeric.component.ts` | Three inline measure lists classified overlapping sets for related reasons and had already drifted — `'D HH:MM:SS'` was in the pre-formatted list but missing from the unit-suppression list, so the widget drew the literal text `D HH:MM:SS` as a unit. | maintainability (0.65) | Named `PRE_FORMATTED_MEASURES` / `PERCENT_MEASURES`; the suppression set is built from them |
| 9 | `widget-gauge-ng-linear.component.ts` | `headerColor` re-derived `theme.contrastDim` in TypeScript when `--skip-contrast-dim-color` already carries it per theme — a computed, a binding and a null branch replacing one CSS line. | maintainability (0.72) | Deleted; the SCSS reads the custom property |
| 10 | `widget-gauge-ng-linear.component.scss` | `@supports not (container-type: size)` can never apply — the build's browserslist floor is above `container-type` support, and the same file already relies on `overflow: clip` at the identical threshold. | maintainability (0.70) | Deleted |
| 11 | `widget-numeric.component.ts` | The `0.1` row fraction was written twice and had to be kept in sync by hand; `labelRowBottom` was documented as an edge but used as a text baseline **and** as the value box top. | maintainability (0.68, 0.62) | One `LABEL_ROW_FRACTION`; `labelRowHeight()` / `labelBaselineY()` with both roles documented |
| 12 | commits | Two subjects were 55 and 58 characters against the repo's 50-character limit. | project-standards (0.88) | History rewritten; all four subjects ≤50 |

### Testing

The original five specs were weak in ways the reviewers demonstrated: the CanvasService fake returned a constant font size and font-blind widths, so every one of the collision scenarios above **passed** the suite. The fake now scales width with size and weight and applies the floor the way `CanvasService` does, and the geometry assertions use absolute numbers instead of restating the implementation.

Added: a `keeps the label clear of the unit at every tile width` invariant across 7 widths, the no-room case, the emoji case, the blank-symbol case, min/max clearance, and `drawValue` position. Mutation-checked — reverting the clamp fails it with `expected 29.2 to be less than or equal to -20.2`, matching the overlap measured in Chrome.

`widget-gauge-ng-linear` had **no spec at all** (flagged by four reviewers). Added one on the radial gauge's harness: title/units stay unset, header renders label and unit, no unit for `''`/`unitless`/`' '`, and the sizing guard — also mutation-checked.

### Coverage

- **Not addressed:** `widget-windsteer` and `widget-windtrends-chart` each carry their own copy of the blank-symbol rule and could now call `getRenderableUnitSymbol`. Left out as unrelated files — worth a follow-up.
- **Residual risk (correctness, maintainability):** dropping `effectiveUnit` from the `buildGaugeOptions` effect is safe only because `adjustedScale` returns a fresh object each recomputation. Giving that computed an `equal:` comparator later would silently stop rebuilding options on a unit change.
- **Operational (project-standards):** a draft release `v1.3.1+1` already exists from #503. Merging this cuts `v1.3.1+2`; publish the **newest** draft, or npm gets 1.3.1 without this change and the version can never be republished.
- **Version bump:** correctly not bumped. `VERSION` is already `1.3.1` against stable `v1.3.0+2`, the cycle was opened at patch level, and this change adds no new capability.
- Untracked and excluded from scope: `perf-harness/{mock-serve,probe,shot}.mjs` (dev-only, deliberately uncommitted).
- `npm run ci`: lint and `snc` clean, **1862 tests pass**. The single failure, `app.component.spec.ts > the same boot NOT under embed DOES fire the self-heal writes`, reproduces on clean `main` in this environment (`localStorage` unavailable without `--localstorage-file`) and is unrelated.

---

**Verdict: ready with fixes — all applied.** Every finding above is fixed on the branch and re-verified in a headless browser across tile sizes in both themes, including the 1-column numeric and 1-row gauge tiles the review found broken.

Issue #515, opened 2026-08-09
Night theme collapses grey and orange, so the racer line's current-line button has no signal


mairas:
`widget-racer-line` distinguishes the currently selected line end with `.line-button-current` (orange fill) against `.line-button` (grey fill). The night palette maps both `grey` and `orange` to `#520000`, so in night mode the two render identically — same fill `rgb(82,0,0)`, same ink `rgb(229,108,91)` after #507. "This is the current line" has no visual signal at all.

Pre-existing palette property, not caused by #507 — that issue made both labels *readable*, which is what exposed it; before, both were invisible.

Fix direction: give the night theme a distinct tone for one of the two, or signal current-ness with something other than fill colour (a border or weight) so it survives a palette where accents collapse.

Found while verifying #507.

Issue #514, opened 2026-08-09
Autopilot mode label renders at 8px on a realistic tile


mairas:
The autopilot control bar's mode button is labelled `Mode {{ apMode() | titlecase }}` — up to `Mode Off-Line`, 13 characters. Its type is bounded by button width (`$label-ctrlbar-width: 11cqw`) so it cannot clip, which means a narrow button makes it small: measured **8.0px on an 81px-wide button** (a 4-of-12-column autopilot tile at a 1200px viewport), and 13.5px at 131px.

8px bold Roboto is below what a helm instrument can be read at from arm's length, which is the complaint that started this whole line of work.

Raising the coefficient does not solve it: at 11cqw the longest label already fills about 79% of the button, so there is roughly 1px of headroom before it clips. The cause is 13 characters in that space, so the fix is one of:

- Shorten the label — the word "Mode" is arguably redundant when the button's job is mode selection, and `Auto` / `Wind` / `Route` alone would roughly halve the width.
- Give the control bar's mode button more of the row.

Both change what the user sees, so neither was done unilaterally.

Related oddity in the same rule: `@supports not (container-type: size)` pins 14px, so a browser *without* container-query support renders this label larger and more legibly than a modern one on the same tile.

Found while verifying #507.

mairas:
Correcting this: measurement in #516's review found the premise partly wrong. The autopilot modes row is floored at 50px by `.ap-modes`'s `minmax(50px, 1fr)`, giving 10.8–16.8px for "Mode True Wind" across 240–800px tile widths, with no clipping in any geometry tested. "Tack Stbd" never clipped either.

The 8.0px figure came from a narrower hypothetical than the grid actually permits. Still small for arm's length on a helm display, so the shorter-label question stands — but it is not the 8px emergency this issue describes.

Issue #513, opened 2026-08-09
Autopilot dodge mode has no visual affordance and never has


mairas:
`widget-autopilot.component.html:98` applies `.ap-btn-active` to the Dodge button when dodge mode is engaged. **No SCSS rule anywhere in `src/` matches that class.**

The other attempt at the same affordance was also inert: `[style.fill]="dodgeModeActive() ? 'currentColor' : 'none'"` on the two chevron paths, where each `d` is a pair of single straight-line subpaths — so the implied close is a zero-area polygon with nothing to fill. Both variants were rendered and the screenshots are byte-identical.

So engaging dodge has never changed the Dodge button's appearance, in any theme. The user infers the state only from side effects elsewhere in the widget: the four route sub-widgets disappear and the ±1°/±10° row appears in `nav`.

The class binding was deliberately left in place as the hook when the widget's labels were converted to text (that conversion removed the dead chevrons, which regressed nothing). What "dodge engaged" should look like is a design decision — an inverted fill, a border, a different accent — so no style was invented for it.

Found while fixing #507 in this widget.

Issue #512, opened 2026-08-09
Material's hover/pressed state layer drops most racing-widget buttons below AA


mairas:
#507 fixed the label ink against each button's **rest** fill. Angular Material also paints a state layer over the fill on hover and press, which changes the effective background — and a rendered measurement across the four racing widgets found **16 of 17 night-theme buttons and 3 light-theme buttons fall below 4.5:1 in those states**, including pairs that score 9.69 at rest.

This is a distinct defect class from #507, which is why it is not folded in: the ink is now correct for the fill it was chosen against, and the state layer is a second, overlaid background nobody picked an ink for. It affects every filled button that overrides `filled-container-color`, not just the racing set.

Fix direction: Material exposes state-layer opacity tokens (`filled-hover-state-layer-opacity`, `filled-pressed-state-layer-opacity`) and a state-layer colour. Either tune the opacity per accent so the composited background stays within range of the chosen ink, or set the state-layer colour explicitly rather than letting it default. Whichever is chosen needs measuring in the composited state, not reasoning about the token values.

Reproduce with the rig left in the scratchpad from that pass (`states.mjs` measures rest/hover/pressed per class; `hover.mjs` validates the model against a real pointer).

Found while verifying #507.

PR #516, opened 2026-08-09
Legible, named and correctly-inked controls across the racing and autopilot widgets

### LINKED ISSUE #502: app.component spec: 'the same boot NOT under embed DOES fire the self-heal writes' fails on localStorage being undefined

### LINKED ISSUE #507: Filled-button label ink is inverted against the widget accent colours (3 of 4 theme/colour pairs fail WCAG AA)

### LINKED ISSUE #509: widget-race-timer and widget-racer-line still use viewBox-scaled labels with no accessible names
`widget-race-timer` and `widget-racer-line` still render their button labels as SVG `<text>` inside a `<mat-icon>`, the technique #508 replaced in `widget-racer-timer`. Both defects come with it:

- **Labels are scaled by empty space.** `preserveAspectRatio="xMidYMid meet"` fits the whole viewBox, and the viewBoxes are far wider than the words (`widget-race-timer.component.scss:101` still sets `font-size: 15px` inside a `0 0 100 14` box). The label shrinks to make room for the emptiness beside it.
- **No control has an accessible name.** Angular Material sets `aria-hidden="true"` on `mat-icon` unless given an explicit value (`icon.mjs:99`), so the label is dropped from the accessibility tree and neither template carries an `aria-label`. Every button reads as unlabelled.

The approach that replaced it in #508: each control is a `container-type: size` container, and its label is ordinary text at `min(<h>cqh, <w>cqw)` so it fits both axes. Worth extracting the shared pieces (the size-container rule and the coefficient set) when the second widget is converted, not before — two call sites is when the shape is clear.

`widget-racer-timer.component.spec.ts` has a test pinning the accessible-name property that can be copied; note it strips `[aria-hidden="true"]` subtrees before reading `textContent`, since `textContent` alone cannot see this defect.

Found in review of #508.

### LINKED ISSUE #510: Racer widgets: four button classes, two rule bodies, and a -small suffix that changes nothing
In `widget-racer-timer.component.scss`, `.main-button` and `.main-small-button` are byte-identical, as are `.other-button` and `.other-small-button` — each pair sets the same `flex` and the same two `mat.button-overrides`. The `-small` suffix promises a size difference no declaration delivers, so editing button sizing in one pair leaves half the modes unchanged and the next reader hunting for the difference.

There are really two roles: the port-coloured primary action and the blue secondary action. Collapsing to two names that say that would touch nine template usages in this widget, and the sibling racing widgets likely carry the same pattern — worth checking them in the same pass.

Left out of #508 deliberately: that PR was already rewriting every line of markup carrying these names, but folding a rename in would have made a layout fix hard to review.

mairas:
## Why

Four issues that turned out to share one root and one file set.

**#507 — the label ink was chosen against the wrong thing.** The racing and autopilot widgets set `filled-label-text-color: var(--skip-contrast-color)`, the ink that contrasts with the widget *surface*, on buttons whose fill is a fixed accent colour. The accent does not track the theme in the same direction — `blue` is dark in the light theme and light in the dark theme — so the pairing inverts. Measured: light blue 3.16, light port 3.01, dark blue 2.97. In the night theme, whose palette collapses `grey`, `orange` and the contrast ink onto `#520000`, several labels rendered in the exact colour of the button behind them: **1.00:1, invisible**.

**#509 — two widgets still drew labels as SVG text inside `mat-icon`,** in a viewBox far wider than the words, so `preserveAspectRatio="meet"` scaled them to fit the emptiness rather than the button. Material marks `mat-icon` `aria-hidden` unless given an explicit value, so every one of those buttons was also unnamed to a screen reader.

**#510 — `.main-button`/`.main-small-button` and `.other-button`/`.other-small-button` were byte-identical pairs.** The `-small` suffix promised a size difference no declaration delivered.

**#502 — Skip could fail to boot where local storage is denied.** Reading the `localStorage` *property* throws `SecurityError` in Firefox and Safari when the origin is denied storage, so the usual `typeof localStorage !== 'undefined'` guard does not help: the throwing getter still runs. The constructor's guard skipped only the *load* path and left every later write dereferencing bare, which is how a dashboard effect reached `setItem` on `undefined`.

## How

Six per-accent on-colour tokens (`--skip-on-{blue,blue-dim,port,starboard,orange,grey}-color`) across the three themes, each chosen by measured WCAG ratio rather than by eye, and applied to every filled button in the four widgets.

The night theme needed care rather than the same treatment. Its accents are dark reds and its contrast ink is one of them, so no darker ink can clear the bar — the new inks are the dimmest tones of night's *own* primary palette that reach 4.5:1, keeping the red cast instead of importing white. Night `starboard` is the exception and needed a brighter tone: it is the one green fill in an all-red palette, and pure black manages only 2.30 against it, so nothing darker can reach even 3:1.

`widget-race-timer` and `widget-racer-line` move to the text-label technique from #508 — each control its own size container, label bounded by *both* axes so a tile taller than it is wide cannot clip the sign off an adjust button.

**`widget-autopilot` could not be fixed by a token at all.** Its labels were SVG inside `mat-icon`, and `.night-theme` includes `mat.icon-overrides((color: var(--skip-contrast-color)))`, which emits `--mat-icon-color` matching the `mat-icon` element directly and beating inheritance from the button. Proven rather than assumed: forcing `--mat-button-filled-label-text-color: #00ff00` rendered green in light and dark and stayed `#520000` in night. Converting its labels to text removes the element that selector matches, which is what lets the fix reach night.

For #502, the 17 dereferences across settings, network init and config upgrade go through one accessor that resolves the `Storage` object first. Absent storage reads as null and writes no-op; a write the store *accepts and then rejects* still logs, because a quota failure is a fault and an unavailable store is not.

## Result

Every button in all four widgets clears 4.5:1 in light, dark and night at rest. The autopilot's night pairs went from 1.15 / 1.41 / 1.69 to 5.60 / 4.78 / 5.37.

Measured in a browser against Angular Material's real filled-button rule extracted from the built bundle, with the theme class on `<body>` where the `--skip-*` tokens are actually scoped, and every alpha composited down the real ancestor chain. Three independent verification passes; the second and third re-measured the first's claims rather than trusting them, and the second found that the issue as filed had missed three of the six accents.

## Verification

`npm run lint`, `npm run snc`, `npm run test:headless` (174 files / **1881** tests), `npm run test:mcp-schema`, and `npm run build:dev` all clean. The build is included deliberately: the other four never compile `src/styles.scss`, so a missing Sass variable would pass all of them and break only at app build.

Deployed to a HALPI2 and serving.

## Found but not fixed

- #512 — Material's hover/pressed state layer drops 16 of 17 night buttons and 3 light ones below AA. A second, overlaid background that no ink was chosen against; affects every filled button that overrides its container colour, not just these.
- #513 — the autopilot's dodge mode has no visual affordance and never has. `.ap-btn-active` matches no rule, and the other attempt (`[style.fill]` on two straight-line subpaths) encloses zero area. Rendered both variants: byte-identical. Needs a design decision, so no style was invented.
- #514 — the autopilot mode label renders at 8px on a realistic tile. `Mode Off-Line` is 13 characters and already fills ~79% of its button, so the coefficient has ~1px of headroom; the fix is a shorter label or a wider button, both product calls.
- #515 — night collapses `grey` and `orange`, so the racer line's current-line button has no signal. Pre-existing; #507 made both labels readable, which is what exposed it.

Closes #502, closes #507, closes #509, closes #510


mairas:
## Code review — PR #516

**Reviewers:** correctness, testing, maintainability, adversarial. Two measured in headless Chrome against the compiled component SCSS, old versus new.

**First, what is clear.** Correctness traced all ten autopilot controls, every racer-line handler and argument, and the racer-timer class renames against `origin/main`: click handlers, `[disabled]` bindings, adjust-button signs and command strings are **identical**. Adversarial separately checked whether a denied storage read could overwrite a server-side profile — it cannot; `startup()` returns early unless the remote context is bootstrapped, and a missing slot routes to the degraded UI rather than a write. The two things I most wanted challenged both came back clean.

### P2 — High

| 1 | `widget-racer-line…scss:39` | **Percentage bottom padding drives the control row to zero height.** `padding-bottom` resolves against the containing block's *width*; the row is `height: 24%` with border-box, so its content box is `0.24H − 0.03W`. Measured old→new control height: 800×300 58→48px, 1200×250 46→24px, 1400×180 29.2→**1.2px**, 1900×166 25.8→**0.0px** — buttons invisible and untappable, and `min(45cqh, 26cqw)` then computes to 0. A full-width racer line on a dense 24-column dashboard lands exactly there. My commit said the percentage "stays proportional"; it is proportional to the wrong axis. `widget-racer-timer` carries the same `3%` from #508. | adversarial, correctness | **0.95** |
| 2 | `widget-racer-line…html:47` | **The line-name button truncates any name past ~8 characters.** Mode 4 renders `getDisplayLineName()` — arbitrary user data from the racer plugin — with the `.label` coefficient tuned for three-character labels. Measured at the widget's own default tile: `Committee` needs 214px in a 174px button, `Windward Gate` needs 312px. Both ellipsise to the same stub, so the helm cannot tell which line they are about to make current. The SVG this replaced used `viewBox="0 0 150 14"` with `meet`, which *scaled* an 18-character name to fit. | adversarial | 0.85 |
| 3 | `settings.service.ts:332` | **A profile switch silently reverts.** `setActiveProfile` updates memory, no-ops the write, then reloads; boot re-reads null and falls back to `default`. The page reloads, so the switch looks like it happened, and it never takes — forever. The previous `setItem` threw, which at least surfaced it. Same shape for `resetConnection`, `setIsRemoteControl`, `setInstanceName`, `setRemoteContextDemand`. | adversarial | 0.80 |
| 4 | `settings.service.ts:597` | **Denied storage mints a new device UUID every load.** "Denied" and "first boot" are now the same state, so `getDefaultConnectionConfig()` runs each load and calls `UUID.create()` while its write no-ops. `SkipUUID` is the remote-control identity published as `self.displays.<uuid>`, so every session that enables Remote Control leaves a permanent orphan entry no UI can remove. | adversarial | 0.72 |
| 5 | `settings.service.ts:106` | **Nothing boots `SettingsService` with storage denied.** The constructor now runs `loadConnectionConfig()` and `startup()` unconditionally — that *is* the #502 fix — and no test exercises it. The util is tested only in isolation. | testing | 0.90 (P1 as filed) |
| 6 | `local-storage.util.ts:24` | **The accessor guards the write but not the read.** `setLocalStorageItem` wraps the call; `getLocalStorageItem` and `removeLocalStorageItem` guard only the property lookup. A store that resolves but throws from its methods still crashes the constructor — and `removeLocalStorageItem('authorization_token')` runs there unconditionally on every boot. That is the exact #502 failure, re-entering through the fix. | adversarial, testing, maintainability | **0.85** |
| 7 | `widget-racer-line…scss:160` | **`.port-small-button` and `.starboard-small-button` are byte-identical to their non-`-small` twins** — the defect #510 describes, in a file this PR edits, while the PR claims to close #510. Both members of each pair were touched here. | maintainability | 0.88 |
| 8 | `widget-race-timer…spec.ts:163` | The five adjusters differ only by sign and magnitude into `addTime`/`removeTime`; the new spec asserts label text and DOM order only. Swapping `+5s` with `-5s` keeps the suite green while the countdown runs the wrong way at a start sequence. `widget-racer-timer` has the table-driven precedent this widget did not get. Same for the autopilot: hand-authored directional names (`"Turn 1 degree to port"`) sit beside hand-retyped handlers with nothing asserting they agree. | testing | 0.88 |

### P3 — Low

| # | Issue | Reviewer | Conf. |
|---|-------|----------|-------|
| 9 | The tack confirmation says `Repeat [Tack Starboard] key to confirm` while the button now reads **Tack Stbd** — under a 5-second countdown it names a control that is not on screen. Before this PR both buttons read "Tack" and the prompt was the only disambiguator. | correctness, adversarial | **0.95** |
| 10 | `isLocalStorageAvailable()` only asks whether the Storage object exists, so a kiosk WebView with a zero quota reports available, skips the boot warning, then drops every write. `SsoRedirectService.budgetStorageWorks()` already does the write/read-back probe for exactly this. | correctness | 0.75 |
| 11 | The v10 migration writes `appConfig` (the version stamp) first, then dashboards and theme, then reloads unconditionally. A swallowed failure on the dashboards write leaves the version stamped, so the upgrade never re-runs and the user lands on an empty dashboard set. The old throw stopped before the reload. | correctness | 0.62 |
| 12 | The label technique is copied into four stylesheets and **has already drifted**: `$label-height` is 45cqh in three and 40cqh in the autopilot, undocumented; the `@supports` block lists two selectors in one file and one in the others. Two sibling comments assert *opposite* conclusions about which axis binds. | maintainability | 0.85 |
| 13 | `accessibleName` is copy-pasted into four specs, and `src/test-helpers/` already exists as the shared home. | testing, maintainability | **0.90** |
| 14 | The autopilot spec pins `length === 10`, which holds only because `runtimeMock` hardcodes `apiVersion: 'v2'`; the shipped `DEFAULT_CONFIG` has `null` and renders 9. | testing, correctness | 0.80 |
| 15 | Comments narrate the change rather than the constraint — "which is the one thing the SVG this replaced got right", "The rewrite re-authored every button by hand", `describe('local-storage.util (#502)')`. | maintainability | 0.78 |
| 16 | Dead: `filled-icon-spacing: 0px` in three widgets now that no button holds an icon; `.autopilot-btn` in an autopilot media query matches nothing. | maintainability | 0.80 |

| # | Issue | Action |
|---|-------|--------|
| 17 | `loadConfigFromLocalStorage` does an unguarded `JSON.parse` in the constructor, so a corrupt value hard-fails boot with no in-app way to clear it. Its sibling `readStoredConnectionConfig` guards the same parse. This is the remaining fatal boot path after #502. | fix-here — trivial, same function the diff rewires |
| 18 | `.timer-adjusters-group` collapses to zero height on short tiles — measured 0.0px at 240×400, 800×300 and 1400×200 in **both** old and new, so the conversion preserved it rather than causing it. | file-issue |
| 19 | A declined reload (`ReloadService` returns without navigating on a 5xx) leaves `storage.sharedConfigName` repointed, so later saves PATCH the newly selected slot with the old profile's dashboards. | file-issue |

### Coverage

- `aria-label` **replaces** the visible label rather than containing it (`"Turn 1 degree to port"` vs `-1°`), which breaks WCAG 2.5.3 Label in Name for voice control — a user saying "click minus one degree" gets nothing.
- Nothing pins the geometry contract. Findings 1 and 2 are both invisible to jsdom, which does no layout and resolves no `cq*` units. A headless-Chrome check asserting a minimum control height and `scrollWidth <= clientWidth` at a few aspect ratios would catch both.
- Nothing guards the token wiring: a token declared in two of three theme blocks renders inherited ink in the third and lint, `snc`, tests and the build all stay green. `contrast-token-usage.spec.ts` is the existing precedent.
- **#514 overstates its case.** The autopilot modes row is floored at 50px by `.ap-modes`'s `minmax(50px, 1fr)`, giving 10.8–16.8px for "Mode True Wind" at 240–800px tiles with no clipping. I will correct that issue.

---

**Verdict: not ready.** Finding 1 makes a control row disappear entirely at a reachable tile aspect, and it is mine from #508 — the percentage I chose resolves against width, so widening a tile shrinks the controls it was meant to space. Finding 6 lets the #502 crash back in through the fix that closes it. Findings 3 and 4 are consequences of making "denied" and "empty" indistinguishable that were not thought through when that call was made.

Fix order: 1 → 6, 17 → 9, 7 → 2 → 3, 4, 5 → 8 → the P3 cleanups. 18 and 19 to issues.

mairas:
## Review findings: critical fixed, rest deferred to issues

Per the decision to take the critical set now and file the rest.

**Fixed in this PR**

| # | Fix |
|---|---|
| 1 | **The percentage padding is gone.** It resolved against the containing block's *width* while the row is a percentage of the tile's *height*, so widening a tile shrank the controls it was meant to space. Re-measured at the same geometries the review used: 1900×166 **0.0 → 31.8px**, 1400×180 **1.2 → 35.2px**, 1200×250 24.0 → 52.0px, 800×300 48.0 → 64.0px, and 160×120 unchanged at 20.8px. Nothing zeroes anywhere. A first attempt at `min(14px, 8cqh)` was discarded — `.buttons-container` has no *ancestor* container, so `cqh` would have resolved against the viewport, which is the same bug in a new coat. |
| 6 | **Every storage operation is total now.** `getItem` and `removeItem` are wrapped like `setItem` was. This mattered: `removeLocalStorageItem('authorization_token')` runs from the `SettingsService` constructor on every boot, so a store that resolves and throws from its methods reintroduced exactly the crash the accessor exists to prevent. |
| 10 | **`isLocalStorageAvailable()` does a real write/read-back probe**, cached per session, so a zero-quota kiosk WebView reports unavailable instead of answering yes and discarding every write. `setLocalStorageItem` returns whether the write landed, which #517 needs. |
| 17 | **A corrupt stored value no longer kills boot.** `loadConfigFromLocalStorage` parses in the constructor with no guard, so a truncated entry failed DI with no in-app way to clear it; its sibling `readStoredConnectionConfig` already guarded the same parse. |
| 9 | **The tack prompt quotes the label on screen** — `[Tack Stbd]`, not `[Tack Starboard]`. |
| 7 | **`.port-small-button` / `.starboard-small-button` removed**, the same duplication #510 removed from the racer timer. #510's scope is now genuinely closed. |
| — | **The degree controls' aria-labels are dropped.** They replaced the visible text rather than containing it, so voice control could not reach a button whose spoken label was absent from its name (WCAG 2.5.3). Their visible text is the name now, and the spec asserts a name exists rather than that it contains letters — `-1°` has none. |

**Deferred, with the decisions recorded**

| Issue | What |
|---|---|
| #517 | Storage denied should **block the change and say so**, not accept and lose it. Covers the silently reverting profile switch and the per-load UUID churn. The session-memory-fallback alternative was considered and rejected — it still loses everything on reload. |
| #518 | The racer line's long line names should **size from measured text length**, as the SVG did, rather than truncating. |
| #519 | Race-timer adjusters collapse to zero height on short tiles — measured identical before and after, so pre-existing. |
| #520 | A declined reload leaves saves pointed at the unloaded profile. Pre-existing. |
| #521 | Extract the label-sizing technique; four copies have already drifted, including two comments asserting opposite conclusions about which axis binds. |
| #522 | Click-through coverage for the hand-rewritten controls, plus the shared `accessibleName` helper and the autopilot spec's mock-pinned control count. |
| #523 | No test boots `SettingsService` with storage denied. |
| #524 | Nothing guards token wiring or rendered geometry — both defect classes have shipped once each. |

I also corrected #514: measurement found its premise partly wrong. The autopilot modes row is floored at 50px by `.ap-modes`'s `minmax(50px, 1fr)`, giving 10.8–16.8px with no clipping, not the 8px I reported.

**Verification.** `npm run lint`, `npm run snc` clean. Full suite 174/174. One run showed `minichart.component.spec.ts` failing; it passes in isolation and the suite is green on re-run — that is #469, the documented flake under full-suite load, not a regression.

Issue #500, opened 2026-08-07
[Bug]: Absolute start-time input in Racer Timer widget flashes closed instantly on iOS (Safari/Chrome)


jrehm:
### Operating System

Linux, Other

### Browser

Safari, Chrome

### Skip Version

1.2

### Signal K Server Version

2.29

### Bug Description

Originally filed this under KIP https://github.com/mxtommy/Kip/issues/1136#issue-5028663038 - not sure if it will be addressed there.

In the Racer Timer widget (widget-racer-timer), the "set absolute start time" mode renders a native <input type="time" step="1"> bound via [(ngModel)]="startAtTimeEdit":

https://github.com/mxtommy/Kip/blob/master/src/app/widgets/widget-racer-timer/widget-racer-timer.component.html#L90

On iOS (confirmed on iPhone and iPad, in both Safari and Chrome for iOS), tapping into this field opens the native time-picker popover for only a fraction of a second before it closes itself, making it impossible to enter an absolute start time on iOS. Expected: the picker stays open until the user picks a time or taps away.

Suspected root cause: KIP's dashboard grid is built on gridstack.js (gridstack/dist/angular), which positions every widget tile via CSS transform: translate(...) on the .grid-stack-item wrapper for GPU-accelerated drag/resize. This puts the native <input type="time"> inside a transform-positioned ancestor chain. iOS/WebKit has a long-standing, well-documented bug where a native date/time input's popover, anchored to an element inside a transform context, cannot be stably positioned — WebKit renders it for a frame and then immediately tears it down. This matches the "flashes open for ~0.1s" symptom exactly, and is specific to iOS/WebKit-based browsers (Safari, Chrome-for-iOS, Firefox-for-iOS all share the WebKit engine on iOS), which lines up with it only being reported on iOS.

This isn't limited to widget-racer-timer in principle — any widget using a native <input type="date|time|datetime-local"> inside the gridstack-positioned dashboard would likely hit the same issue on iOS. I also confirmed the identical markup/bug is present in the HALOS-branded fork (@halos-org/skip v1.1.0), which is a downstream fork of this codebase — consistent with a structural bug rather than something environment-specific.

[...]

Issue #469, opened 2026-07-30
Flaky: minichart spec 'plots a backfill batch…' sees 0 chart points under full-suite load


mairas:
## Symptom

`minichart.component.spec.ts` → **"plots a backfill batch and then a live point off the history stream"** intermittently fails:

```
AssertionError: expected +0 to be 3
  src/app/widgets/minichart/minichart.component.spec.ts:123
    expect(chartValueData(component).length).toBe(3);
```

The synchronous assertion at line 123 sometimes sees **0** chart points instead of 3 after pushing a 2-point backfill batch + 1 live point through a mocked history stream.

## Evidence it's a test-determinism flake (not product code)

- **Passes 4/4 in isolation** (`ng test --include=…/minichart.component.spec.ts`).
- **Intermittent only under full-suite load** — observed ~2 of 4 full `npm test` runs on a loaded dev box (halpi, running live services + builds); CI on the less-loaded Node 24 runner was green.
- `vitest` runs with `isolate: true` (no config override), so it is **not** cross-file global leakage.
- Matches the documented **order-dependent chart.js-mock** class (self-contained chart.js spec mocks must export the full member set the util registers / reset state between tests, else order-dependent failures). The prior test in the file (line 100, `HISTORY_UNAVAILABLE`) shares the module-level chart.js mock; residual mock state across the 4 tests is the likely trigger.

## Suggested fix direction

- Make the assertion deterministic: flush the stream / run change detection (or `await` a tick) before asserting `chartValueData(component).length`, rather than assuming the backfill→chart update is synchronous.
- And/or reset the chart.js mock state in `beforeEach` and confirm the mock exports the full member set the streaming path writes/reads.

## Provenance

Surfaced during the PR #468 review (windsteer animation fix) full-suite verification. Unrelated to that change — it touches none of the same files. Filing so the flake is tracked rather than dropped.

mairas:
Hit this again on PR #516's CI: `AssertionError: expected +0 to be 3` in the full-suite run, passing on a plain re-run of the same commit with no code change ([failed job](https://github.com/halos-org/skip/actions/runs/31326791349/job/93278339160), [passing re-run](https://github.com/halos-org/skip/actions/runs/31326791349/job/93278567324)).

Also reproduced locally in the same shape earlier today: failed once in `npm run test:headless`, passed both in isolation via `--include` and on a second full run. So it is load-dependent rather than order-dependent on any particular sibling spec.

[...]

PR #525, opened 2026-08-09
Leave native form controls to the browser in the gesture recogniser

mairas:
## Why

Tapping the Racer Timer's absolute start-time field on an old iPad does nothing at all — the field never focuses and no picker opens. That is a stronger form of #500, where the picker reportedly opens for about a tenth of a second and closes.

`GestureDirective` sits on every widget card (`widget-host2.component.html`, `[enablePress]="dashboardStaticView()"`, true in normal use). On a touch pointer it does two things to whatever you touched:

- `capEl.setPointerCapture(ev.pointerId)` on the **original event target** — the `<input>` itself.
- `ev.preventDefault()` on `pointerdown`, registered `{ passive: false, capture: true }`, to suppress the UA long-press hint.

On WebKit that `preventDefault()` also suppresses the control's default action, so the field never focuses and its picker never opens. Neither behaviour is wanted on a form control: a long press on a text field should select text, not open the widget's options.

This is the mechanism I flagged when #500 was first triaged and never got to test, because I had no iOS device. An iPad turned up and showed it directly.

## How

A pointerdown whose target is inside `input, select, textarea, option, [contenteditable]` bails before any tracking is set up, so the control keeps its own pointer handling. Gestures starting anywhere else on the card are untouched.

The blast radius is deliberately narrow — only the tapped control — but it is not limited to this widget: any Material form field reachable under the directive was subject to the same interference on touch.

## Verification

Three tests in `gesture.directive.spec.ts`, each mutation-checked by disabling the exemption:

- a pointerdown on an input is not `preventDefault`ed
- the pointer is not captured on an input
- a hold that starts on an input fires no `press`

`npm run lint` clean, `npm run snc` clean, suite 176/176. (One run showed the known `minichart` flake, #469; green on re-run.)

## What this does not settle

I still cannot verify #500 itself — no iOS device here, and Playwright's desktop WebKit does not use the iOS picker popover. This removes one of three candidate causes. The other two are untouched: the reporter's gridstack `transform` anchoring theory, and — now largely ruled out by the Mac-side fix in #516 — the input being destroyed mid-picker by a `(change)`-triggered mode switch.

Worth retesting on the iPad after this lands. If the field then works, #500 can close with the mechanism recorded, and it is worth reporting upstream: mxtommy/Kip#1136 is the same code.


mairas:
## Code review — PR #525

**Reviewers:** correctness, testing, maintainability, adversarial. Two measured in Chromium against the compiled component CSS.

**The pointerId question is settled, and the answer is "no bug".** Three reviewers flagged that the new guard returns after `_activePointers.add(ev.pointerId)`. The adversarial pass traced it: `onPointerUp` (line 831) and `onPointerCancel` (840) both call `cleanupActivePointer` unconditionally in their `!this.tracking` branch, and every ancestor host — widget card, gridstack, `.skip-app-shell` — receives the same bubbling pointerup, so whichever fires first removes the id. Five pre-existing early returns depend on the identical contract. No change needed; a test to pin it is worth having, since nothing covers it.

### P1

| 1 | `…racer-timer.component.ts:240` | **The Set button is dead after the failure it just reported.** `setStartTime()` returns early when the value equals `lastSubmittedStartTime`, which is assigned *before* the PUT and never reset — not on refusal, not on timeout. So the error toast asks the helm to retry, and the retry sends nothing, shows nothing, and changes nothing. Escape is altering the time by a second or reloading. The dedupe existed only because blur fired on every exit from the field; this PR removed blur, so it now blocks nothing but deliberate presses. Same dead end after a successful set followed by Reset. | adversarial, correctness, testing, maintainability | **0.95** |

### P2

| 2 | `…racer-timer.component.scss:133` | **The Set button squeezes the time value into truncation.** `.select-button` 18% + `.set-start-button` 22% + two 2% gaps is 44% of the row now fixed and unshrinkable, leaving the field 56% where it had 80%. Measured first non-clipping tile width: **104px on main → 140px with this PR**; at 120×120 the field gets 58.3px and needs 62.9px, rendering `23.45.0`. A 3-column widget on a 1024px dashboard is ~128px — inside the new clipping band. | adversarial | 0.88 |
| 3 | `…racer-timer.component.ts:244` | **Set does nothing, silently, when the seconds segment is unfilled.** `step="1"` gives three segments; a keyboard user types `2040`, sees `20:40`, presses Set — and `input.value` is still `''` until seconds are filled, so the parse guard bails with no request and no message. Hits exactly the keyboard users the Enter path serves. | adversarial | 0.72 |
| 4 | `gesture.directive.spec.ts` | **All three new tests assert absence**, so they pass equally if the directive stopped working entirely. Nothing pins that a *non*-control still captures and presses, so widening the selector to something widget cards are full of (`button`, `span`) would disable long-press-to-open-options with the suite green. Only `input` is exercised — `select`, `textarea`, `contenteditable` and the nested-target case that justifies `closest()` over `matches()` are untested. | testing | 0.85 |
| 5 | `…racer-timer.component.spec.ts` | The editing guard is tested only for `{editing, null value}` — one of four corners. The branch that **fills** the field from the server (`if (!editing) startAtTimeEdit.set(...)`) is executed by no test, so pinning `editing` to `true` would blank the pre-fill on every device and stay green. | testing, correctness | 0.85 |

### P3

| # | Issue | Reviewer | Conf. |
|---|-------|----------|-------|
| 6 | `FORM_CONTROL_SELECTOR` matches `[contenteditable=""]` and `="true"` literally, missing the valid `plaintext-only`; `option` is redundant since any `option` receiving a pointerdown is inside a `select` that `closest()` already matches. | maintainability | 0.68 |
| 7 | The changelog claims **"a long press selects text"** — false. The directive still sets `user-select: none` and `-webkit-touch-callout: none` on the host (inherited into the input) and preventDefaults `contextmenu` for every target. | correctness | 0.68 |
| 8 | The changelog narrates blur-commit, a behaviour introduced *and* removed inside this same unreleased block, so it documents a development step rather than a delta users will see. | maintainability | 0.80 |
| 9 | The SCSS comment on `.set-start-button` explains why the button exists rather than why the rule is what it is, and is the fourth copy of that rationale. | maintainability | 0.72 |
| 10 | `if (!this.pendingStartTimeRequest)` is unreachable — `putRequest` returns null only for an undefined value or empty path, neither possible at this call site. It also skips `clearPendingStartTimeTimer`, making the ordering look deliberate. | maintainability | 0.75 |
| 11 | `editing = this.mode() === 4` couples the observer to a magic number; `mode` is an untyped signal whose five meanings live only in `@case` branches. | maintainability | 0.78 |
| 12 | A start time set from another device is hidden while this one sits in mode 4 — the field freezes on *entering* the mode, not on focusing the input, and the next Set overwrites the other device's value. | adversarial | 0.65 |

### Coverage

- The iOS behaviour all of this exists for remains device-only. jsdom has no pointer-capture implementation and no focus consequence for `preventDefault`, so the specs prove the guard fires, not that it fixes the iPad.
- `isNativeFormControl` matches `input[type=range]`, so dragging the brightness slider no longer feeds the shell's swipe recogniser — a latent misfire before, unverified now.
- The guard enumerates tag names rather than "has a native default action". `<video controls>` in widget-video retargets `ev.target` to the `<video>`, so shadow-DOM controls stay unexempt — same class as the bug being fixed, not verified broken.
- Set measures ~23×21px at a 120px tile, under the 44px minimum. Shared with every other button in this widget, so pre-existing, but Set is new.

---

**Verdict: not ready.** Finding 1 makes the only commit path refuse the retry its own error message asks for, and findings 2 and 3 are two more ways to press Set and have nothing happen. Fix order: 1 → 3 → 2 → 4, 5 → 6 → 7, 8, 9, 10. Findings 11 and 12 to issues.

mairas:
## Review findings addressed

All P1s and P2s fixed, P3s except the two that became issues.

| # | Fix |
|---|---|
| 1 | **The dedupe guard is gone.** It existed only because blur fired on every exit from the field; with Set and Enter as the only commit paths, every press is deliberate. A test now presses Set again after a 403 and asserts a second request — mutation-checked by reinstating the guard, which fails it. |
| 3 | **A half-entered time says why.** A time input reads as empty until every segment is filled, so pressing Set after typing `20:40` without seconds hit the parse guard and returned silently. It toasts now, with a test. |
| 2 | **Set gives up width instead of the field**: `flex: 0 1 auto` with `min-width: 0`, so a narrow tile shrinks the button rather than truncating the time the user is about to commit. |
| 4 | **The guard's upper bound is pinned.** A non-control sibling asserts the mirror image — preventDefault, capture and press all still happen. Mutation-checked by widening the selector to include `button`, which now fails instead of silently disabling long-press on real widget content. The absence-assertions are table-driven over `input`, `textarea`, `[contenteditable]` and a *nested* target, which is what justifies `closest()` over `matches()`. |
| 5 | **Both branches of the editing guard are covered**: the field fills from the server while the form is closed, and a server value does not overwrite what is being typed while it is open. |
| 6 | Selector is `input, select, textarea, [contenteditable]:not([contenteditable="false"])` — picks up `plaintext-only`, drops the redundant `option`. |
| 7 | The changelog no longer claims a long press selects text. Correct catch: the host still sets `user-select: none` and `-webkit-touch-callout: none`, which the field inherits, and still preventDefaults `contextmenu` for every target. |
| 8, 9 | Changelog no longer narrates blur-commit, a behaviour introduced and removed inside this same unreleased block; the SCSS comment explains the rule rather than the button's existence. |
| 10 | The unreachable `!pendingStartTimeRequest` branch is gone and `clearPendingStartTimeTimer()` moved above the request, so a previously armed timer is cleared regardless of outcome. |
| — | Added a test that the shared `_activePointers` set drains after pointerdown+pointerup on a control, pinning the contract the adversarial trace relied on. Five older early returns depend on it and none of them pinned it. |

**No change for the pointerId question.** Three reviewers flagged it; the adversarial pass traced it and found `onPointerUp`/`onPointerCancel` call `cleanupActivePointer` unconditionally in their `!tracking` branch, with every ancestor host receiving the bubbling pointerup. It is now covered by a test rather than by argument.

**Deferred:** #526 (mode is an untyped number with five meanings — the named constant added here is a patch, not a fix), #527 (another device's start time is hidden while the form is open — gates on the mode rather than on the field being touched), #528 (the exemption enumerates tag names rather than native behaviour, leaving `<video controls>` shadow-DOM controls and `mat-select` outside it).

`npm run lint`, `npm run snc` clean; suite 176/176. Gesture spec 13 → 18 tests, racer timer 19 → 23.

PR #529, opened 2026-08-10
Fix video camera v2 URL fallback

macjl:
## Summary

- Keep using `endpoint.httpServiceUrlV2` for Video camera resource calls when the server advertises it.
- Fall back to the discovered v1 API URL when `httpServiceUrlV2` is missing.
- Fall back to the configured Signal K server URL if no discovered HTTP endpoint is available.
- Cover the camera save path with a test where only the v1 endpoint is discovered.

## Root Cause

The Video widget could probe cameras through the SK Video plugin because plugin URLs already use the configured Signal K base URL as a fallback. Saving a camera uses the Signal K Resources API instead, and the component only passed `endpoint.httpServiceUrlV2`. On Signal K servers that expose the v2 API but do not advertise `endpoints.v2` in `/signalk`, that value is `null`, so saving fails before the `PUT /signalk/v2/api/resources/cameras/:id` request is made.

## Validation

- `npm run snc`
- `npm run test:headless -- --include src/app/core/utils/signalk-plugin-url.util.spec.ts --include src/app/widget-config/video-camera-setup/video-camera-setup.component.spec.ts`

Note: the targeted test run emits existing Angular extended diagnostics warnings from unrelated templates, but the selected specs pass: 35 tests passed.

### DIFF
--- a/src/app/core/utils/signalk-plugin-url.util.spec.ts
+++ b/src/app/core/utils/signalk-plugin-url.util.spec.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest';
-import { resolveSignalKPluginBaseUrl } from './signalk-plugin-url.util';
+import { resolveSignalKPluginBaseUrl, resolveSignalKV2ApiBaseUrl } from './signalk-plugin-url.util';
 
 describe('resolveSignalKPluginBaseUrl', () => {
   const HTTP_V1 = 'http://host:3000/signalk/v1/api/';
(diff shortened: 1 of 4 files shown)


mairas:
## Code Review — multi-persona

**Scope:** 2 files, +8/−3, base `83e5da3c`, head `994b018c`.
**Reviewers:** correctness, testing, maintainability, project-standards (always-on) + api-contract (v1→v2 versioning derivation), adversarial (silent-failure surface). Each ran in an isolated context against the PR head.
**CI:** all checks pass (`tests`, `version-bump-check`, `version-check`).

The fix works for the case it targets. On a server advertising `http://h:3000/signalk/v1/api/`, branch 2 produces `http://h:3000/signalk/v2/api` and the PUT goes out. Everything below is about the branches around that one.

### P2 — Moderate

|---|---|---|---|---|
| 1 | `video-camera-setup.component.ts:72` | `String.replace` returns its input unchanged on a miss, and `??` only falls through on null/undefined — so any non-null `httpServiceUrl` that is not exactly `.../signalk/v1/api[/]` becomes the v2 base verbatim, and branch 3 can never run. `resolveSignalKPluginBaseUrl` ten lines up already strips `/signalk/v2/api`, `/signalk/v1/api`, `/signalk/v2`, `/signalk/v1` and bare `/signalk`, and its spec pins `https://boat.local/signalk` as a real input — so the codebase assumes shapes this regex does not cover. `CamerasResourceClient.collectionUrl` appends `/resources/cameras` to whatever it is handed. | correctness, testing, maintainability, api-contract, adversarial | 0.95 |
| 2 | `video-camera-setup.component.spec.ts:131` | The diff removed `httpServiceUrlV2` from the camera-mode fixture rather than adding a v1-only one. That describe block is the only place asserting a base URL passed to `resources.save`/`list`, so branch 1 — the path taken on every server that *does* advertise v2 — is now covered by nothing. The assertions still pass because the rewritten v1 URL is byte-identical to the removed v2 value, so deleting `this.endpoint()?.httpServiceUrlV2 ??` from the component keeps CI green. | testing, correctness, maintainability, api-contract, adversarial, project-standards | 1.00 |
| 3 | `video-camera-setup.component.ts:67‑70` | The two adjacent resolvers disagree on which server root wins: `resolveSignalKPluginBaseUrl` prefers the user-configured Signal K URL, `v2BaseUrl` prefers the discovered endpoint. `addCamera()` (lines 521‑525) PUTs the camera resource via `v2BaseUrl` and then POSTs credentials via `pluginBaseUrl` in one `try`, with no rollback. When the stored `signalKUrl` differs from the page origin, the resource lands on one host and the credentials call goes to another — cross-origin, `fetch` default `same-origin` credentials, so no cookie. The user sees "Could not save the camera" after the camera was already written. | maintainability, api-contract, adversarial | 0.80 |
| 4 | `cameras-resource-client.ts:28,61` | The fallback replaces a `null` base (which threw before any request) with a plausible URL, so requests now go out where none did. `list()` maps 404 → `[]`, so a wrong base shows "no saved cameras" and auto-opens the add form; `remove()` maps 404 → success, so delete clears the widget's `cameraId` and reports nothing wrong while the resource survives on the server. Pre-existing client behavior, made reachable by this diff. | adversarial | 0.68 |

### P3 — Low

|---|---|---|---|---|
| 5 | `video-camera-setup.component.ts:73` | Branch 3 is unreachable whenever the app is connected. `processEndpointResponse` throws if the server omits the v1 endpoint (`signalk-connection.service.ts:250`), so a Connected endpoint always carries `httpServiceUrl`; the branch only fires in Connecting/Disconnected/Error. The PR's third bullet describes a path that never runs for a connected server. | adversarial | 0.78 |

### Pre-existing

| # | File | Issue | Route |
|---|---|---|---|
| 6 | `history-api-client.service.ts:78` | `httpServiceUrl.replace('/v1/', '/v2/')` — unanchored, replaces the first `/v1/` anywhere, no configured-URL fallback. Same `endpoints.v2`-not-advertised bug class this PR fixes, in a second consumer. Counting the autopilot components' hardcoded `/signalk/v2/api` constants, the repo has four different rules for deriving a v2 base. | file-issue |
| 7 | `video-camera-setup.component.spec.ts:130,350,395` | All three `SignalKConnectionService` fakes omit `serverVersion$`, which CLAUDE.md requires. Passes today because nothing in this injection tree subscribes to it; `StorageService` already does. | fix-here |

### Confirmed non-issues

Trailing slashes are safe end to end (`baseOrThrow` strips one before joining). The reverse-proxy path prefix works (`https://host/sk/signalk/v1/api/` → `https://host/sk/signalk/v2/api`). Branch 1 still wins when the server advertises v2, so no previously-working configuration regresses. No `VERSION` bump is needed — main is at 1.4.0, latest stable is v1.3.0+2, so a prior PR already opened the cycle.

### Untested branches

v2 advertised (finding 2); v1 present but non-matching (finding 1); the configured-URL fallback, including a URL already ending in `signalk/` which would yield `.../signalk/signalk/v2/api`; and the all-null terminal. Every fixture in the spec sets `signalKURL: { url: null }`.

---

**Verdict: ready with fixes.** No P0/P1. Fixes for findings 1, 2 and 3 follow as a commit pushed to this branch.

mairas:
## Fixes for review findings 1, 2 and 3 pushed (7e8dd944)

`resolveSignalKV2ApiBaseUrl` now lives in `signalk-plugin-url.util.ts` next to `resolveSignalKPluginBaseUrl`, and both resolve the server root through one shared suffix ladder.

**Finding 1 — the `replace()` pass-through.** The ladder already strips `/signalk/v2/api`, `/signalk/v1/api`, `/signalk/v2`, `/signalk/v1` and bare `/signalk`, so shapes the old regex missed now yield the v2 base instead of falling through as a v1 URL. The end-anchoring is kept, so a reverse-proxy path prefix still survives (`https://halos.local/sk/signalk/v1/api/` → `https://halos.local/sk/signalk/v2/api`) — there is a test for that.

**Finding 3 — the precedence split.** Both resolvers now take the configured Signal K URL over the discovered endpoint, which is the rule `resolveSignalKPluginBaseUrl` already documented. A save can no longer write the camera resource to one host and its credentials to another. The advertised `httpServiceUrlV2` still wins over everything when the server publishes it, so the PR's first bullet is unchanged.

Two consequences worth naming:

- A configured URL now goes through the same suffix strip as the discovered endpoint, so one ending in `/signalk/` no longer produces `.../signalk/plugins/sk-video/` or `.../signalk/signalk/v2/api`. That changes `resolveSignalKPluginBaseUrl` output for that input, but only from a URL that was already wrong.
- Whether configured-first is the right global rule is a separate question. Skip's session cookie is same-origin only and `initializeConnection` runs with proxy mode on, so the discovered endpoint is always the page origin while a stored `signalKUrl` can be stale. If configured-first is wrong it is wrong for every plugin client, not just this one — that is its own change, not this PR's.

**Finding 2 — coverage.** The branch matrix moved to `signalk-plugin-url.util.spec.ts`, where each URL shape is a plain function call: advertised endpoint used as-is, trailing slash stripped, derivation from `/signalk/v1/api`, `/signalk/v1`, `/signalk/v2` and bare `/signalk`, reverse-proxy prefix, configured-URL precedence, doubled `/signalk` segment, blank configured URL, and the null terminal. The component spec keeps your v1-only block and gains `camera mode, advertised v2 endpoint`, whose fixture advertises `http://v2.h:3000/signalk/v2/api` — a host the derivation cannot produce, so the assertions fail if the advertised value stops winning.

Both guards were checked by removing the implementation: dropping the advertised-endpoint branch fails exactly the 2 new component tests, and restoring the original regex derivation fails 3 of the new util tests.

`npm run ci` passes locally — lint, snc, 1937 tests across 176 files, 33 MCP schema tests.

Not addressed here: finding 4, now tracked as [#564](https://github.com/halos-org/skip/issues/564); finding 5, which the precedence change above makes moot (the configured URL is checked before the discovered endpoint, so that branch is reachable while connected); and finding 7, a three-line spec change. Finding 6 is unrelated to this branch and is tracked as [#563](https://github.com/halos-org/skip/issues/563).

Issue #521, opened 2026-08-09
Extract the widget label-sizing technique; four copies have already drifted


mairas:
Four widget stylesheets carry byte-identical copies of the `.label` rule, the `@supports not (container-type: size)` fallback and its comment: widget-autopilot, widget-race-timer, widget-racer-line, widget-racer-timer.

They have already diverged within a single PR (#516): `$label-height` is `45cqh` in three files and `40cqh` in the autopilot with no comment explaining why, and the `@supports` block lists two selectors in the autopilot and one in the other three. Two sibling comments assert **opposite** conclusions about which axis binds — "width is the binding axis on all ordinary tiles" versus "height is what actually binds on an ordinary tile".

The coefficients themselves (`26cqw`, `18cqw`, `11cqw`, …) name the label that drove them but not how to re-derive one, and `$label-word-width: 26cqw` carries a five-character label in one file and three-character labels in another, so the same variable name means a different margin per file.

Fix direction: a `src/styles-common/_label-sizing.scss` partial exposing a mixin that takes the width coefficient and emits the `.label` base plus the fallback; each widget `@use`s it and declares only its own coefficients. `src/styles-common/` already exists. Record the derivation — character count and per-character width factor — beside each coefficient so a new label can be sized arithmetically instead of by re-measuring in a browser.

Found in review of #516.

Issue #360, opened 2026-07-19
Electrical widgets: layout-constant static-capture renders NaN geometry in combined test bundles


mairas:
## Summary

The electrical widgets capture the shared layout scalars from `widgets/shared/electrical-card-layout.constants.ts` into **class-static fields**:

```ts
private static readonly VIEWBOX_WIDTH = ELECTRICAL_DIRECT_CARD_VIEWBOX_WIDTH;
private static readonly CARD_HEIGHT   = ELECTRICAL_DIRECT_CARD_HEIGHT;
// ...CARD_GAP, COMPACT_CARD_HEIGHT
```

When **two or more electrical widget specs are bundled into one vitest/esbuild test run**, these statics read `undefined`, so the widget renders `viewBox="0 0 undefined NaN"` and NaN card geometry. Each spec **alone** renders correctly. The **production app is unaffected** — it is a single deterministic bundle that orders the constants first (verified on-device in Stage 2).

## Reproduction

```
# each alone: correct geometry
ng test --include='.../widget-charger.component.spec.ts'

# bundled: a viewBox probe prints  "0 0 undefined NaN"
ng test --include='.../widget-charger...' --include='.../widget-alternator...' \
        --include='.../widget-inverter...' --include='.../widget-ac...'
```

Originally surfaced as three trio golden-DOM snapshots failing when run together; confirmed on the still-static-capturing `widget-charger` with a temporary `viewBox` probe (alone → well-formed; bundled with the trio → `0 0 undefined NaN`).

## Root cause

Not an ESM import cycle — `electrical-card-layout.constants.ts` is a dependency-free leaf and cannot be in one. The trigger is **esbuild bundle-linearization order**: with cycles elsewhere in the app graph, esbuild does not guarantee the leaf constants module's assignments are emitted before a consumer's **class-static field capture**, so the capture reads `undefined`. Reading the same constant at **draw/call time** is always correct — which is why the layout *object* (read in `render()`) was always fine even in the NaN'd output.

## Fix

Read `ELECTRICAL_DIRECT_CARD_*` at draw/call time; drop the static captures.

## Status

- **Trio (alternator / inverter / ac): FIXED** in Stage 3a (`refactor/4-electrical-direct-card-draw`) — the shared `drawDirectCards()` / `initDirectCardSvg()` read the constants at draw time; golden-DOM snapshots now pass in a combined run.
[...]

Issue #30, opened 2026-06-30
[ANG-03] WidgetRuntimeDirective.options computed has side effects: dead output emission and mutable fields


mairas:
**Severity:** low | **Confidence:** confirmed | **Area:** angular | **Source:** codebase review 2026-06-30, finding `ANG-03`

## Problem

The `options` computed in `WidgetRuntimeDirective` performs side effects inside its body, violating Angular's requirement that computeds be pure. Specifically, it writes three memoization fields (`this.lastBaseRef`, `this.lastUserRef`, `this.lastMergedRef`) and calls `this.runtimeConfig.emit(merged)` on every evaluation. Because a computed is lazily evaluated and memoized, the emission only fires when `options()` is first read after an input change and not at all if no consumer reads it — making the emission timing non-deterministic. A repo-wide grep confirms that `runtimeConfig` has no template binding and no TypeScript consumer; the output declaration and its single `.emit` call are entirely dead code. The reference-stability caching itself (keeping object identity stable across merges) is a legitimate technique, but it should not co-host an output emission.

## Impact

Emitting an Angular output from inside a reactive computation can cause subtle ordering bugs and surprising test behaviour: the emission fires during a reactive read rather than as a discrete event, and it silently drops if the signal happens not to be read in a given change-detection cycle. While the dead-code nature of `runtimeConfig` means no current feature is broken, the pattern is a risk surface if the output is wired up in the future without realising it carries this behaviour. Removing or relocating the emission cleans up dead code and makes the computed safe to reason about.

## Affected code

- `src/app/core/directives/widget-runtime.directive.ts:32-54`

## Suggested direction

Remove the `runtimeConfig` output entirely — it has no consumers. If an emission is ever needed in the future, move it into a dedicated `effect()` that reads `options()` so the emission happens as an explicit side-effect step, not inside a computed. The reference-stability caching (lines 24-26, 36-38, 49-51) can stay, but the computed body should contain no observable side effects.

## Verification

Verified against the codebase: all factual sub-claims (output declaration, emit call, absence of consumers) confirmed in `src/app/core/directives/widget-runtime.directive.ts` (confidence: confirmed).

mairas:
[...]

PR #530, opened 2026-08-10
Widgets with no data show the instrument, not a blank card

mairas:
Widgets whose Signal K path never reports went blank or near-blank. A radial gauge on an
absent RPM path drew nothing at all — a white (or black) rectangle where an instrument
should be. This makes every affected widget degrade to something a helm can read: the
instrument is still there, and the missing reading says `--`.

## Why the gauges were blank

`gaugePath` sets `suppressBootstrapNull: true`, so the stream never emits for a path with
no data, so `value()` never left `undefined`, so the `@if (shouldRenderGauge())` wrapping
the entire template never became true. The three ng-canvas gauges (radial, linear,
compass) now mount as soon as config and theme are known and convey "no reading" by
suppressing the needle and progress bar, with `--` in the value box.

One trap worth recording: the linear gauge cannot express that by turning `barProgress`
off. canvas-gauges computes the bar geometry (`barDimensions`) inside the progress-bar
pass, and every later draw step reads it, so `barProgress: false` at construction makes
`ngOnInit` throw `Cannot read properties of undefined (reading 'isVertical')` and the
gauge renders as a dead canvas — with data as well as without. The bar is painted
transparent instead.

## The rest

- **Position** printed an empty string per coordinate, so an unfed widget was its label
  and nothing else. Now `--`, bounded to the size a real coordinate gets (`drawText`
  otherwise sizes two dashes to fill the card).
- **Autopilot** gated every control row on a real mode, so an off-line or unconfigured
  pilot collapsed to two buttons over an opaque panel. The rows stay laid out and
  disabled; the message sits on a bordered plate over a transparent layer so the AP screen
  reads through; heading, wind angle and target stay `null` rather than being coerced to
  `0`, so the screen shows `--` instead of a fabricated north. The message also used
  `--mat-sys-error-container` — a *background* token, pale pink — as text colour directly
  on the grey overlay; it now uses `--mat-sys-error` on the plate (measured ≈8.9:1 dark,
  ≈5.3:1 light). Engage is disabled while an error is showing: the v2 branch returned
  "enabled" unconditionally, leaving a live Engage button on a widget whose own message
  said it was not configured.
- **All six electrical widgets** (solar, BMS, alternator, inverter, AC, charger) covered
  themselves with a 92%-opaque scrim over an empty SVG. Each now draws one dimmed all-`--`
  card behind a translucent message plate. Alternator, inverter and AC share
  `drawDirectCards`, so that placeholder lives there once; solar, BMS and charger have
  their own renders and their own.

## Verification

Reproduced and re-checked against the perf-harness mock with streaming off, plus a
streaming run to confirm nothing regressed when data *is* present — that second run is
what caught the linear-gauge crash above. `npm run ci` passes at 1938 tests; nine are new,
covering the no-data render for the radial gauge, the position placeholder, the autopilot
control rows and engage guard, and the placeholder card in each of the six electrical
widgets. Deployed to a test device and exercised there.

The full-suite run intermittently fails `minichart.component.spec.ts`; that reproduces on
`main` with this branch stashed and passes in isolation. Issue #469 covers it — not
addressed here.

### DIFF
--- a/src/app/widgets/shared/electrical-card-layout.constants.ts
+++ b/src/app/widgets/shared/electrical-card-layout.constants.ts
@@ -57,3 +57,9 @@ export const ELECTRICAL_DIRECT_CARD_FULL_LAYOUT: IElectricalDirectCardLayout = {
 export const ELECTRICAL_DIRECT_CARD_COMPACT_LAYOUT: IElectricalDirectCardLayout = {
   ...ELECTRICAL_DIRECT_CARD_FULL_LAYOUT
 };
+
+/**
+ * Opacity of the all-'--' card the electrical widgets draw while nothing has reported. One value
+ * so the family's empty state cannot render at two intensities.
+ */
+export const ELECTRICAL_NO_DATA_OPACITY = 0.6;
--- a/src/app/widgets/shared/electrical-direct-card-draw.ts
+++ b/src/app/widgets/shared/electrical-direct-card-draw.ts
@@ -6,8 +6,7 @@ import {
   ELECTRICAL_DIRECT_CARD_GAP,
   ELECTRICAL_DIRECT_CARD_HEIGHT,
   ELECTRICAL_DIRECT_CARD_VIEWBOX_WIDTH,
-  ELECTRICAL_DIRECT_COMPACT_CARD_HEIGHT
-} from './electrical-card-layout.constants';
+  ELECTRICAL_DIRECT_COMPACT_CARD_HEIGHT, ELECTRICAL_NO_DATA_OPACITY } from './electrical-card-layout.constants';
 
 /**
  * The per-card display fields the shared direct-card draw reads. The trio's own
@@ -54,7 +53,6 @@ interface DirectCard<TEntity> {
 }
 
 const PLACEHOLDER_KEY = '__no-electrical-data__';
-const PLACEHOLDER_OPACITY = 0.6;
 
 /** Dimmed all-'--' card drawn while nothing has reported, so the widget keeps its card shape. */
 const PLACEHOLDER_MODEL: DirectCardDisplayModel = {
@@ -142,7 +140,7 @@ export function drawDirectCards<TEntity extends IElectricalTopologySnapshotCore>
   const merged = enter.merge(selection as Selection<SVGGElement, DirectCard<TEntity>, SVGGElement, unknown>);
 
   merged.attr('transform', item => `translate(0, ${item.y})`);
-  merged.attr('opacity', isPlaceholder ? PLACEHOLDER_OPACITY : null);
+  merged.attr('opacity', isPlaceholder ? ELECTRICAL_NO_DATA_OPACITY : null);
 
   if (includeCardBg) {
     merged.select(`rect.${classPrefix}-card-bg`)
--- a/src/app/widgets/shared/electrical-direct-card-draw.ts
+++ b/src/app/widgets/shared/electrical-direct-card-draw.ts
@@ -53,6 +53,24 @@ interface DirectCard<TEntity> {
   y: number;
 }
 
+const PLACEHOLDER_KEY = '__no-electrical-data__';
+const PLACEHOLDER_OPACITY = 0.6;
+
+/** Dimmed all-'--' card drawn while nothing has reported, so the widget keeps its card shape. */
+const PLACEHOLDER_MODEL: DirectCardDisplayModel = {
+  id: '',
+  titleText: '',
+  modeText: '--',
+  busText: '--',
+  metricsLineOne: '--',
+  metricsLineTwo: '--',
+  stateBarColor: 'var(--skip-contrast-dim-color)',
+  titleTextColor: 'var(--skip-contrast-dim-color)',
+  metaTextColor: 'var(--skip-contrast-dim-color)',
+  primaryMetricsTextColor: 'var(--skip-contrast-dim-color)',
+  secondaryMetricsTextColor: 'var(--skip-contrast-dim-color)'
+};
+
 /**
  * Select the direct-card `<svg>`, seed the shared viewBox/role/aria-label, and
  * append the stacked-card layer group. Reads the layout constants at call time
@@ -89,11 +107,18 @@ export function drawDirectCards<TEntity extends IElectricalTopologySnapshotCore>
 
   const layout = compact ? ELECTRICAL_DIRECT_CARD_COMPACT_LAYOUT : ELECTRICAL_DIRECT_CARD_FULL_LAYOUT;
   const cardHeight = compact ? ELECTRICAL_DIRECT_COMPACT_CARD_HEIGHT : ELECTRICAL_DIRECT_CARD_HEIGHT;
-  const cards: DirectCard<TEntity>[] = entities.map((entity, index) => ({
-    key: entity.deviceKey ?? entity.id,
-    entity,
-    y: index * (cardHeight + ELECTRICAL_DIRECT_CARD_GAP)
-  }));
+  // With nothing to show, draw one all-'--' card so the widget keeps its shape behind the
+  // empty-state message instead of collapsing to a bare panel and a sentence. The synthetic
+  // entity only ever reaches the id text, which the placeholder model blanks anyway.
+  const isPlaceholder = entities.length === 0;
+  const models = isPlaceholder ? { [PLACEHOLDER_KEY]: PLACEHOLDER_MODEL } : displayModels;
+  const cards: DirectCard<TEntity>[] = isPlaceholder
+    ? [{ key: PLACEHOLDER_KEY, entity: { id: '' } as TEntity, y: 0 }]
+    : entities.map((entity, index) => ({
+        key: entity.deviceKey ?? entity.id,
+        entity,
+        y: index * (cardHeight + ELECTRICAL_DIRECT_CARD_GAP)
+      }));
 
   const contentHeight = cards.length ? cards[cards.length - 1].y + cardHeight : cardHeight;
   svg.attr('viewBox', `0 0 ${ELECTRICAL_DIRECT_CARD_VIEWBOX_WIDTH} ${contentHeight}`);
@@ -117,6 +142,7 @@ export function drawDirectCards<TEntity extends IElectricalTopologySnapshotCore>
   const merged = enter.merge(selection as Selection<SVGGElement, DirectCard<TEntity>, SVGGElement, unknown>);
 
   merged.attr('transform', item => `translate(0, ${item.y})`);
+  merged.attr('opacity', isPlaceholder ? PLACEHOLDER_OPACITY : null);
 
   if (includeCardBg) {
     merged.select(`rect.${classPrefix}-card-bg`)
@@ -138,15 +164,15 @@ export function drawDirectCards<TEntity extends IElectricalTopologySnapshotCore>
     .attr('ry', layout.stateBarCornerRadius)
     .attr('width', 3)
     .attr('height', cardHeight - 3)
-    .attr('fill', item => displayModels[item.key]?.stateBarColor ?? widgetColors.dim);
+    .attr('fill', item => models[item.key]?.stateBarColor ?? widgetColors.dim);
 
   if (entities.length > 1) {
     merged.select(`text.${classPrefix}-title`)
       .attr('x', layout.titleX)
       .attr('y', layout.titleY)
       .attr('font-size', layout.titleFontSize)
-      .attr('fill', item => displayModels[item.key]?.titleTextColor ?? 'var(--skip-contrast-color)')
-      .text(item => displayModels[item.key]?.titleText ?? titleFallback(item.entity));
+      .attr('fill', item => models[item.key]?.titleTextColor ?? 'var(--skip-contrast-color)')
+      .text(item => models[item.key]?.titleText ?? titleFallback(item.entity));
   } else {
     merged.select(`text.${classPrefix}-title`).text('');
   }
@@ -164,8 +190,8 @@ export function drawDirectCards<TEntity extends IElectricalTopologySnapshotCore>
     .attr('y', layout.metaY)
     .attr('font-size', layout.metaFontSize)
     .attr('opacity', 0.8)
-    .attr('fill', item => displayModels[item.key]?.metaTextColor ?? 'var(--skip-contrast-dim-color)')
-    .text(item => displayModels[item.key]?.modeText ?? '');
+    .attr('fill', item => models[item.key]?.metaTextColor ?? 'var(--skip-contrast-dim-color)')
+    .text(item => models[item.key]?.modeText ?? '');
 
   merged.select(`text.${classPrefix}-bus`)
     .attr('x', layout.metaRightX)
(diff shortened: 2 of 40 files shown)


mairas:
## Code review — PR #530

**Scope:** `83e5da3c..HEAD`, 36 files, 733 insertions / 175 deletions.
**Intent:** every widget whose Signal K path reports nothing degrades to a readable instrument — layout still drawn, missing reading as `--` — with no behaviour change when data is present.
**Reviewers:** correctness, testing, maintainability, project-standards (always-on); reliability (diff reworks error-overlay handling and the gauge failure path); adversarial (diff far over the 50-line threshold).

The no-data rendering does what it claims. What the reviewers converged on is the other half of the intent: **three control-path and live-data regressions that only appear when data IS present**, two of them on the autopilot.

### P0 — Critical

None.

### P1 — High

|---|---|---|---|---|---|
| 1 | `widget-autopilot.component.ts:321` | **A failed command locks out Disengage for 6s.** `apEngageBtnDisabled` gates the engage/disengage *toggle*, not an engage-only control. Any non-200 PUT or failed v2 REST call shows the error overlay for `ERROR_DISPLAY_DURATION` (6000 ms), during which the helm cannot take a live pilot to standby — including after a failed Disengage, which is exactly when it matters. Each further failing press re-arms the timer, and `setPersistentError` holds it open for the widget's lifetime. V1 has no standby entry in the mode menu, so there is no fallback. | correctness, adversarial | 0.95 | manual → human |
| 2 | `widget-autopilot.component.ts:326` | **Engage stays live while the pilot is off-line.** The `apModeKnown()` gate was added to `apBtnDisabled` but not to `apEngageBtnDisabled`. Under v1, `apState` is deliberately `null`, so the `['standby','off-line']` check is always false; the v2 branch returns `false` unconditionally. A mode reported `off-line` raises no error overlay, so neither new gate fires — the helm can press Disengage on a pilot the widget already knows is unreachable, see nothing happen, and believe it is off. | reliability | 0.85 | gated_auto → human |
| 3 | `svg-autopilot.component.ts:96` | **Heading dropout now freezes the dial on a stale reading.** Removing the `?? 0` coercion made `null` reachable, but the effect early-returns on null and nothing ever sets `compassAngle` back — so `{{ compassAngle() ?? '--' }}` only ever shows `--` before the *first* heading. After that, a dead compass renders identically to a live one. This is a regression in kind: previously the dial snapped to north (visibly wrong), now it holds a plausible number. Same shape for `appWindAngle` and the rudder rects. | reliability, adversarial | 0.95 | gated_auto → human |

Findings 1 and 2 are one design decision, not two patches: the gate should key on reachability (`apModeKnown()`) and on direction (engage vs disengage), not on a 6-second display timer. `errorActive` is a message-visibility signal being used as a reachability signal.

### P2 — Moderate

|---|---|---|---|---|---|
| 4 | `widget-gauge-ng-linear.component.ts:233` | **The no-data effect applies zone colours even with `ignoreZones: true`.** It calls `zoneColor(...currentState())` without the guard the sibling state effect has (`if (... cfg.ignoreZones) return`), so the first datapoint in an alarm zone paints the bar red and the effect that would correct it never runs. Live-data behaviour change. | correctness, adversarial | 0.92 | safe_auto → review-fixer |
| 5 | `widget-ac/-alternator/-inverter/-charger .scss:22` | **Empty-state message paints *under* the placeholder card.** `.X-svg` carries `z-index: 20`; `.X-empty` has none, so it resolves to `auto`. Only `.solar-empty` got `z-index: 30` in this diff, and BMS is saved by DOM order. Invisible before the diff because the SVG drew nothing when empty. Verified: 4 of 6 stylesheets are missing it. | correctness | 0.85 | safe_auto → review-fixer |
| 6 | three gauge specs | **The linear and compass no-data paths have no tests.** The compass has no spec file at all despite getting the same rework; the linear spec exists and was not touched. The linear's transparent-bar workaround — the one thing the PR body calls a trap — has nothing pinning it, so a future "cleanup" to `barProgress: false` reproduces the dead canvas with CI green. | testing ×2, reliability, project-standards, adversarial | 0.95 | safe_auto → review-fixer |
| 7 | `svg-autopilot.component.ts:159` | **Wind mode still prints a fabricated `0°`.** The diff's own comment claims a missing wind angle now shows `--`, but the mode-value effect reinstates `: 0` one component downstream, so a wind-hold pilot with a dead vane displays a bare `0°`. The PR description is wrong on this point. | reliability, adversarial | 0.90 | gated_auto → downstream-resolver |
| 8 | `widget-gauge-ng-{radial,linear,compass}` | **`dataAvailable` latches true when the gauge path is reconfigured.** It is written only inside the stream callback, and all three set `suppressBootstrapNull`, so a rebuilt subscription on a silent path never fires it. Re-point a gauge at a dead path from the options dialog and it keeps the old needle position and number, presented as a live reading of the new path. | correctness, reliability | 0.85 | manual → downstream-resolver |
| 9 | six `.scss` files | **Six byte-identical `.X-empty-panel` blocks** (plus a seventh near-copy in `.ap-error-panel`). A theme tweak has to land seven times. `src/styles-common/pathWarningHint.scss` is the existing precedent for exactly this, and its comment gives the reason. Issue [#521](https://github.com/halos-org/skip/issues/521) already tracks four drifted copies of another such technique. | maintainability, project-standards | 0.95 | gated_auto → review-fixer |
| 10 | four render modules | **`PLACEHOLDER_OPACITY = 0.6` declared four times**, plus six hard-coded `'0.6'` spec assertions. Ten sites for one visual decision; all four modules already import from `electrical-card-layout.constants.ts`. | maintainability, project-standards | 0.90 | safe_auto → review-fixer |
| 11 | `electrical-direct-card-draw.ts:116`, `widget-charger.component.ts:508` | **Synthetic placeholder entity typed by assertion.** `{ id: '' } as TEntity` claims a full snapshot from one field; any future field read on `item.entity` compiles and returns `undefined` on the placeholder path only. The charger copy already flows into `resolveShapeMode(item.charger)` and survives only because that function happens to `?? ''` both fields it touches. | maintainability, project-standards | 0.85 | gated_auto → review-fixer |
| 12 | `widget-autopilot.component.ts:770` | **The null heading/wind/target passthrough has no test.** Four handlers dropped their coercion; restoring any `?? 0` puts a fabricated reading on a marine autopilot display and CI stays green. | testing | 0.85 | safe_auto → review-fixer |
| 13 | `widget-autopilot.component.spec.ts:130` | **The blank-widget fix itself is untested.** The widget went blank because `[style.display]="apGrid()"` collapsed the grid container. The new test asserts `adjustHdgBtnVisibility()`/`tackBtnVisibility()` — different signals that never fed that binding. Re-introducing any gate on `.autopilot-grid-container` leaves all three new tests green. A fixture-based `render()`/`controls()` harness already exists in the file. | testing | 0.80 | safe_auto → review-fixer |
| 14 | `widget-bms.component.ts:719` | **Message and placeholder use non-equivalent empty conditions.** Plate on `!hasBatteries()` (visible batteries), placeholder on `banks.length === 0 && batteries.length === 0`. Verified reachable: `bankSummaries` maps over configured banks unconditionally while `visibleBatteries` filters to ids present in the data map, so a configured bank awaiting data puts "No batteries detected" over full-opacity bank cards. | maintainability, correctness | 0.72 | gated_auto → review-fixer |
| 15 | `widget-bms.component.html:2` | **BMS loses full-bleed layout when batteries have no bank.** Dropping the `hasBanks()` title gate and the `bms-svg--full` modifier changes a live, data-present configuration: batteries reported without a bank now render 15% shorter with a title row that was previously suppressed. The placeholder work did not need to touch that case. | adversarial | 0.70 | manual → human |
| 16 | three gauges | **Silent catch hides failed needle suppression.** The suppression effects are the only thing keeping an unfed gauge from showing an indicator at the scale minimum, and their `update()` is `catch { /* ignore */ }`. canvas-gauges rethrows real errors, so a future option combination leaves a needle up next to a `--` readout with nothing in the console — the same silent mode as the crash this PR works around. | reliability | 0.72 | safe_auto → review-fixer |
| 17 | three gauges | **The needle/bar suppression rule is written 2–3 times per gauge**, split between `buildGaugeOptions` and the new effect. Both sites are load-bearing (options rebuild on theme change, effect on data change), so missing the `buildGaugeOptions` copy resurrects the needle on an unfed gauge after a theme switch. | maintainability | 0.75 | safe_auto → review-fixer |
| 18 | PR body | **Contrast ratios quoted with no rendered images.** Every claim here is about what a widget looks like; the body cites "≈8.9:1 dark, ≈5.3:1 light" and "92%-opaque scrim" with zero image embeds, so none of it is auditable. | project-standards | 0.78 | manual → human |

### P3 — Low

|---|---|---|---|---|---|
| 19 | `widget-bms.component.ts:295` | `hasBanks` is dead — this diff removed both readers. Verified: one occurrence, its own declaration. Interacts with #15; restoring the modifier gives it a reader again. | maintainability | 0.90 | safe_auto → review-fixer |
| 20 | (git history) | All six commit subjects run 59–67 chars against the repo's 50-char cap. Type, scope and mood are correct; length is the only violation. | project-standards | 0.90 | manual → human |
| 21 | `widget-position.component.spec.ts:144` | Test renamed from "blanks both coordinates" to "shows a placeholder" but its assertion is unchanged — it only checks no drawn string contains `:`. Reverting `formatCoordinate` to `''` keeps it green. Its three siblings all got a real `'--'` assertion. | testing | 0.90 | safe_auto → review-fixer |
| 22 | `widget-bms.component.spec.ts:288` | The new test was inserted between the `Guards #360` comment and the test it explains, so the next reader gets the wrong rationale and the #360 note looks deletable. | project-standards | 0.88 | safe_auto → review-fixer |
| 23 | `widget-gauge-ng-linear.component.ts:309` | The `barProgress` comment — the only record of why the linear hides its bar by transparency while the radial toggles the flag — has an unparseable middle clause ("computes the bar geometry every other draw step reads (barDimensions) inside the progress-bar pass"). | maintainability | 0.80 | safe_auto → review-fixer |
| 24 | `widget-autopilot.component.scss:58` | The countdown/confirmation overlay shares `.ap-screen-overlay` and lost its opaque backing when that class went translucent for the error case. It has no plate of its own, so the confirmation prompt now reads over a rotating compass. | correctness | 0.75 | safe_auto → review-fixer |
| 25 | `widget-position.component.ts:154` | `valueTextHeight` exists only to bound the `--` placeholder's size; the spec inspects only `drawText` argument 1, so deleting the helper or its guard breaks nothing. | testing | 0.80 | safe_auto → review-fixer |
| 26 | `widget-autopilot.component.ts:312` | `errorActive` reads the literal string `'visible'` out of a signal that exists to drive `[style.visibility]`. Converting the overlay to `@if` or a class binding silently re-enables the engage gate. Invert it: make `errorActive` the writable signal and derive the visibility string. | maintainability | 0.70 | gated_auto → review-fixer |
| 27 | `widget-gauge-ng-linear.component.ts:163` | `textValue` is `dataAvailable ? '' : '--'` expressed as a signal with two setters and a read-your-own-write guard, while the radial expresses the same rule as a one-line `computed` in this same PR. | maintainability | 0.70 | safe_auto → review-fixer |
| 28 | `widget-gauge-ng-linear.component.ts:230,268` | `const opt: LinearGaugeOptions = {...} as LinearGaugeOptions` annotates and asserts the same literal; the assertion disables excess-property checking, so a misspelled gauge option compiles into a silent no-op. The pre-image had the annotation alone. | maintainability | 0.66 | safe_auto → review-fixer |
| 29 | `widget-charger.component.ts:508` | `as ChargerSnapshot` is redundant — `{ id: '' }` already satisfies it structurally — and it silences the compiler if the type later gains a required field. (The sibling `as TEntity` is genuinely unavoidable against an unresolved generic.) | project-standards | 0.62 | safe_auto → review-fixer |
| 30 | PR body | `#469` written bare rather than as a markdown link to the full URL. The surrounding wording correctly avoids a closing keyword — `closingIssuesReferences` is `[]`. | project-standards | 0.70 | safe_auto → review-fixer |

| File | Issue | Route |
|---|---|---|
| `widget-autopilot.component.ts:363` | `adjustHdgBtnVisibility` and `tackBtnVisibility` are character-for-character identical, including a duplicated six-mode allowlist. This PR had to add the same guard line to both — the drift risk demonstrating itself. | **fix-here** — trivial, and the diff already touches both |

### Coverage

- **The gauge test shim has no `update()` method.** `vitest.config.ts` aliases `@godind/ng-canvas-gauges` to a shim exposing only `options` and `value`, so every `this.ngGauge()?.update(...)` throws `update is not a function` in tests and is swallowed by the surrounding catch. Needle suppression, `barProgress`, and the transparent-bar colour are all unobservable under test — the new radial tests assert signal state only. A misspelled option key or a null option value passes CI. This is the single largest gap behind findings 4, 6, 16 and 17.
- **Untracked file excluded from scope:** `perf-harness/mock-serve.mjs` (pre-existing scratch, not part of this branch).
- **Residual risks:** every `dataAvailable` toggle calls `update()` → `redraw()` → `SmartCanvas.init()`, forcing a full static-layer redraw (plate, borders, ticks, numbers); on a flapping path with `enableTimeout` on that is a full redraw per TTL cycle on the Pi-class target, with no perf-harness numbers attached. `BaseGauge.update()` calls `.toString()` per property, so one `undefined` value throws mid-loop and silently drops every remaining option in that push. The radial pushes width/height with no positive-size guard, while the linear added `updateGaugeSize()` precisely because the library throws `IndexSizeError` from inside its own MutationObserver on a non-positive size — the radial is one zero-height layout pass from the same failure. The linear's transparent-bar workaround depends on canvas-gauges' internal draw order; a library upgrade that moves `barDimensions` re-breaks it silently.
- **Confirmed compliant:** VERSION 1.3.1 against latest stable v1.3.0+2 — the cycle is already open at patch level and this is patch work, so no re-bump is due. `debian/changelog` policy does not apply (npm-only, no `debian/`). No closing-keyword bindings (`closingIssuesReferences` is `[]`).

---

### Verdict: **Not ready**

The no-data work itself is sound and the diagnosis in the PR body holds up. What blocks it is the second half of the stated intent — "must not change behaviour when data IS present" — which three findings contradict outright: the Disengage lockout (#1), the frozen heading (#3), and the linear gauge's zone colouring under `ignoreZones` (#4). The BMS layout change (#15) is a fourth, milder instance.

Suggested order:

1. **#1 + #2 together** — one decision about what the engage/disengage gate keys on. This is a safety control on a marine autopilot; it needs a deliberate answer, not a patch.
2. **#3, #7** — return `compassAngle`/`appWindAngle` to null on dropout and stop coercing the wind-hold readout, so `--` means "no data" everywhere rather than "no data yet".
3. **#4, #5** — two-line fixes, both live-behaviour or visible-layering.
4. **#6, #12, #13** — the tests that would have caught #1, #3 and #4.
5. The consolidation set (#9, #10, #17) and the P3s at your discretion.

mairas:
## Review findings addressed

Pushed as `dddae60d`, `b7151878`, `c46d4ed4`, `95c70232`. `npm run ci` passes at **1952 tests / 178 files** (was 1938 / 176).

### P1

**#1 + #2 — the engage/disengage gate.** Resolved as one decision rather than two patches: a v2 REST API that answers while its delta paths are silent is not a real state, so **reachability is the only gate** and it now covers every control including the Mode button. `errorActive` is gone. Consequences, both intended:

- A refused command no longer disables anything. The Disengage lockout is fixed at the root — the overlay already reports the failure, and locking the button that takes a misbehaving pilot off was the wrong response to it.
- An off-line or unknown mode disables the whole widget, which also fixes the original complaint (a live Engage button next to "500 - Autopilot widget not configured") by a better mechanism than the error gate it replaces.

Verified in a browser against the mock: 9 of 9 buttons disabled, Mode and Disengage included. New tests cover disabled-when-unknown, enabled-when-known, and **still enabled after a refused command with the label reading Disengage** — the case that was missing.

**#3 — frozen heading.** `compassAngle` now returns to null on a null or non-finite heading and clears `compassInitialized`, so the next real heading lands without sweeping from a stale angle. This was the sharpest finding in the review: dropping the `?? 0` coercion made `--` reachable at boot but not after, trading a visibly-wrong reading (dial snapping north) for a plausible one (dial frozen). New `svg-autopilot.component.spec.ts` — the component had no spec at all — covers absent-at-boot, dropout-returns-to-absent, and re-acquire.

### P2

- **#7** wind-hold prints `--` instead of a fabricated `0°`. The PR body's claim was true of the widget signal but not of the SVG readout; both now agree.
- **#4** linear gauge honours `ignoreZones`. The fix drove a small refactor: the colour decision was unobservable in tests (the gauge shim has no `update()`), so it is now one `barColor()` used by the no-data effect, the zone-state effect and the option builder — which also collapses finding **#17**'s duplicated suppression rule. Tested directly.
- **#5** `z-index: 30` on ac, alternator, inverter and charger. Verified in the browser: message plate at 30 over the card SVG at 20.
- **#15** BMS full-bleed layout restored, keyed `hasBatteries() && !hasBanks()` so only the true empty state takes the inset layout. This gives **#19**'s dead `hasBanks` a reader again.
- **#14** BMS message and placeholder now share one `nothingToShow()` predicate, so the plate cannot land over live bank cards.
- **#6** the gap five reviewers converged on: `widget-gauge-ng-compass.component.spec.ts` created (no spec existed), and the linear spec gained a no-data block that pins `barProgress: true` + transparent colour — the trap the PR body describes and nothing enforced.
- **#13** the blank-widget fix is now tested at the DOM level: grid container present, not `display: none`, all controls disabled. The previous test asserted row-visibility signals that never fed that binding.
- **#10** one `ELECTRICAL_NO_DATA_OPACITY` in the shared constants replaces four copies.
- **#21, #22, #23, #24, #28, #29** applied.

### Not addressed

- **#9** six duplicated `.X-empty-panel` SCSS blocks. Real, and it doubles a surface that was already duplicated six ways before this PR. It wants a global class alongside `styles-common/pathWarningHint.scss` and touches all six widgets plus the autopilot's near-copy — a consolidation better done on its own, not folded into a bugfix branch.
- **#8** `dataAvailable` latching across a path reconfigure. Reachable only by re-pointing a gauge at a dead path from the options dialog. The safe reset is conditional on the configured path actually changing, since an unconditional one blinks the needle off on every theme change.
- **#16** logging the swallowed `update()` errors, **#26** inverting `errorActive`/visibility (moot — `errorActive` is deleted), **#11** typing the synthetic entity as nullable, **#20** the over-length commit subjects, **#25**, **#27**, **#30**.
- **#18** screenshots: see below.

### Screenshots

The `gh` CLI has no supported way to upload images to a PR comment, so these have to be dragged in through the web UI. They are on the review machine at `.playwright-mcp/` in the workspace root:

| File | Shows |
|---|---|
| `nodata-before.png` / `nodata-after.png` | matched main-vs-branch pair, same dashboard and theme, streaming off |
| `electrical-family-after.png` | all six electrical widgets with no data |
| `ap-error-after.png` / `ap-error-light.png` | the autopilot error plate, dark and light — the source of the ≈8.9:1 and ≈5.3:1 figures |
| `review-fixes.png` | after this round: 9 of 9 autopilot controls disabled, message plate above the card |
| `withdata-after.png` | streaming run confirming the data-present path, which is what caught the linear-gauge crash |

The contrast figures were measured off the rendered page (`getComputedStyle` on the plate and text), not estimated.

Issue #13, opened 2026-06-30
[HF-07] History batch delivery accesses undocumented RxJS ReplaySubject._buffer internal


mairas:
**Severity:** medium | **Confidence:** confirmed | **Area:** history | **Source:** codebase review 2026-06-30, finding `HF-07`

## Problem

`getDatasetBatchThenLiveObservable()` in `dataset-stream.service.ts` snapshots the internal `_buffer` field of an RxJS `ReplaySubject` using a cast through `any` (`(subject as any)._buffer`), emits the captured history-seeded datapoints as a single batch, and then calls `skip(buffer.length)` on the live observable to avoid re-emitting those points. The `_buffer` field is an undocumented RxJS private implementation detail and is not part of any public API. An RxJS major-version upgrade that renames or restructures this field would hit the fallback (`? ... : []`) and silently produce an empty history batch — wiping history from all chart widgets without any build-time or runtime error. Additionally, because the buffer length is snapshotted synchronously before `concat` subscribes to the live stream, new values that arrive between snapshot and subscription can cause the `skip(buffer.length)` count to mismatch, leading to dropped or double-emitted datapoints.

## Impact

Every chart widget that consumes history-seeded data relies on this delivery path: `widget-data-chart`, `widget-windtrends-chart`, and `minichart` all call `getDatasetBatchThenLiveObservable()`. A broken seeding path silently renders historical context invisible in charts — particularly relevant for marine use where trend data (wind, speed, depth) is essential for situational awareness. The bug surface grows with each RxJS upgrade and is invisible in tests because it degrades to empty data rather than throwing.

## Affected code

- `src/app/core/services/dataset-stream.service.ts:583-603`

## Suggested direction

Remove the dependency on `_buffer` by tracking the history-seeded datapoints explicitly inside the service — for example, keeping an owned array alongside the `ReplaySubject` that is populated at the same time the subject is seeded from the History API response. Build the batch + live observable from that owned array and the subject's live emissions, eliminating the private-field access and the racy `skip(buffer.length)` entirely. A regression test asserting that the batch observable emits the seeded history points would prevent silent regressions on future RxJS upgrades.

## Verification

Verified against the codebase (confidence: confirmed): `dataset-stream.service.ts` lines 591–602 contain the exact `(subject as any)._buffer` access as described, the subject is a `ReplaySubject` seeded from the History API, and all three consumer widgets are confirmed.

mairas:
[...]

PR #538, opened 2026-08-11
Draw Realtime Data Plot scale ticks inside the plot area

### LINKED ISSUE #539: Data Plot: toggling Vertical Data Graph never rebuilds the chart
Found during the review of #538 (pre-existing, not introduced there).

`WidgetDataChartComponent.lastVerticalChart` is initialized to `null` and its only assignment sits inside the branch that requires it to already be non-null:

```
private lastVerticalChart: boolean | null | undefined = null;   // line 122
...
const verticalChanged = this.lastVerticalChart !== null && this.lastVerticalChart !== cfg.verticalChart;  // line 181
if (verticalChanged) {
  this.lastVerticalChart = cfg.verticalChart;   // line 183 — unreachable
  this.rebuildForDataset(cfg);
}
```

So `verticalChanged` is false for the component's whole lifetime, and `verticalChart` is not part of `pathSignature` either (line 154), so the other rebuild path does not fire on an orientation change.

## What breaks

Toggling **Vertical Data Graph** on a live widget swaps the two scale blocks in place via `setChartOptions()` + `update('none')` instead of recreating the chart:

- `transformDatasetRows()` picks `{x: value, y: timestamp}` or `{x: timestamp, y: value}` per row at push time, and already-buffered points are never re-mapped. Existing data plots transposed until the whole window scrolls out — up to 365 days on a day-scale config.
- The realtime scale changes `type` between the two blocks. Chart.js `buildOrUpdateScales` replaces a scale whose type changed and overwrites `scales[id]` without destroying the old one; `RealTimeScale.destroy()` is what cancels the streaming plugin's `setInterval` and `requestAnimationFrame` loop, so the orphaned scale keeps ticking against the live chart.

Config edits reach the widget in place (`widget-host2.component.ts` `applyRuntimeConfig` / `reconfigure`), so this is the normal path, not an edge case.

## Fix direction

Add `cfg.verticalChart` to `pathSignature` so the orientation goes through the same rebuild path as every other structural change, and drop the dead `lastVerticalChart` flag. Alternatively seed the flag in `rebuildForDataset()`. Either way it needs a spec: no test currently sets `verticalChart: true` for this widget, so both orientation branches of `setChartOptions()` are uncovered.

Raised independently by the correctness and adversarial reviewers on #538.

mairas:
## Why

Turning on either scale in the Realtime Data Plot cost a lot of the plot. The value axis reserved a
label gutter, and the time axis reserved a label row *plus* a title row for "Last 30 second" — on a
small tile that left the trace squeezed into a band. Separately, in the widget settings the Source
and "Angle display range" select boxes shared a border, because the angle field sits outside the
`.flex-container` blocks that carry the form's 10px rhythm and `subscriptSizing` is `dynamic`
app-wide, so a field with no hint reserves nothing below it.

## How

Chart.js 4.5.1 does all of this natively — no plugin, no custom draw pass:

- `ticks.mirror` draws the labels inside the grid **and** makes `Scale.fit()` reserve zero space for
  them. The docs call it vertical-only; the implementation applies it to horizontal axes too, and it
  works on the bottom time axis.
- Tick labels are layer `z=0`, grid is `z=-1`, and everything at `z<=0` draws before the datasets —
  so labels land above the grid and below the plot lines with no extra work.
- `ticks.textStrokeColor` / `textStrokeWidth` stroke behind the fill, giving the Numeric widget's
  card-coloured halo: it knocks out the grid line behind the glyphs without hiding the value line
  that crosses in front.
- `grid.drawTicks: false` drops the 8px tick-mark gutter.

The plot window moves off the axis title onto the widget label as a compact suffix — `SOG (30 s)`.
Legacy `TimeScaleFormat` members have no abbreviation, so a config still carrying one gets the bare
label rather than a wrong one.

## Verified

Rendered before/after in the perf-harness mock server across all three tile shapes (wide, vertical,
small); the grid now runs to the tile edge in each. `npm run lint`, `npm run snc` and the full suite
(178 files, 1952 tests) pass.

## Known, not introduced here

On narrow tiles the large value readout is drawn over the widget label — the `title` and `subtitle`
plugins are overlaid on one row (`subtitle.padding.top: -35`) with no width negotiation. That
already happened with the bare label; the `(30 s)` suffix makes it start on wider tiles too.

Two smaller rough edges from the mirrored layout: the last time label and the bottom value label can
nearly touch in the corner (two axes, no cross-axis collision avoidance), and a `position: 'right'`
value axis cannot be fully inset because its label anchor is `scale.left + ticks.padding`, which
pushes outward.

## Merged in from #543

Fixes #539 — toggling Vertical Data Graph never rebuilt the chart. `lastVerticalChart` was
initialized to `null` and assigned only inside the branch that required it non-null, and
`verticalChart` was absent from `pathSignature`, so an orientation change swapped the scale blocks
in place: buffered points kept the old x/y mapping and plotted transposed until the window scrolled
out, and the replaced realtime scale was dropped without `destroy()`, leaving its refresh interval
and rAF loop running. `cfg.verticalChart` now sits in `pathSignature` and the dead flag is gone.

That work landed here rather than in its own PR to `main` because it touches the same component and
was stacked on this branch.

### DIFF
--- a/src/app/widget-config/dataset-chart-options/dataset-chart-options.component.html
+++ b/src/app/widget-config/dataset-chart-options/dataset-chart-options.component.html
@@ -63,7 +63,7 @@
     </div>
   </div>
   @if (datachartAngleRange(); as angleRangeCtrl) {
-    <mat-form-field class="full-width">
+    <mat-form-field class="full-width" style="margin-top: 10px;">
       <mat-label>Angle display range</mat-label>
       <mat-select [formControl]="angleRangeCtrl" name="datachartAngleRange">
         <mat-option [value]="null">Automatic (based on path)</mat-option>
--- a/src/app/widgets/widget-data-chart/widget-data-chart.component.spec.ts
+++ b/src/app/widgets/widget-data-chart/widget-data-chart.component.spec.ts
@@ -29,6 +29,7 @@ vi.mock('chartjs-plugin-annotation', () => ({ default: {} }));
 vi.mock('chartjs-adapter-date-fns', () => ({}));
 vi.mock('@aziham/chartjs-plugin-streaming', () => ({ default: {} }));
 
+import { signal } from '@angular/core';
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { EMPTY, Subject } from 'rxjs';
 import { WidgetDataChartComponent } from './widget-data-chart.component';
@@ -53,13 +54,17 @@ const makeConfig = (overrides: Partial<IWidgetSvcConfig> = {}): IWidgetSvcConfig
 describe('WidgetDataChartComponent', () => {
   let fixture: ComponentFixture<WidgetDataChartComponent>;
 
-  const runtimeMock = { options: vi.fn() };
+  // A real signal, not a vi.fn: the component tracks runtime.options() inside computed/effect, so a
+  // config edit only reaches the rebuild path when the source is reactive — which is what the live
+  // WidgetRuntimeDirective provides.
+  const options = signal<IWidgetSvcConfig | undefined>(undefined);
+  const runtimeMock = { options };
   const historyMock = { getBackfillThenLive: vi.fn() };
   const unitsMock = { convertToUnit: (_unit: string, value: number) => value, getUnitDisplaySymbol: (measure: string) => measure, resolvePathMeasure: () => 'knots' };
   const canvasMock = { releaseCanvas: vi.fn() };
 
   const setup = async (config: IWidgetSvcConfig): Promise<void> => {
-    runtimeMock.options.mockReturnValue(config);
+    options.set(config);
 
     await TestBed.configureTestingModule({
       imports: [WidgetDataChartComponent],
@@ -83,7 +88,7 @@ describe('WidgetDataChartComponent', () => {
   beforeEach(() => {
     vi.spyOn(HTMLCanvasElement.prototype, 'getContext')
       .mockReturnValue({} as unknown as CanvasRenderingContext2D);
-    runtimeMock.options.mockReset();
+    options.set(undefined);
     historyMock.getBackfillThenLive.mockReset().mockReturnValue(EMPTY);
   });
 
@@ -192,6 +197,7 @@ describe('WidgetDataChartComponent', () => {
     }).subtitle;
 
   interface AxisState {
+    type?: string;
     ticks?: { mirror?: boolean; padding?: number; textStrokeColor?: string; textStrokeWidth?: number; color?: string };
     grid?: { display?: boolean; drawTicks?: boolean };
     title?: { display?: boolean };
@@ -243,6 +249,29 @@ describe('WidgetDataChartComponent', () => {
       expect(axis?.grid?.display).toBe(true);
       expect(axis?.grid?.drawTicks).toBe(false);
     }
+  });
+
+  it('rebuilds the chart when the orientation is toggled', async () => {
+    const emissions$ = new Subject<IDatasetServiceDatapoint>();
+    historyMock.getBackfillThenLive.mockReturnValue(emissions$);
+
+    await setup(makeConfig({ verticalChart: false, showTimeScale: true, showYScale: true }));
+    emissions$.next({ timestamp: 1000, data: { value: 5 } });
+
+    expect(readAxis('x')?.type).toBe('realtime');
+    expect(fixture.componentInstance.lineChartData.datasets[0].data.length).toBeGreaterThan(0);
+    const streamsBefore = historyMock.getBackfillThenLive.mock.calls.length;
+
+    options.set(makeConfig({ verticalChart: true, showTimeScale: true, showYScale: true }));
+    fixture.detectChanges();
+    await fixture.whenStable();
+
+    // Swapping the orientation swaps which axis carries time. transformDatasetRows only transposes
+    // rows as they arrive, so anything already buffered plots as garbage — up to a 365-day window —
+    // unless the toggle goes through the full rebuild rather than an in-place options update.
+    expect(readAxis('y')?.type).toBe('realtime');
+    expect(fixture.componentInstance.lineChartData.datasets[0].data.length).toBe(0);
+    expect(historyMock.getBackfillThenLive.mock.calls.length).toBeGreaterThan(streamsBefore);
   });
 
   it('keeps the per-axis tick colour when the shared inside-tick options are applied', async () => {
--- a/src/app/widgets/widget-data-chart/widget-data-chart.component.spec.ts
+++ b/src/app/widgets/widget-data-chart/widget-data-chart.component.spec.ts
@@ -185,6 +185,81 @@ describe('WidgetDataChartComponent', () => {
     expect(title).not.toContain('celsius');
   });
 
+  interface SubtitleState { display?: boolean; text?: string }
+  const readSubtitle = (): SubtitleState | undefined =>
+    (fixture.componentInstance.lineChartOptions.plugins as unknown as {
+      subtitle?: SubtitleState;
+    }).subtitle;
+
+  interface AxisState {
+    ticks?: { mirror?: boolean; padding?: number; textStrokeColor?: string; textStrokeWidth?: number; color?: string };
+    grid?: { display?: boolean; drawTicks?: boolean };
+    title?: { display?: boolean };
+  }
+  const readAxis = (id: 'x' | 'y'): AxisState | undefined =>
+    (fixture.componentInstance.lineChartOptions.scales as unknown as Record<string, AxisState>)[id];
+
+  it('states the plot window on the widget label, not on a time-axis title', async () => {
+    await setup(makeConfig({ displayName: 'SOG', timeScale: 'second', period: 30, showTimeScale: true }));
+
+    // The axis title cost the plot a whole row to say this; the label says it in four characters.
+    // Nothing else carries the window, so the suffix is the only indicator the user has left.
+    expect(readSubtitle()?.text).toContain('SOG (30 s)');
+    expect(readAxis('x')?.title?.display ?? false).toBe(false);
+  });
+
+  it('keeps the plot window on screen when the widget label is hidden', async () => {
+    // The removed axis title rendered independently of Show Label, so hiding the name must not
+    // take the window with it.
+    await setup(makeConfig({ displayName: 'SOG', timeScale: 'minute', period: 10, showLabel: false }));
+
+    const subtitle = readSubtitle();
+    expect(subtitle?.display).toBe(true);
+    expect(subtitle?.text?.trim()).toBe('(10 min)');
+    expect(subtitle?.text).not.toContain('SOG');
+  });
+
+  it('falls back to the bare label for a legacy time scale with no abbreviation', async () => {
+    // Stored configs can still carry the pre-migration TimeScaleFormat members. A full Record or a
+    // `?? format` default would render "SOG (30 Last 30 Minutes)" on those.
+    await setup(makeConfig({ displayName: 'SOG', timeScale: 'Last 30 Minutes', period: 30 }));
+
+    expect(readSubtitle()?.text?.trim()).toBe('SOG');
+  });
+
+  it.each([
+    { orientation: 'horizontal', verticalChart: false },
+    { orientation: 'vertical', verticalChart: true }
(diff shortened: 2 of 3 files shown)


mairas:
# Code review

**Scope:** `feat/chart-ticks-inside-plot` vs `main` — 2 files, 39 insertions, 34 deletions.
**Intent:** Reclaim plot area by drawing both axes' tick labels inside the grid (Chart.js `ticks.mirror`), keep them readable with a card-coloured text halo, move the plot-window text from the time-axis title into the widget label, and add vertical margin above the settings "Angle display range" field.
**Mode:** interactive.

**Reviewers:** correctness, testing, maintainability, project-standards (always-on) + adversarial (73 changed lines in a rendering path with four near-duplicate axis config blocks). Security, api-contract and reliability were not selected — no auth, input handling, exported signatures, or error/retry paths in the diff.

## P2 — High

| # | File | Issue | Reviewer(s) | Conf. | Route |
|---|------|-------|-------------|-------|-------|
| 1 | `widget-data-chart.component.ts:181` | **Toggling Vertical Data Graph never rebuilds the chart.** `lastVerticalChart` starts `null` and is assigned only inside the branch that requires it non-null, so `verticalChanged` is false for the component's lifetime; `verticalChart` is absent from `pathSignature` too. Buffered points keep the old x/y orientation until the window scrolls out, and the replaced `RealTimeScale` is dropped without `destroy()`, leaving its interval and rAF loop running. | correctness, adversarial | 0.90 | pre-existing → filed as #539 |
| 2 | `widget-data-chart.component.ts:332` | **Vertical orientation: rotated mirrored value labels escape the plot.** With `verticalChart` on, the value axis is a plain linear top axis that still rotates wide labels (`RealTimeScale` neutralises rotation only for the time axis). `mirror` moves the anchor inside the plot *and* removes the clip — `_computeLabelArea()` returns `undefined` whenever `mirror` is set — so a rotated run of text is painted, halo and all, over the widget label and the value readout. Reachable via `enableMinMaxScaleLimit` over a wide range, or `numDecimal` at its max. | adversarial | 0.70 | gated_auto → human, needs verification |
| 3 | `widget-data-chart.component.ts:266` | **Both axes on: time and value labels collide in the corner.** They used to occupy disjoint gutters. The value label draws after the time label (`Scale._layers` collapses the realtime scale into one layer at a lower `_idx`), so its 3px card-coloured stroke erases part of the newest timestamp. | adversarial | 0.70 | gated_auto → human, needs verification |
| 4 | `widget-data-chart.component.ts:415` | **The window suffix pushes the widget label into the value readout.** Title (32px, align end) and subtitle (22px, align start) already share one row via `padding.top: -35`, both boxes full width, neither constraining the other. The suffix adds up to nine characters (`" (60 min)"`, `" (365 d)"`), and canvas `maxWidth` condenses glyphs rather than truncating. The PR body acknowledges the collision as pre-existing; this quantifies how much sooner it now triggers. | adversarial | 0.70 | gated_auto → human, needs verification |
| 5 | `widget-data-chart.component.ts:415` | **No test covers the plot-window suffix.** The removed axis title was the only other carrier, so the suffix is now the sole indicator of the plot window and nothing asserts it. No spec reads `plugins.subtitle` anywhere in the repo. | testing (also raised as a gap by maintainability, project-standards, adversarial) | 0.90 | safe_auto → review-fixer |
| 6 | `VERSION:1` | **`VERSION` sits at patch level for a `feat`-typed cycle.** Latest stable is `v1.3.0+2`; `VERSION` is 1.3.1, opened by the patch bump in 35d9c5f6. Per the workspace AGENTS.md, `VERSION` must reflect the cycle's highest semver level, so a minor-level change in a patch-opened cycle re-bumps once. The level is author/reviewer judgment — the version-bump check only enforces that *a* bump occurred. Note the same mismatch predates this PR (5dc770b9, 8d7ba8cc were both `feat` with no re-bump). | project-standards | 0.68 | gated_auto → human |
| 7 | `widget-data-chart.component.ts:278` | **The two per-orientation `scales` literals duplicate 29 identical lines.** This PR paid the cost directly — one change, eight parallel edits — and hoisted consts only for the part it touched, so the method now mixes both conventions. | maintainability | 0.72 | pre-existing → filed as #540 |

## P3 — Low

| # | File | Issue | Reviewer(s) | Conf. | Route |
|---|------|-------|-------------|-------|-------|
| 8 | `widget-data-chart.component.ts:409` | **Turning off Show Label now also hides the plot window.** The removed time-axis title rendered whenever the axis was visible, independent of `showLabel`; the replacement rides on the subtitle, which is gated on `cfg.showLabel`. A user with the label off and the time scale on sees bare `30"` ticks with nothing saying how far back the plot reaches. Concrete fix: `display: cfg.showLabel \|\| !!windowSuffix`, with the text falling back to the suffix alone. | correctness, adversarial | 0.95 | gated_auto → human |
| 9 | `widget-data-chart.component.ts:269` | **Night theme: the dataset fill covers the mirrored tick labels.** The halo protects labels from the grid (drawn before them) but not from the datasets (drawn after). Dark and light fill at 16% alpha (`#ffffff29` / `#00000029`) so it barely shows; night `contrast-dimmer` is `#260000c5` — 77% alpha — over dark-red-on-black label ink. | correctness | 0.62 | advisory → human |
| 10 | `widget-data-chart.component.ts:276` | **The legacy-`TimeScaleFormat` fallback branch is untested.** `TIME_SCALE_SUFFIX` is a `Partial<Record<...>>`, so the three legacy members fall to the empty-suffix branch. Nothing pins that, so a later switch to a full `Record` or a `?? format` default would render `SOG (30 Last 30 Minutes)` on stored configs. | testing | 0.85 | safe_auto → review-fixer |
| 11 | `widget-data-chart.component.ts:305` | **The tick/grid options are unasserted and the vertical orientation is uncovered.** `...insideTicks` is spread into four separate blocks and `grid: insideGrid` assigned in four places; no spec sets `verticalChart: true` or reads `lineChartOptions.scales`, so one block losing `mirror` would reintroduce the gutter for that orientation only. | testing, adversarial | 0.85 | safe_auto → review-fixer |
| 12 | `widget-data-chart.component.ts:305` | **`...insideTicks` is spread last, so the shared const wins collisions.** No collision today, but it already carries text-styling keys (`textStrokeColor`, `textStrokeWidth`); adding `color` or `font` to it later silently overrides the per-axis `color` in four places with no compile error. Spread it first instead. | maintainability | 0.65 | safe_auto → review-fixer |
| 13 | `widget-data-chart.component.ts:272` | **`insideGrid` is assigned by reference while `insideTicks` is spread.** One object aliased into all four axes. Safe today — Chart.js reads grid options through resolver proxies — but runtime mutation of `chart.options.scales` is an established pattern in this codebase (`widget-windtrends-chart` does it), and the mixed copy/alias treatment of two adjacent consts misleads. Use `grid: { ...insideGrid }`. | maintainability (echoed as a residual risk by correctness and project-standards) | 0.76 | safe_auto → review-fixer |
| 14 | `widget-data-chart.component.ts:319` | **Dead `title: { display: false, text: "Value Axis" }` on both value axes.** Hard-coded off, never rendered, and now the only `title` blocks left in the file — which reads as if axis titles are still a live feature. | maintainability | 0.85 | pre-existing → fix-here (trivial, same method) |
| 15 | `widget-data-chart.component.ts:283` | **Constant ternaries inside the `verticalChart` branch.** `position: cfg.verticalChart ? "right" : "left"` (and `"top" : "bottom"` at 312) sit inside `if (cfg.verticalChart)`, so the false arms are unreachable. | maintainability | 0.88 | pre-existing → fix-here (trivial, same method) |

## Pre-existing (action taken)

- **#1** → https://github.com/halos-org/skip/issues/539 (orientation toggle never rebuilds)
- **#7** → https://github.com/halos-org/skip/issues/540 (duplicated per-orientation scale literals, with #14 and #15 noted as fold-ins)
- **#14, #15** → trivial and inside the method this PR rewrites; fix here.

## Coverage

- Suppressed: 0 findings below the 0.60 confidence gate.
- Untracked and therefore out of scope: `perf-harness/shot-chart.mjs`, `perf-harness/shot-chart-config.mjs` (the screenshot harness used to verify this change), and the pre-existing stale `perf-harness/mock-serve.mjs`.
- **Residual risk the unit suite cannot reach:** the spec mocks Chart.js wholesale, so every claim in the PR body about `mirror`, `fit()` and draw order is verified only by the manual perf-harness renders, which are not in CI. Findings 2, 3, 4 and 9 are all visual and would surface immediately in screenshots of the combinations the PR skipped — `verticalChart`, `inverseYAxis`, `enableMinMaxScaleLimit` over a wide range at `numDecimal` 5, day/hour scales, night theme, and a 1×1 tile.
- Mirrored labels are unclipped in general (`_computeLabelArea` returns `undefined` under `mirror`). A 70–80px value label on a small tile is a large fraction of the plot and, on the narrowest tiles, can overrun `chartArea.left`.
- The min/max/average annotation labels draw at `afterDatasetsDraw` with `position: "start"`, landing in the same corners the mirrored tick labels now occupy.
- A `displayName` that already names the window renders it twice — `SOG (30 s) (30 s)`. Existing configs are not migrated.
- No reviewer disputed the core Chart.js claims; the adversarial pass verified each against `node_modules/chart.js/dist/chart.js` and confirmed them, then found where they break down (findings 2 and 3).

---

**Verdict: ready with fixes.**

Fix order: #8 (the one behaviour regression this PR introduces), then #5, #10, #11 (the test gaps that would have caught it), then #12, #13 (trivial hardening) and #14, #15 (trivial pre-existing cleanup in the same method). #2, #3, #4 and #9 are visual and need a screenshot pass over the uncovered config combinations before they can be sized — they may be acceptable as-is, but that is a judgement call, not a verified one. #6 is a release-level decision: bump `VERSION` to 1.4.0, or retype the change as patch-level.

mairas:
Review findings addressed in 7ea2f3e6.

**#8 — plot window hidden when Show Label is off.** Fixed. The subtitle now renders whenever the label is on *or* there is a window suffix, and falls back to the suffix alone when the name is hidden. Verified in the mock harness at a 4-wide tile with `showLabel: false`: the plot window still reads `(30 s)`.

**#5, #10, #11 — missing coverage.** Six spec cases added to `widget-data-chart.component.spec.ts`:

- the plot window rides on the label, and the time axis carries no displayed title
- the window survives `showLabel: false`
- a legacy `TimeScaleFormat` yields the bare label with no parenthesised suffix
- `mirror`, `textStrokeColor`, `textStrokeWidth`, `grid.display` and `grid.drawTicks` land on both axes, in both orientations (table-driven, so the vertical branch is exercised for the first time)
- the per-axis tick colour survives the shared options, under a theme that names each key so ink and halo are distinguishable

Both mutations were checked: reverting the #8 fix, and dropping `...insideTicks` from the vertical value axis, each fail exactly one of the new tests.

**#12, #13 — spread order and the aliased grid object.** Fixed. `...insideTicks` is spread first in all four blocks; `grid: { ...insideGrid }` copies per axis.

**#14, #15 — pre-existing, routed fix-here.** Both dead `title: { display: false, text: "Value Axis" }` blocks deleted, and the two unreachable `position` ternaries inside `if (cfg.verticalChart)` folded to constants.

**#6 — VERSION.** Staying at 1.3.1. This is a layout fix, not new functionality, so the cycle keeps its patch level. (The `feat(data-chart)` commit type on d28f4775 overstates it by that reading.)

**#2, #3, #4, #9 — the four visual findings — are not addressed and are not verified either way.** They need a screenshot pass over the config combinations this PR never rendered: `verticalChart` with wide rotated value labels, `enableMinMaxScaleLimit` over a wide range at `numDecimal` 5, day/hour scales, the night theme's 77%-alpha fill, and a 1×1 tile. #4 in particular is a pre-existing collision between the title and subtitle plugins that the suffix makes trigger sooner.

`npm run lint`, `npm run snc` and the full suite (178 files, 1958 tests) pass.

mairas:
Triage of the four remaining review findings:

- **#1** (orientation toggle never rebuilds) — real, fixed in #543, stacked on this branch.
- **#2** (rotated mirrored value labels escape the plot in the vertical layout) — not reproducible on the device. No action.
- **#3** (time and value labels collide in the shared corner) — accepted as-is. No action.
- **#4** (the window suffix widens the label/value overlap) — accepted as-is. No action.

#9 (night-theme fill over the mirrored labels) is untouched and still unverified.

Deployed to a test device and exercised before this triage.

Issue #544, opened 2026-08-11
Flaky: minichart backfill spec intermittently sees zero points in the full suite


mairas:
`minichart.component.spec.ts > plots a backfill batch and then a live point off the history stream`
fails intermittently in the full suite:

```
AssertionError: expected +0 to be 3 // Object.is equality
 ❯ src/app/widgets/minichart/minichart.component.spec.ts:123:46
```

Zero points means the subscription callback in `MinichartComponent.startStreaming()` never appended,
even though the test pushes a two-point batch and a live point through the Subject the history mock
returns.

## Observed

- Failed CI on the #538 merge run ([31519006100](https://github.com/halos-org/skip/actions/runs/31519006100)); a re-run of the same commit passed.
- Failed once locally on `main` (`npm run test:headless`), then passed on the next run of the same commit.
- Passes when run in isolation (`ng test --include=.../minichart.component.spec.ts`), and passes when
  paired with `widget-data-chart.component.spec.ts`.
- Full suite passed twice on the #538 branch before merge, so it is not a straightforward
  consequence of that change — but a single green run at the pre-merge commit `30ab45a2` is not
  enough to rule out that the change perturbed timing or ordering. Unestablished either way.

## Ruled out

The `ctx` guard in the subscription callback (`if (!chartWithCtx.ctx) return;`). That spec's
`MockChart` hardcodes `public ctx = {}` precisely so the guard passes, so a null canvas context is
not the path.

## Still open

The remaining early-return paths in the callback are `this.isDestroyed`, `!this.chart`, and
`startStreaming()` bailing at `if (!info || !dataPath) return;` before subscribing at all — a bail
there would leave nobody subscribed to the Subject, which matches a length of 0 exactly. Which one
fires, and why only sometimes and only under the full suite, is not yet established.

Worth reproducing with a full-suite loop that logs `component.streamSub`, `this.chart` and
`dataSourceInfo` at the point of the push.

mairas:
Hit rate is higher than when filed: failed 2 of 3 consecutive local `npm run test:headless` runs on a branch that touches only `dashboard.component.ts`/`.spec.ts` (unrelated to minichart). Frequent enough now to block unrelated PRs on a first CI attempt.

PR #546, opened 2026-08-11
Shorten the dashboard page transition to a slide-and-fade

mairas:
## Why

The page transition read as page → slide → blank → slide → page. The blank was structural, not
stylistic: there is one `<gridstack>` inside one `.page-slide` wrapper, so `runPageChange()` has to
slide that single wrapper fully off-screen, reload the same grid with the next page's widgets, and
slide it back. Travelling 100% of the wrapper's width is what hid the swap, and it exposed an empty
viewport twice on the way.

## How

Travel drops to 15% and both phases animate opacity, so the page is transparent well before it gets
far. The swap now hides behind opacity 0 rather than behind the viewport edge, and no empty sweep is
ever shown. Per-phase duration drops 180 ms → 150 ms to match the shorter distance.

That moves the load-bearing invariant, which is the part worth reviewing: **the exit phase must end
at opacity 0.** If a later change drops the fade and keeps the short travel, a half-loaded page
flashes in place mid-transition. The method comment says so and the spec asserts the opacity
endpoints alongside the phase ordering, so that regression fails a test rather than shipping.

A direct page → slide → page transition was considered and rejected: it needs two live grids, which
puts the incoming page's widget mount inside the animation window — trading a clean blank for jank
on exactly the frames being improved, on Pi-class hardware.

## How this variant was chosen

Two candidates were built and deployed to a HALPI2 for side-by-side judgement:

- this one — 15% travel, both phases fade, 150 ms each
- slide the outgoing page fully off-screen, fade the incoming one up in place

The second still shows the full-width empty sweep on the way out. This one was preferred after
testing both on the device.

## Verified

`npm run lint`, `npm run snc`, and the dashboard spec (38 tests). Deployed to a HALPI2 and exercised
by swipe and by arrow keys.

Note: the full suite currently trips the known minichart flake (#544), which is unrelated to this
change — it touches only `dashboard.component.ts` and its spec.

`VERSION` stays at 1.3.1: the cycle is already open (latest stable is `v1.3.0+2`) and this is a
transition-style change, not new functionality.

### DIFF
--- a/src/app/core/components/dashboard/dashboard.component.spec.ts
+++ b/src/app/core/components/dashboard/dashboard.component.spec.ts
@@ -482,13 +482,14 @@ describe('DashboardComponent', () => {
         interface RunPageChangeApi {
             runPageChange: () => Promise<void>;
             animatePhase: (el: HTMLElement, from: SlidePose, to: SlidePose, easing: string) => Promise<void>;
+            ngOnDestroy?: () => void;
             loadDashboard: (dashboardId: number) => void;
             prefersReducedMotion: () => boolean;
         }
 
         // Drive the sequence without real Web Animations: record phase/load ordering
         // (load carries the index) and stub loadDashboard so it does not touch the grid.
-        function instrument(): { api: RunPageChangeApi; order: string[]; animate: Mock } {
+        function instrument(): { api: RunPageChangeApi; order: string[]; animate: Mock; slide: HTMLElement } {
             const api = component as unknown as RunPageChangeApi;
             const order: string[] = [];
             const slide = document.createElement('div');
@@ -499,7 +500,7 @@ describe('DashboardComponent', () => {
                 return Promise.resolve();
             }) as unknown as Mock;
             vi.spyOn(api, 'loadDashboard').mockImplementation((id) => { order.push(`load:${id}`); });
-            return { api, order, animate };
+            return { api, order, animate, slide };
         }
 
         function deferred(): { promise: Promise<void>; resolve: () => void } {
@@ -508,17 +509,30 @@ describe('DashboardComponent', () => {
             return { promise, resolve };
         }
 
-        it('slides out then in, loading the current page off-screen at the midpoint, for next', async () => {
+        it('slides out then in, loading the current page while transparent at the midpoint, for next', async () => {
             mockDashboardService.activeDashboard.set(1);
             (mockDashboardService.consumePendingPageDirection as Mock).mockReturnValue('next');
-            const { api, order } = instrument();
+            const { api, order, slide } = instrument();
+            let pointerEventsDuringExit: string | undefined;
+            (api.animatePhase as unknown as Mock).mockImplementationOnce((el, from, to) => {
+                order.push(`animate:${from.pct}@${from.opacity}->${to.pct}@${to.opacity}`);
+                pointerEventsDuringExit = (el as HTMLElement).style.pointerEvents;
+                return Promise.resolve();
+            });
 
             await api.runPageChange();
 
+            // A transparent page is still hit-testable and, at this travel, still inside the host's
+            // clip box: input must be off for the whole transition or a tap reaches an unseen control.
+            expect(pointerEventsDuringExit).toBe('none');
             // The exit must end fully transparent: opacity, not distance, is what hides the swap.
             expect(order).toEqual(['animate:0@1->-15@0', 'load:1', 'animate:15@0->0@1']);
             expect(mockDashboardService.beginPageTransition).toHaveBeenCalledTimes(1);
             expect(mockDashboardService.endPageTransition).toHaveBeenCalledTimes(1);
+            // A missed restore now leaves the dashboard invisible and deaf to taps, not just displaced.
+            expect(slide.style.opacity).toBe('');
+            expect(slide.style.transform).toBe('');
+            expect(slide.style.pointerEvents).toBe('');
         });
 
         it('mirrors the slide direction for a previous navigation', async () => {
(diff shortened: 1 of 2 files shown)


mairas:
# Code review

**Scope:** `feat/page-transition-short-slide-fade` vs `main` — 2 files, 42 insertions, 22 deletions.
**Intent:** Remove the empty-viewport sweep from the dashboard page transition by shortening travel to 15% and hiding the grid swap behind opacity instead of distance.
**Mode:** interactive.

**Reviewers:** correctness, testing, maintainability, project-standards (always-on). Adversarial was not selected (42 changed lines, below the threshold, no auth/payments/data-mutation surface); security, api-contract and reliability do not apply to this diff.

## P2 — High

| # | File | Issue | Reviewer(s) | Conf. | Route |
|---|------|-------|-------------|-------|-------|
| 1 | `dashboard.component.ts:318` | **The faded wrapper stays hit-testable for the whole transition.** `opacity: 0` does not stop hit-testing, and 15% travel keeps the wrapper inside `:host { overflow: hidden }` — where the old ±100% put it outside the clip box, and clipping clips hit-testing too. So for ~300 ms a tap can actuate a control the user cannot see, and after the midpoint that control belongs to the *page they have not seen yet*. `widget-autopilot`, `widget-boolean-switch`, `widget-multi-state-switch` and `widget-slider` all issue Signal K PUTs and stay live in locked mode. Nothing gates widget input on the transition: `isPageTransitioning` is read only by the navigate guards. Verified: `dashboard.component.scss:3-9`, and grep of `isPageTransitioning` consumers. | correctness, project-standards | 0.82 | gated_auto → needs a decision |
| 2 | `dashboard.component.ts:342` | **`animatePhase`'s body never executes in any test.** Every transition spec stubs it, so the keyframes, the `setSlidePose(el, to)`-then-`cancel()` ordering, and the durations are unverified. Drop `opacity` from the keyframes, or reorder the persist and the cancel, and the wrapper snaps back to opacity 1 before `loadDashboard` runs — the exact half-loaded flash this PR claims to prevent — with all 38 tests still green. jsdom has no `Element.animate`, so a fake element is needed. | testing (echoed as a gap by correctness and maintainability) | 0.92 | safe_auto |
| 3 | `dashboard.component.spec.ts:491` | **Nothing asserts the wrapper's inline pose is restored.** `instrument()` builds a real `div` but never returns it, so no test can read `slide.style`. Dropping the opacity line from `setSlidePose`, or inverting its `=== 1` guard, breaks nothing. This matters more than it did before: a missed restore now leaves the dashboard permanently invisible rather than merely displaced. | testing (echoed by correctness and maintainability) | 0.88 | safe_auto |

## P3 — Low

| # | File | Issue | Reviewer(s) | Conf. | Route |
|---|------|-------|-------------|-------|-------|
| 4 | `dashboard.component.ts:31` | **The comment's rationale is wrong.** It says the page is "transparent well before it gets far", but `animatePhase` passes one `easing` in the options object, which is effect easing — `transform` and `opacity` ride the identical curve. Under `ease-in` the exit is at ~0.69 opacity and under 5% displacement at its halfway point; opacity only reaches 0 at the very end. The endpoint invariant still holds, but the stated reason it is safe does not, and someone could extend travel on the strength of it. | correctness | 0.78 | manual → human |
| 5 | `dashboard.component.spec.ts:511` | **Test name still says "off-screen"** while the body it guards now asserts the opposite (`animate:0@1->-15@0` with a comment saying opacity, not distance, hides the swap). A reader trusting the name would restore a large travel value. | maintainability, project-standards | 0.95 | safe_auto |
| 6 | `dashboard.component.ts:324` | **The enter pose literal is duplicated on adjacent lines** — `setSlidePose` seeds it and the next `animatePhase` declares the same `from`. They must be identical or the enter starts with a visible jump; nothing enforces it, and the spec stubs `animatePhase` so a desync would only show on a device. | maintainability | 0.72 | safe_auto |
| 7 | `dashboard.component.ts:321` | **The destroy-mid-slide bail-out is untested.** The existing destroy test calls `ngOnDestroy()` with no slide in flight, so it exercises the flag clearing, not the guard — and this is the abort path most likely to strand the wrapper at opacity 0. | testing | 0.85 | safe_auto |
| 8 | `dashboard.component.ts:365` | **`prefersReducedMotion` itself is never exercised** — the test mocks the method, so the media-query string is never evaluated. A typo would silently animate for users who asked not to be. | testing | 0.72 | pre-existing → fix-here (trivial, same method) |
| 9 | `VERSION:1` | **The open cycle carries four `feat`-typed commits at a patch bump.** No bump is needed for this PR — the cycle is open at 1.3.1 against stable `v1.3.0+2` — but AGENTS.md wants `VERSION` to reflect the cycle's highest level, and that is an author judgement. The maintainer has ruled these layout/transition changes patch-level. | project-standards | 0.60 | advisory → human, no action |

## Coverage

- Suppressed: 0 findings below the 0.60 confidence gate.
- **Everything is verified at the mock boundary.** Ordering and endpoints are pinned; the Web Animations path is not run anywhere in CI, so the claim that the empty sweep is gone rests on the on-device check and the screencast, not on a test.
- The spec pins the tuning value (`15` appears literally in the order strings) rather than the invariant, so retuning travel forces a spec edit.
- `widget-heel-gauge` uses `mix-blend-mode: soft-light`/`overlay`; animating the wrapper's opacity composites that subtree against a group buffer at fractional alpha for the transition. Not visually checked.
- Compositing an opacity animation over a subtree containing canvas gauges and live iframes (`widget-freeboardsk`, `widget-iframe`) has a different cost profile than transform-only on Pi-class hardware. No perf-harness numbers were taken.
- The method comment's "no partially-loaded page is ever visible mid-transition" is stronger than the code guarantees: `ease-out` reaches ~0.2 opacity within about a frame of the grid load, and widget content renders on a later change-detection tick. Not a regression — the old version leaked the same content through the page's leading edge — but the comment overstates it.

---

**Verdict: ready with fixes.**

Finding 1 is the one that matters and is not a test-only concern: it is a new way for a tap to reach an invisible control that sends a PUT. Fix order: 1, then 2 and 3 (the coverage that would have caught a regression in this very mechanism), then 4 to 8.

mairas:
Review findings addressed in 3ae241e8.

**#1 — the faded page stayed tappable.** Fixed: input is disabled on the wrapper for the animated path and restored in the same `finally` as the pose. The instant-swap and reduced-motion paths return before that point and are unaffected. A test now captures `pointerEvents` during the exit phase, so removing the guard fails.

**#4 — the comment's rationale was wrong.** Reworded rather than re-engineered. `animatePhase` passes one effect easing, so opacity and transform ride the same curve and opacity only reaches 0 at the end of the exit; the comment now says that, and says explicitly that extending the travel on the old reading would expose the swap.

**#2, #3, #7, #8 — coverage.** Four tests added:

- `animatePhase` run for real against a fake element (jsdom has no `Element.animate`): asserts the transform+opacity keyframe pair, the animation options, that `el.style` already carries the end pose when `cancel()` fires, and the persisted transform.
- the wrapper's pose and input are restored after a normal transition
- both are restored when the component is destroyed mid-transition, and the load is skipped
- `prefersReducedMotion` evaluates the real media query instead of being mocked away

Mutation-checked, one at a time: dropping the opacity keyframes, dropping the pointer-events guard, and dropping the pose restore each fail exactly one test. Before this commit all three passed silently.

**#5, #6 — naming and duplication.** The test no longer claims the page goes off-screen, and the enter pose is bound once instead of being written as identical literals on adjacent lines.

**#9 — VERSION.** No change: the cycle is open at 1.3.1 and the maintainer has ruled this level of change patch-level.

Not addressed, and unverified either way: the `mix-blend-mode` compositing on `widget-heel-gauge` at fractional alpha, and the cost of an opacity animation over canvas gauges and live iframes on Pi-class hardware. Both are in the review's coverage notes.

`npm run lint`, `npm run snc` and the full suite (178 files, 1962 tests) pass.

Issue #416, opened 2026-07-27
Heel-gauge & horizon: repair stored attitude path settings (same dead-picker bug as #414)


mairas:
## Problem

`widget-heel-gauge` and `widget-horizon` carry the same broken-settings bug that #414 fixed for the position widget: they read a sub-field (roll / pitch) off the compound `navigation.attitude` object leaf, but a stored config's path entry is `pathType: 'number'` with `isPathConfigurable: true`, so the settings dialog shows a **configurable number path picker that can't select the object leaf** — a dead row.

Confirmed on the boat: the heel-gauge stored config (now at configVersion 15) has `angle: { path: 'self.navigation.attitude', pathType: 'number', isPathConfigurable: true }`.

**Why they weren't fixed by #415:** the v14→v15 migration only rewrote `widget-position`, but it stamped the whole config to v15. So heel-gauge/horizon are now stranded at v15 with the old shape and no migration reaches them.

**Decision:** keep these paths **hidden and fixed** (not editable like position — attitude is effectively never retargeted). Their `DEFAULT_CONFIG` and the shipped `DefaultDashboard` seed are *already* correct (`isPathConfigurable: false`, whole-leaf path). Only stored configs are broken, so **no widget or seed change is needed** — just a migration to repair stored configs.

## Fix

- **Config migration v15 → v16** (`upgradeConfigV15toV16`, scoped to `widget-heel-gauge` + `widget-horizon`): for each attitude path entry, set `path` to `self.navigation.attitude` and `isPathConfigurable: false`. Repairs the stranded v15 configs (and any stale `.roll`/`.pitch` sub-field paths). Keep the two horizon entries (hidden) — no collapse, per the hidden-fixed decision.
- **Version-bump couplings** (per the checklist): `LATEST_APP_CONFIG_VERSION → 16`; boot gate adds `liveVersion === 15`; `case 15` in the dispatch switch + `version === 15` remote-slot branch; `perf-harness/lib/skip-config.mjs` seed → 16; regenerate `skip-dashboard-schema.json` (meta.configVersion only — widget shapes unchanged); fixed `V16_MIGRATION_OUTPUT_VERSION = 16`.
- **Tests:** a v15→v16 migration test (heel-gauge + horizon repaired, non-attitude widget untouched, non-v15 slot skipped); a v15 boot-gate test; extend the seed guard to assert the attitude widgets are `isPathConfigurable: false`.

## Verification

[...]

PR #549, opened 2026-08-11
feat(config): ship the onboarding dashboards as the default seed

### LINKED ISSUE #548: feat(profiles): ship the onboarding dashboards as the default seed
## Problem

Every new Skip profile is seeded from `DefaultDashboard` (`src/default-config/config.blank.dashboard.ts`) — `ProfileService.buildBlankConfig()` for the **New** button, and the `DashboardService` constructor for any profile that boots with zero dashboards. That content is older than the curated `onboarding` profile now running on the halpi.hurma test device, which is the configuration we actually want people to start from.

Separately, **New** already hands the user the current release's factory configuration, but nothing says so. The dialog only asks for a name (`config.component.ts:73`), so it reads as "make an empty profile". An existing user who wants to see what the latest shipped configuration looks like has no reason to press it.

## Source

`onboarding` in `applicationData/users/admin-e0c818c4/skip/11.0.0.json` on halpi.hurma. Its `configVersion` is 19, matching `LATEST_APP_CONFIG_VERSION`, so no migration is involved. Four pages: Demo (12 widgets), Environment (8), Motoring (14), `Sailing ` (9). All 15 widget types it uses are registered in `WidgetService`, and every widget already satisfies the gridstack `id === widgetProperties.uuid` invariant the seed file's header requires.

## Scope

**Seed refresh**

- R1. Replace the `DefaultDashboard` array with the onboarding profile's `dashboards`.
- R2. Reset the three boat-specific `$source` bindings to `default`: `trueWindAngle` and `trueWindSpeed` on `can0.c0aa8200ffce8b18`, and `courseOverGround` on `can0.c0789100ffc00001`. Every other source in the export is already `default`.
- R3. Strip the trailing space from the page name `Sailing `.
- R4. Keep the `Demo` page and its name. It is a deliberate mixed-widget first page.

**Test guard**

- R5. Remove `widget-horizon` from the seed-guard loop in `src/default-config/config.blank.dashboard.spec.ts`. The onboarding profile has no Pitch & Roll widget, and that loop asserts the path shape of *seeded* widgets — over a widget the seed does not contain, it asserts nothing. The profile's two heel gauges keep the `navigation.attitude` family covered. Every other assertion in that spec must keep passing against the new seed; position, wind-steer and autopilot are all present.

**Discoverability**

- R6. Say in the New profile dialog that the profile starts from the pages shipped with this version of Skip.

## Non-goals

- No read-only "Factory" profile row in the profile list. Considered and dropped: **New** already produces a profile seeded from the factory configuration, so a synthetic row would add a concept without adding a capability.
- No change to the reserved `default` profile name, and no change to **Advanced → Default**, which stays an empty-dashboards reset.

## Done when

- A profile created with **New** shows the four onboarding pages.
- No `can0.*` source appears anywhere in the seed.
- `npm run ci` passes.

mairas:
Closes #548.

## Why

Every new profile is seeded from `DefaultDashboard` — `ProfileService.buildBlankConfig()` for the **New** button, and the `DashboardService` constructor for any profile that boots with no dashboards. That content was older than the curated `onboarding` profile on the halpi.hurma test device, which is the configuration we actually want people to start from.

**New** has always handed the user the configuration shipped with the running version, but the dialog only asked for a name, so it read as "make an empty profile". That is why an existing user wanting the latest shipped pages had no reason to press it.

## What

The seed is now the onboarding profile's four pages — a Demo page plus Environment, Motoring and Sailing (12 / 8 / 14 / 9 widgets).

A live export carries its vessel in more than its `$source` values. The sanitization covers four kinds:

- **Identity** — three CAN-bound sources; `trackedDevices` and `optionsById` on the five electrical widgets (which pinned one boat's Victron MAC, its `engine.XX` alternator, and its 240/320 W panel ratings); and `autopilot.instanceId`, which came back as `raymarineN2K`. That last one matters most: with a concrete instance id the widget builds v2 REST endpoints and presents live Engage/Tack/Dodge controls that 404 on every other vessel, instead of taking its explicit "not configured" branch.
- **Calibration** — an engine gauge left on a 0–100 scale reading an RPM path (the same seed's other RPM gauge uses 0–3600), and a fuel gauge scaled to this boat's 60 L tank. A clamped gauge states a wrong number rather than showing no data, so it reads as a working instrument. The fuel gauge now reads `currentLevel` as a percentage, which is capacity-independent.
- **Behavior** — three pages carried auto-switch triggers. The old seed had none. Shipping them opts every new profile into `PageSwitchService` moving the view on its own: a moored boat would boot straight past the curated first page, and the view would jump on each nav or engine state change.
- **Consistency and cosmetics** — a bearing left on `unitless` (renders raw radians, ~3, beside the same bearing as 172 on another page), wind speeds in m/s where the rest of the seed uses knots, a wind dial driven at 100 ms against a 1000 ms default, and stray trailing spaces in one page name and three widget names.

## Guards

The header used to carry the sanitization recipe as prose. Every item above is now asserted, because the recipe is what failed:

- The vessel-agnostic scan in `dashboard.service.spec.ts` gains `optionsById` and `instanceId`, and walks the whole page rather than just `configuration` so a page-level trigger cannot hide outside it. A separate assertion rejects any seeded trigger.
- `config.blank.dashboard.spec.ts` gains unit agreement across pages (this one caught the m/s wind speeds after the first round of fixes) and a name-whitespace check.
- `widget-horizon` leaves the seed loop — the seed carries none, so that assertion had nothing left to guard — and its #416 path shape is asserted against `WidgetHorizonComponent.DEFAULT_CONFIG` instead, alongside the heel gauge. Dropping a widget from a future seed can no longer drop its guard.

`VERSION` goes to 1.4.0 via `./run bumpversion minor`: the open cycle was opened by a patch bump and this is a minor-level change.

## Verification

`npm run ci` passes: lint, `snc`, 1968 tests across 178 files, 33 MCP-schema tests.

`perf-harness/shot-default-seed.mjs` (added here) boots the real production bundle against an empty-dashboards profile so `DashboardService` seeds `DefaultDashboard` exactly as on a fresh install, and shoots every page plus the dialog.

It streams a realistic SI value for every path the seed subscribes to. That is the point: the first version of this script ran with the mock's delta stream off, so every widget rendered blank and **none of the defects above could appear in it**. A dataless screenshot verifies nothing worth verifying. Page and widget counts now come from the seed source rather than a flag, and a mismatch exits non-zero.

## Known issue, not introduced here

#556 — `widget-gauge-steel` renders `0.00` when its path had a subscriber on a previously-visited page. Reached directly the same gauge reads correctly; reached via another page that also subscribes to `self.propulsion.main.revolutions` it sticks at zero. Every other widget on a shared path recovers, so it is confined to that widget type. It is pre-existing — any config with two widgets on one path across two pages hits it — but this seed makes it reachable out of the box, because it places an RPM gauge on both the Demo and Motoring pages.

No additional operational monitoring required: this changes bundled seed data, one dialog string, and test guards, with no runtime service, endpoint, or persistence behavior involved. Existing profiles are untouched — the seed is only read when a profile is created or boots with no dashboards.

On the first device that takes this build, create a profile with **New** and confirm the four pages appear, that the autopilot widget reports "not configured" rather than presenting live controls, and that no widget is bound to a foreign `$source`. If a page came out wrong the fix is another re-export, not a rollback, since nothing persisted changes shape.


mairas:
## Code review — 5 personas

Scope: `origin/main..feat/onboarding-default-seed` — 2620 lines of re-exported seed data plus ~110 lines of code.
Team: correctness, testing, maintainability, project-standards (always-on) + adversarial (large diff that changes the default configuration every install receives).

Findings were verified against the data before acting; every claim below was reproduced independently. **Fixed in this branch unless marked otherwise.**

### P1 — shipped a wrong or vessel-specific value

| # | Finding | Reviewers |
|---|---|---|
| 1 | `autopilot.instanceId` came back as `raymarineN2K`; the old seed had `null`. With a concrete id the widget builds v2 REST endpoints and presents live Engage/Tack/Dodge controls that 404 on every other vessel, instead of its explicit not-configured branch. | correctness, project-standards, adversarial |
| 2 | Demo engine gauge on an RPM path with `displayScale` 0–100 — the same seed's other RPM gauge uses 0–3600. The widget clamps, so an idling engine pegs the needle and prints 100 forever. | correctness, adversarial |
| 3 | Motoring fuel gauge scaled 0–60 L, the source boat's tank. The clamped value is also the printed label, so a 200 L tank reads "60". Now reads `currentLevel` as a percentage, which is capacity-independent. | adversarial |
| 4 | The verification harness ran with the mock's delta stream off, so every widget rendered blank and none of #1–#3 could appear in it. The screenshots proved nothing. Now streams a realistic SI value per seeded path. | adversarial |

### P2 — behavior and consistency

| # | Finding | Reviewers |
|---|---|---|
| 5 | Three of four pages carried auto-switch triggers; the old seed had none. Every new profile would opt into `PageSwitchService` moving the view — a moored boat boots past the curated first page, and the view jumps on each nav/engine state change. Stripped. | correctness, adversarial |
| 6 | Sailing's bearing left on `unitless` renders raw radians (~3) beside the same bearing as 172 on Motoring. | correctness, adversarial |
| 7 | Sailing's wind speeds in m/s where the Demo dial and the trends chart use knots (`DEFAULT_CONFIG` is knots). Caught by the new unit-agreement guard after the first fix round. | correctness |
| 8 | Wind dial at `updateInterval: 100` against a 1000 ms default, with 11 subscribed slots, on Pi-class targets. | adversarial |
| 9 | `widget-horizon` was deleted from the seed guard rather than relocated, establishing "drop the widget, drop its guard" as the precedent. Its #416 shape is now asserted against `DEFAULT_CONFIG`, alongside the heel gauge. | testing, maintainability, adversarial |
| 10 | The sanitization recipe lived in a header comment and was asserted by nothing — which is how #1 shipped past a green CI. The vessel-agnostic scan now covers `optionsById` and `instanceId` and walks the whole page so a trigger cannot hide outside `configuration`; the seed spec gains unit agreement and a name-whitespace check. | testing, maintainability, project-standards |
| 11 | `VERSION` stayed at a patch bump for a minor-level change. Now 1.4.0 via `./run bumpversion minor`. | project-standards |
| 12 | Nothing asserted that `createProfile` passes the description — deleting the line broke no test. Added, plus its rename counterpart. | testing |

### P3

| # | Finding | Reviewers |
|---|---|---|
| 13 | The shot script logged widget counts instead of asserting them, and hardcoded `--pages 4`. Counts now come from the seed source and a mismatch exits non-zero. | testing, maintainability, project-standards |
| 14 | Trailing whitespace in one page name and three widget display names — the signature of a hand transform. | maintainability, adversarial |
| 15 | `VMG` was the only seeded numeric with `enableTimeout` on, on a path most boats never publish, so it retries and logs forever. | correctness |
| 16 | No README entry for the new probe. Added. | project-standards, maintainability |

### Not fixed

- **#556** — `widget-gauge-steel` reads `0.00` when its path had a subscriber on a previously-visited page. Reproduced and filed. Pre-existing, but this seed makes it reachable out of the box by placing an RPM gauge on two pages. **Worth deciding before merge** — see the PR description.
- **Generic seed-vs-`DEFAULT_CONFIG` guard** (testing, P2). 11 of 15 seeded widget types have no shape guard, four of them new to the seed. An audit found zero drift today, so this is future-regression coverage; the targeted guards added here cover the failure modes that have actually shipped.
- **Perf measurement of the new seed** (adversarial, P2, advisory). Widget count went 30 → 43 while chart count went 11 → 7, and only the active page mounts. No before/after at 10× throttle was run.
- **`.dialog-description` vs the `.dialogue-` prefix** used by five sibling dialog stylesheets (maintainability, P3, 0.62 confidence).
- **Widget uuids are not regenerated per profile**, only page ids, so profiles seeded from the default share widget uuids. Pre-existing; no consumer keys off them today.

### Verdict

Ready with fixes applied — `npm run ci` green (lint, `snc`, 1970 tests / 178 files, 33 MCP-schema tests), pending the #556 decision.

PR #557, opened 2026-08-11
fix(gauge-steel): seed a freshly built gauge with its value

### LINKED ISSUE #556: widget-gauge-steel reads 0.00 when its path had a subscriber on a previously-visited page
`widget-gauge-steel` renders `0.00` and a zeroed needle when the page it is on is reached *after* a page that subscribed to the same Signal K path. Reached directly, the same widget on the same path with the same data renders correctly.

## Reproduce

With `perf-harness/shot-default-seed.mjs` on `feat/onboarding-default-seed` (whose seed puts a radial gauge on `self.propulsion.main.revolutions` on page 0 and a steel gauge on the same path on page 2), streaming `propulsion.main.revolutions = 30` (Hz, i.e. 1800 rpm):

| Navigation | Steel gauge reads |
|---|---|
| straight to `#/dashboard/2` | `1800.00`, needle correct |
| `#/dashboard/0`, wait, then `#/dashboard/2` | `0.00`, needle at zero |

Nothing else differs — same bundle, same seed, same values, 5 s settle in both cases. No console errors and no page errors.

## Scope

Only the steel gauge fails. On that same Motoring page, every other widget on a path that is also subscribed from an earlier page recovers its value normally — speed over ground, course over ground, bearing to waypoint, depth, coolant temperature and fuel level all render correctly in the failing run. 19 of the seed's 38 paths are subscribed from more than one page, so the trigger is common; the symptom is confined to this widget type.

## Where to look

`widget-gauge-steel.component.ts:98-113`. The observe callback sets `dataValue` to `lower` when `raw == null`, which matches the observed zero, so the widget is either receiving only a bootstrap null or not re-establishing its subscription when the path already had a subscriber that went away. Unlike the radial gauge on the same path, its seeded `gaugePath` carries no `suppressBootstrapNull` — worth checking whether that is the whole story or a coincidence.

## Why it matters

Pre-existing: any config with two widgets on one path across two pages hits this today. It becomes reachable out of the box once the refreshed default seed lands, because that seed ships an RPM gauge on page 0 and page 2. A gauge pinned at zero reads as a working instrument reporting zero, not as a broken one.

mairas:
Closes #556.

## The bug

A steelseries gauge only ever learned a value through `ngOnChanges`, and that handler drops two classes of change:

```
if (!this.gaugeStarted) { return; }                  // every change before the canvas exists
if (changes.value && !changes.value.firstChange)     // and the first change, always
```

Meanwhile `startGauge()` constructed `new steelseries.Radial(...)` and never applied the value the component already held, so a fresh gauge started at the library default and waited for a *subsequent* change to paint.

While a reading moves that is invisible — the next change lands a moment later and paints the right number. It sticks when no further change ever arrives, which is exactly what a steady reading produces. A path that a previously-visited page already subscribed emits its cached value on subscribe, i.e. before the canvas is built, so that one emission was dropped and nothing followed it. The needle stayed at zero indefinitely while every other widget on the page read correctly.

Rebuilds fell into the same hole: `onResized` and a zone change both call `startGauge(true)`, replacing the gauge object, and the new one was equally unseeded.

## Reproduction

Two pages, same Signal K path, one radial gauge and one steel gauge, streaming a constant value:

| Navigation | Before | After |
|---|---|---|
| direct to the steel page | 1800.00 | 1800.00 |
| via the radial page first | **0.00** | 1800.00 |

Both the constant value and the prior subscriber are required. An oscillating value masks it completely, which is why a first minimal reproduction attempt with a moving value did not show it.

## The fix

Seed the gauge from `this.value()` right after constructing it. That covers the initial build and every rebuild, and it needs no change to the two `ngOnChanges` guards — whatever they drop, the next build re-reads the current value.

## Verification

`npm run ci` passes: lint, `snc`, 1965 tests across 178 files, 33 MCP-schema tests.

Three regression tests in `gauge-steel.component.spec.ts`. The two that assert seeding fail against the unfixed component with `expected [] to deeply equal [ 1800 ]`; the third asserts a gauge with no value yet is left alone, and passes either way so it cannot mask a regression. Confirmed end-to-end in a real browser against the production bundle with the reproduction above.

No additional operational monitoring required: one widget's initial paint, no service, endpoint, or persistence behavior involved.

Worth an eye on the first device that takes this build: open a dashboard with a Classic Steel gauge on a path another page also shows, navigate between them, and confirm the needle tracks rather than sitting at zero.

### DIFF
--- a/src/app/widgets/gauge-steel/gauge-steel.component.spec.ts
+++ b/src/app/widgets/gauge-steel/gauge-steel.component.spec.ts
@@ -146,4 +146,62 @@ describe('GaugeSteelComponent', () => {
     // rather than staying collapsed/off-scale from the boot-time '' unit.
     expect(internals.gaugeOptions.section).toEqual([{ lower: 11.5, upper: 12.5, color: 'red' }]);
   });
+  // #556: the gauge only ever learns a value through ngOnChanges, which drops every change that
+  // lands before the canvas exists (and the first change always). A path a previously-visited page
+  // already subscribed emits on subscribe -- before the canvas is built -- so with a constant
+  // reading no further change ever arrives and the needle stayed at zero for good.
+  interface FakeGauge { setValue: (v: number) => void; setValueAnimated: (v: number) => void }
+
+  const stubRadial = (seeded: number[]): void => {
+    const steel = (globalThis as unknown as { steelseries: Record<string, unknown> }).steelseries;
+    steel.Radial = vi.fn(function (this: FakeGauge) {
+      this.setValue = (v: number) => { seeded.push(v); };
+      this.setValueAnimated = vi.fn();
+    });
+  };
+
+  it('seeds a freshly built gauge with the value it already holds', () => {
+    const seeded: number[] = [];
+    stubRadial(seeded);
+
+    fixture.componentRef.setInput('widgetUUID', 'uuid-warm-path');
+    fixture.componentRef.setInput('subType', 'radial');
+    fixture.componentRef.setInput('minValue', 0);
+    fixture.componentRef.setInput('maxValue', 3600);
+    // The value arrives before the gauge is built, which is the whole point.
+    fixture.componentRef.setInput('value', 1800);
+
+    (component as unknown as { startGauge: (f?: boolean) => void }).startGauge(true);
+
+    expect(seeded).toEqual([1800]);
+  });
+
+  it('re-seeds the value on a rebuild, so a resize or zone change does not blank the needle', () => {
+    const seeded: number[] = [];
+    stubRadial(seeded);
+
+    fixture.componentRef.setInput('widgetUUID', 'uuid-rebuild');
+    fixture.componentRef.setInput('subType', 'radial');
+    fixture.componentRef.setInput('minValue', 0);
+    fixture.componentRef.setInput('maxValue', 3600);
+    fixture.componentRef.setInput('value', 1800);
+
+    const internals = component as unknown as { startGauge: (f?: boolean) => void };
+    internals.startGauge(true);
+    internals.startGauge(true); // a resize or zone rebuild replaces the gauge object
+
+    expect(seeded).toEqual([1800, 1800]);
+  });
+
+  it('leaves a gauge unseeded when no value has arrived yet', () => {
+    const seeded: number[] = [];
+    stubRadial(seeded);
+
+    fixture.componentRef.setInput('widgetUUID', 'uuid-no-value');
+    fixture.componentRef.setInput('subType', 'radial');
+
+    (component as unknown as { startGauge: (f?: boolean) => void }).startGauge(true);
+
+    expect(seeded).toEqual([]);
+  });
 });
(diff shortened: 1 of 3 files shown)


mairas:
## Code review — correctness + adversarial

Both reviewers independently re-derived the diagnosis rather than accepting it, and both confirmed it: `DataService.subscribePath` seeds a new registration from the `_skData` cache so a warm path emits synchronously on subscribe, `unsubscribePath` only completes at refCount 0, and every churning consumer releases through the idempotent `acquirePath` handle — so no live subscription was being torn down under the widget. A value genuinely should not have arrived after the canvas was built; the component dropped the only one there was. `setValue` exists on all three shipped constructors (verified by loading the real library in jsdom), and no wrong-unit first paint is reachable through the seed. The adversarial reviewer also found the corroborating precedent: `widget-horizon.component.ts:283` already re-applies its last known values after every rebuild, which is the shape this PR adds.

They agreed on one real gap, now fixed in `9f40968c`.

### P2 — fixed

**A rebuild in the same change batch discards a gauge mid-animation, and the orphan wins the last frame.** `ngOnChanges` applied `changes.value` before the structural branches that call `startGauge(true)`, so a batch carrying both started a tween on the outgoing gauge and then replaced it. The library exposes no tween cancel — only a later `setValueAnimated` on the same instance stops one — so the discarded gauge kept repainting the *same canvas* from its own snapshot of min/max, sections and size. Under a steady reading nothing repaints afterwards, so the stale face stays on screen: #556 again by another route.

Not hypothetical. It was measured — an orphaned instance issued 285 further `drawImage` calls into the shared canvas over 1.5 s after its replacement was built and seeded — and it is triggered by the ordinary boot sequence, where the server-resolved measure lands together with the value it re-converted, changing value + units + min + max in one pass.

The fix skips the animated set when a rebuild follows in the same batch; the rebuild's seed paints the value. Two specs cover it, and the first fails without the guard (`expected [ 1800 ] to deeply equal []`).

### P3 — fixed

- **The seed fabricated a reading before any data arrived.** The parent's `dataValue` started at `0`, never null, so `current != null` was always true and every build seeded 0 — parking the needle at zero on a scale with a negative minimum (rudder angle, temperature in °C) where the library would have shown its start-at-minimum face. `dataValue` is now `number | null`, which also makes the third spec's contract reachable instead of decorative.
- **The teardown comment was wrong.** `setValue` does not stop a running `setValueAnimated` tween; nothing in the shipped library does. Comment corrected rather than left asserting something false.

### Not fixed — filed

- Config edits to `subType`, `barGauge` (Digital Meter) and `decimals` have no `ngOnChanges` branch at all, so they apply only when an unrelated event happens to rebuild. Same dropped-change class this PR fixes for `value`, different inputs.
- `gaugeStarted` is set true before the `subType` switch, so an unknown subtype leaves `gauge` undefined while the flag says otherwise; the next `ngOnChanges` then throws on an unguarded dereference and aborts the rest of the handler.

### Noted, not actioned

A single `ngOnChanges` batch can still construct the gauge two or three times, because `zones`, `radialSize` and `units`/`min`/`max` each call `startGauge(true)` independently — now with one extra repaint per build from the seed. Worth a perf-harness number on Pi-class hardware before this compounds, but it predates this PR and the seeding adds one paint, not one construction.

### Verdict

Ready to merge. `npm run ci` green — lint, `snc`, 1967 tests across 178 files, 33 MCP-schema tests. Re-verified end-to-end in a real browser after the follow-up: the reproduction that read `0.00` now reads `1800.00`.

Issue #555, opened 2026-08-11
Tracking: anonymous read-only access on secured Signal K servers


mairas:
## Anonymous read-only access on secured Signal K servers (tracking)

Skip sends every visitor of a security-enabled Signal K server to a login page, even when the server grants anonymous read access. That blocks Skip from being bundled as Signal K's default instrument webapp: [SignalK/signalk-server#2931](https://github.com/SignalK/signalk-server/pull/2931) is held up because Skip is unusable at demo.signalk.org, where KIP works.

The ask from that PR has two halves — Skip usable read-only on a secured server with no sign-in, and a way for a new visitor to arrive at a curated dashboard.

## What is actually broken

The redirect is Skip's own code (commit `2195cbf2`), not something Signal K does to webapps: webapps mount as bare `express.static` (`signalk-server/src/interfaces/webapps.ts`), the catch-all middleware at `tokensecurity.ts:694` only annotates the request, and no caller passes `redirect: true` to `http_authorize`. Freeboard-SK reads the same status field and merely offers a Login menu item.

Two defects compound. Skip gates on `authenticationRequired`, which the server hardcodes to `true` under `tokensecurity` regardless of `allow_readonly` (`tokensecurity.ts:940`), and ignores `readOnlyAccess`, the field that actually reports anonymous read. And Skip no longer has any config source other than a logged-in user's applicationData slot, so even suppressing the redirect leaves the visitor on an app with no dashboards — which is also why Skip is already broken on a server with security switched off.

## Deployments that must not change

- **HaLOS.** Signal K is not behind a Traefik gate (`halos-marine-containers/apps/signalk-server/metadata.yaml:49`); it authenticates natively against Authelia with `SIGNALK_OIDC_AUTO_LOGIN: "true"` (`metadata.yaml:76`). Because auto-login takes priority over anonymous read in the new branch ordering, a session-less visitor still bounces to Authelia and comes back to their own profile.
- **A plain Signal K server with `allow_readonly` off** — the upstream default (`tokensecurity.ts:245`). The API answers 401 to an anonymous visitor there, so the sign-in redirect stays correct and stays.

## Sub-issues

- #550 — honor `readOnlyAccess` in the bootstrap auth decision
- #551 — give an anonymous session a configuration to render (global scope, built-in default fallback)
- #552 — suppress applicationData writes in a session that cannot write
- #553 — document publishing the anonymous default dashboard

[...]

Issue #553, opened 2026-08-11
docs: publishing the anonymous default dashboard to the global applicationData scope


mairas:
## Problem

Once an anonymous session reads its dashboard from the global applicationData scope, somebody has to put one there. Skip has no UI that writes to the global scope — `PROFILE_SCOPE` is `'user'`, and `'global'` appears only in `listConfigs` (`src/app/core/services/storage.service.ts:295`) and the legacy upgrade path. Signal K's admin UI does not edit applicationData either.

This is the second half of the request on [SignalK/signalk-server#2931](https://github.com/SignalK/signalk-server/pull/2931): "a way to configure things so that a new user at demo.signalk.org gets a demo dashboard".

## Scope

- R1. Document the publishing step: an admin POSTs a Skip configuration to `/signalk/v1/applicationData/global/<appid>/<version>`, with the concrete appid and file version Skip reads, and a worked example. Writes there are admin-only (`signalk-server/src/tokensecurity.ts:887-892`), reads are not, which is what makes the anonymous fetch work.
- R2. State the prerequisite plainly: the global read only succeeds while `allow_readonly` is on, because the anonymous request needs the `AUTO` principal to get past `http_authorize`.

## Non-goals

- No "publish this profile as the anonymous default" action in Skip's UI for now. It is friendlier than a documented `curl`, but it is a new admin capability rather than part of fixing the bug; open it separately if demo.signalk.org turns out to need repeated re-publishing.

PR #560, opened 2026-08-12
Anonymous read-only access on secured Signal K servers

### LINKED ISSUE #550: fix(auth): honor readOnlyAccess so a server with anonymous read is not forced through sign-in
## Problem

Skip sends the browser to sign-in on every security-enabled Signal K server, including servers that grant anonymous read access. `AppNetworkInitService.handleCookieAuth` branches on `status.authenticationRequired` (`src/app/core/services/app-initNetwork.service.ts:85`), but the server hardcodes that field to `true` whenever `tokensecurity` is active (`signalk-server/src/tokensecurity.ts:940`), independent of `allow_readonly`. The field that actually reports anonymous read is `readOnlyAccess` (`tokensecurity.ts:939`), and Skip never parses it: it is absent from `ILoginStatus` (`src/app/core/services/authentication.service.ts:13-22`) and appears only in a comment at `:116`.

A secondary path has the same effect on servers with no OIDC at all: `oidcAutoLogin` is then `undefined`, and the guard at `app-initNetwork.service.ts:98` tests `!== false`, so an absent field is read as "redirect".

The redirect is Skip's own code (commit `2195cbf2`), not something Signal K does to webapps. Webapps mount as bare `express.static` (`signalk-server/src/interfaces/webapps.ts`, `mountWebModules`), the catch-all middleware at `tokensecurity.ts:694` only annotates the request and always calls `next()`, and no caller passes `redirect: true` to `http_authorize`, so the server's own `res.redirect('/admin/#/login')` branches are unreachable. Freeboard-SK reads the same `authenticationRequired` field and does nothing but show a Login menu item.

Raised upstream on [SignalK/signalk-server#2931](https://github.com/SignalK/signalk-server/pull/2931): Skip cannot be the bundled default webapp while it forces sign-in at demo.signalk.org.

## Scope

- R1. Add `readOnlyAccess` to `ILoginStatus`.
- R2. Reorder the not-logged-in branch of `handleCookieAuth`: `oidcAutoLogin === true` redirects as today; otherwise `readOnlyAccess === true` yields a new `anonymous` outcome; otherwise the current redirect / auth-blocked path stands.
- R3. The `anonymous` outcome spends no SSO redirect budget and sets no auth-blocked bootstrap issue.

Two deployments must be unaffected, and both fall out of R2 rather than needing a flag. HaLOS sets `SIGNALK_OIDC_AUTO_LOGIN: "true"` (`halos-marine-containers/apps/signalk-server/metadata.yaml:76`), so HaLOS devices take the first branch exactly as now. A server with security enabled and `allow_readonly` off — the Signal K default (`tokensecurity.ts:245`) — takes the third and keeps its `/admin/#/login` redirect, which is correct there because the API genuinely answers 401.

## Non-goals

- No change to `authenticationRequired` in signalk-server. Deriving it from `allow_readonly` would repair every webapp at once but silently changes what other consumers display; that is a conversation for the upstream PR, not a unilateral edit.

## Done when

- The four not-logged-in combinations (`oidcAutoLogin` on/off × `readOnlyAccess` on/off) are covered by unit tests on `handleCookieAuth`.
- A Signal K server with security enabled, `allow_readonly: true` and no OIDC boots Skip without a redirect.
- A HaLOS device still bounces to Authelia on a session-less load.

This change alone leaves the anonymous visitor on an empty app — the config source is the sibling issue.

### LINKED ISSUE #551: feat(storage): give an anonymous read-only session a configuration to render
## Problem

An anonymous Skip session has nowhere to get dashboards from, so removing the sign-in redirect is not on its own enough to make one usable.

`SettingsService.startup()` refuses to publish any configuration unless `storage.isRemoteContextBootstrapped()` is true and `initConfig.app` is present (`src/app/core/services/settings.service.ts:110-122`); otherwise it logs "Shared configuration enabled but remote bootstrap handoff is missing or empty. Waiting for explicit recovery action." and returns. The only caller of `bootstrapRemoteContext` sits inside `if (this.isLoggedIn)` in `app-initNetwork.service.ts:213`, and `isLoggedIn` is true only for `status === 'loggedIn'` (`authentication.service.ts:108`).

This already bites today, independent of the read-only work: on a Signal K server with security switched off entirely, `authenticationRequired` is `false` (`signalk-server/src/dummysecurity.ts:34-39`), Skip proceeds past auth as designed — and then renders with no dashboards, because nothing bootstrapped the storage context.

The **global** applicationData scope is the natural source. Its GET is not admin-gated — `addAdminWriteMiddleware` wraps only PUT/POST/DELETE (`signalk-server/src/tokensecurity.ts:887-892`) — so a request carrying the anonymous `AUTO` readonly principal (`tokensecurity.ts:1665-1668`) can read it whenever `allow_readonly` is on. `StorageService.getConfig` already takes `'global'` (`src/app/core/services/storage.service.ts:327`).

## Scope

- R1. On an anonymous bootstrap, read the configuration from the global scope, and fall back to the built-in `DefaultDashboard` when the server has none published.
- R2. Bootstrap the storage context from whichever source won, so `SettingsService.startup()` publishes settings and `DashboardService` renders.
- R3. Carry writable-vs-read-only in the bootstrap context. `dashboard.service.ts:128` currently reads `isRemoteContextBootstrapped()` to decide whether to persist per-device subscribe demand; a read-only bootstrap must not make it attempt that write.
- R4. The same path covers the security-disabled server, which today reaches the "waiting for explicit recovery action" dead end.

## Done when

- A Signal K server with security enabled, `allow_readonly: true` and a config published to the global scope shows that dashboard to an anonymous visitor.
- The same server with nothing published shows the built-in default dashboard rather than a recovery state.
- A server with security disabled renders dashboards instead of logging the bootstrap-handoff warning.
- A logged-in session is byte-for-byte unchanged: it still loads its own user-scope profile.

### LINKED ISSUE #552: fix(storage): suppress applicationData writes in a session that cannot write
## Problem

Once Skip boots an anonymous read-only session, every routine save still fires at the server and comes back 401. Settings and dashboard changes go through `storage.patchConfig` (`src/app/core/services/settings.service.ts:285,290`), profile create/rename/delete through `storage.setConfig` (`profile.service.ts:76,97`), and the migration path writes unconditionally (`configuration-upgrade.service.ts:172-458`). The write queue already anticipates the failure — its `catchError` comment names "a read-only session's 401" (`storage.service.ts:129-140`) — but each failure still increments `_patchFailures` and pushes `_patchFailure$`, so a visitor who so much as drags a widget collects error toasts for writes that were never going to land.

All applicationData writes funnel through one serial queue (`storage.service.ts:126-140`), which makes this a single choke point rather than a hunt through call sites.

## Scope

- R1. Suppress writes at the queue entry when the session cannot persist, covering `patchConfig`, `setConfig` and `removeItem` together. Suppressed means "no request, no failure event", not "request that fails quietly".
- R2. Skip the `configuration-upgrade.service` migrations in a session that cannot write — a migration that cannot be persisted must not run at all rather than run and fail per item.
- R3. Confirm the write affordances are already hidden. `canWriteUserData$` requires `status === 'loggedIn'` (`authentication.service.ts:65`), so an anonymous session should surface no save, profile or config-write control; fix any that do not key off it.
- R4. Say so in the connection panel. `connection-status.component.html:9-15` already renders "Not signed in" with a working Sign in button for exactly this state (and `SsoRedirectService.manualSignIn` already bypasses the redirect budget) — it needs wording that distinguishes "live data, read-only" from "not connected".

## Done when

- Editing a dashboard in an anonymous session produces no network write and no error toast.
- Local-only state that legitimately belongs to the device (the connection config in localStorage) still persists.
- A logged-in read-write session saves exactly as before.

mairas:
## Why

Skip sent every visitor of a security-enabled Signal K server to a sign-in page, even servers that serve anonymous reads happily. That is what blocks [SignalK/signalk-server#2931](https://github.com/SignalK/signalk-server/pull/2931): Skip cannot be Signal K's bundled default webapp while it is unusable at demo.signalk.org, where KIP works.

The redirect is Skip's own code (`2195cbf2`), not something the server does to webapps — webapps mount as bare `express.static`, and no caller passes `redirect: true` to the server's `http_authorize`.

Two defects compounded. Skip gated on `authenticationRequired`, which the server hardcodes to `true` whenever its security strategy is active regardless of `allow_readonly`, and ignored `readOnlyAccess`, the field that actually reports anonymous read. And Skip had no config source other than a signed-in user's applicationData slot, so suppressing the redirect alone would have left the visitor on an app with no dashboards — the dead end a security-disabled server already reached.

## How

The bootstrap's not-logged-in branch is ordered `oidcAutoLogin === true` → redirect, else `readOnlyAccess` → anonymous, else today's behaviour. Both deployments that must not change fall out of that ordering rather than a new flag: HaLOS sets `SIGNALK_OIDC_AUTO_LOGIN=true`, so its devices still bounce to Authelia; a server with `allow_readonly` off — the upstream default — keeps its sign-in redirect, which is correct there because the API genuinely answers 401.

An anonymous session reads its dashboard from the shared **global** applicationData slot, whose GET Signal K does not admin-gate, and falls back to the dashboards shipped in this release when nothing is published. The bootstrap context now records that the config is not the session's own, which force-locks the layout the way embed mode is locked, keeps the visitor from writing per-device state derived from someone else's dashboards, and skips the config migration and its instructions dialog.

Storage readiness moved off "is signed in" onto "is reachable", with the write verbs guarded on write capability instead. That is the honest split — Signal K admin-gates only the write verbs on applicationData — and it fixes a second case in passing: a signed-in `readonly` user's saves used to reach the server purely to come back 401.

## Notes for review

- `buildDefaultConfig()` went into a new `src/default-config/config.default.factory.ts` rather than `config.blank.const.ts`. The const module would have had to import the dashboard seed, which type-imports `dashboard.service`, closing an import cycle onto the constants it exports. `ProfileService.buildBlankConfig` delegates to it.
- Three spec files needed a local `AuthenticationService` stub: with the real root service they carry a null login status, so the new write guard refused writes those suites depend on.
- `VERSION` 1.3.1 → 1.4.0. The cycle was open at patch level against v1.3.0; this is feature-level, so it re-bumps once to the highest level in the cycle.

## Verification

`npm run ci` clean: lint, `snc`, 1983 unit tests, 33 schema tests. Not yet exercised against a live Signal K server with `allow_readonly: true` and no OIDC, nor against a HaLOS device.

Closes #550
Closes #551
Closes #552

Tracking issue #555 stays open — documenting how an operator publishes the global config (#553) is not in this branch.


mairas:
# Code review — anonymous read-only access

**Scope:** `479bfcb1..HEAD`, 19 files, +411/−36. Clean worktree, no untracked files excluded.
**Intent:** let a session-less visitor of a Signal K server that grants anonymous read land on a working read-only dashboard, without changing what an auto-login deployment (HaLOS) or a server without anonymous read does.

**Reviewers:** correctness, testing, maintainability, project-standards (always-on) · security (auth decision + permission gate) · api-contract (`StorageService` failure behavior, `ILoginStatus`, new exported union) · adversarial (411 added lines, auth-touching).

The server-side premises the change rests on were verified against the signalk-server source and all hold: `readOnlyAccess` mirrors `allow_readonly`, a tokenless request on such a server gets an `AUTO`/`readonly` principal, and GET on the global applicationData scope is not admin-gated.

---

## P1 — Critical

|---|---|---|---|---|---|
| 1 | `src/app/core/services/storage.service.ts:225` | `canPersist()` ignores the read-only context, so a session acquired mid-visit writes the shared config into the user's own profile | correctness, adversarial, api-contract, maintainability | 0.92 | gated_auto → needs decision |

Four reviewers reached this independently, by three different routes. `_isReadOnlyContext` is latched at bootstrap and consulted only by the UI locks; `canPersist()` is a live read of `loginStatus` and consulted only by the write guards. They can disagree, and both directions are broken:

**Corruption path.** A visitor boots anonymously: `sharedConfigName = 'default'` (the anonymous constant, `app-initNetwork.service.ts:307`) and `initConfig` = the *global* published config. `'default'` is also the reserved real user profile (`profile.service.ts:15`). The browser then acquires a Signal K session out of band — another tab signs in, which the new "Sign in" affordance actively invites. `SignalKDeltaService` re-probes `loginStatus` on the next WebSocket open (`signalk-delta.service.ts:156`), routine on a boat. `canPersist()` flips true; `_isReadOnlyContext` does not. Any Settings toggle now calls `patchConfig`, which builds its URL as `user/<appid>/<ver>` with path `/default/…` (`storage.service.ts:444`) — writing content derived from the shared global slot into the signed-in user's own profile. If the global config carries an older `configVersion`, the user's profile is downgraded and their next boot re-runs the migration chain over already-migrated data.

**Inverse.** A signed-in `readonly` user is not a read-only *context*, so `isReadOnlyContext()` is false for them: they can unlock the layout, edit, and have every save silently dropped by the new `patchConfig` early return — with no `patchFailure$` event, which is exactly the "save lost with no signal" state #552 set out to prevent.

Fix: `canPersist()` must be `this._auth.canWriteUserData() && !this._isReadOnlyContext`. A config loaded as someone else's shared view must not become writable because the session changed underneath it. Then route the UI locks (`dashboard.service.ts:548`, `app.component.ts:108`) through the same predicate so the two halves cannot drift, and consider leaving `sharedConfigName` unset in a read-only bootstrap so the existing empty-slot guard fires as a second line.

## P2 — High

|---|---|---|---|---|---|
| 2 | `VERSION` | Duplicate bump — `origin/main` is already at 1.4.0 | project-standards | 1.00 (verified) | manual |
| 3 | `src/app/core/services/dashboard.service.ts:547` | #552 R3 unmet: only the lock toggle is gated; every dashboard mutator is live and two unlock controls are now silently dead | adversarial, testing | 0.85 | manual |
| 4 | `src/app/core/services/app-initNetwork.service.ts:96` | No anonymous fallback when the auto-login redirect is refused — locked out exactly when SSO is down | correctness, adversarial | 0.85 | manual |
| 5 | `src/app/core/services/storage.service.ts:527` | `removeItem` refuses synchronously while `setConfig` rejects, so ProfileService's `.catch()` never attaches | api-contract | 0.82 | gated_auto |
| 6 | `src/app/core/services/app-initNetwork.service.ts:239` | Anonymous boot burns the one-shot v12→v13 migration, permanently losing the device's remote-control identity | adversarial | 0.75 | manual |
| 7 | `CHANGELOG.md` | User-visible behavior change ships in a minor bump with no entry | api-contract | 0.78 | manual |
| 8 | `CLAUDE.md:63,83` | Two architecture claims are now false | project-standards | 0.72 | gated_auto |
| 9 | `src/app/core/components/settings/display/display.component.ts:225` | Anonymous session can enable/reconfigure Signal K plugins — "read-only" holds for applicationData only | security | 0.62 | manual · **pre-existing → fix-here + file-issue** |

**#2** — `origin/main` moved to 1.4.0 nine commits ago (`60e74533`, PR #549). The workspace rule is "if VERSION already differs from the stable release, a prior PR opened the cycle — do not bump unless yours is a higher semver level." A minor in a minor-opened cycle is not higher. Commit `738614ea` rewrites the same four files main already changed; it conflicts on rebase and invites a resolution that re-runs `bumpversion` and ships a spurious 1.5.0. CI only emits a `::notice::` here, so nothing catches it.

**#3** — `isReadOnlySession()` gates `toggleStaticDashboard`/`setStaticDashboard` and nothing else. `add`, `update`, `delete`, `duplicate` and the editor's drag-reorder write the `dashboards` signal ungated, and the page-manager button that drives them is on an ungated toolbar. An anonymous visitor can delete every page of the shared dashboard they were sent to look at, get no indication nothing was saved, and no undo but a guessed reload. Meanwhile the Edit-page button and the bare `e` hotkey now call a swallowed `setStaticDashboard(false)` — they do nothing and say nothing. Embed never exposed this because embed unmounts the whole toolbar; this is the first force-locked context that still renders chrome. Combined with #1, each in-memory mutation becomes a real write the moment a session appears.

**#4** — `readOnlyAccess` is consulted only *before* the redirect attempt. With `oidcAutoLogin: true` and `allow_readonly: true`, a refused redirect (framed, `sessionStorage` untrackable, or budget spent after 3 bounces) falls to `auth-blocked` and a degraded shell. Concretely: Authelia is down, the display burns three redirect attempts, and the boat shows a sign-in recovery screen instead of the instruments the server would have served read-only. This is the failure mode the feature exists for.

**#9** — reachable before this change (via the degraded shell), but the change turns it from an error state into a supported session where "read-only" is the advertised contract. `PluginConfigClientService` has no write gate, and `POST /plugins/:id/config` has no authorization middleware server-side, so the `AUTO` readonly principal passes. Gate the client calls here; the server-side hole needs an upstream issue.

## P3 — Moderate

|---|---|---|---|---|---|
| 10 | `app-initNetwork.service.ts:318` | `loadAnonymousConfig` swallows 5xx/timeout as "nothing published" — a wrong dashboard presented as `ready` | correctness | 0.70 | gated_auto |
| 11 | `app-initNetwork.service.ts:321` | Published config accepted on `app` alone; a missing `notificationConfig` throws in `ToastService`, a non-array `dashboards` renders blank | adversarial | 0.70 | manual |
| 12 | `config.component.html:88` | Reset-to-default is the one ungated control in the card → persistent "resolve this before Skip can be used reliably" toast | correctness, adversarial | 0.80 | safe_auto |
| 13 | `app-initNetwork.service.ts:154` | Subscribe scope is decided pre-auth from the device's own profile and never recomputed → a shared dashboard's AIS/DSC widgets can stay silently empty | correctness | 0.65 | manual |
| 14 | `app.component.ts:108` | A stale published config is rendered unmigrated with *both* upgrade signals suppressed and no operator-visible warning | correctness, adversarial | 0.68 | manual |
| 15 | `storage.service.ts:162` | `storageServiceReady$` silently changed meaning; JSDoc stale, and `resetSettings` now reports a generic save failure instead of the real reason | api-contract | 0.70 | gated_auto |
| 16 | `app-initNetwork.service.ts:301` | `?profile=` is honored for a signed-in user and silently dropped for an anonymous one | api-contract | 0.63 | advisory |

**Test quality** (testing, maintainability):

| # | File | Issue | Conf. | Route |
|---|---|---|---|---|
| 17 | `app-initNetwork.service.spec.ts:472` | "never offers the missing-shared-config recovery" passes with the whole feature reverted — its one assertion is satisfied by the redirect path too | 0.85 | manual |
| 18 | `app-initNetwork.service.spec.ts:433` | The `#551` describe installs a `refreshLoginStatus` implementation that outlives it (`mockClear()` resets calls, not the implementation) | 0.70 | gated_auto |
| 19 | `app.component.ts:108` | The read-only upgrade gate — the only thing suppressing the instructions dialog — has no test; the sibling embed suite is a two-line extension | 0.85 | manual |
| 20 | `app-initNetwork.service.spec.ts:424` | `expect(resetBudget).not.toHaveBeenCalled()` is vacuous — `handleCookieAuth` has no path that calls it; the real anonymous boot *does* call it in the `finally` | 0.80 | manual |
| 21 | `app-initNetwork.service.spec.ts:432` | Dead setup: `setConnConfig(...)` is overwritten by `loadLocalStorageConfig()` on the next line of execution | 0.75 | safe_auto |
| 22 | `connection-status.component.spec.ts:21` | Reaches the real service's private `applyLoginStatus` when the public `refreshLoginStatus` + HTTP flush seam exists | 0.62 | advisory |

**Design//clarity** (maintainability):

| # | File | Issue | Conf. | Route |
|---|---|---|---|---|
| 23 | `config.default.factory.ts:11` | The import-cycle justification does not survive type erasure — `config.blank.dashboard`'s only edge to `dashboard.service` is a type-only import, elided at compile time, and no `import/no-cycle` rule exists. The module may still be right; the stated reason is not checkable | 0.70 | advisory |
| 24 | `storage.service.ts:225` | `canPersist()` is a pure alias for `canWriteUserData()`; with `isReadOnlyContext` and `isReadOnlySession` that is four names for two ideas | 0.68 | advisory |
| 25 | `profile.service.ts:198` | `buildBlankConfig()` is now a one-line delegate with one caller that re-clones an already-fresh copy | 0.80 | gated_auto |
| 26 | `storage.service.ts:64` | `_isLoggedIn` and its subscription are vestigial; the readiness log now prints `Authenticated false` on a fully ready service | 0.75 | gated_auto |
| 27 | `settings.service.spec.ts:137` | Comment still claims global app-service stubs, which CLAUDE.md says do not exist — and this PR added the local provider four lines below, proving it | 0.65 | safe_auto · pre-existing → fix-here |

## Requirements completeness

Plan source: the three closing-keyword issues (explicit).

- **#550** R1 ✅ · R2 ✅ · R3 ✅ — the branch table is genuinely covered including the fail-closed absent-field case. One cell is unpinned: `readOnlyAccess: true` + `oidcAutoLogin: true` with a *refused* redirect, which is finding #4.
- **#551** R1 ✅ · R2 ✅ · R3 ✅ · R4 ⚠️ partially — the security-disabled server still takes the `isLoggedIn` path and gets no bootstrap; it is no longer a dead end only because nothing else changed for it. Verify before claiming R4.
- **#552** R1 ✅ · R2 ✅ · **R3 ❌ not met** — R3 asked to "fix any [affordance] that does not key off it". The profile controls were already gated, but the dashboard mutators, the page-manager button, the Edit-page button, the `e` hotkey and the reset-to-default button are not (findings #3, #12). · R4 ✅

## Coverage

- Suppressed below the confidence gate: 0 findings.
- **Residual risks:** no code path in Skip writes the global scope, so the published anonymous dashboard has no in-app affordance and no documented recipe — most visitors will see the shipped defaults (this is what #553 is for). · `patchConfig`'s silent drop also fires for a fully authorized admin whenever `loginStatus` is transiently nulled by an authoritative non-2xx, e.g. a proxy 502 during a Signal K restart. · `RemoteDashboardsService` is gated on embed only, so an anonymous session still advertises remote-control paths under the device's UUID (the PUTs 403 and are console-only). · The HaLOS SSO bounce survives on one env var; a device provisioned without `SIGNALK_OIDC_AUTO_LOGIN` flips from "SSO or nothing" to "anonymous dashboard".
- **Testing gaps:** no test drives `SettingsService`/`DashboardService` from a read-only bootstrap, so "the visitor renders a working read-only dashboard" is asserted only as `bootstrapRemoteContext` argument shape. · No coverage of a signed-in `readonly` user through the dashboard path. · No coverage of a non-404 anonymous read failure. · No coverage of an anonymous boot in embed mode, the path the Freeboard panel iframes take — `handleCookieAuth` now returns `anonymous` before the `framed` branch that previously governed it. · The perf-harness mock answers `loggedIn`, so no scenario boots the anonymous path.

---

## Verdict: not ready

Fix order:

1. **#1** — one-line predicate fix plus routing the locks through it; this is a write into a real user's profile.
2. **#2** — rebase onto `origin/main` and drop `738614ea`. Do not resolve the conflict by re-running `bumpversion`.
3. **#3** — decide gate-vs-hide for the dashboard mutators and the two now-inert unlock controls; #552 R3 is not met without it.
4. **#4, #5, #6** — behavior fixes with clear shapes.
5. **#9** — gate the plugin client calls; file the signalk-server authorization issue.
6. **#12, #17, #18, #20, #21, #27** — small, mechanical.
7. **#7, #8** — CHANGELOG entry and the two false CLAUDE.md claims.

The P3 remainder is fine as follow-up issues, except #10 and #11, which decide what an operator sees when their published config is broken or unreachable — worth settling while the code is fresh.

mairas:
**Correction to finding #9.** The claim that the server does not refuse `POST /plugins/:id/config` from an anonymous readonly principal is wrong, and no upstream signalk-server issue is warranted.

The finding cited `src/interfaces/plugins.ts:928`, where the per-plugin router registers `post('/config')` with no middleware. It missed the app-level gate: `src/tokensecurity.ts:745` wraps `${SERVERROUTESPREFIX}/plugins` and `:759` wraps the legacy `/plugins`, both with `adminAuthenticationMiddleware(false)` via `app.use`, which applies to every method on those subtrees. An `AUTO`/`readonly` principal fails `hasAdminAccess` and gets a 403.

What survives is a UX defect, not a vulnerability: the Display settings form offers an action that will 403 for a read-only visitor, in the same family as the ungated reset-to-default button (#12). Re-rated **P3**, no upstream issue, no security impact.

mairas:
Review addressed. Force-pushed: the branch is rebased onto current `main` (which is why the SHAs changed) and carries one fix commit on top.

**P1 #1 — fixed.** `canPersist()` is now `canWriteUserData() && !isReadOnlyContext`, and every lock and affordance reads one signal, `DashboardService.isReadOnlySession`, rather than picking one of the two predicates. That closes the corruption path (an anonymous tab acquiring a session and writing the shared config into the user's `default` slot) and the inverse case (a signed-in `readonly` user editing a layout whose saves were dropped). Making it a signal rather than a method also fixed a bug the fix itself introduced: as a plain method the toolbar's hidden controls would not have re-evaluated under OnPush when the session changed.

**P2 — all fixed** except #9, corrected above.

- **#2** — `738614ea` dropped; the branch is rebased and carries main's 1.4.0. The cycle was already open at minor level, so no bump belongs here.
- **#3** — the dashboard mutators refuse in a read-only session, and the page-management and edit controls are hidden rather than left visible-but-inert. Covered by a toolbar test that asserts both controls are present for a writable session first, so the absence is the guard and not the fixture.
- **#4** — a refused auto-login redirect (framed, budget spent, untrackable budget) now falls through to read-only when the server grants anonymous read, with the auth-blocked issue cleared. Three tests: budget-exhausted, framed, and the negative case where the server grants no anonymous read.
- **#5** — `removeItem` is `async`, so the `.catch()` its ProfileService callers attach actually attaches. Spec updated to `rejects`.
- **#6** — the one-shot remote-control migration is skipped on the anonymous outcome, with a test asserting the stored `configVersion` stays at 12.
- **#7, #8** — CHANGELOG entries under New Features and Behavior changes; the two false CLAUDE.md claims corrected, and the Auth & profiles paragraph now documents the third session shape and both write gates.

**P3 — fixed:** #10 (a non-404 fetch failure is rethrown instead of passing the shipped dashboards off as the published one), #11 (the published config is screened with the same shape predicate as an imported profile, now shared in `config-shape.util.ts`), #12 (reset-to-default gated), #15 (readiness JSDoc says reachable, not permitted), #16 (an ignored `?profile` says so), #17/#18/#20/#21 (the vacuous assertions, the leaking mock implementation, and the dead setup), #19 (two AppComponent tests for the guard, plus the positive case), #23 (the cycle claim replaced with the reason that is actually checkable), #25/#26/#27 (the one-line delegate, the vestigial login tracking and its misleading log line, the stale spec comment).

**Deferred:** #13 → [#561](https://github.com/halos-org/skip/issues/561). The subscribe scope is chosen before the auth outcome is known, so fixing it means either failing open to `all` or re-initializing the connection after the anonymous config loads — a change to the connection lifecycle, not to the read-only path.

**Not taken:** #14 in the form suggested. Migrating the published config in memory needs `ConfigurationUpgradeService`, which injects `SettingsService` — that edge pulls settings into the bootstrap's DI graph and broke 20 specs when I tried it. A stale published config is now rendered with a warning naming its version and the current one, which tells the operator what to republish without the read-only session rewriting anything. #22 and #24 left alone: #22's private-method access matches the pattern used elsewhere in the suite, and #24 dissolved once `canPersist()` gained its second clause.

`npm run ci` clean: lint, strict-null gate, 2006 unit tests (up from 1983), 33 schema tests. Still not exercised against a live server with `allow_readonly: true` and no OIDC, nor on a HaLOS device — that is the remaining verification before merge.

PR #565, opened 2026-08-12
fix(auth): send the login return address as redirect

### LINKED ISSUE #554: fix(auth): login redirect sends returnTo, but Signal K reads redirect
## Problem

Skip's login redirect carries a `returnTo` parameter that no Signal K server reads, so a user who signs in from Skip is dropped on the Signal K admin root instead of coming back to the page they left.

`buildLoginRedirectUrl` emits `returnTo` (`src/app/core/utils/login-redirect.util.ts`), and both login endpoints read `redirect`:

- the OIDC login endpoint takes `req.query.redirect` and falls back to `'/'` when it is absent (`signalk-server/src/oidc/oidc-auth.ts:281-284`), and that value is what the callback restores;
- the admin login page reads a `redirect` param out of the hash query (`packages/server-admin-ui/src/views/security/Login.tsx:41-43`, helpers in `views/security/loginRedirect.ts`).

The string `returnTo` does not occur anywhere in signalk-server. `noAutoLogin`, the other parameter Skip sends, is read (`Login.tsx:89`), which is presumably why the mismatch went unnoticed.

This hits HaLOS today, on the ordinary path: a user without a Signal K session opens Skip, gets bounced to Authelia, signs in, and lands on the admin UI. Not tested on a device — this is what the code says.

## Scope

- R1. Emit `redirect` instead of `returnTo`, in the same component of the URL the current code already picks (query for the OIDC endpoint, hash query for `/admin/#/login`).
- R2. Keep `noAutoLogin` as-is.
- R3. Keep Skip's own `isSafeReturnTo` validation. Both servers validate too (`isSafeRelativeUrl`, `isSafeRedirect`), but an open-redirect guard on the constructing side is worth its few lines.

## Done when

- Signing in from Skip on an OIDC server returns to the Skip page the user started on.
- Signing in from Skip via `/admin/#/login` does the same.
- The self-route loop guard still rejects Skip's own `/login` as a target.

Independent of the anonymous read-only work; separate branch and PR.

mairas:
## Why

Skip's sign-in redirect carried the page to return to as `returnTo`, a name neither Signal K login endpoint reads. Both read `redirect`: the OIDC login endpoint takes it from the query string (`src/oidc/oidc-auth.ts`, `req.query.redirect`), and the admin login page reads it out of its hash query (`packages/server-admin-ui/src/views/security/Login.tsx`), honoring it after both a password login and an SSO bounce. The parameter Skip sent was therefore dropped, the server fell back to `'/'`, and the user landed on the Signal K admin root rather than the Skip page they left.

This hits HaLOS on the ordinary path: a user without a Signal K session opens Skip, is bounced to Authelia, signs in, and ends up in the admin UI.

## How

`buildLoginRedirectUrl` emits the value under the server's wire name, held in one constant next to the comment explaining where each endpoint reads it. `returnTo` stays as Skip's internal vocabulary for the option and its validator — the change is the name on the wire, not the concept. Placement is unchanged and still correct for both endpoint shapes: query for the OIDC URL, hash query for `/admin/#/login`.

`noAutoLogin` and `isSafeReturnTo` are untouched, per the issue's R2 and R3.

## Verification

`npm run ci` clean: lint, strict-null gate, 2019 unit tests, 33 schema tests. The redirect spec now pins the wire name on both endpoint shapes, including a new case for the admin hash route with a realistic Skip return path.

Not exercised end-to-end against a live server — verified by reading both endpoints in `signalk-server` (the admin-side `redirect` support is commit `7c84f297`, present in `upstream/master`).

Fixes #554


mairas:
## Code review — PR #565

**Scope:** `fa8c6e9c..963aaec1`, 4 files (2 source, 2 docs). Untracked `perf-harness/*.mjs` excluded — pre-existing, unrelated.
**Intent:** rename the wire parameter Skip sends on its sign-in redirect from `returnTo` (which no Signal K endpoint reads) to `redirect` (which both do), without changing placement, validation, or `noAutoLogin`.

**Reviewers:** correctness, testing, maintainability, project-standards (always-on); security (a user-influenced return path, and a previously-inert parameter becoming live); api-contract (an unversioned cross-repo parameter); adversarial (auth code).

All fixes below are applied in `3247b9f1` and `b5ea7f14`.

### P2 — High

|---|---|---|---|---|---|
| 1 | `docs/signalk-auth-plan.md:185` | The plan still records `returnTo` as a verified Signal K fact, twice as a decision, and once as an open question asking someone to confirm it. That claim is what produced #554. The PR corrected the sequence diagram and left the prose contradicting it. | maintainability, project-standards, api-contract, correctness, adversarial | 0.90 | safe_auto → **fixed** |
| 2 | `sso-redirect.service.ts:116` | A framed sign-in navigates the top window but took its return address from the iframe's URL, which carries `?embed=1`. Making the parameter live turns a wrong destination into a dead end: no toolbar, dashboard locked, and `location.replace` leaves no history entry back to Freeboard-SK. Previously the ignored parameter left the user on the admin root, which is at least navigable. | correctness, adversarial (security concurs) | 0.90 | manual → **fixed** |
| 3 | `login-redirect.util.ts:1` | The comment records the correct name but not that Signal K's own OIDC reference documents the wrong one. Anyone re-checking the constant against the official docs concludes it is wrong and has a citation for reverting it. | maintainability | 0.82 | safe_auto → **fixed** |

### P3 — Low

|---|---|---|---|---|---|
| 4 | `sso-redirect.service.spec.ts:43` | Nothing asserted the return address at the caller. This service is the only place the real `window.location` is composed and handed to the builder, and the spec checked the login path and `noAutoLogin` alone — which is how #554 shipped with a green suite. | testing (correctness, security, api-contract flag the same gap) | 0.85 | safe_auto → **fixed** |
| 5 | `CHANGELOG.md:14` | "Applies to both the SSO login and the server's own login page" was unqualified. The admin half needs server 2.27.0; the OIDC half works back to 2.20.0. | api-contract | 0.80 | safe_auto → **fixed** |
| 6 | `login-redirect.util.spec.ts:58` | Comment narrated what the code used to send. The durable rationale already sits next to the constant. | project-standards | 0.70 | safe_auto → **fixed** |

| # | File | Issue | Reviewers | Conf. | Action |
|---|---|---|---|---|---|
| 7 | `sso-redirect.service.ts:117` | Same defect class as #554, one parameter over: `noAutoLogin` is read only by the admin login page, so on an OIDC server the recovery sign-in has no server-side loop protection — and `manualSignIn()` resets Skip's own budget on every call. | maintainability, api-contract | 0.90 | **file-issue** → [#566](https://github.com/halos-org/skip/issues/566). A comment recording which endpoint reads it is in this PR; the behavior question is not. |
| 8 | `signalk-server/docs/oidc.md:389` | The upstream root cause. The OIDC reference documents the login parameter as `returnTo`; the endpoint reads `redirect`. `docs/develop/webapps.md:302` documents it correctly, so the two upstream pages disagree — and this repo's plan cited the wrong one. | api-contract, maintainability | 0.90 | **owner: human.** `SignalK/signalk-server` is a repo we do not own; filing needs an explicit go-ahead. Recorded in the plan and the code comment meanwhile. |

### Requirements completeness (issue #554, `plan_source: explicit`)

- **R1** — emit `redirect` in the same URL component: met. Query for the OIDC URL, hash query for `/admin/#/login`, placement unchanged.
- **R2** — keep `noAutoLogin` as-is: met. Behavior untouched; finding 7 is an observation, not a change.
- **R3** — keep `isSafeReturnTo`: met. Untouched.

Of the three "done when" criteria, the self-route guard is covered by unit tests. The two sign-in round trips are cross-repo and cannot be unit-tested — they need a device check.

### Coverage

Security returned zero findings. It traced the value from `window.location` through `isSafeReturnTo`, `encodeURIComponent`, and both server sinks, and confirmed that making the parameter live opens no redirect: the OIDC endpoint stores it in an encrypted state cookie and ends at a same-origin `res.redirect`, and the admin page's `isSafeRedirect` requires a leading `/`, blocking `javascript:`. The value never reaches the identity provider.

Adversarial checked the loop-guard near misses. `#/login/` and `#/Login` do not match Skip's route (they fall through to `**`), so the exact-match guard holds; `#/login;x=1` would pass it, but nothing produces such a URL.

Compatibility, for the record: the change degrades, it never regresses. `req.query.redirect` has been in the OIDC endpoint since `f7924867`, the commit that introduced OIDC (v2.20.0); the admin page has read it since `7c84f297` (v2.27.0). No version ever read `returnTo`, so a server that reads neither falls back to `/` — exactly today's behavior. Emitting both names was considered and rejected: it would add a permanently dead parameter and re-create the ambiguity that caused this bug.

**Residual risks** (not actioned):

- Skip's `isSafeReturnTo` and the server's `isSafeRedirect` are independent implementations of the same rule and have already drifted — Skip rejects `0x7f`, the server does not; the server rejects the admin login route by regex, Skip by an exact-match list. They agree on every target Skip currently emits, and nothing keeps them aligned.
- The return address now travels inside the OIDC state cookie. A multi-kilobyte return URL could push it past the browser's 4 KB limit and fail the login rather than just losing the return. No Skip path generates one near that size.
- `manualSignIn` falls back to `/admin/#/login` when OIDC is off. If the session cookie is in fact still valid, that page renders the admin dashboard and never acts on `redirect`. Server-side and unchanged here, but this PR is what makes the missed return noticeable.

**Testing gaps** (not closed): nothing verifies Skip's emitted value against the server's own validators, or that the inner `#` of a hash-routed target survives the OIDC round trip. Both are cross-repo and only checkable on a device.

### Verdict

**Ready with fixes** — all six actionable findings are applied, and `npm run ci` is clean: lint, strict-null gate, 2023 unit tests (up from 2019), 33 schema tests.

Two things remain open and neither blocks this PR: [#566](https://github.com/halos-org/skip/issues/566) for the `noAutoLogin` behavior question, and a decision on whether to report the upstream documentation bug to `SignalK/signalk-server`.

Before merge, the device check is worth doing: sign in from Skip on the OIDC path and confirm you return to the page you left with its hash route intact.

PR #567, opened 2026-08-12
docs(help): publishing the anonymous default dashboard

### LINKED ISSUE #553: docs: publishing the anonymous default dashboard to the global applicationData scope

mairas:
## Why

[#560](https://github.com/halos-org/skip/pull/560) made a session-less visitor a supported case: on a server that grants anonymous read, Skip loads the `default` slot from the global application-data scope instead of redirecting to a login page. Nothing in Skip writes that slot — deliberately, since anything that could write it from a browser session could also overwrite it from one — so an operator has to put it there by hand, and nothing documented how.

This is the second half of the request on [SignalK/signalk-server#2931](https://github.com/SignalK/signalk-server/pull/2931): "a way to configure things so that a new user at demo.signalk.org gets a demo dashboard."

## What's here

A new help page, **Read-Only Access Without Signing In**, under Advanced Features. It covers the two server prerequisites (`Allow Readonly Access` on; OIDC auto-login off, or Skip sends the visitor to sign in instead), the publishing recipe, what a read-only session refuses, and when to republish.

The recipe is export-and-upload rather than hand-authored JSON: **Settings > Configurations > Advanced > Download** already emits exactly the `{app, dashboards, theme}` shape the slot expects, so an operator builds the dashboard in Skip as an ordinary user and POSTs the file. The four fixed URL segments (`global` / `skip` / `11` / `default`) are spelled out in a table, since all four have to match or the visitor's fetch quietly returns nothing. The verification step notes that a never-created slot answers `{}` rather than 404, which otherwise reads as a published-but-empty dashboard.

**Login & Configurations** carried a claim that [#560](https://github.com/halos-org/skip/pull/560) made false — that without a signed-in account, server-stored configuration is unavailable and Skip runs on connection settings alone. Corrected, and pointed at the new page.

## Verification

The recipe is the one confirmed on hardware while testing [#560](https://github.com/halos-org/skip/pull/560): an admin POST to `/signalk/v1/applicationData/global/skip/11/default` returned 200, an anonymous GET of the same path returned the document, and an anonymous write returned 401. Re-checked against the code for this PR: `addAdminWriteMiddleware` is applied to the global applicationData URLs (`src/interfaces/applicationData.js`), so reads are ungated and writes are admin-only; the JSON login endpoint returns `{timeToLive, token}`; the UI label is `Allow Readonly Access` in the server's Security > Settings.

A new spec checks every `menu.json` entry against a file on disk — a mistyped entry otherwise fails only when a reader opens that page, as an empty panel. Mutation-checked: renaming the new entry to a nonexistent file fails the spec.

`npm run ci` clean: lint, strict-null gate, 2019 unit tests across 179 files, 33 schema tests.

Closes #553

### DIFF
--- a/src/app/core/components/app-help/help-docs-menu.spec.ts
+++ b/src/app/core/components/app-help/help-docs-menu.spec.ts
@@ -1,0 +1,33 @@
+import { describe, it, expect } from 'vitest';
+import { existsSync, readFileSync } from 'node:fs';
+import { join } from 'node:path';
+import { cwd } from 'node:process';
+
+/**
+ * The help menu names its pages by filename, and the viewer fetches them at runtime. A renamed or
+ * mistyped entry therefore fails only when a reader opens that page, as an empty panel — nothing in
+ * the build or the app notices. Check the two sides against each other here instead.
+ */
+const HELP_DOCS_DIR = join(cwd(), 'src/assets/help-docs');
+
+interface MenuEntry {
+  title: string;
+  file?: string;
+  items?: MenuEntry[];
+}
+
+function collectFiles(entries: MenuEntry[]): string[] {
+  return entries.flatMap(entry => [
+    ...(entry.file ? [entry.file] : []),
+    ...(entry.items ? collectFiles(entry.items) : [])
+  ]);
+}
+
+describe('help-docs menu.json', () => {
+  const menu: MenuEntry[] = JSON.parse(readFileSync(join(HELP_DOCS_DIR, 'menu.json'), 'utf8'));
+
+  it('names a markdown file that exists for every entry', () => {
+    const missing = collectFiles(menu).filter(file => !existsSync(join(HELP_DOCS_DIR, file)));
+    expect(missing).toEqual([]);
+  });
+});
--- a/src/app/core/services/app-initNetwork.service.ts
+++ b/src/app/core/services/app-initNetwork.service.ts
@@ -225,8 +225,8 @@ export class AppNetworkInitService implements OnDestroy {
             throw error;
           }
           if (!remoteConfig?.app) {
-            // SK answers a never-created slot with 200 {} rather than a 404, so an appless config is
-            // the real "no shared configuration yet" signal.
+            // A slot that was never written 404s, handled above; an appless 200 is the other shape
+            // "no shared configuration yet" arrives in, so it lands in the same degraded state.
             startupDegraded = true;
             this._bootstrapIssue$.next({
               reason: 'missing-shared-config',
@@ -335,8 +335,8 @@ export class AppNetworkInitService implements OnDestroy {
   }
 
   /**
-   * The published global config, or the shipped defaults. SK answers a never-created slot with
-   * 200 {} rather than a 404, so an appless body means "nothing published" just as a 404 does —
+   * The published global config, or the shipped defaults. A slot that was never written 404s, and
+   * an appless 200 body means "nothing published" just as that 404 does —
    * but nothing else does: a 5xx or a timeout means the operator's dashboard exists and could not
    * be fetched, and silently showing a different one as if it were the boat's is worse than the
    * recovery state, so it is rethrown to the bootstrap's handler.
--- a/src/assets/help-docs/anonymous-default.md
+++ b/src/assets/help-docs/anonymous-default.md
@@ -47,7 +47,7 @@ curl -s -X POST "$SERVER/signalk/v1/applicationData/global/skip/11/default" \
   --data-binary @SkipConfig.json
 ```
 
-The URL is fixed in three places, and all three have to match or Skip will not find what you published:
+Every segment of that URL is fixed, and all four have to match or Skip will not find what you published:
 
 | Segment | Value | Meaning |
 |---|---|---|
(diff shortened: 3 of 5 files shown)


mairas:
## Code review — PR #567

**Scope:** `fa8c6e9c..5df72140`, 4 files. **Intent:** document how an operator publishes the dashboard an anonymous visitor sees, and correct a claim [#560](https://github.com/halos-org/skip/pull/560) made false.

**Reviewers:** correctness (fact-checking every claim against the implementation — wrong documentation here fails silently: the operator publishes to a URL Skip never reads and sees nothing) and project-standards. Right-sized down from the usual team for a docs change.

All findings are applied in `75b740c6`.

### P2 — High

| # | File | Issue | Reviewer | Conf. | Route |
|---|---|---|---|---|---|
| 1 | `anonymous-default.md:67` | The verification step said a never-published slot answers `{}`. It answers **404 with an empty body**. An operator who publishes to the wrong URL gets a 404 and reads it as a malformed request, and one who gets an empty line cannot tell which case they are in. | correctness | 0.85 | **fixed** |

Checked on hardware rather than taken on trust, because it contradicted a comment this repo merged last week:

```
GET .../applicationData/global/skip/11/default   -> HTTP 404, 0 bytes   (unauthenticated and with an admin token)
GET .../applicationData/global/skip/11           -> HTTP 200, {}
```

The `200 {}` is real, but it belongs to the **version file** one level up, not to the named slot. Two comments in `app-initNetwork.service.ts` from [#560](https://github.com/halos-org/skip/pull/560) stated it of the slot; both are corrected here. The code was never wrong — `loadAnonymousConfig` catches the 404 and falls back, and `app-initNetwork.service.spec.ts:477` already asserts exactly that — so this was a description defect, not a behavior one. (The reviewer listed the missing 404 spec as a testing gap; that spec exists.)

### P3 — Low

|---|---|---|---|---|---|
| 2 | `anonymous-default.md:73` | "Settings changes … are refused rather than silently dropped" overstated it. Page and layout mutations are refused (`DashboardService.refuseMutation`), but the autosave path drops silently by design (`storage.service.ts:433`, and the comment there says so). | correctness | 0.65 | **fixed** |
| 3 | `anonymous-default.md:50` | The sentence promised three fixed URL segments; the table lists four. A reader checking their URL cannot tell which one is optional. | correctness, project-standards | 0.88 | **fixed** |
| 4 | `configuration.md:21` | The cross-reference named the new page in bold and told the reader where to look. Six existing pages link with `[Title](#/help/file.md)`, which the viewer renders as a working link. | project-standards | 0.72 | **fixed** |
| 5 | `anonymous-default.md:77` | "exactly as before" is development history — a reader who never used a pre-anonymous Skip has no before. | project-standards | 0.66 | **fixed** |

### Verified correct (no change needed)

The correctness reviewer traced these to their source and they hold: the published URL and all four of its segments against `storage.service.ts`, `SERVER_CONFIG_APPID` and `REMOTE_CONFIG_FILE_VERSION`; the access model (`addAdminWriteMiddleware` on the global URLs makes reads ungated and writes admin-only, and with readonly off `http_authorize` answers 401, so the stated prerequisite is real); the login recipe and its `{timeToLive, token}` response; that **Download** emits exactly the `{app, dashboards, theme}` shape `isValidConfigShape` accepts; the two prerequisites against what `handleCookieAuth` branches on; the `?profile=` warning; the version-mismatch behavior; the `Allow Readonly Access` label; and that the new menu spec fails when an entry names a missing file.

One nuance the page does not mention and does not need to: with auto-login on, an exhausted redirect budget also falls through to anonymous when readonly is set. The first visit still redirects, so the page's guidance holds.

### Coverage

No version bump is owed. `VERSION` is 1.4.0 against the latest stable `v1.3.0+2`, so a prior PR opened the cycle; independently, `version-bump-check.yml` excludes `.*\.md$` and `.*\.spec\.[^/]+$` from package-affecting files.

**Residual risk:** the page hardcodes `.../global/skip/11/default`, but `11` is `REMOTE_CONFIG_FILE_VERSION` and `skip` is `SERVER_CONFIG_APPID`. A file-version bump makes the published instructions wrong with nothing failing to catch it.

**Testing gap left open:** `help-docs-menu.spec.ts` checks menu → file only. A page on disk but absent from the menu is unreachable in the app and still passes. Both directions currently agree (verified: no page is missing from the menu, no entry is missing a page), so this is a guard that could be tightened rather than a present defect.

### Verdict

**Ready with fixes** — all five findings applied. `npm run ci` clean: lint, strict-null gate, 2019 unit tests across 179 files, 33 schema tests.

PR #568, opened 2026-08-12
fix(deploy): verify the device serves the build just synced

### LINKED ISSUE #562: fix(deploy): deploy-halos reports success while the device still serves the previous build
## Problem

`./run deploy-halos <host>` can report success while leaving the *previous* build being served, and it fails outright on a device where the framework owns `system-plugins/`.

Observed on halosdev.hal deploying from a feature branch:

```
==> Linking, registering the dependency, and (re)starting Signal K if needed...
  registered @halos-org/skip -> file:system-plugins/skip
  @halos-org/skip already mounted; skipping restart (express.static serves the new build on next request)
  OK: @halos-org/skip is served by Signal K (listed in /skServer/webapps)
  https://halosdev.hal:4430/@halos-org/skip/ -> HTTP 200
```

After that "OK", the served page still referenced the old bundle:

```
local  public/index.html          -> main-PWJFROYF.js
served /@halos-org/skip/          -> main-AJ4JYROG.js
       /@halos-org/skip/main-PWJFROYF.js -> 404
```

`systemctl restart marine-signalk-server-container.service` fixed it: the same URL then served `main-PWJFROYF.js`. So the restart-skip heuristic was wrong here, and every check the script performs still passed — the package was registered, the name appeared in `/skServer/webapps`, and the URL returned 200. A deploy that silently leaves the old code running is worse than one that fails, because the next thing anyone does is test the old build and draw conclusions from it.

Two defects:

1. **The restart heuristic trusts the wrong signal.** "Listed in `/skServer/webapps`" says a module with that name is known, not that the live `express.static` mount resolves to the payload just synced.
2. **`mkdir -p "$sysdir"` runs as the invoking user.** On a device where `system-plugins/` was created by the framework (root-owned, holding `signalk-halpi`), the PREP step dies with `Permission denied` before anything is synced. Pre-creating `system-plugins/skip` as `pi` worked around it.

## Scope

- R1. Verify the deploy by comparing the hashed entry bundle in the local `public/index.html` against the one the device serves, and restart when they differ. The current 200-check cannot tell a fresh build from a stale one.
- R2. Create the target directory with the privilege the parent actually requires, rather than assuming the invoking user owns `system-plugins/`.
- R3. Fail loudly when the post-deploy verification does not match, instead of printing OK.

## Done when

- A deploy onto a device already serving an older Skip ends with the device serving the new bundle, or with a non-zero exit and a message saying it did not.
- A first deploy onto a device with a root-owned `system-plugins/` succeeds without manual preparation.

mairas:
## Why

`./run deploy-halos` reported success on halosdev while the device kept serving the previous bundle — local `public/index.html` referenced `main-PWJFROYF.js`, the device served `main-AJ4JYROG.js`, and the new bundle 404'd. Every check the script performs passed, because every one of them is also true of the build being replaced: the package stays registered, `/skServer/webapps` keeps listing it, and the URL keeps answering 200. A restart fixed it.

Second defect, same function: `mkdir -p "$sysdir"` runs as the invoking user, and `system-plugins/` belongs to the container-app framework, which creates it as root on a device that ships plugins there (`signalk-halpi`). The PREP step died with `Permission denied` before anything synced.

A deploy that silently leaves the old code running is worse than one that fails — the next thing anyone does is test the old build and draw conclusions from it.

## How

The hashed entry bundle is the one thing that names *this* build, so the script now asks the device for it (`serves_this_build`) after the sync. Matching means the deploy landed and no restart is needed; not matching restarts Signal K and re-checks; still not matching exits non-zero. It also fetches the bundle itself, since an index naming a file is not the same as the file being fetchable.

The restart-skip heuristic is gone. It was right about the mechanism — `express.static` reads `public/` from disk per request, so a content-only update usually needs no restart — but "usually" was standing in for a check.

Two things fixed along the way:

- The restart went through `sudo docker restart signalk-server`, against the workspace rule: container apps restart through their systemd unit, which owns the provision/prestart step and the dependency ordering. Now `marine-signalk-server-container.service`.
- The target directory is created with the privilege its parent actually needs, then handed to the invoking user so the rsync that follows needs none.

## Verification

All three paths exercised against `halosdev.hal`:

1. **Normal deploy** — synced a fresh build, device already served it (no restart needed), verified `main-PMMECG3L.js`, exit 0.
2. **Fail-loud path** — with the expected bundle name replaced by `main-NOSUCHBUILD.js`, the script restarted Signal K, re-checked, printed `FAIL: ... still does not reference main-NOSUCHBUILD.js after a restart`, and exited **1**. The first attempt at this exposed a real bug in the fix: the return code was captured and never checked, so it printed FAIL and exited 0.
3. **Root-owned parent** — moved `system-plugins/skip` aside so the deploy had to create it under the root-owned parent. It did, chowned it to `pi:pi`, synced, and verified. The parent is still `root:root` afterwards, as it should be.

Device left as found: test backups removed, `skip.pretest` deleted, serving `main-PMMECG3L.js`.

Closes #562


mairas:
## Code review — PR #568

**Scope:** `fa8c6e9c..af1d09ec`, the `deploy-halos` function in `run`. **Reviewer:** correctness, on the shell hazards specifically — errexit/pipefail interaction, the quoted heredocs, and the sudo path. Right-sized to one persona for dev tooling; the change had already been exercised on `halosdev.hal` along its three main paths.

It found three things those runs could not have exposed. All are fixed in `b8d0b4e4`.

### P2 — High

| # | Line | Issue | Conf. | Route |
|---|---|---|---|---|
| 1 | `run:114` | **The guard this PR added was dead code.** `entry=$(grep ... \| head -1)` as a bare statement exits the script under `errexit` when grep matches nothing, so the `if [[ -z "$entry" ]]` diagnostic below it could never run. The operator would see the deploy stop after "Syncing build" with no output and a bare exit, `public/` already synced and everything after it skipped. | 0.95 | **fixed** |
| 2 | `run:93` | `mkdir -p` returns 0 for a directory that already exists, whatever owns it. The fallback covered the shape I tested — root-owned parent, absent leaf — but not a root-owned `system-plugins/skip` itself, which is what a device that received Skip through the container-app framework has. There `mkdir -p` succeeds, the chown is skipped, and the rsync two lines later fails. | 0.70 | **fixed** |

Finding 1 is confirmed rather than reasoned about:

```
$ bash -c 'set -eo pipefail; entry=$(grep -o zzz /etc/hostname | head -1); echo "GUARD REACHED"'
grep: /etc/hostname: No such file or directory
exit=2                       # "GUARD REACHED" never prints
```

The assignment is the `if` condition now, which suspends `errexit`, and the same construction reaches the guard and returns 1 with its message.

Finding 2 is fixed by testing writability rather than creation, and re-verified on hardware: with `system-plugins/skip` chowned to `root:root`, the deploy now takes the sudo branch, hands the directory back to `pi`, and verifies — `drwxr-xr-x root root` before, `drwxr-xr-x pi pi` after, exit 0.

### P3 — Low

| # | Line | Issue | Conf. | Route |
|---|---|---|---|---|
| 3 | `run:94` | `sudo mkdir -p` creates every missing parent, so a mistyped hostname or an unprovisioned device would get a root-owned tree plus a full copy of the build, failing only later at `mkdir node_modules`. Before this PR that case failed immediately on the unprivileged `mkdir`. | 0.75 | **fixed** |
| 4 | `run:208` | The closing summary named a stale bundle as the cause for every nonzero `rc`, including a failed dependency registration, a failed restart, or a missing CA file. | 0.85 | **fixed** |

Finding 3 restores fail-fast: the remote step refuses a host with no `/var/lib/container-apps/marine-signalk-server-container/data/data` before any `sudo` runs. Verified standalone against a nonexistent path — it exits 1 with the message and creates nothing.

Also taken from the residual risks: `serves_this_build` could not distinguish "the device serves a different build" from "the check could not run", so a missing CA or an unresolvable hostname would restart the marine stack and then report a stale bundle. It returns a distinct status for an unfetchable page now, and the post-restart failure says which happened. Capturing the body instead of piping into `grep -q` also removes the `pipefail`/SIGPIPE interaction the reviewer flagged, which was only safe by accident of where the bundle reference sits in `index.html`.

### Verification

Re-run on `halosdev.hal` after the fixes: normal deploy verified and exit 0; root-owned leaf handled; host guard and entry guard confirmed standalone. The earlier fail-loud test (deliberately wrong expected bundle → restart → FAIL → exit 1) still holds.

Device left as found: test backups removed, serving `main-PMMECG3L.js`.

### Coverage

**Testing gaps left open**, all needing a device shape I do not have: a run against a host lacking the container-app directory over ssh (checked standalone instead), and a remote-step failure for a reason other than a stale bundle.

**Residual risk:** the post-restart check runs once, immediately after `/skServer/webapps` answers, assuming static mounts are live no later than that endpoint. True for the Signal K versions in use; if it stopped being true, a good deploy would hard-fail with no retry.

### Verdict

**Ready with fixes.** No `VERSION` bump: `run` is dev tooling, excluded from package-affecting files, and the cycle is already open at 1.4.0.

Issue #570, opened 2026-08-12
Mass and area paths resolve to unitless: no conversion group exists for kg or m2


mairas:
Split out of the #536 fix, which completed `SERVER_TARGET_UNIT_ALIASES` against the Signal K server's built-in unit-preference presets. Two categories could not be completed there, because the gap is not an alias — Skip has no conversion group for them at all:

| category | SI base | preset targets |
| --- | --- | --- |
| `mass` | `kg` | `kg`, `pound` |
| `area` | `m2` | `m2`, `sqft` |

`getConversionsForPath` filters `_conversionList` for a group containing a measure equal to the path's SI unit. No group contains `kg` or `m2`, so the filter comes back empty and the path degrades to `unitless` — even for the identity target, and on every preset including metric. `design.displacement` (kg) and a sail area (m2) therefore render a raw number with no label.

Both are small additions to `_conversionList` plus `unitConversionFunctions` entries (js-quantities handles `kg -> lbs` and `m^2 -> ft^2`), and the existing preset-vocabulary spec case in `units.service.spec.ts` documents them as the deliberate exclusions — extend that table in the same change.

Note `skBaseUnits` already lists both `kg` (Mass) and `m2` (Area) as valid Signal K units, so the type surface expects them; only the conversion table is missing.

PR #571, opened 2026-08-12
Honour every unit the Signal K server's presets ask for

### LINKED ISSUE #536: [Bug]: Fuel rate has no display format
### Operating System

Raspberry Pi OS

### Browser

Chromium

### Skip Version

1.3.0

### Signal K Server Version

2.30.0

### Bug Description

After creating a widget for showing fuel rate it shows '**unitless**' and there is no way to configure it like was possible in KIP:

Image

My Metadata seems ok (to me):

Image

### Steps to Reproduce

Create widget with path "propulsion.0.fuel.rate".

Image

My apologies upfront if this is my mistake and not a bug at all.

mairas:
Fuel rate rendered as `0.00 unitless` instead of `7.2 l/h`, with no way to pick a display format.

## Why

Since the Phase 2 units work, a path's display measure comes from the Signal K server's `meta.displayUnits.targetUnit`, mapped onto a Skip measure through `SERVER_TARGET_UNIT_ALIASES` and then checked for membership in the path's conversion group. A target that fails that check degrades the path to `unitless` — the raw SI value with the literal word "unitless" as its label, and no per-widget override left to correct it.

The table carried five aliases. The server's six built-in presets emit far more than five names, and the metric ones publish the `volumeRate` category as **`L/h`** where Skip's Flow measure is `l/h` — so every fuel rate on a metric preset resolved to nothing. `km/h`, `F`, `kilometer`, `mile` and `foot` were broken the same way; `gallon-imp`, `gal-imp/h` and `btu` had no Skip measure at all.

## What changed

- Completed the alias table against every `(baseUnit, targetUnit)` pair the built-in presets can emit, read from signalk-server `unitpreferences/presets/*.json`.
- Added the three missing measures with their conversions: imperial gallon, imperial gallons per hour, BTU.
- Gave the Unitless group an empty symbol, so a path the server states no preference for renders its SI value with no label rather than the word "unitless". This reaches the steel, linear and radial gauges, the data-chart title and the history dialog's axis, which all label themselves from `getUnitDisplaySymbol` with a resolved measure.
- **README gained a Display Units section**: where unit preferences live in the Signal K server admin UI, what each of the six built-in presets selects, how categories carry one choice across hundreds of paths, how to override a single path, and what Skip does when the server names a unit it cannot convert.
- Mock Signal K server gained meta emission (`control.selfMeta`) — it sent values but never meta, so nothing unit-dependent could be exercised in a browser — plus a `shot-fuel-rate.mjs` probe for this case.

Review follow-ups routed to issues rather than fixed here: halos-org/skip#574 (a path whose Signal K unit is `deg` is scaled by 57.3 — pre-existing, and why the `angleDegrees` row is absent from the vocabulary table), halos-org/skip#575 (a Numeric widget keeps stale Min/Max across a live preference switch), halos-org/skip#576 (no component spec can observe a rendered unit label), and SignalK/signalk-server#2951 (the server's own `gal/h` and `gal-imp/h` formulas are 3600x too small, so Skip and the SK admin UI disagree on imperial fuel rate — Skip's factors are the correct ones).

Mass (`kg`) and area (`m2`) paths stay unresolved: Skip has no conversion group for either, so even the identity target fails. Tracked in halos-org/skip#570.

## Verification

`npm run ci` green: lint, snc, 2033 unit tests, 33 schema tests. A new spec case asserts every preset pair resolves to a group-valid measure with a working conversion and a non-empty symbol, so a future preset addition fails loudly.

Rendered end-to-end against the mock server streaming `propulsion.0.fuel.rate` at 2e-6 m³/s (= 7.2 L/h) with the metric preset's `volumeRate` meta — steel gauge, linear gauge, numeric tile:

| before | after |
| --- | --- |
| `0.00` labelled `unitless` | `7.20` labelled `l/h` |

Fixes #536


mairas:
# Code review — server unit-preference vocabulary

**Scope:** `851353b4..1d437018` (6 files at review time; 9 after fixes). **Mode:** interactive, fixes applied on the branch.

**Intent:** make Skip honour every display unit the Signal K server asks for, so a fuel rate stops rendering as `0.00 unitless`.

**Team:** correctness, testing, maintainability, project-standards (always-on) + adversarial (diff ~175 non-generated lines; numeric conversions a skipper acts on) + api-contract (regenerated MCP schema artifact, a persisted measure-key enum, and a documented public seam changing behaviour).

Findings below are post-merge across the six reviewers; confidence is boosted where independent reviewers converged. 4 findings were suppressed under the 0.60 gate.

## P2 — High

| --- | --- | --- | --- | --- | --- |
| 1 | `units.service.ts:786` | Alias table consulted before the group check, so a target that *is* a measure loses its meaning: a charge path asking for `C` (the identity target the server's `base` category emits) became Celsius, missed the Charge group, and dropped its label | correctness, maintainability, adversarial, api-contract | 0.93 | **fixed** `334cb27e` |
| 2 | `units.service.ts:107` | Vocabulary scoped to the six presets, but a per-path override or custom preset selects from the whole definitions file — `L/min` sat one line from the `L/h` that caused #536, same case-only mismatch, plus `meter`, `radian`, `gradian`, `hertz`, `watt`, `second`, `minute`, `day` | correctness, maintainability, adversarial, api-contract | 0.95 | **fixed** `334cb27e` |
| 3 | `units.service.ts:799` | An unhonourable target still degraded silently — the same shape as #536, which is why #536 needed a user's bug report to find. The neighbouring unsupported-base-unit branch already logs | maintainability, correctness, adversarial, api-contract | 0.95 | **fixed** `334cb27e` — one console line per distinct target |
| 4 | `CHANGELOG.md` | Understated the upgrade: newly-resolvable targets override each widget's stored unit (`widget-streams.directive.ts:216` uses the stored measure only while the resolved one is `unitless`), and on Metric presets uncategorised metre paths — XTE, antenna altitude, anchor distance — become kilometres, so a 25 m XTE reads 0.03 km | api-contract, adversarial | 0.85 | **fixed** `ba1731e9` |
| 5 | component specs | No test can see a rendered unit label: ~12 component specs stub `getUnitDisplaySymbol` as identity, which is precisely the pre-fix behaviour. A regression reprinting "unitless" on the steel/radial gauge, chart title or history axis passes CI | testing, api-contract | 0.88 | issue #576 |
| 6 | `perf-harness/lib/server.mjs:202` | Meta latched per connection, so swapping `selfMeta` on a live page kept streaming the first frame — a preference-switch probe would have measured the old units and looked green | adversarial | 0.80 | **fixed** `1d437018` |
| 7 | `units.service.ts:494` | Skip's `gal/h` and `gal-imp/h` are 3600× the server's own formulas. Skip is right (1 m³/s = 951019.39 US gal/h); the server's factors are gallons per *second*. Same labelled quantity, two answers, on every imperial preset | adversarial, correctness | 0.90 | upstream SignalK/signalk-server#2951 |

## P3 — Low

| --- | --- | --- | --- | --- | --- |
| 8 | `units.service.spec.ts:190` | The `angleDegrees deg -> deg` row blessed a dimension error: Skip's only `deg` measure converts *from* radians, so honouring it scales an already-degrees value by 57.3 (a 137° heading renders 7849.5°). Finiteness is not correctness | adversarial, correctness | 0.85 | row removed; hazard filed as #574 |
| 9 | `units.service.spec.ts:281` | The Unitless exemption skipped the conversion assertion too, not just the symbol one — and `unitless` is the measure every unresolved path lands on | testing, api-contract | 0.90 | **fixed**: positive pin (`symbol === ''`) instead of a skip |
| 10 | `units.service.ts:708` | `getRenderableUnitSymbol` became a pure alias once the Unitless measures carried an empty symbol, and its doc still claimed the other method "returns the measure key itself — reserving room for … the word unitless" | maintainability, api-contract | 0.85 | doc corrected; collapsing the two methods (~20 call sites) left out of this PR |
| 11 | `units.service.spec.ts:229` | The vocabulary loop's `not.toBe('')` cannot fail independently, and the three new symbols were asserted nowhere; the `gal/h` tolerance pinned js-quantities' rounding 8× tighter than the exact US gallon | testing | 0.83 | **fixed**: explicit symbol assertions, physical tolerances |
| 12 | `widget-numeric.component.ts:98` | `subscriptionSignature` keys on the stored unit, not the resolved measure, so a live preference switch leaves Min/Max in the old unit — Celsius extremes under a Fahrenheit heading, and the stale minimum sticks for the session. Pre-existing, but this PR makes nearly every category flip live | adversarial | 0.75 | issue #575 |
| 13 | `perf-harness/README.md`, `lib/server.mjs:125` | New committed probe undocumented where every other probe is; `control`'s JSDoc omits the `selfMeta` this PR adds | project-standards, maintainability | 0.80 | **fixed** `1d437018` |
| 14 | `shot-fuel-rate.mjs:34` | Probe copied the lib's private `node()` and re-declared factories instead of extending `lib/skip-config.mjs`, so it rots silently when the persisted widget shape moves | maintainability | 0.75 | **fixed** `1d437018` |
| 15 | `units.service.spec.ts:243`, `:186`, `units.service.ts:68` | Test name narrates the diff ("previously had no measure for") and covers BTU, which is neither imperial nor volume; `(tracked separately)` names no issue; `IUnit.symbol`'s doc does not cover `''` as a meaningful value | maintainability, project-standards, api-contract | 0.70 | **fixed** |
| 16 | commit `bafdb375` | Subject is 57 chars against the repo's 50-char convention | project-standards | 0.85 | **declined** — rewriting pushed history mid-review costs more than the 7 characters |
| 17 | `shot-fuel-rate.mjs` | New `.mjs` under skip/CLAUDE.md's "no new JavaScript source files" rule; perf-harness has no TS toolchain and all ten existing probes are `.mjs`, so the written rule and the practice disagree | project-standards | 0.60 | your call — either carve the harness out in CLAUDE.md or file the migration |

## Pre-existing (action required)

| Finding | Route |
| --- | --- |
| `C`-alias collision (finding 1) | fix-here — fixed in this PR |
| `deg`-base paths scaled by 57.3 (finding 8) | file-issue — #574 |
| Numeric Min/Max across a live measure change (finding 12) | file-issue — #575 |
| Mass/area paths have no conversion group | file-issue — #570 (filed with the original PR) |

## Coverage

- **Verified clean by project-standards:** VERSION policy (1.4.0 vs latest stable v1.3.0+2 — cycle already open by a minor bump, no re-bump owed; `version-bump-check` passes), the regenerated MCP schema artifact matches `buildSchema()`, eslint clean, changelog placement and voice, and `closingIssuesReferences` is exactly #536.
- **Residual risks:** the alias table is a hand-copied snapshot, so it detects Skip-side drift but not the server adding a preset — the new console warning is the mitigation, not a test. Skip still ignores the server's `formula`/`symbol`, which is what keeps it correct despite the upstream `gal/h` bug but means the table can only grow. New measure keys persist into saved dashboards, so a config authored on 1.4.0+ shows nothing for those widgets if rolled back to 1.3.x.
- **Testing gaps:** no render-site assertion (#576); no coverage of a live preset switch on a mounted widget; the probe asserts tile count only, so the reading and label it exists to prove are a human eyeball.
- **Excluded from scope:** four untracked `perf-harness/*.mjs` dev scripts predating this branch.

---

**Verdict: ready with fixes — all applied.** `npm run ci` green (2036 unit + 33 schema tests), CI green on `1d437018`, and the fuel-rate probe re-shot after the refactor still reads 7.20 / 7.2 L/H / 7.2 with `l/h` labels. Four findings were routed to issues (#574, #575, #576, plus upstream SignalK/signalk-server#2951) rather than fixed here; two were declined with reasons above.

PR #572, opened 2026-08-12
Start Compact Linear unfiltered instead of preset to voltage

### LINKED ISSUE #541: Compact Linear defaults its path Filter to Voltage instead of All
Adding a Compact Linear widget opens its settings with the **Filter** dropdown preset to `Electrical: Volt`, so the Signal K Path autocomplete only offers voltage paths. A generic gauge should start unfiltered — the Filter's `All` option.

The default comes from `WidgetSimpleLinearComponent.DEFAULT_CONFIG` (`src/app/widgets/widget-simple-linear/widget-simple-linear.component.ts:36-37`):

```
pathSkUnitsFilter: 'V',
convertUnitTo: 'V'
```

Every other general-purpose gauge ships `pathSkUnitsFilter: null` + `convertUnitTo: 'unitless'`, which is what renders as `All`: Linear (`widget-gauge-ng-linear.component.ts:61-63`), Radial (`widget-gauge-ng-radial.component.ts:56-58`), Steel (`widget-gauge-steel.component.ts:42-44`), Numeric (`widget-numeric.component.ts:47-49`). The widgets that legitimately preset a filter (Wind Steer, Race Steer, Racer Timer) are all fixed-purpose. Compact Linear is the outlier.

`convertUnitTo` and `displayScale` carry the same voltage assumption (`0..15`, versus Steel's neutral `0..100`), so the fix should decide whether to align all three or only the filter.

Existing widget instances store their own config, so a default change affects only newly added widgets; no migration is needed.

mairas:
## Why

Adding a Compact Linear widget opened its settings with the **Filter** dropdown preset to `Electrical: Volt`, so the Signal K Path autocomplete only offered voltage paths. Every other general-purpose gauge — Linear, Radial, Steel, Numeric — ships `pathSkUnitsFilter: null` + `convertUnitTo: 'unitless'`, which renders as `All`. The widgets that legitimately preset a filter (Wind Steer, Race Steer, Racer Timer) are fixed-purpose; Compact Linear is not.

## What

`WidgetSimpleLinearComponent.DEFAULT_CONFIG` now matches its siblings: filter `null`, `convertUnitTo: 'unitless'`, and `displayScale` 0–100 rather than the matching voltage-shaped 0–15. The `displayUpper` fallback moved to 100 for the same reason — it re-encoded the same assumption one line over.

Existing widget instances carry their own stored config, so this affects only newly added widgets and needs no migration.

`src/assets/skip-dashboard-schema.json` is the regenerated artifact (`npm run gen:mcp-schema`).

Fixes #541


mairas:
## Code review — 4 personas, findings applied

Reviewers: correctness, testing, maintainability, project-standards. Each ran in a fresh context against an isolated worktree of this branch.

### The fix was incomplete, and two reviewers found the same hole independently

Correcting `DEFAULT_CONFIG` only reaches widgets a user adds from now on. The pages Skip ships are stored config too — and four of their Compact Linear gauges carried `pathSkUnitsFilter: "V"` in their own saved settings, on paths that are not voltage:

| Line | Path | Filter | convertUnitTo |
|---|---|---|---|
| 447 | `self.tanks.fuel.main.currentLevel` | `"V"` | `percent` |
| 1470 | `self.propulsion.main.oilTemperature` | `"V"` | `celsius` |
| 1566 | `self.propulsion.main.coolantTemperature` | `"V"` | `celsius` |
| 1651 | `self.tanks.fuel.main.currentLevel` | `"V"` | `percent` |

`WidgetRuntimeDirective.options()` merges the default under the saved config, so the seeded `"V"` beats the new `null`. The correctness reviewer traced the consequence: `path-control-config.component.ts:71` preselects the Filter dropdown from the stored value and `:183` restricts the autocomplete to `meta.units === 'V'`. So on a fresh install, opening the shipped Fuel gauge's settings showed a path list with its own bound path missing from it — the symptom this PR closes, on the widgets a new user opens first. The seed is stamped at `LATEST_APP_CONFIG_VERSION`, so no migration was ever going to correct it.

Fixed in 49ad05ef: the four filters are now `null`, and the CHANGELOG says so.

### Both halves are now guarded

Neither half had a regression test, and the schema artifact does not substitute — `schema-artifact.spec.ts` byte-compares the committed JSON against generator output, so a revert plus a regeneration passes. Two guards added, each following a precedent that already exists:

- `config.blank.dashboard.spec.ts` — seeded Compact Linear slots must match `DEFAULT_CONFIG`'s filter, following the wind-steer seed test in the same file. Mutation-checked: restoring `"V"` fails it with `expected 'V' to be null`.
- `widget-schema.spec.ts` — pins the `gaugePath` slot at `defaultConvertUnitTo: 'unitless'`, `expectedSkUnit: null`, alongside the existing `widget-numeric` case.

### Cleared

The `?? 15` → `?? 100` change is not a back-compat hazard. `options()` merges the default under the saved config and `serialize()` persists the merged object, so a placed widget's stored `displayScale.upper` always wins; the fallback is unreachable through Host2. Against the siblings, the gaugePath slot is now byte-identical to `widget-gauge-ng-linear` and `widget-gauge-ng-radial`, and `displayScale` matches. `numDecimal` still differs (2 vs 0), deliberately, for a compact readout.

Version reasoning confirmed correct: latest stable is `v1.3.0+2`, `VERSION` is already 1.4.0, so the cycle is open and a patch-level fix owes no re-bump.

### Not actioned

Two P3s, both pre-existing and both the repo's house pattern rather than debt this PR introduces:

- `?? 100` at line 87 restates `DEFAULT_CONFIG.displayScale.upper` at line 41. `widget-gauge-ng-linear`, `widget-gauge-ng-radial` and `gauge-steel` all do the same beside their own defaults. Changing it here alone would make this widget the odd one out; changing it repo-wide is a separate PR.
- `widget-simple-linear` does not set `suppressBootstrapNull: true` where `ng-linear`, `ng-radial` and `numeric` do, so it renders the bootstrap null as `--`. Unrelated to this issue and a behaviour choice, not a defect.

Gate on the head commit: 2037 tests pass, `snc` clean, `test:mcp-schema` 34 pass.

PR #580, opened 2026-08-12
Add the mass and area conversion groups

### LINKED ISSUE #570: Mass and area paths resolve to unitless: no conversion group exists for kg or m2

mairas:
> **Stacked on [#571](https://github.com/halos-org/skip/pull/571)** (`fix/server-unit-preference-vocabulary`), which is where the preset-vocabulary spec table lives. Retarget to `main` once #571 merges.

## Why

`getConversionsForPath` filters `_conversionList` for a group containing a measure equal to the path's SI unit. No group contained `kg` or `m2`, so the filter came back empty and the path degraded to `unitless` — even for the identity target, and on every preset including metric. `design.displacement` and a sail area rendered a raw number with no label.

This was split out of the #536 fix, which could not complete these two categories because the gap is not an alias: Skip had no conversion group for them at all.

## What

| category | SI base | targets |
| --- | --- | --- |
| Mass | `kg` | `kg`, `lbs` |
| Area | `m2` | `m2`, `sqft` |

Plus the matching `unitConversionFunctions` entries (js-quantities handles `kg -> lbs` and `m^2 -> ft^2`), and one alias — the server preset names its imperial mass target `pound`, following the same shape as the existing `foot: 'feet'`. `sqft` needs none; the server's target name and Skip's measure agree.

`skBaseUnits` already listed both `kg` (Mass) and `m2` (Area) as valid Signal K units, so the type surface expected them; only the conversion table was missing.

The two categories move out of the "deliberately absent" notes in both `SERVER_TARGET_UNIT_ALIASES` and the spec's preset-vocabulary table, and into the table itself as four new rows.

`src/assets/skip-dashboard-schema.json` is the regenerated artifact (`npm run gen:mcp-schema`).

## Tests

Three added, plus the four new vocabulary rows. All four cases fail without the groups (the existing preset-vocabulary test included).

Full suite: 2039 passing.

Fixes #570


mairas:
Closed automatically by GitHub when the base branch `fix/server-unit-preference-vocabulary` was deleted on merging [#571](https://github.com/halos-org/skip/pull/571). A closed PR whose base is gone cannot be retargeted or reopened, so this continues as [#583](https://github.com/halos-org/skip/pull/583) — same branch and same commit, rebased onto `main`.

PR #607, opened 2026-09-02
Detect Autopilot API v2 providers instead of requiring a known autopilot plugin

### LINKED ISSUE #605: Autopilot Head should detect Autopilot API v2 providers without requiring specific plugins
## Problem

The Autopilot Head widget currently appears to be gated by a fixed plugin dependency list, such as the legacy Signal K autopilot plugin or a specific provider plugin. This prevents the widget from being added when Signal K has an active Autopilot API v2 provider registered by another plugin.

This is a problem because the Signal K Autopilot API v2 is provider-based. A compatible autopilot can be exposed through:

`/signalk/v2/api/vessels/self/autopilots`

without the historical `autopilot` plugin being active.

## Expected behavior

When adding or configuring the Autopilot Head widget, Skip should consider the widget available if the server exposes at least one Autopilot API v2 provider at:

`/signalk/v2/api/vessels/self/autopilots`

The widget should not require a hard-coded plugin id if the v2 API already reports an available provider.

## Current behavior

The widget can be blocked before its own Autopilot API discovery/configuration flow runs, with a message indicating that one of the known autopilot plugins is installed but none is active.

## Suggested direction

The Autopilot Head dependency check could be relaxed or changed from a hard-coded plugin-id requirement to an API capability check:

1. Prefer Autopilot API v2 provider discovery via `/signalk/v2/api/vessels/self/autopilots`.
2. Allow the widget when that endpoint returns one or more providers.
3. Keep legacy plugin checks only as a fallback for v1 compatibility.

This would make the widget work with any compatible Signal K Autopilot API v2 provider, while preserving support for existing provider plugins.

mairas:
Skip gated the Autopilot Head widget on two plugin ids — `autopilot` and `pypilot-autopilot-provider`. Signal K's Autopilot API v2 is provider-based, so a pilot registered by any other plugin left the widget greyed out in the Add Widget list, or blocked the add with a toast saying an autopilot plugin is installed but inactive. The widget's own config dialog already discovers providers over `/signalk/v2/api/vessels/self/autopilots` and never checks a plugin id, so the gate was the only thing in the way.

A widget definition can now name endpoints that satisfy its any-of plugin gate on their own (`anyOfApis: string[]`), and Autopilot Head names the v2 provider registry. An endpoint counts as satisfied when it answers with a non-empty provider collection; an empty registry, a 404 and a 401 all read as no provider. `requiredPlugins` still hard-gates — an API never stands in for a required plugin — and the endpoint is probed only when the plugin lists would otherwise block the widget, so the usual paths gain no round trip.

Verified in a browser against the production bundle, driving the real Add Widget list with the endpoint answering as a third-party provider plugin (`{"my-pilot":{"provider":"some-third-party-autopilot-plugin","isDefault":true}}`) and, for the negative case, `{}`. Card enabled with a provider (`aria-disabled` absent), still blocked without one (`aria-disabled=true`) — the two states stay visually distinct, red for blocked and struck-through grey for satisfied-anyway. `npm run ci` is green (2137 + 34 tests).

Closes #605


mairas:
## Code review — 7 personas

**Scope:** `ba8080a4..83acfe44` (branch `fix/autopilot-api-v2-detection`). **Model:** Opus 5, inherited from the session.

**Intent:** stop gating Autopilot Head on two hard-coded plugin ids; allow it whenever the Autopilot API v2 provider registry reports a provider, with required plugins still hard-gating and no added round trip on paths the plugin lists already satisfied.

**Team:** correctness, testing, maintainability, project-standards (always-on) · api-contract (new exported field on the `WidgetDescription` registry interface) · reliability (network probe added to a UI gate path) · adversarial (~80 changed non-test lines gating what a user may add).

### P2 — Moderate

|---|---|---|---|---|---|
| 1 | `tools/gen-mcp-schema/generate.ts:346` | The generated widget catalog dropped `anyOfApis`. `toCatalogEntry` copies named fields, so the artifact still described Autopilot Head as gated on the two plugin ids — the same wrong conclusion this PR fixes in the UI — and CI could not catch it, because the test compares the committed file against a fresh generation and both omitted the field. | project-standards, api-contract, correctness | 0.95 | fixed |
| 2 | `src/app/core/services/widget.service.ts:849` | The probe had no deadline. Angular's `HttpClient` never times out on its own, so a wedged provider endpoint held the whole Add Widget list open — every widget unreachable, no spinner, no error — and on the add path swallowed the click with neither a widget nor a toast. Every other awaited fetch in `core/services` bounds itself. | reliability, adversarial | 0.90 | fixed |
| 3 | `src/app/core/components/widgets-list/widgets-list.component.ts:65` | `onSelectWidget` rebuilds `WidgetDescription` field by field, so that one spread is the only thing carrying `anyOfApis` to the add-time gate. Dropping it would leave the card available and the add refused — this exact bug, silently reintroduced — with the whole suite green. | testing, maintainability, api-contract, reliability, adversarial | 0.85 | fixed |

### P3 — Low

|---|---|---|---|---|---|
| 4 | `src/app/core/services/widget.service.ts:800` | `anyOfApis` only ever ran for widgets that also declared `anyOfPlugins`: a widget naming an endpoint alone was never a probe candidate and shipped ungated, with no compile or test signal. The field name invites exactly that declaration. | adversarial, maintainability | 0.85 | fixed |
| 5 | `src/app/core/services/widget.service.ts:852` | The bare `catch { return false }` collapsed a 5xx, a network drop and an empty registry into one answer with nothing logged, so a transient blip flips the widget into the "activate the plugin" toast and no field report can tell the cases apart. | reliability | 0.70 | fixed |
| 6 | `src/app/core/services/widget.service.ts:494` | The endpoint literal appeared in three modules. Whoever updates the two `API_PATHS` constants has no reason to look in the widget registry. | maintainability | 0.75 | fixed (partly — see below) |
| 7 | `src/app/core/services/widget.service.ts:107` | The `isDependencyValid` doc still said "installed/available dependency checks"; the flag can now be true with no `pluginsStatus` entry enabled. | api-contract | 0.68 | fixed |
| 8 | `src/app/core/services/widget.service.ts:800` | The post-pass linked two passes through a `Set` of widget object references. Any later edit that copies a widget between them breaks the lookup with no type error and no obvious symptom. | maintainability | 0.62 | fixed |
| 9 | `src/app/core/services/widget.service.spec.ts:28` | New HTTP spec with no `httpMock.verify()` — a stray or unflushed probe would pass unnoticed. Sibling service specs all verify. | project-standards | 0.66 | fixed |
| 10 | `src/app/core/components/dashboard/dashboard.component.spec.ts:420` | Both gate tests stubbed `hasAnyApiProvider` with a fixed value and asserted only the outcome; passing the wrong field or `undefined` would leave them green. | testing, correctness | 0.80 | fixed |
| 11 | `src/app/core/services/widget.service.ts:850` | The array-response branch had no reachable caller and no test — a guess at an API shape no declared endpoint returns. | testing, maintainability | 0.75 | fixed (branch removed) |
| 12 | `src/app/core/services/widget.service.ts:849` | The probe uses a bare origin-relative path while the plugin half routes through `PluginConfigClientService.toServerUrl`, which prefixes the configured Signal K URL. | api-contract | 0.62 | documented, not changed |

### Applied fixes

Two commits, `f5632113` and `83acfe44`:

- **`anyOfApis` reaches the MCP catalog.** Added to `WidgetCatalogEntry` and `toCatalogEntry`, artifact regenerated. `widget-autopilot` now carries both the plugin list and the API alternative.
- **The probe is bounded.** `timeout(5000)` — the value `authentication.service.ts` and `storage.service.ts` already use for a fetch that gates a user-facing decision — and the catch logs the endpoint and error before returning `false`.
- **The any-of gate is the union of both lists.** A widget declaring `anyOfApis` alone is now gated on it rather than silently ungated. Computed in one pass, so no widget identity is carried between passes. No behavior change for Autopilot Head, which declares both.
- **Endpoint path shared** between `select-autopilot` and `widget-autopilot` via `V2_AUTOPILOTS_PATH`. The widget registry entry deliberately keeps a literal: `gen-mcp-schema` parses `_widgetDefinition` through the TypeScript AST and rejects anything that is not a static literal (it fails with "Unsupported (non-literal) syntax"). That constraint, and the requirement to mirror new gating fields into the catalog, is now written into `CLAUDE.md` — it is not discoverable from the registry itself.
- **Tests:** the `onSelectWidget` passthrough (both lists carried, absent lists omitted, invalid widget ignored), the probe argument on the add path, the no-probe-when-a-plugin-is-active short-circuit, a probe that never answers, and `httpMock.verify()`.

Finding 3's test was checked against a deliberately broken build: removing the `anyOfApis` spread fails it (`expected "close" to be called with …`), and it passes once restored.

### Verification

Headless Chromium against the production bundle, driving the real Add Widget dialog, with the endpoint intercepted:

| Endpoint answers | List renders | Autopilot Head |
|---|---|---|
| `{"my-pilot":{"provider":"some-third-party-autopilot-plugin","isDefault":true}}` | 578 ms | enabled (`aria-disabled` absent) |
| `{}` | 578 ms | blocked (`aria-disabled=true`) |
| never (connection held open) | 5364 ms | blocked, list still renders |

The third row is the fix for finding 2: before it, that request had no deadline at all.

`npm run ci` green — lint, `snc`, 180 spec files, 7 MCP-schema files.

### Pre-existing (filed)

- [#608](https://github.com/halos-org/skip/issues/608) — the card gate reads plugin **installed**, the add gate reads **enabled**, so an installed-but-disabled plugin gives an enabled-looking card that refuses the add. Predates this PR; it now costs a round trip before the refusal.
- [#609](https://github.com/halos-org/skip/issues/609) — the Add Widget dialog renders an empty grid while its dependency scan runs, with no loading or error state, and the plugin lookups themselves are still unbounded.

### Coverage

Suppressed below the 0.60 confidence gate: none reported. No reviewer failed or timed out. Untracked files excluded from scope: four `perf-harness/shot-*.mjs` scratch scripts from earlier sessions.

Known gap, unfixed on purpose: four reviewers noted that the "a provider never stands in for a required plugin" guard is untested. No shipped widget pairs `requiredPlugins` with `anyOfApis` and `_widgetDefinition` is private, so covering it would mean opening the registry for injection. The guard is stated in both paths and reachable only by a future widget that combines the two.

**Verdict: ready to merge.** All P2s fixed and verified, both pre-existing findings filed.

